Issue #2, opened 2015-04-25
Single Page Apps


benaadams:
```trackPageView``` correctly uses ```performance.timing```, however for SPA using xhr this is always first page load time. Use of [Resource Timing](http://www.w3.org/TR/resource-timing/) can get round this.

PerformanceTiming 
```javascript
{connectEnd: 1429922346337
connectStart: 1429922346216
domComplete: 1429922378350
domContentLoadedEventEnd: 1429922378038
domContentLoadedEventStart: 1429922378028
domInteractive: 1429922378028
domLoading: 1429922377408
domainLookupEnd: 1429922346216
domainLookupStart: 1429922346216
fetchStart: 1429922346215
loadEventEnd: 1429922378361
loadEventStart: 1429922378350
navigationStart: 1429922345941
redirectEnd: 0
redirectStart: 0
requestStart: 1429922346337
responseEnd: 1429922377776
responseStart: 1429922377406
secureConnectionStart: 0
unloadEventEnd: 0
unloadEventStart: 0}
```
PerformanceResourceTiming
```javascript
{connectEnd: 0
connectStart: 0
domainLookupEnd: 0
domainLookupStart: 0
duration: 13486.342999996617
entryType: "resource"
fetchStart: 10071187.829000002
initiatorType: "xmlhttprequest"
name: "http://..../"
redirectEnd: 0
redirectStart: 0
requestStart: 0
responseEnd: 10084674.171999998
responseStart: 0
secureConnectionStart: 0
startTime: 10071187.829000002}
```

However that's also a little messy in that you need to search though it  to find your xhr page load and it has everything in it, images, xhr (including ```dc.services.visualstudio.com/v2/track```) etc so you need to pick out the correct one. Also you need to periodically clear the buffer when it gets full to add more recent stuff (e.g. 150 items in chrome):
```javascript
if (window.performance) {
    if (performance.clearResourceTimings) {
        performance.clearResourceTimings();
    } else if (performance.webkitClearResourceTimings) {
        performance.webkitClearResourceTimings()
    }
}
```
Equally, the user will need to manually calculate the domProcessing, perhaps using ```setImmedate``` or polyfill for it. So there isn't an immediate "this is the right way solution" that I can think of that can be automated, without bundling it into the xhr loader + page updater.

[...]

PR #3, opened 2015-04-25
Example change for SPA

benaadams:
Mock up example for #2

### DIFF
--- a/JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
+++ b/JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
@@ -49,7 +49,7 @@ module Microsoft.ApplicationInsights.Telemetry {
              *  |---network---||---request---|---response---|---dom---|
              *  |--------------------------total----------------------|
              */
-            var total: number;
+            var total: number = 0;
             var network: number;
             var request: number;
             var response: number;
--- a/JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
+++ b/JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
@@ -49,55 +49,55 @@ module Microsoft.ApplicationInsights.Telemetry {
              *  |---network---||---request---|---response---|---dom---|
              *  |--------------------------total----------------------|
              */
+            var total: number;
+            var network: number;
+            var request: number;
+            var response: number;
+            var dom: number;
 
             if (timings) {
-                durationMs = timings.duration;
-                if (timings.perfTotal) {
-                    this.perfTotal = timings.perfTotal;
-                    this.networkConnect = timings.networkConnect;
-                    this.sentRequest = timings.sentRequest;
-                    this.receivedResponse = timings.receivedResponse;
-                    this.domProcessing = timings.domProcessing;
-                }
+                total = timings.total;
+                network = timings.network;
+                request = timings.request;
+                response = timings.response;
+                dom = timings.dom;
             } else {
-
                 var timing = PageViewPerformance.getPerformanceTiming();
                 if (timing) {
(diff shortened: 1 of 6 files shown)


benaadams:
Example how you might use it for Jquery SPAs: https://gist.github.com/benaadams/f7d1e9136d8c60939467

MaxShehovtsov:
@benaadams Hi Ben, I've created a new branch https://github.com/Microsoft/ApplicationInsights-JS/tree/spaTimings where I ported your changes just to make is compile and pass all tests (your's became outdated). If you're still interested in bringing it in please merge it to your branch so that the pull request gets updated. At this point the feedback I have is to refactor it to reduce code duplication, after it's done I'll be happy to merge it.

Thanks!

benaadams:
@MaxShehovtsov thank you for the sync update

Ignoring whitespace I think its now merged ok: https://github.com/Microsoft/ApplicationInsights-JS/pull/3/files?w=1

JavaScript/JavaScriptSDK/Telemetry/Timings.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        constructor(durationMs: number, networkMs: number, sendMs: number, receiveMs: number, domProcessingMs: number) {
+            this.duration = 0;
+            var total = Timings.getDuration(durationMs);
+            var network = Timings.getDuration(networkMs);
+            var request = Timings.getDuration(sendMs);
+            var response = Timings.getDuration(receiveMs);
+            var dom = Timings.getDuration(domProcessingMs);
+            
+            if (total < Math.floor(network) + Math.floor(request) + Math.floor(response) + Math.floor(dom)) {
```
MaxShehovtsov: Looks like Timings acts as a container which you fill out and pass to trackPageView. So it shouldn't do any checks here as they are done in PageViewPerformance.

JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
```diff
@@ -49,43 +49,57 @@ module Microsoft.ApplicationInsights.Telemetry {
...
+                    this.perfTotal = timings.perfTotal;
+                    this.networkConnect = timings.networkConnect;
+                    this.sentRequest = timings.sentRequest;
+                    this.receivedResponse = timings.receivedResponse;
+                    this.domProcessing = timings.domProcessing;
+                }
+            } else {
+
+                var timing = PageViewPerformance.getPerformanceTiming();
+                if (timing) {
```
MaxShehovtsov: This is common for both cases (with and without custom timings) - getting the values from timings "provider" and checking for total (== 0 and >= than a sum of it's parts). Could you pls refactor it?

MaxShehovtsov:
Also looks like unit tests need to be extended to account for timings case.

benaadams:
Should now be reusing same logic, and just acting as overriding container

MaxShehovtsov:
Cool! looking forward to seeing the unit tests! :+1:

benaadams:
Might take a little while, need to learn a few bits around the testing methodology first

MaxShehovtsov:
For this we need just unit tests (we also have E2E tests, perf tests which are not needed for this pull requests). It's under JavaScriptSDK.Tests/CheckinTests. We run it with Chutzpah (from the visual studio), another way is to open the test harness html made up for selenium tests - JavaScriptSDK.Tests/Selenium/Tests.html (in the same folder - checkinTests.ts which compiles all checkin tests together).
Also we had to tweak chutzpah and the tests are compiled "manually" i.e. with powershell - see inside JavaScriptSDK.Tests.csproj, at the end of the proj definition - this is in case you need to dig into it (hopefully not).

Hope this helps

benaadams:
Not realistically going to get round to this.

PR #8, opened 2015-05-27
- Stop sending "Script error." - Adding try/catch blocks to public api

MaxShehovtsov:
Because of CORS some users are getting "Script error." with no details (url/stack trace/file name). This is not actionable. In this fix "Script error." 's are filtered out.
For our own exceptions (coming from the script on CDN - ai.0.js) a tracing has been added to report meaningful errors when they happen. Otherwise we have the same "Script error." problem.

### DIFF
--- a/JavaScript/JavaScriptSDK/Sender.ts
+++ b/JavaScript/JavaScriptSDK/Sender.ts
@@ -87,13 +87,13 @@ module Microsoft.ApplicationInsights {
         
                 // validate input
                 if (!envelope) {
-                    _InternalLogging.throwInternalUserActionable(LoggingSeverity.WARNING, "Cannot send empty telemetry");
+                    _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.CRITICAL, "Cannot send empty telemetry");
                     return;
                 }
 
                 // ensure a sender was constructed
                 if (!this._sender) {
-                    _InternalLogging.warn("No sender could be constructed for this environment, payload will be added to buffer." + Serializer.serialize(envelope));
+                    _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.CRITICAL, "Sender was not initialized");
                     return;
                 }
             
--- a/JavaScript/JavaScriptSDK/Sender.ts
+++ b/JavaScript/JavaScriptSDK/Sender.ts
@@ -78,42 +78,45 @@ module Microsoft.ApplicationInsights {
          * Add a telemetry item to the send buffer
          */
         public send(envelope: Telemetry.Common.Envelope) {
-            
-            // if master off switch is set, don't send any data
-            if (this._config.disableTelemetry()) {
-                // Do not send/save data
-                return;
-            }
+            try {
+                // if master off switch is set, don't send any data
+                if (this._config.disableTelemetry()) {
+                    // Do not send/save data
+                    return;
+                }
         
-            // validate input
-            if (!envelope) {
-                _InternalLogging.throwInternalUserActionable(LoggingSeverity.WARNING, "Cannot send empty telemetry");
-                return;
-            }
+                // validate input
+                if (!envelope) {
+                    _InternalLogging.throwInternalUserActionable(LoggingSeverity.WARNING, "Cannot send empty telemetry");
+                    return;
+                }
 
-            // ensure a sender was constructed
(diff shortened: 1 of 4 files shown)


benaadams:
As aside:

https://azure.microsoft.com/en-gb/documentation/articles/app-insights-web-failures-exceptions/

> In a web browser; if you include script files from CDNs or other domains, ensure your script tag has the attribute crossorigin="anonymous" and the server sends CORS headers to get a stack trace and detail for unhandled javascript execptions from these resources.

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -78,42 +78,45 @@ module Microsoft.ApplicationInsights {
...
+                }
 
-            // ensure a sender was constructed
-            if (!this._sender) {
-                _InternalLogging.warn("No sender could be constructed for this environment, payload will be added to buffer." + Serializer.serialize(envelope));
-                return;
-            }
+                // ensure a sender was constructed
+                if (!this._sender) {
+                    _InternalLogging.warn("No sender could be constructed for this environment, payload will be added to buffer." + Serializer.serialize(envelope));
```
scsouthw: This warning looks outdated, the payload doesn't get added to the buffer in this case.

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -80,6 +80,18 @@
         }
 
         /**
+        * Checks if error has no meaningful data inside. Ususally such errors are received by window.onerror when error
+        * happens in a script from other domain (cross origin, CORS).
+        */
+        public static isMeaninglessError(message: string, url: string, lineNumber: number, columnNumber: number, error: Error): boolean {
+            return message == "Script error."
```
AlexBulankou: Is "Script error" not browser specific? Also what about localization? Is there any way we can check for it without checking for text?

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -80,6 +80,18 @@
         }
 
         /**
+        * Checks if error has no meaningful data inside. Ususally such errors are received by window.onerror when error
+        * happens in a script from other domain (cross origin, CORS).
+        */
+        public static isMeaninglessError(message: string, url: string, lineNumber: number, columnNumber: number, error: Error): boolean {
```
AlexBulankou: Perhaps a better name for this API would be isCrossOriginError

scsouthw: I think this should be configurable. Although the browser strips most information out of this, the fact that a script on the site is throwing errors could still be useful information.

AlexBulankou:
:shipit:

MaxShehovtsov:
@benaadams Thanks Ben! I was actually considering crossorigin attribute too - we probably need to apply it to our snippet. The only reason I decided to go with try/catch is that there are already customers who receive lots of "Script error." - fixing it directly in script will apply the change without any snippet modification.

JavaScript/JavaScriptSDK/Sender.ts
MaxShehovtsov: Good catch!

JavaScript/JavaScriptSDK/Util.ts
MaxShehovtsov: Based on telemetry it's the same text for different counties and doesn't have localized versions. Let's have it as is until we learn something else.

JavaScript/JavaScriptSDK/Util.ts
MaxShehovtsov: @scsouthw agree - will add it in the next iteration

benaadams:
I do get some useful errors from "Script Error"; get custom properties, get browser versions etc Some I get partial stack traces (using the crossorigin ); get number of users having it, so know where to look and how important it is to address.

So:

capture

Is mostly caused by 1 user with a dodgy toolbar installed which is interacting badly with our app.

capture2

So would probably want it as a configurable property as there is useful information there.

MaxShehovtsov:
@benaadams Right Ben, that's useful because you have stack, file name, line number.

benaadams:
@MaxShehovtsov ah, gochta missed the extra conditions :+1:

PR #15, opened 2015-06-17
Making message for 'missing required field spec' more specific and ma…

AlexBulankou:
…king message and typeName fields on the contract not required

### DIFF
--- a/JavaScript/JavaScriptSDK/Telemetry/Exception.ts
+++ b/JavaScript/JavaScriptSDK/Telemetry/Exception.ts
@@ -10,7 +10,7 @@ module Microsoft.ApplicationInsights.Telemetry {
         public static envelopeType = "Microsoft.ApplicationInsights.Exception";
         public static dataType = "ExceptionData";
 
-        
+
         public aiDataContract = {
             ver: true,
             handledAt: true,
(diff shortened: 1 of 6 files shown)


JavaScript/JavaScriptSDK/Telemetry/Trace.ts
```diff
@@ -11,7 +11,7 @@ module Microsoft.ApplicationInsights.Telemetry {
 
         public aiDataContract = {
             ver: true,
-            message: true,
+            message: false,
```
MaxShehovtsov: What's the reasoning behind this change?

JavaScript/JavaScriptSDK/Telemetry/Exception.ts
```diff
@@ -72,8 +72,8 @@ module Microsoft.ApplicationInsights.Telemetry {
         public aiDataContract = {
             id: false,
             outerId: false,
-            typeName: true,
-            message: true,
+            typeName: false,
```
MaxShehovtsov: I think we need to double check on what's expected by the backend/DC.

AlexBulankou:
Agreed fixing it

PR #16, opened 2015-06-23
Improve diagnostics of unexpected exceptions in AppInsights._onerror

olegsych:
This pull request attempts to improve diagnostics information we log in `AppInsithgs._onerror` method when encountering unexpected exceptions to help us understand the cause of the following error: 

```
AI (Internal): _onerror threw an exception: {} while logging error: null
```

In addition to dumping the names and the properties of the unexpected exception, the `_onerror` method now also writes `toString` representation of the exception object.

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
@@ -655,6 +655,48 @@ class AppInsightsTests extends TestClass {
             }
         });
 
+        this.testCase({
+            name: "AppInsights._onerror creates a dump of unexpected error thrown by trackException for logging",
+            test: () => {
+                var sut = new Microsoft.ApplicationInsights.AppInsights(this.getAppInsightsSnippet());
+                var dumpSpy = sinon.spy(Microsoft.ApplicationInsights.Util, "dump")
+                try {
+                    var unexpectedError = new Error();
+                    sinon.stub(sut, "trackException").throws(unexpectedError);               
+
+                    sut._onerror("any message", "any://url", 420, 42, new Error());
+
+                    Assert.ok(dumpSpy.calledWith(unexpectedError));
+                }
+                finally {
+                    dumpSpy.restore();
+                }
+            }
+        });
+
+        this.testCase({
+            name: "AppInsights._onerror logs dump of unexpected error thrown by trackException for diagnostics",
+            test: () => {
+                var sut = new Microsoft.ApplicationInsights.AppInsights(this.getAppInsightsSnippet());
+                var throwInternalNonUserActionableSpy = sinon.spy(Microsoft.ApplicationInsights._InternalLogging, "throwInternalNonUserActionable");
+                var dumpStub = sinon.stub(Microsoft.ApplicationInsights.Util, "dump");
+                try {
+                    sinon.stub(sut, "trackException").throws(new Error());
+                    var expectedErrorDump: string = "test error";
+                    dumpStub.returns(expectedErrorDump);
+
+                    sut._onerror("any message", "any://url", 420, 42, new Error());
+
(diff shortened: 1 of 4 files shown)


JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -655,6 +655,48 @@ class AppInsightsTests extends TestClass {
             }
         });
 
+        this.testCase({
```
MaxShehovtsov: I'd add another case for null error, and maybe some another exotic case with "", 0, etc instead of error object.

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -148,5 +148,14 @@
...
                 && error == null;
         }
+
+        /**
+        * Returns string representation of an object suitable for diagnostics logging.
+        */
+        public static dump(object: any): string {
+            var objectTypeDump: string = Object.prototype.toString.call(object);
+            var propertyValueDump: string = JSON.stringify(object);
+            return objectTypeDump + propertyValueDump;
```
MaxShehovtsov: will these 2 strings be separated by something?

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
olegsych: It feels `AppInsights._onerror` shouldn't care what the error is, only that it's dumped. I'll add these tests for `Util.dump` and make sure it handles any special cases.

JavaScript/JavaScriptSDK/Util.ts
olegsych: It would look like `[object Object]{property:"value"}` right now. I'm happy to add any additional formatting you feel would be useful.

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -148,5 +148,14 @@
                 && columnNumber == 0
                 && error == null;
         }
+
+        /**
+        * Returns string representation of an object suitable for diagnostics logging.
+        */
+        public static dump(object: any): string {
+            var objectTypeDump: string = Object.prototype.toString.call(object);
+            var propertyValueDump: string = JSON.stringify(object);
```
scsouthw: I think stringify can fail on some object types and it could potentially be very large.

MaxShehovtsov: I think in this case it makes sense to trace the object size and shrink it. And maybe indicate that the object was cut off to make it easier for later telemetry analysis of how many large errors are there.

scsouthw: yeah, tracking the stringified length or truncating would solve the size issue.

For objects with circular structures we would still cause an error to be thrown here. For example try calling dump(window).

MaxShehovtsov: I think this pseudo code makes sense - if error is big it gets cut off, if there's any problem (like circ ref) we'll report it as well and then will be able to analyze it later - it perfectly follows our current strategy data-driven decisions for fixing/improving the code.

try { 
  res = stringify(...) 

  size = size(res)
  if (tooBig(size)) {
    cutOff = true;
    res = "[object too big(" + size + ")]:" + shink(res);
  }

  return res;
} 
catch (e) { 
  InternalLogging.Critical("couldn't stringify"); 
}

MaxShehovtsov: Since this code is only invoked from the window.onerror the cases where an arbitrary large/"unsafe" object can come to this function is limited to:
- no CORS \ arbitrary object is thrown from the script
- arbitary object is thrown from inline script on the page

This outside of scope of this bug. 

Since the result of this fix will be consumed by us as part of weekly telemetry review, the "big arbitrary object" issue will come up and be noticed - and we'll address it if it meets the bar.

PR #20, opened 2015-06-30
E2e html

MaxShehovtsov:
Changing E2E tests to make them runnable in browser. Running with chutzpah got broken with new chutzpah version.
Snippet tests are still to be fixed.

Changes:
- created qunit's test harness htmls 1 per each test class
- made each e2e test class compile in a single js file
- switched from deprecated qunit.asyncTest to qunit.test + assert.async
- removed polling loop as our delay is guaranteed to be greater than our sender's batch interval

### DIFF
--- a/JavaScript/.gitignore
+++ b/JavaScript/.gitignore
@@ -26,5 +26,6 @@ _Chutzpah.*
 /*/E2ETests/ai.js
 /*/E2ETests/sprint*Snippet.js
 /*/E2ETests/testSnippet.js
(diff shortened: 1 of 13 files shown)


olegsych:
At some point I'd like to understand if we actually need all of these end-to-end tests or if they duplicate unit tests without the benefit of serving as [customer tests](http://www.jamesshore.com/Agile-Book/customer_tests.html).

:shipit:

PR #25, opened 2015-07-02
Sampling - initial feedback

MaxShehovtsov:
Collecting initial feedback on sampling implementation.

- added sampling logic
- added unit tests


JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -128,13 +130,20 @@ module Microsoft.ApplicationInsights {
             this._applyDeviceContext(envelope, this.device);
             this._applyInternalContext(envelope, this.internal);
             this._applyLocationContext(envelope, this.location);
-            this._applyOperationContext(envelope, this.operation);
             this._applySampleContext(envelope, this.sample);
             this._applyUserContext(envelope, this.user);
+            this._applyOperationContext(envelope, this.operation);
```
ameyagholkar: Maybe I am missing something.. but why the change in order? If the order IS important, mention that in a comment?

JavaScript/JavaScriptSDK/Context/Sample.ts
```diff
@@ -2,6 +2,48 @@
     "use strict";
 
     export class Sample {
-        public sampleRate: string;
+        public sampleRate: number;
+
+        public INT_MAX_VALUE: number = Math.pow(2, 32) - 1;
```
ameyagholkar: NUMBER.MAX_VALUE?

JavaScript/JavaScriptSDK/Context/Sample.ts
```diff
@@ -2,6 +2,48 @@
...
+            if (sampleRate > 100 || sampleRate < 0) {
+                _InternalLogging.throwInternalUserActionable(LoggingSeverity.CRITICAL, "Sampling rate is out of range (0..100): '" + sampleRate
+                    + "'. Sampling will be disabled, you may be sending too much data which may affect your AI service level.");
+                this.sampleRate = 100;
+            }
+
+            this.sampleRate = sampleRate;
+        }
+
+        public IsSampledIn(envelope: Telemetry.Common.Envelope): boolean {
```
ameyagholkar: I would rather call it - "isSampled".

JavaScript/JavaScriptSDK/Context/Sample.ts
```diff
@@ -2,6 +2,48 @@
...
+                score = Math.random()
+            }
+            
+            return score * 100 < this.sampleRate;
+        }
+
+        public static GetSamplingHashCode(input: string): number {
+            if (input == "") { return 0; }
+
+            var hash: number = 5381;
```
ameyagholkar: For my curiosity, why this specific number?

JavaScript/JavaScriptSDK/Context/Sample.ts
ameyagholkar: JSDoc this to explain about the function

JavaScript/JavaScriptSDK/Context/Sample.ts
ameyagholkar: If this is the seed, is it worth randomizing this as well?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Sample.tests.ts
```diff
@@ -0,0 +1,133 @@
...
+                sample.IsSampledIn(envelope1);
+
+                // assert
+                Assert.ok(mathRandomSpy.calledOnce);
+                mathRandomSpy.restore();
+            }
+        });
+    }
+
+    private GetEnvelope(): Microsoft.ApplicationInsights.Telemetry.Common.Envelope {
```
ameyagholkar: JavaScript semantics? _getEnvelope()

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Sample.tests.ts
```diff
@@ -0,0 +1,133 @@
...
+                                
+                // act
+                var isSampledIn1 = sample.IsSampledIn(envelope1);
+
+                // assert
+                Assert.ok(isSampledIn1);
+            }
+        });
+
+        this.testCase({
```
ameyagholkar: One more test worth adding would be to check what happens if a sample with a illegal sample value is created.

JavaScript/JavaScriptSDK/Context/Sample.ts
MaxShehovtsov: It's too big (2^53), while I'm working with 32 bit integers (FYI - bitwise operations in javascript result with 32 bit signed integers - http://stackoverflow.com/questions/307179/what-is-javascripts-highest-integer-value-that-a-number-can-go-to-without-losin)

JavaScript/JavaScriptSDK/Context/Sample.ts
MaxShehovtsov: isSampledIn is better - meaning that the package is still in after applying sampling logic. Actually naming is taken from the server-side SDK implementation by Vitaly Filimonov.

JavaScript/JavaScriptSDK/TelemetryContext.ts
MaxShehovtsov: Accidentally :)

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Sample.tests.ts
MaxShehovtsov: Thanks!

JavaScript/JavaScriptSDK/Context/Sample.ts
```diff
@@ -2,6 +2,57 @@
...
+        }
+
+        public static getSamplingHashCode(input: string): number {
+            if (input == "") { return 0; }
+
+            var hash: number = 5381;
+
+            for (var i: number = 0; i < input.length; ++i) {
+                hash = ((hash << 5) + hash) + input.charCodeAt(i);
+                hash = hash & hash;
```
Dmitry-Matveev: Should this be XOR instead of AND?

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -128,13 +130,21 @@ module Microsoft.ApplicationInsights {
...
 
             envelope.iKey = this._config.instrumentationKey();
 
-            this._sender.send(envelope);
+            if (this.sample.isSampledIn(envelope)) {
```
Dmitry-Matveev: Some telemetry types are not supposed to be sampled (like SessionState) - are they processed in some other method but not track(envelope)?

JavaScript/JavaScriptSDK/Context/Sample.ts
MaxShehovtsov: The whole point is to make hash a 32 bit int (without changing first 32 bits to be in sync with our server side implementation) - in JS number is 64 bit "thing" that can be both floating point and 53 bit int, BUT bitwise operations always make it 32 bit int. Now, I'm thinking I could avoid hash=hash&hash by deviding the result (in getSamplingScore) by 2^52 instead of 2^31 as I do now... 
I should definitely put a comment. :)

JavaScript/JavaScriptSDK/TelemetryContext.ts
MaxShehovtsov: Excellent point! I should definitely filter out metrics (however session state is throttled as well)

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -65,7 +66,8 @@ module Microsoft.ApplicationInsights {
                 emitLineDelimitedJson: () => this.config.emitLineDelimitedJson,
                 maxBatchSizeInBytes: () => this.config.maxBatchSizeInBytes,
                 maxBatchInterval: () => this.config.maxBatchInterval,
-                disableTelemetry: () => this.config.disableTelemetry
+                disableTelemetry: () => this.config.disableTelemetry,
+                sampleRate: () => this.config.samplingPercentage
```
AlexBulankou: why not have them synchronized - sampleRate and sampleRate or samplingPercentage and samplingPercentage

JavaScript/JavaScriptSDK/Initialization.ts
```diff
@@ -130,6 +130,7 @@ module Microsoft.ApplicationInsights {
             config.verboseLogging = Util.stringToBoolOrDefault(config.verboseLogging);
             config.emitLineDelimitedJson = Util.stringToBoolOrDefault(config.emitLineDelimitedJson);
             config.diagnosticLogInterval = config.diagnosticLogInterval || 10000;
+            config.samplingPercentage = config.samplingPercentage || 100;
```
AlexBulankou: if customer sets sampling percentage 0, they get 100, because 0||100 is 100 is this intentional?

JavaScript/JavaScriptSDK/Context/Sample.ts
```diff
@@ -2,6 +2,57 @@
...
+                score = Math.random()
+            }
+
+            return score;
+        }
+
+        public static getSamplingHashCode(input: string): number {
+            if (input == "") { return 0; }
+
+            var hash: number = 5381;
```
AlexBulankou: comment on why 5381?

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -128,13 +130,21 @@ module Microsoft.ApplicationInsights {
...
+                this._sender.send(envelope);
+            }
+            else {
+                _InternalLogging.logInternalMessage(LoggingSeverity.WARNING,
+                    "Telemetry is sampled and not sent to the AI service. SampleRate is " + this.sample.sampleRate);
```
AlexBulankou: Do we really need this, unclear what customer benefit is. Will this appear in browser console, probably not what we want

JavaScript/JavaScriptSDK/appInsights.ts
MaxShehovtsov: in schema - sampleRate which doesn't make sense, so it's samplingPercentage in config. I can't change schema. :)

JavaScript/JavaScriptSDK/TelemetryContext.ts
MaxShehovtsov: I think I need AI (internal) for that.

JavaScript/JavaScriptSDK/Context/Sample.ts
vitalyf007: Alex, Bing it! :) http://stackoverflow.com/questions/1579721/why-are-5381-and-33-so-important-in-the-djb2-algorithm

JavaScript/JavaScriptSDK/Initialization.ts
vitalyf007: In .Net Sdk there is no check on sampling percentage set by customer. If it is set to a value <=0 that is equivalent to disabling telemetry collection. Value of > 100 gives the same behavior as 100. We probably want to have that in sync...

JavaScript/JavaScriptSDK/TelemetryContext.ts
MaxShehovtsov: checked with Vitaly that both metrics and session state types should not be sampled out.

JavaScript/JavaScriptSDK/Initialization.ts
MaxShehovtsov: Alex is right, 0 will evaluate to false and 100 will be used, totally unintentionally. Vitaly, I'll sync it with server side logic.

JavaScript/JavaScriptSDK/TelemetryContext.ts
MaxShehovtsov: Actually AI (internal) is not a good idea - sending more traces when telemetry gets sampled. All we get to do is to write to browser console and hope that someone will benefit from it.

MaxShehovtsov:
addressed feedback

MaxShehovtsov:
@guys - waiting for :sheep: :fork_and_knife:

ameyagholkar:
:sheep: it

MaxShehovtsov:
@AlexBulankou @vitalyf007 @Dmitry-Matveev - I've addressed the feedback. Thank you!

Dmitry-Matveev:
:sheep: :fork_and_knife:

AlexBulankou:
.
               ,;:'`'::'
            __||
      _____/LLLL\_
      \__________"|
    ~^~^~^~^~^~^~^~^~^~
it

MaxShehovtsov:
@vitalyf007 added score generator tests. They do pass with.net SDK's input.

PR #26, opened 2015-07-06
Initial changes for supporting flushing of events on before page unload

ameyagholkar:
- Hooks up with the onbeforeunload event and attaches a callback that will try to flush all pending events
- This callback will handle sending of these events asynchronously for now

### DIFF
--- a/JavaScript/JavaScriptSDK/Init.ts
+++ b/JavaScript/JavaScriptSDK/Init.ts
@@ -28,9 +28,7 @@ function initializeAppInsights() {
 
             // Add callback to push events when the user navigates away
             // Note: This will not work for browsers < Opera 12 && IE 10
-            if ('onbeforeunload' in window) {
-                console.log('Adding onbeforeunload');
-                
+            if ('onbeforeunload' in window) {                
                 // Callback to flush all events
                 var flushAllEvents = function() {
                     appInsightsLocal.trackEvent('AI onbeforeunload: Flushing all events');
--- a/JavaScript/JavaScriptSDK/Init.ts
+++ b/JavaScript/JavaScriptSDK/Init.ts
@@ -27,19 +27,23 @@ function initializeAppInsights() {
             init.pollInteralLogs(appInsightsLocal);
 
             // Add callback to push events when the user navigates away
-            // Note: This will not work for browsers < Opera 12 && IE 10
+            // Note: This approach tries to push an async request with all the pending events onbeforeunload.
+            //       Firefox does not respect this. Other browsers DO push out the call with < 100% hit rate.
+            //       Telemetry here will help us analyze how effective this approach is.
+            //       Another approach would be to make this call sync with a acceptable timeout to reduce the 
+            //       impact on user experience.
             if ('onbeforeunload' in window) {                
                 // Callback to flush all events
                 var flushAllEvents = function() {
-                    appInsightsLocal.trackEvent('AI onbeforeunload: Flushing all events');
-                    appInsightsLocal.context._sender.triggerSend(false /* async */);
+                    appInsightsLocal.trackEvent('AI (Internal): Flushing all events onbeforeunload');
+                    appInsightsLocal.context._sender.triggerSend();
                 };
                 
-                // check if the browser is from IE family, if yes use addEventListener
-                if (Microsoft.ApplicationInsights.Util.isBrowserIE(window.navigator.userAgent)) {
-                    window.addEventListener("beforeunload", flushAllEvents);
-                } else {
-                    window.onbeforeunload = flushAllEvents;
+                // check if addEventListener is available
+                if (window.addEventListener) {
+                    window.addEventListener("beforeunload", flushAllEvents, false);
+                } else if (window.attachEvent) {
+                    window.addEventListener("onbeforeunload", flushAllEvents);
                 }
             }
         }
--- a/JavaScript/JavaScriptSDK/Init.ts
+++ b/JavaScript/JavaScriptSDK/Init.ts
@@ -39,11 +39,8 @@ function initializeAppInsights() {
                     appInsightsLocal.context._sender.triggerSend();
                 };
                 
-                // check if addEventListener is available
-                if (window.addEventListener) {
-                    window.addEventListener("beforeunload", flushAllEvents, false);
-                } else if (window.attachEvent) {
-                    window.addEventListener("onbeforeunload", flushAllEvents);
+                if (!Microsoft.ApplicationInsights.Util.addEventHandler('beforeunload', flushAllEvents)) {
+                    Microsoft.ApplicationInsights._InternalLogging.throwInternalNonUserActionable(Microsoft.ApplicationInsights.LoggingSeverity.CRITICAL, 'Could not add handler for beforeunload');
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


ameyagholkar:
@MaxShehovtsov 
@olegsych 
@scsouthw

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -135,9 +135,18 @@ module Microsoft.ApplicationInsights {
...
+         * Immediately send buffered data
+         * @param async {boolean} - Indicates if the events should be sent asynchronously (Optional, Defaults to true)
          */
-        public triggerSend() {
+        public triggerSend(async?: boolean) {
+            // We are async by default
+            var isAsync = true;
+            
+            // Respect the parameter passed to the func
+            if (async !== undefined && async !== null) {
```
scsouthw: You could instead check `if (typeof async === "boolean")`

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -157,5 +157,19 @@
...
             return objectTypeDump + propertyValueDump;
         }
+
+        /**
+         * Returns a boolean indicating whether the browser is IE or not.
+         * @param {string} userAgent - The user agent string for the browser
+         */
+        public static isBrowserIE(userAgent: string): boolean {
+            if (!userAgent) {
+                throw new Error('No user agent provided');
```
scsouthw: Seems harsh to throw here. Probably better to just return false since IE would never hide the user agent but other browsers might.

scsouthw: As a corollary it's better practice to check for the features you need instead of looking for a specific browser. For example, if a future version IE supports the feature things will just work.

JavaScript/JavaScriptSDK/Init.ts
```diff
@@ -25,6 +25,25 @@ function initializeAppInsights() {
...
+                console.log('Adding onbeforeunload');
+                
+                // Callback to flush all events
+                var flushAllEvents = function() {
+                    appInsightsLocal.trackEvent('AI onbeforeunload: Flushing all events');
+                    appInsightsLocal.context._sender.triggerSend(false /* async */);
+                };
+                
+                // check if the browser is from IE family, if yes use addEventListener
+                if (Microsoft.ApplicationInsights.Util.isBrowserIE(window.navigator.userAgent)) {
```
SergKanz: Use ```if ($window.addEventListener)``` instead? Also consider using ```attachEvent``` for older browsers. Please minimize number of times you override customer's onbeforeunload callback. Maybe check on null? Or do not set at all if addEventListener function doesn't exists.

JavaScript/JavaScriptSDK/Sender.ts
ameyagholkar: Yup! Changed.

JavaScript/JavaScriptSDK/Init.ts
ameyagholkar: Nice catch. I shouldn't have done that. Will switch to using addEventListener/attachEvent instead

JavaScript/JavaScriptSDK/Init.ts
```diff
@@ -25,6 +25,23 @@ function initializeAppInsights() {
             init.emptyQueue();
 
             init.pollInteralLogs(appInsightsLocal);
+
+            // Add callback to push events when the user navigates away
+            // Note: This will not work for browsers < Opera 12 && IE 10
+            if ('onbeforeunload' in window) {                
+                // Callback to flush all events
+                var flushAllEvents = function() {
+                    appInsightsLocal.trackEvent('AI onbeforeunload: Flushing all events');
```
MaxShehovtsov: I'd do "AI (Internal)" - to keep "naming" convention for our traces consistent

JavaScript/JavaScriptSDK/Util.ts
ameyagholkar: Makes sense. I was looking at the problem other way round. Thanks for the feedback. I'll change this.

JavaScript/JavaScriptSDK/Init.ts
ameyagholkar: Done.

ameyagholkar:
Addressed changes.
@MaxShehovtsov @scsouthw @SergKanz

JavaScript/JavaScriptSDK/Init.ts
```diff
@@ -25,6 +25,27 @@ function initializeAppInsights() {
...
+            //       impact on user experience.
+            if ('onbeforeunload' in window) {                
+                // Callback to flush all events
+                var flushAllEvents = function() {
+                    appInsightsLocal.trackEvent('AI (Internal): Flushing all events onbeforeunload');
+                    appInsightsLocal.context._sender.triggerSend();
+                };
+                
+                // check if addEventListener is available
+                if (window.addEventListener) {
```
AlexBulankou: if (addEventListener)... else attachEvent is such a fundamental event triggering functionality, that I think we want to make it a utility function.

AlexBulankou: I mean something like: addEvent = function(eventName, callback)
{
 if (window.addEventListener){window.addEventListener(eventName, callback, false);}else
 if (window.attachEvent){window.attachEvent("on"+eventName, callback);}else
window["on"+eventName] = callback;
}

JavaScript/JavaScriptSDK/Init.ts
```diff
@@ -25,6 +25,27 @@ function initializeAppInsights() {
...
+                    window.addEventListener("beforeunload", flushAllEvents, false);
+                } else if (window.attachEvent) {
+                    window.addEventListener("onbeforeunload", flushAllEvents);
```
AlexBulankou: you meant window.attachEvent here

JavaScript/JavaScriptSDK/Init.ts
```diff
@@ -25,6 +25,27 @@ function initializeAppInsights() {
...
 
             init.pollInteralLogs(appInsightsLocal);
+
+            // Add callback to push events when the user navigates away
+            // Note: This approach tries to push an async request with all the pending events onbeforeunload.
+            //       Firefox does not respect this. Other browsers DO push out the call with < 100% hit rate.
+            //       Telemetry here will help us analyze how effective this approach is.
+            //       Another approach would be to make this call sync with a acceptable timeout to reduce the 
+            //       impact on user experience.
+            if ('onbeforeunload' in window) {
```
AlexBulankou: Unit test to verify that we subscribe for onbeforeunload?

JavaScript/JavaScriptSDK/Init.ts
ameyagholkar: Thanks.

JavaScript/JavaScriptSDK/Init.ts
ameyagholkar: Moved this logic to Initialization class and tested the new function.

ameyagholkar:
Addressed changes.

@MaxShehovtsov @AlexBulankou

MaxShehovtsov:
:sheep: :hamburger:

AlexBulankou:
:shipit:

PR #27, opened 2015-07-09
Add authId API + persist to cookie and server-side

Assaf-Neufeld:
Note: I left as is all fields generated by the BOND.


JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -29,12 +34,34 @@ module Microsoft.ApplicationInsights.Context {
...
         public storeRegion: string;
-        
+
+        static cookieSeparator: string = '|';
+
+        public setAuthId(id: string, accountId? : string) {
+            this.authId = id;
+            var authCookie = this.authId;
+
+            if (accountId) {
```
MaxShehovtsov: just FYI if(0) will not pass - is it what you want (impossible to set accountid to 0)?

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -50,10 +77,22 @@ module Microsoft.ApplicationInsights.Context {
...
-                Util.setCookie('ai_user', newCookie.join('|') + ';expires=' + date.toUTCString());
+                Util.setCookie('ai_user', newCookie.join(User.cookieSeparator) + ';expires=' + date.toUTCString());
             }
 
             this.accountId = accountId;
+
+            // get authId and accountId from cookie
+            var authCookie = Util.getCookie('ai_authUser');
+            if (authCookie) {
+                var kvps = authCookie.split(User.cookieSeparator);
```
MaxShehovtsov: what is 'kvps'?

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -29,12 +34,34 @@ module Microsoft.ApplicationInsights.Context {
...
+                authCookie = [this.authId, this.accountId].join(User.cookieSeparator);
+            }
+            
+            // Set the suth id cookie. No expiration date because this is a session cookie (expires when browser closed).
+            Util.setCookie('ai_authUser', authCookie);
+        }
+
+        public clearAuthId() {
+            this.authId = null;
+            this.accountId = null;
```
MaxShehovtsov: accountid is not only for auth id, can you double check that you're supposed to delete it when you clear auth id?

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -50,10 +77,22 @@ module Microsoft.ApplicationInsights.Context {
...
+
+            // get authId and accountId from cookie
+            var authCookie = Util.getCookie('ai_authUser');
+            if (authCookie) {
+                var kvps = authCookie.split(User.cookieSeparator);
+                if (kvps.length > 0) {
+                    this.authId = kvps[0];
+                }
+                if (kvps.length > 1) {
+                    this.accountId = kvps[1];
```
MaxShehovtsov: The constructor takes account id, but you don't honor it if there is one in cookies. This needs to be clarified

JavaScript/JavaScriptSDK.Tests/Contracts/Generated/RequestData.unittests.ts
```diff
@@ -12,6 +12,7 @@
 /// <reference path="../../../JavaScriptSDK/Contracts/Generated/PageViewData.ts" />
 /// <reference path="../../../JavaScriptSDK/Contracts/Generated/PageViewPerfData.ts" />
 /// <reference path="../../../JavaScriptSDK/Contracts/Generated/AjaxCallData.ts" />
+/// <reference path="../../../JavaScriptSDK/Contracts/Generated/AvailabilityData.ts" />
```
MaxShehovtsov: why do you need it? are you adding code using this file?

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -513,6 +513,74 @@ class AppInsightsTests extends TestClass {
         });
 
         this.testCase({
+            name: "AppInsightsTests: set authID context is applied",
+            test: () => {
+                // setup
+                var appInsights = new Microsoft.ApplicationInsights.AppInsights(this.getAppInsightsSnippet());
+
+                appInsights.setAuthId("10001");
```
MaxShehovtsov: Do you want to add another test for setAuthId testing for accountId?

MaxShehovtsov: Do you want to add another test case testing setting up account id?

Assaf-Neufeld: Added

JavaScript/JavaScriptSDK.Tests/Contracts/Generated/RequestData.unittests.ts
Assaf-Neufeld: This file was auto-generated by the BOND generator.
I'm not using this file, but I left all changes intact so everything will be insync with the same schema.

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: done

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -11,6 +11,11 @@ module Microsoft.ApplicationInsights.Context {
         public id: string;
 
         /**
+         * Authorized user id
+         */
+        public authId: string;
```
ameyagholkar: I think this should be `private` if we need cookie to be updated when `authId` is updated.

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -311,6 +311,14 @@ module Microsoft.ApplicationInsights {
             }
         }
 
+        public setAuthId(authId: string, accountId?: string) {
```
ameyagholkar: Please jsdoc `setAuthId` and `clearAuthId`

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -40,6 +40,11 @@
             return value;
         }
 
+        public static deleteCookie(name: string) {
```
ameyagholkar: JsDoc

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -29,12 +34,34 @@ module Microsoft.ApplicationInsights.Context {
          * The store region.
          */
         public storeRegion: string;
-        
+
+        static cookieSeparator: string = '|';
+
+        public setAuthId(id: string, accountId? : string) {
+            this.authId = id;
```
ameyagholkar: I would check whether the type of `accountId` is a `string` or not. Since this function is called from the public API and its JS, we could potentially see non-string values.

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -29,12 +34,34 @@ module Microsoft.ApplicationInsights.Context {
          * The store region.
          */
         public storeRegion: string;
-        
+
+        static cookieSeparator: string = '|';
+
+        public setAuthId(id: string, accountId? : string) {
```
ameyagholkar: JsDoc

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -29,12 +34,34 @@ module Microsoft.ApplicationInsights.Context {
...
+            this.authId = id;
+            var authCookie = this.authId;
+
+            if (accountId) {
+                this.accountId = accountId;
+                authCookie = [this.authId, this.accountId].join(User.cookieSeparator);
+            }
+            
+            // Set the suth id cookie. No expiration date because this is a session cookie (expires when browser closed).
+            Util.setCookie('ai_authUser', authCookie);
```
ameyagholkar: Consider moving the `ai_authUser` to a `static` field like the cookie separator since this string is used at multiple places.

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -50,10 +77,22 @@ module Microsoft.ApplicationInsights.Context {
...
+                Util.setCookie('ai_user', newCookie.join(User.cookieSeparator) + ';expires=' + date.toUTCString());
             }
 
             this.accountId = accountId;
+
+            // get authId and accountId from cookie
+            var authCookie = Util.getCookie('ai_authUser');
+            if (authCookie) {
+                var kvps = authCookie.split(User.cookieSeparator);
+                if (kvps.length > 0) {
```
ameyagholkar: I don't think you need to check for this condition. `split()` will return a new array of length 1 by default if it's not able to split the string.

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: This comment got me thinking about the types of input we may get from our users:
-	We can’t allow control characters – “;,= ” and our own “|” because it might mess with the cookie 
-	It’s unclear whether you can have non-ASCII characters in the cookie. This worked for me in chrome, FF and IE, but it’s not officially supported by the latest spec.

I decided to reject input with control characters, and encodeURI to avoid unicode characters.
(my initial thought was to encode in base64, But the out-of-the-box btoa() function isn’t supported in IE9 and earlier)

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: I left it public to be inline with all other properties. This is true for user id and session id as well, and they are open.

JavaScript/JavaScriptSDK/Contracts/Generated/ContextTagKeys.ts
```diff
@@ -35,20 +38,38 @@ module AI
         public userAgent: string;
         public userId: string;
         public userStoreRegion: string;
+        public userAuthUserId: string;
+        public userAnonymousUserAcquisitionDate: string;
+        public userAuthenticatedUserAcquisitionDate: string;
         public sampleRate: string;
         public internalSdkVersion: string;
         public internalAgentVersion: string;
+        public internalDataCollectorReceivedTime: string;
```
SergKanz: why do we need internal fields here? Let's remove them

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -50,10 +97,38 @@ module Microsoft.ApplicationInsights.Context {
...
+        }
+
+        private validateUserInput(id: string) {
+            // Validate:
+            // 1. Id is a non-empty string.
+            // 2. It does not contain special characters for cookies.
+            if (typeof id !== 'string' ||
+                !id ||
+                id.match(/,|;|=| |\|/)) {
+                throw new Error("User auth id and account id should be of type string. They should not contain commas, semi-colons, equal signs, spaces, or vertical-bars.");
```
SergKanz: one of SDK principles was not to throw errors, provide defaults instead, Same as sanitization works in SDK. Let's just replace bad symbols to the underscores or something like this

JavaScript/JavaScriptSDK.Tests/Contracts/Generated/RequestData.unittests.ts
MaxShehovtsov: Makes sense

JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -35,6 +35,9 @@
         createLazyMethod("track" + method.pop());
     }
 
+    createLazyMethod("setAuthUserContext");
```
SergKanz: Why do we need to change a snippet and we do not change the AppInsights object itself in the actual script and do not add those methods there? It will be more logical to add those methods into the actual AppInsights object so people with the old snippet can still use those method, when new customers may benefit of earlier loading of them.

JavaScript/JavaScriptSDK/Context/User.ts
ameyagholkar: +1
I would even say that the function name `validateUserInput` suggests that it returns a `boolean` which would indicate whether the string is valid or not. Based on this return value, you can decide whether you should set the id or not.

JavaScript/JavaScriptSDK/Context/User.ts
ameyagholkar: But since with the `authId` we are expecting the server side to respect whatever `string` the user might set to get a more accurate count, i think that this should not be exposed publicly.

JavaScript/JavaScriptSDK/Contracts/Generated/ContextTagKeys.ts
Assaf-Neufeld: I agree the internals have no purpose here - but this file is auto-generated. Consulting with Max we thought we should leave them as is - I dont think it's harmful. 

But if you think otherwise I will manually remove them..

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: I originally wrote this as replace, but the PM's didnt like the idea of manipulating the user input - the user might search for the exact string later in the UI and not find it.

If we're rejecting bad input, then we have to log it somehow - just ignoring it is not a good experience IMO.
Would you prefer this to be:
_InternalLogging.throwInternalUserActionable(LoggingSeverity.WARNING, "Setting auth user context failed..... ");
?

MaxShehovtsov: My recommendation - go either way and collect user's feedback. We can't guess what users will/won't like.

JavaScript/JavaScriptSDK/Contracts/Generated/ContextTagKeys.ts
MaxShehovtsov: I think if there IS a feedback it's on schema itself not on your implementation, Assaf. Shouldn't be a blocker for this pull request but please follow up with sergkanz@microsoft.com to address his concerns.

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: ok, I made it private with a public getter

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: Ok - made validate return bool, and logged warning + rejected the input according to result.

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -29,12 +33,60 @@ module Microsoft.ApplicationInsights.Context {
...
+            return this.authId;
+        }
+
+        /**
+         * Sets the autheticated user id and the account id in this session.
+         * @param id {string} - The autheticated user id
+         * @param accountId {string} - The account id.
+         * @returns {} 
+         */
+        public setAuthUserContext(id: string, accountId?: string) {
```
olegsych: The term `Auth` is ambiguous in this context. It could mean `AuthenticatedUser` or `AuthorizedUser`. Can we spell it out instead of using an abbreviation?

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -29,12 +33,60 @@ module Microsoft.ApplicationInsights.Context {
          * The store region.
          */
         public storeRegion: string;
-        
+
+        /**
+         * Authorized user id
+         */
+        private authId: string;
+        public getAuthId() {
```
olegsych: As a person reading this API for the first time, I find it confusing. What does `AuthId` mean? Is this an authentication token, an authorization token or something else? With JavaScript's support for doc comments and IntelliSense being very limited this method would also look confusing at the point of call and require users to refer to the API documentation. 

Can we spell this out as `AuthenticatedUserId`?

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -312,6 +312,32 @@ module Microsoft.ApplicationInsights {
         }
 
         /**
+         * Sets the autheticated user id and the account id in this session.
+         * User auth id and account id should be of type string. They should not contain commas, semi-colons, equal signs, spaces, or vertical-bars.
+         *   
+         * @param authId {string} - The authenticated user id. A unique and persistent string that represents each authenticated user in the service.
+         * @param accountId {string} - An optional string to represent the account associated with the authenticated user.
+         */
+        public setAuthUserContext(authId: string, accountId?: string) {
```
olegsych: Can we spell out `AuthenticatedUser` here instead of using the `Auth` abbreviation to make the API less ambiguous?

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -29,12 +33,60 @@ module Microsoft.ApplicationInsights.Context {
          * The store region.
          */
         public storeRegion: string;
-        
+
+        /**
+         * Authorized user id
+         */
+        private authId: string;
```
olegsych: We have a "naked" `id` property. Why are we introducing `authId` with a getter?

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -513,6 +513,142 @@ class AppInsightsTests extends TestClass {
         });
 
         this.testCase({
+            name: "AppInsightsTests: set authID context is applied",
```
olegsych: Can we take advantage of sinon and only verify that `AppInsights` methods call the respective `UserContext` methods, which are tested separately?

JavaScript/JavaScriptSDK/Context/User.ts
olegsych: How common do we expect the usage of `accountId` to be? Is it going to be common enough to justify making this method more difficult to understand? Would two separate methods, `setAuthenticatedId` and `setAccountId` be easier to understand?

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -29,12 +33,60 @@ module Microsoft.ApplicationInsights.Context {
...
+            // Set the cookie. No expiration date because this is a session cookie (expires when browser closed).
+            // Encoding the cookie to handle unexpected unicode characters.
+            Util.setCookie(User.authUserCookieName, encodeURI(authCookie));
+        }
+
+        /**
+         * Clears the authenticated user id and the account id from the user context.
+         * @returns {} 
+         */
+        public clearAuthUserContext() {
```
olegsych: Would customers ever need to clear the `authId` without clearing the `accountId` or vice versa?

JavaScript/JavaScriptSDK/appInsights.ts
Assaf-Neufeld: done

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: We want to encourage people to update this field only through the API - that way it will be persisted + get the server-side. I wouldnt like users to directly update the field so I made it private. 
Regarding Id - that field could probably be private as well, but at this point it will be a breaking change for existing users who are already updating it directly (same for accountId).

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: We debated this in the design, and in the end decided that having a single setting method is better.
- Having 2 sets and 2 clears complicates the API as users will wonder when and how to use each.
- It complicates things as it introduces potentially buggy scenarios where one is set or cleared before/without the other. This way auth context is always set or cleared in one stroke.
- The account id param is clearly documented as an optional parameter in the code. And we will of course show this in the quick-start guide.

All that being said, I cant say how common accountId is going to be..

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: I dont see a need for this currently. Generally speaking, we want to call set on log-in, and clear on log-off. Clearing just one of the values would be a "partial" log-off scenario.

We do support scenarios where in 1 session the same user switches between accounts, or switching users within the same account - it's possible to just call set again with the new context.

JavaScript/JavaScriptSDK/Context/User.ts
olegsych: I like the idea of hiding internal implementation details behind proper abstractions. In this particular case, I am concerned that 
- `private` visibility is strictly a TypeScript concept. Our API is predominantly consumed from JavaScript, where the `authId` property will appear as a regular public property. A common workaround for "hide" private members in JavaScript is a naming convention, such as starting internal member names with an underscore.
- We are introducing an inconsistency into the previously established API, making it more confusing than necessary.

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -312,6 +312,32 @@ module Microsoft.ApplicationInsights {
         }
 
         /**
+         * Sets the autheticated user id and the account id in this session.
+         * User auth id and account id should be of type string. They should not contain commas, semi-colons, equal signs, spaces, or vertical-bars.
+         *   
+         * @param authenticatedUserId {string} - The authenticated user id. A unique and persistent string that represents each authenticated user in the service.
+         * @param accountId {string} - An optional string to represent the account associated with the authenticated user.
+         */
+        public setAuthenticatedUserContext(authenticatedUserId: string, accountId?: string) {
```
olegsych: In a previous thread, @Assaf-Neufeld mentioned that one of the scenarios we want to support is when the user switches between accounts, which brings me back to my earlier question regarding splitting the set method. Having used languages with first-class property concept for a long time, I'd also find working with two methods per property, like `getAuthenticatedUserId` and `clearAuthenticatedUserId`, confusing. Since we are already discussing the merits of encapsulating `User.authId` property with `User.getAuthId` method, what do you think about the single-method convention for encapsulating properties? 

```TypeScript
public authenticatedUserId(value?: string) {
  // if typeof value === undefined: return current value
  // else: store new value in cookie
  //    special case: if value === "", remove current value from cookie
}

// Get example
var currentAuthenticatedUserId = appInsights.authenticatedUserId();

// Set example
appInsights.authenticatedUserId("JohnDoe@mycompany.com");

// Clear example
appInsights.authenticatedUserId("");
```

This is common property pattern in C++ and also in some JavaScript frameworks, like [knockout](http://knockoutjs.com/documentation/observables.html). In this case we would still have only two methods, `authenticatedUserId()` and `userAccountId`.

JavaScript/JavaScriptSDK/Context/User.ts
MaxShehovtsov: @Assaf-Neufeld  +1 to Oleg's comments - please don't add getAuthId(), this introduces API inconsistency. Instead let's wait until it's obvious that this getter needs to be added.

JavaScript/JavaScriptSDK/appInsights.ts
MaxShehovtsov: @olegsych to me personally that's more confusing than setXXX. In knockout () means observable property, I don't think there's anything similar in our api yet.
I like the idea in general though and maybe that's the way we need to go for all context-level properties. Do you think we need a separate issue/user story?

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -312,6 +312,32 @@ module Microsoft.ApplicationInsights {
...
+            try {
+                this.context.user.setAuthenticatedUserContext(authenticatedUserId, accountId);
+            } catch (e) {
+                _InternalLogging.throwInternalUserActionable(LoggingSeverity.WARNING, "Setting auth user context failed. " + JSON.stringify(e));
```
MaxShehovtsov: Switch to Util.dump() as json stringify doesn't stringify errors :smile_cat:

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -312,6 +312,32 @@ module Microsoft.ApplicationInsights {
...
+        }
+
+        /**
+         * Clears the authenticated user id and the account id from the user context.
+         */
+        public clearAuthenticatedUserContext() {
+            try {
+                this.context.user.clearAuthenticatedUserContext();
+            } catch (e) {
+                _InternalLogging.throwInternalUserActionable(LoggingSeverity.WARNING, "Clearing auth user context failed. " + JSON.stringify(e));
```
MaxShehovtsov: Switch to Util.dump()

JavaScript/JavaScriptSDK/appInsights.ts
olegsych: I tend to like this pattern because it is succinct. I was familiar with it before knockout came around, so using `observable()` is a property get and `observable(value)` is a property set didn't feel confusing. However, I can certainly see how this can be confusing for someone not familiar with this pattern in other JavaScript frameworks. Anyway, I don't have a strong preference; I just wanted us to consider it before settling on the get/set method pattern.

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: You're right, fixed it

JavaScript/JavaScriptSDK/Context/User.ts
Assaf-Neufeld: Fixed.

JavaScript/JavaScriptSDK/appInsights.ts
Assaf-Neufeld: Agree with Max here - I think the setXXX/clearXXX api is clearer, even if less concise.

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -513,6 +513,142 @@ class AppInsightsTests extends TestClass {
...
+                var trackStub = sinon.stub(appInsights.context._sender, "send");
+
+                // verify
+                var test = (action) => {
+                    action();
+                    this.clock.tick(1);
+                    var envelope = this.getFirstResult(action, trackStub);
+                    var contextKeys = new AI.ContextTagKeys();
+                    Assert.equal("myuser中国话", envelope.tags[contextKeys.userAuthUserId], "user.authenticatedId is correct, special characters removed");
+                    Assert.equal("اللغةالعربيةهيجميلةעבריתזהנחמד", envelope.tags[contextKeys.userAccountId], "user.accountIdis correct, special characters removed");
```
MaxShehovtsov: that's hard to maintain, could you please switch to English alphabet or use special characters explicitly (e.g. with '\x' syntax).

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/User.tests.ts
```diff
@@ -105,6 +109,295 @@ class UserContextTests extends TestClass {
...
+                cookieStub.restore();
+                loggingStub.restore();
+            }
+        });
+
+        this.testCase({
+            name: "setAuthenticatedUserContext: handles unicode characters correctly",
+            test: () => {
+                // setup
+                var authAndAccountId = ['שלום', 'להתראות'];
```
MaxShehovtsov: here and in other places - for unicode specific tests, could you use '\xxx' syntax?

PR #28, opened 2015-07-14
Scripterror

MaxShehovtsov:
1) Removing ai.0.js from script error stack.
2) Reporting document url as a custom properties to make script error more user friendly. This might go away once we implement operation name for javascript.


MaxShehovtsov:
need to fix unit test

PR #30, opened 2015-07-15
Measuring domProcessing from responseEnd

AlexBulankou:
Some modern browsers (specifically Chrome, but  sometimes IE 10 too) issues domLoading event before responseEnd event, so responseStart->responseEnd interval overlaps with domLoading->loadEventEnd interval causing subcomponent times to be larger than navigationStart->loadEventEnd total, causing user confusion when they look in the UI. To account for this situation I'm changing the measurement of domProcessing  interval to start with responseEnd event.


MaxShehovtsov:
:sheep: :it:

MaxShehovtsov:
Agreed to follow up on UI change - the "dom loading" is not a dom loading anymore, only part of it after response end.

AlexBulankou:
BTW looks like with the new version of Chrome (that was released today?) domLoading is back to where it was (after responseEnd)

PR #35, opened 2015-07-17
Fix bug with sending expired session ids

OsvaldoRosado:
These changes fix a bug where we incorrectly send the wrong session id on the first request after a session has expired due to the session cookie persisting for a very long time. 

Most of the additions are to fix a bug that pops up after expiring the session cookie where we need to keep the old session id to terminate the session.

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
@@ -166,7 +166,8 @@ class SessionContextTests extends TestClass {
                 // A more sophisticated approach of emulating document.cookie is needed to support multiple cookies
                 Microsoft.ApplicationInsights.Util["document"] = this.generateEmulatedCookieStore();
 
-                var user = new Microsoft.ApplicationInsights.Context.User(undefined);
+                // Initialize our user and session cookies, as well as local storage
+                new Microsoft.ApplicationInsights.Context.User(undefined);
                 var sessionManager = new Microsoft.ApplicationInsights.Context._SessionManager(null,() => { });
                 sessionManager.update();
                 sessionManager.backup();
@@ -174,6 +175,7 @@ class SessionContextTests extends TestClass {
 
                 // Lose the session cookie but not the user cookie
                 Microsoft.ApplicationInsights.Util["document"].cookie = "ai_session=; expires=Thu, 01 Jan 1970 00:00:01 GMT;";
+                new Microsoft.ApplicationInsights.Context.User(undefined);
                 var sessionManager = new Microsoft.ApplicationInsights.Context._SessionManager(null,() => { });
                 sessionManager.update();
                 sessionManager.backup();
@@ -192,6 +194,7 @@ class SessionContextTests extends TestClass {
                 // Local storage and non-local storage should act the same
                 Microsoft.ApplicationInsights.Util["document"].cookie = "ai_session=; expires=Thu, 01 Jan 1970 00:00:01 GMT;";
                 Microsoft.ApplicationInsights.Util["document"].cookie = "ai_user=; expires=Thu, 01 Jan 1970 00:00:01 GMT;";
+                new Microsoft.ApplicationInsights.Context.User(undefined);
                 var sessionManager = new Microsoft.ApplicationInsights.Context._SessionManager(null,() => { });
                 sessionManager.update();
                 sessionManager.backup();
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
@@ -1,4 +1,5 @@
 ﻿/// <reference path="../../../JavaScriptSDK/context/session.ts" />
+/// <reference path="../../../JavaScriptSDK/context/user.ts" />
 /// <reference path="../../testframework/common.ts" />
 /// <reference path="../Util.tests.ts"/>
 
@@ -10,19 +11,13 @@ class SessionContextTests extends TestClass {
     /** Method called before the start of each test method */
     public testInitialize() {
         this.results = [];
-
-        if (window.localStorage) {
-            window.localStorage.clear();
-        }
+        this.resetStorage();
     }
 
     /** Method called after each test method has completed */
     public testCleanup() {
         this.results = [];
-        
-        if (window.localStorage) {
-            window.localStorage.clear();
-        }
+        this.resetStorage();
     }
 
     public registerTests() {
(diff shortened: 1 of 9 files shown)


JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
```diff
@@ -10,11 +10,19 @@ class SessionContextTests extends TestClass {
     /** Method called before the start of each test method */
     public testInitialize() {
         this.results = [];
+
+        if (window.localStorage) {
```
ameyagholkar: (Aesthetics)
Move these three lines to a separate func so that you can reuse below?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
```diff
@@ -376,5 +466,67 @@ class SessionContextTests extends TestClass {
     private restoreFakeCookie() {
         Microsoft.ApplicationInsights.Util["document"] = this.originalDocument;
     }
+
+    // Make a fake document.cookie property that correctly handles its unique getter/setter functionality.
+    // Document.cookie lets you read the entire set of cookies, but on set acts essentially as +=, with
+    // some logic to overwrite existing cookies with the same name.
+    // Document.cookie also silently removes expired cookies.
+    private generateEmulatedCookieStore() {
```
ameyagholkar: Nice stuff. :+1:

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
```diff
@@ -120,6 +128,88 @@ class SessionContextTests extends TestClass {
...
+                var sessionId = sessionManager.automaticSession.id;
+
+                // Lose the session cookie but not the user cookie
+                Microsoft.ApplicationInsights.Util["document"].cookie = "ai_session=; expires=Thu, 01 Jan 1970 00:00:01 GMT;";
+                new Microsoft.ApplicationInsights.Context.User(undefined);
+                var sessionManager = new Microsoft.ApplicationInsights.Context._SessionManager(null,() => { });
+                sessionManager.update();
+                sessionManager.backup();
+
+                if (window.localStorage) {
```
ameyagholkar: Can you eliminate this `if` condition by emulating a test case where localStorage would not exist?

JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -87,32 +87,22 @@ module Microsoft.ApplicationInsights.Context {
             }
         }
 
+        public backup() {
```
ameyagholkar: (Aesthetics) JsDoc? Below as well.

JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -51,6 +51,12 @@ module Microsoft.ApplicationInsights.Context {
                 date.setTime(date.getTime() + 31536000000);
                 var newCookie = [this.id, acqStr];
                 Util.setCookie('ai_user', newCookie.join('|') + ';expires=' + date.toUTCString());
+
+                // If we have an ai_session in local storage this means the user actively removed our cookies.
+                // We should respect their wishes and clear ourselves from local storage
+                if (window.localStorage && localStorage['ai_session']) {
+                    localStorage.removeItem('ai_session');
```
AlexBulankou: I think we want to abstract localStorage usage in separate function. Also as we discussed I think we should add try/catch around accessing localStorage to catch unexpected IO errors and let execution flow proceed

JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -122,6 +112,32 @@ module Microsoft.ApplicationInsights.Context {
             }
         }
 
+        private initializeAutomaticSessionWithData(params) {
```
ameyagholkar: JsDoc &
Strongly type params to `[]: string`?

JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -122,6 +112,32 @@ module Microsoft.ApplicationInsights.Context {
...
+            if (params.length > 0) {
```
ameyagholkar: I think you don't need this. `split()` returns array of [1] by default

JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -122,6 +112,32 @@ module Microsoft.ApplicationInsights.Context {
...
+                    this.automaticSession.acquisitionDate = this.automaticSession.acquisitionDate > 0 ? this.automaticSession.acquisitionDate : 0;
+                }
+
+                if (params.length > 2) {
+                    var renewal = +params[2];
+                    this.automaticSession.renewalDate = +new Date(renewal);
+                    this.automaticSession.renewalDate = this.automaticSession.renewalDate > 0 ? this.automaticSession.renewalDate : 0;
+                }
+            } catch (e) {
+                _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.WARNING, "Error parsing ai_session cookie, session will be reset: " + Util.dump(e));
```
ameyagholkar: @MaxShehovtsov Should this be CRITICAL?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
```diff
@@ -120,6 +128,88 @@ class SessionContextTests extends TestClass {
...
+                    Assert.equal(3, localStorage["ai_session"].split('|').length, "Cookie value before expiration should include user id, acq date and renew date");
+                    Assert.equal("newGuid", localStorage["ai_session"].split('|')[0], "First part of cookie value should be new user id guid");
+
+                    // cleanup
+                    getCookieStub.restore();
+                    setCookieStub.restore();
+                    newGuidStub.restore();
+                } else {
+                    // this might happen on IE when using a file:// url
+                    Assert.ok(true, "browser does not support local storage in current environment");
```
olegsych: Do we run these tests on browsers that don't support localStorage?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
OsvaldoRosado: This if is actually so that browsers that do not support local storage, like IE when using file:// urls, do not cause improper test failures

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
```diff
@@ -120,6 +128,88 @@ class SessionContextTests extends TestClass {
...
+                }
+            }
+        });
+
+        this.testCase({
+            name: "SessionContext: session manager can recover old session id and isFirst state from lost cookies when localStorage is available",
+            test: () => {
+                // A more sophisticated approach of emulating document.cookie is needed to support multiple cookies
+                Microsoft.ApplicationInsights.Util["document"] = this.generateEmulatedCookieStore();
```
olegsych: For a person looking at this test for the first time, it feels [obscure](http://xunitpatterns.com/Obscure%20Test.html) because it is long, complex. It also appears to contain [conditional test logic](http://xunitpatterns.com/Conditional%20Test%20Logic.html) and multiple test conditions. I'm wondering if it could be split in two or more smaller tests.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
ameyagholkar: True. If you can temporarily stub localStorage with undefined, maybe you can split this year into two tests - one which verifies your code works for browsers with local storage and one for which verifies expected behaviour in cases where local storage does not exist.

ameyagholkar: Actually, looking at Alex's comment on encapsulating localStorage behind a set of funcs would be useful here. You can have a func which checks whether local storage exists for the browser - which you can stub to return false/true for these cases.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
```diff
@@ -376,5 +466,67 @@ class SessionContextTests extends TestClass {
...
+
+    // Make a fake document.cookie property that correctly handles its unique getter/setter functionality.
+    // Document.cookie lets you read the entire set of cookies, but on set acts essentially as +=, with
+    // some logic to overwrite existing cookies with the same name.
+    // Document.cookie also silently removes expired cookies.
+    private generateEmulatedCookieStore() {
+        var cookieStore = {
+            _cookies: <string[]>[]
+        };
+        Object.defineProperty(cookieStore, "cookie", {
```
olegsych: This looks like well written code. Do you feel its appropriate to create an emulator for use in a single test? Emulators, such as this, are complex enough to prevent the tests from being [self-checking](http://xunitpatterns.com/Goals%20of%20Test%20Automation.html#Self-Checking Test) and often require separate testing of the emulation logic itself. Have you considered using plain stubs and spies instead?

JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -87,32 +87,22 @@ module Microsoft.ApplicationInsights.Context {
...
-
-                if (this.automaticSession.renewalDate == 0) {
-                    _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.WARNING, "AI sessoin renewal date is 0, session will be reset.");
+                this.initializeAutomaticSessionWithData(cookie.split("|"));
+            } else {
+                // There's no cookie, but we might have session data in local storage
+                // This can happen if the session expired or the user actively deleted the cookie
+                // We only want to recover data if the cookie is missing from expiry. We should respect the user's wishes if the cookie was deleted actively.
+                // The User class handles this for us and deletes our local storage object if the persistent user cookie was removed.
+                var userCookie = Util.getCookie('ai_user');
```
ameyagholkar: I don't see where `userCookie` is used/checked?

JavaScript/JavaScriptSDK/Context/Session.ts
OsvaldoRosado: Perhaps. This was the severity it was originally(I refactored this code into its own function, but the content was unchanged)

JavaScript/JavaScriptSDK/Context/Session.ts
ameyagholkar: Should line 117 be ideally inside the try catch? We are setting the session id to `params[0]` and then we could possibly be failing to parse the cookie in the `try - catch` below. Is that expected behavior?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
OsvaldoRosado: Yes. IE wont support it if running the tests from the powershell script in the root directory, as it uses file:// urls

JavaScript/JavaScriptSDK/Context/Session.ts
ameyagholkar: Considering that we now have logic to pull cookie out of local storage, we have just increased our points of failure. I would think we would want to track these failures. My suggestion would be to change this to CRITICAL. 
@MaxShehovtsov @olegsych thoughts?

JavaScript/JavaScriptSDK/Context/User.ts
ameyagholkar: :+1:

JavaScript/JavaScriptSDK/Context/Session.ts
olegsych: How is the `userCookie` used?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
OsvaldoRosado: It's possible to stub the functions that actually write to the cookie, though it seemed like it would leave the greatest amount of the system unchanged, and by my thinking more fully tested, by emulating the browser api and not impacting any of the internal functions directly. The current way we test cookies in the other tests will generate confusing discrepancies in all but the most simple cases as we replace document.cookie with a simple string. I think we should consider actually replacing the document.cookie replacements we do in other places with this to avoid other cases where our tests will behave differently than real browsers.

JavaScript/JavaScriptSDK/Context/Session.ts
OsvaldoRosado: This is a remnant from when I did checking for that cookie here. I'll remove that

JavaScript/JavaScriptSDK/Context/Session.ts
OsvaldoRosado: I'm not sure, I refactored this code into this function but I didn't actually edit it. This was the structure that existed before.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
olegsych: It's hard for me to tell whether emulation approach is better in general for the JavaScript SDK. There are other ways to isolate tests from a dependencies like the `cookie` property as well, like creating a [port](http://blogs.msdn.com/b/ericgu/archive/2014/12/01/unit-test-success-using-ports-adapters-and-simulators.aspx). Subjectively, emulation feels like an overkill in the context of this set of changes.

JavaScript/JavaScriptSDK/Context/Session.ts
olegsych: Is that the difference between `WARNING` and `CRITICAL`? 
I don't have a strong preference. We can change it back to `WARNING` later, if we decide that `CRITICAL` notifications are too noisy.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
olegsych: Are you referring to the [RunTestsInBrowser.ps1](https://github.com/Microsoft/ApplicationInsights-JS/blob/master/RunTestsInBrowser.ps1)? I think it starts IISExpress and opens a browser to a `http://localhost` port. 

The reason I'm asking is that [conditional test logic](http://xunitpatterns.com/Conditional%20Test%20Logic.html) is a pretty bad [code smell](http://xunitpatterns.com/code%20smell.html). Unless we _really_ need it, I would like us to eliminate this if/else statement.

OsvaldoRosado: Sorry, its running the tests via chutzpah, not the powershell script. It opens the html of the test directly in the browser without a server in between.

OsvaldoRosado:
These new commits address your comments @MaxShehovtsov @ameyagholkar @olegsych @AlexBulankou

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -1,9 +1,59 @@
...
 
     export class Util {
         private static document: any = typeof document !== "undefined" ? document : {};
-
         public static NotSpecified = "not_specified";
+
+        /*
+         * helper methods to access local storage
+         */
+        public static _getStorageObject():Storage {
```
olegsych: Does this need to be public?

OsvaldoRosado: This is accessed on line 172 in the renew function in Session.ts. It's used to directly test if storage is available so we can fire an internal log if it's not. That way we can check in kusto what browsers+versions and how many users are unable to use local storage.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Util.tests.ts
```diff
@@ -5,6 +5,84 @@ class UtilTests extends TestClass {
 
     public registerTests() {
         var Util = Microsoft.ApplicationInsights.Util;
+
```
olegsych: I like these new tests. They are small, easy to understand and test a single concern. :+1:

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
olegsych: OK, It sounds like we can't avoid the conditional logic here. :cry:

JavaScript/JavaScriptSDK/Util.ts
olegsych: Ok, just checking :smile:

olegsych:
LGTM

[LGTM](http://lgtm.in/i/oMkinunEn)

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -1,9 +1,59 @@
...
+            if (window.localStorage) {
+                return window.localStorage;
+            } else {
+                return null;
+            }
+        }
+
+        public static getStorage(name:string):string {
```
ameyagholkar: JsDoc :smile:

ameyagholkar:
:sheep: :+1:

PR #40, opened 2015-07-22
Try catch patches

ameyagholkar:
- try catches for localStorage
- try catches for the two entry points of appInsights creation

### DIFF
--- a/JavaScript/JavaScriptSDK/appInsights.ts
+++ b/JavaScript/JavaScriptSDK/appInsights.ts
@@ -41,7 +41,7 @@ module Microsoft.ApplicationInsights {
 
         constructor(config: IConfig) {
             this.config = config || <IConfig>{};
-    
+
             // load default values if specified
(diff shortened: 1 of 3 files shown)


JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -40,54 +40,58 @@ module Microsoft.ApplicationInsights {
...
-            this.config = config || <IConfig>{};
-
-            // load default values if specified
-            var defaults: IConfig = AppInsights.defaultConfig;
-            if (defaults !== undefined) {
-                for (var field in defaults) {
-                    // for each unspecified field, set the default value
-                    if (this.config[field] === undefined) {
-                        this.config[field] = defaults[field];
+            try {
```
MaxShehovtsov: We don't want it in the constructor - let it fail, it'll be unexpected to see object constructed with undefined state.

ameyagholkar:
Closing this pull request. Will send out a new one.

PR #44, opened 2015-07-31
README.md - add information about AINow

AlexBulankou:
Adding link to ainow to help customers try Application Insights with JS SDK without having to create an Azure account


MaxShehovtsov:
@AlexBulankou Thanks Alex!

PR #47, opened 2015-08-03
Features/page visit time

giladsu:
Initial page visit time feature implementation.
First time "typescripting" so want to get initial feedback before adding tests.

Spec for feature can be found here: https://microsoft-my.sharepoint.com/personal/yossiy_microsoft_com/Documents/Functional Spec - Page visit time.docx?d=w088104dc384f4ac8b944948f5dc3d1c3 

Changes include:

1. Added trackPageVisitTime() method to send metric of time spent viewing a page.  Currently sending a CustomMetric, but will send a dedicated metric in next phases.

2. Added ability to auto-track page visit time. If auto-track is enabled (for phase 1 default is disabled):
    a. When trackPageView() is called, the SDK also calls trackPageVisitTime() for the previous page visited (if exists).

     b. Added PageVisitTimeManager class which is in charge of storing pages visited using session storage in order to achieve 2a.

      c. PageVisitTimeManager exposes 3 methods: restart, stop, and start.  Stop returns previous page viewed (if exists in session storage) and clears storage.  Start adds page passed in to session storage (along with current time for start time).  Restart calls stop, stores return value, calls start, and returns the value stop returned.


JavaScript/JavaScriptSDK/Context/PageVisitTimeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+    export class PageVisitTimeManager {
+
+        private prevPageVisitDataKeyName = "prevPageVisitData";
+
+        /**
+         * Stops timing of current page (if exists) and starts timing for duration of visit to pageName
+         * @param pageName Name of page to begin timing visit duration
+         * @returns {PageVisitData} Page visit data (including duration) of pageName from last call to start or restart, if exists. Null if not. 
+         */
+        restartPageVisitTimer(pageName: string, pageUrl: string) {
```
MaxShehovtsov: private/public and return type to all methods please :)

JavaScript/JavaScriptSDK/Context/PageVisitTimeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+         */
+        startPageVisitTimer(pageName: string, pageUrl: string) {
+            if (this.checkSessionStorageSupported()) {
+                if (window.sessionStorage.getItem(this.prevPageVisitDataKeyName) != null) {
+                    throw new Error("Cannot call startPageVisit consecutively without first calling stopPageVisit");
+                }
+
+                var currPageVisitData = new PageVisitData(pageName, pageUrl);
+                var currPageVisitDataStr = JSON.stringify(currPageVisitData);
+                window.sessionStorage.setItem(this.prevPageVisitDataKeyName, currPageVisitDataStr);
```
MaxShehovtsov: Instead of direct call to window.sessionStorage please use Util class. Look at https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/Util.ts#L30 and https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/Util.ts#L59

JavaScript/JavaScriptSDK/Context/PageVisitTimeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+                    return prevPageVisitData;
+                } else {
+                    return null;
+                }
+
+            }
+            return null;
+        }
+
+        private checkSessionStorageSupported() {
```
MaxShehovtsov: Use Util class instead of this method

JavaScript/JavaScriptSDK/Context/PageVisitTimeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        /**
+         * Starts timing visit duration of pageName
+         * @param pageName 
+         * @returns {} 
+         */
+        startPageVisitTimer(pageName: string, pageUrl: string) {
+            if (this.checkSessionStorageSupported()) {
+                if (window.sessionStorage.getItem(this.prevPageVisitDataKeyName) != null) {
+                    throw new Error("Cannot call startPageVisit consecutively without first calling stopPageVisit");
```
MaxShehovtsov: What happens when you throw from here? Sorry for not having concrete suggestions - please see the similar code like PageViewPerformance.ts when we handle absense of some API. 
Please make sure that
1) this code doesn't affect other functionality
2) we get meaningful telemetry about not sending session events
3) if customer sees errors then its actionable, otherwise send it as not actionable (see Logging.throwXXX() methods)

JavaScript/JavaScriptSDK/Initialization.ts
```diff
@@ -154,6 +154,9 @@ module Microsoft.ApplicationInsights {
             config.verboseLogging = Util.stringToBoolOrDefault(config.verboseLogging);
             config.emitLineDelimitedJson = Util.stringToBoolOrDefault(config.emitLineDelimitedJson);
             config.diagnosticLogInterval = config.diagnosticLogInterval || 10000;
+            config.autoTrackPageVisitTime = (config.autoTrackPageVisitTime !== undefined && config.autoTrackPageVisitTime !== null) ?
```
MaxShehovtsov: I think "(config.autoTrackPageVisitTime !== undefined && config.autoTrackPageVisitTime !== null)" will be hadled by Util.stringToBoolOrDefault() so you don't need to do this check yourself

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -89,6 +94,8 @@ module Microsoft.ApplicationInsights {
 
                 this.context.track(envelope);
             }
+
+            this._pageVisitTimeManager = new Microsoft.ApplicationInsights.Context.PageVisitTimeManager();
```
MaxShehovtsov: I think it should be in TelemetryContext not in AppInsights class.

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -33,6 +37,7 @@ module Microsoft.ApplicationInsights {
 
         private _eventTracking: Timing;
         private _pageTracking: Timing;
+        private _pageVisitTimeManager: Microsoft.ApplicationInsights.Context.PageVisitTimeManager;
```
MaxShehovtsov: I think it should live in TelemetryContext not in AppInsights class.

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -301,7 +318,31 @@ module Microsoft.ApplicationInsights {
...
+        * Log a page visit time
+        * @param    pageName    Name of page
+        * @param    pageVisitDuration Duration of visit to the page in milleseconds
+        */
+        public trackPageVisitTime(pageName: string, pageUrl: string, pageVisitTime: number) {
+            var properties = { PageName: pageName, PageUrl: pageUrl };
+            this.trackMetric("PageVisitTime", pageVisitTime, 1, pageVisitTime, pageVisitTime, properties);
+        }
+
+        private autoTrackPageVisitTime(pageName: string, pageUrl: string) {
```
MaxShehovtsov: That feels to belong more to PageVisitTimeManager class than to AppInsights

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -301,7 +318,31 @@ module Microsoft.ApplicationInsights {
...
+                _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.WARNING, "trackTrace failed, trace will not be collected: " + Util.dump(e));
+            }
+        }
+
+        /**
+        * Log a page visit time
+        * @param    pageName    Name of page
+        * @param    pageVisitDuration Duration of visit to the page in milleseconds
+        */
+        public trackPageVisitTime(pageName: string, pageUrl: string, pageVisitTime: number) {
```
MaxShehovtsov: That needs to go to PageVitisTimeManager, do you agree?

JavaScript/JavaScriptSDK/Context/PageVisitTimeManager.ts
```diff
@@ -0,0 +1,96 @@
+﻿
```
MaxShehovtsov: I don't think this should live in the "Context" folder as it is not a context, it's a separate peice of logic specific to PageViews. Context on the other hand means a data attached to any kind of telemetry (page views, metrics, events, exception, ...)

MaxShehovtsov:
Thank you Gilad! Looking forward to seeing tests in place and the whole feature checked in and going live!

olegsych:
Thanks for contributing, @giladsu! Could you add details about the proposed feature and API usage scenarios to the description of this pull request?

JavaScript/JavaScriptSDK/appInsights.ts
giladsu: trackPageVisitTime()? I wanted it to be part of AppInsights, so that if the user decides not to auto-track, they can call this method themselves whenever they see fit. What do you think?

JavaScript/JavaScriptSDK/appInsights.ts
giladsu: See my previous comment and let me know what you think....

JavaScript/JavaScriptSDK/appInsights.ts
MaxShehovtsov: I wouldn't do it until we get a user ask/need.

JavaScript/JavaScriptSDK/Context/PageVisitTimeManager.ts
giladsu: Good point. Is there a folder you think that is a good match? I couldnt find one that felt like a good fit - I can also add a new one. Let me know what you prefer...

MaxShehovtsov: Don't know - feel free to use your judgement :)

olegsych: Do we expect this type to be called directly by consumers of the AI API?

MaxShehovtsov: I would not expect that.

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -276,9 +293,9 @@ module Microsoft.ApplicationInsights {
          * @param   min The smallest measurement in the sample. Defaults to the average.
          * @param   max The largest measurement in the sample. Defaults to the average.
          */
-        public trackMetric(name: string, average: number, sampleCount?: number, min?: number, max?: number) {
+        public trackMetric(name: string, average: number, sampleCount?: number, min?: number, max?: number, properties?:Object) {
```
olegsych: Do we actually support properties for metric telemetry items?

JavaScript/JavaScriptSDK/appInsights.ts
olegsych: At this time `trackPageVisitTime` appears to be a convenience shortcut for calling `trackMetric` with a predefined set of known and custom properties. Do we expect it to be used by _most_ AI customers?

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -126,6 +133,11 @@ module Microsoft.ApplicationInsights {
                 }
 
                 this._pageTracking.stop(name, url, properties, measurements);
+
+                if (this.config.autoTrackPageVisitTime) {
+                    this.autoTrackPageVisitTime(name, url);
```
olegsych: Why are we using a _custom metric_ to track page view duration in addition to the _page view_ telemetry already tracked with duration here?

JavaScript/JavaScriptSDK/appInsights.ts
olegsych: I agree that moving this logic from `AppInsights` to `PageVisitTimeManager` is a good idea. In `AppInsights` this logic requires exposing the low-level `restartPageVisitTimer` method which is hard to understand here. Implemented in the `PageVisitTimeManager` class, this logic could be consumed here as a simple `trackPageVisit(pageName, url):void` method.

JavaScript/JavaScriptSDK/Context/PageVisitTimeManager.ts
giladsu: Done :)

JavaScript/JavaScriptSDK/Context/PageVisitTimeManager.ts
giladsu: Good question. Right now (phase 1) this exception will never be thrown since restart is the only public method, and it always calls stop before start.

In phase 2 when we add the Page Visibility API and the onunload use, this will be relevant and the calling code will need to handle this correctly.

JavaScript/JavaScriptSDK/appInsights.ts
giladsu: OK - I misunderstood which code we were talking about.  Yes - I agree, will move that logic there.

I am also changing trackPageVisitTime() to private until we go "live" with this feature and are using the new standard metric.

JavaScript/JavaScriptSDK/appInsights.ts
giladsu: Yes.  .NET SDK already has this, I checked and this is working from the UI....

JavaScript/JavaScriptSDK/appInsights.ts
giladsu: This is just for "phase 1" to unblock the development of this feature, until the page visit time standard metric is supported by the backed...

JavaScript/JavaScriptSDK/Context/PageVisitTimeManager.ts
giladsu: Yes - for sure not, this should only be for internal use.

JavaScript/JavaScriptSDK/appInsights.ts
giladsu: So these lines do this:
1. Track PageView for the current page.
2. Track Page Visit Time for the previous page (since we are on a new page, we know that the visit time of the previous page is over).

We are working on moving the custom metric to a standard page visit time metric, but there is still work to be done in the backend.

JavaScript/JavaScriptSDK/appInsights.ts
giladsu: I made this change by adding a method: restartTrackPageVisitTime to the PageTimeVisitManager which sends telemetry for the previous page and starts tracking the new page visit.

Let me know if you have ideas of a better method name, but this seems to be consistent with the other start/stop track method names...

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -89,6 +89,89 @@ module Microsoft.ApplicationInsights {
...
+
+        /**
+         *  Check if the browser supports local storage.
+         *
+         *  @returns {boolean} True if local storage is supported.
+         */
+        public static canUseSessionStorage(): boolean {
+            return !!Util._getSessionStorageObject();
+        }
+
```
giladsu: As we discussed - I have mixed feelings about the code duplication here. On one hand it isnt code that is expected to change and this is safest in terms of preventing regressions.  On the other, it would probably be cleaner to add some helper methods and have the same general code path for both local and session storage.

I'm OK with this - but let me know if you have other thoughts now that you are looking at it.

JavaScript/JavaScriptSDK.Tests/CheckinTests/PageVisitTimeManager.tests.ts
```diff
@@ -0,0 +1,201 @@
...
+            name: "PageVisitTimeManager: consecutive calls to restart returns expected information",
+            test: () => {
+
+                //setup
+                // Mock storage so this will work in all browsers for tests
+                var storage = this.getMockStorage();
+                var getStorageObjectStub = sinon.stub(Microsoft.ApplicationInsights.Util, "_getSessionStorageObject",() => storage);
+
+                var object = { method: function () { } };
+                var spy = sinon.spy(object, "method");
```
MaxShehovtsov: use method instead of spy

JavaScript/JavaScriptSDK.Tests/CheckinTests/PageVisitTimeManager.tests.ts
```diff
@@ -0,0 +1,201 @@
...
+                var pageVisitTimeManager = new Microsoft.ApplicationInsights.Telemetry.PageVisitTimeManager(spy);
+
+                //act
+                var nullPageData = pageVisitTimeManager.restartPageVisitTimer(testValues.page1Name, testValues.page1Url);
```
MaxShehovtsov: separate case (that it returns null)

JavaScript/JavaScriptSDK.Tests/CheckinTests/PageVisitTimeManager.tests.ts
```diff
@@ -0,0 +1,201 @@
...
+        });
+
+        this.testCase({
+            name: "PageVisitTimeManager: startPageVisitTime fails if called twice without a call to stop",
+            test: () => {
+
+                //setup
+                // Mock storage so this will work in all browsers for tests
+                var storage = this.getMockStorage();
+                var getStorageObjectStub = sinon.stub(Microsoft.ApplicationInsights.Util, "_getSessionStorageObject",() => storage);
```
MaxShehovtsov: logically belongs to testInitialize()

JavaScript/JavaScriptSDK/Telemetry/PageVisitTimeManager.ts
```diff
@@ -0,0 +1,118 @@
...
+            return prevPageVisitData;
+        }
+
+        /**
+         * Starts timing visit duration of pageName
+         * @param pageName 
+         * @returns {} 
+         */
+        public startPageVisitTimer(pageName: string, pageUrl: string) {
+            if (Util.canUseSessionStorage()) {
```
MaxShehovtsov: add try/catch to all publics

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -89,6 +89,89 @@ module Microsoft.ApplicationInsights {
...
+         *  @param {string} name - the name of the object to get from storage
+         *  @returns {string} The contents of the storage object with the given name. Null if storage is not supported.
+         */
+        public static getSessionStorage(name: string): string {
+            var storage = Util._getSessionStorageObject();
+            if (storage !== null) {
+                try {
+                    return storage.getItem(name);
+                } catch (e) {
+                    _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.CRITICAL, "Browser failed read of local storage.");
```
MaxShehovtsov: can you dump the exception to internal logging as well

JavaScript/JavaScriptSDK.Tests/CheckinTests/PageVisitTimeManager.tests.ts
```diff
@@ -0,0 +1,185 @@
...
+            page3ViewTime: 30
+        };
+
+        this.testCase({
+            name: "PageVisitTimeManager: When trackPreviousPageVisit is called once, the tracking delegate is not called since there are no previous pages",
+            test: () => {
+
+                //setup
+                var object = { method: function () { } };
+                var spy = sinon.spy(object, "method");
```
MaxShehovtsov: any way to get rid of this spy?

MaxShehovtsov: a better naming for "method" might help to understand its purpose in the test (like "trackPageVisitCallbackStub")

MaxShehovtsov:
:shipit:

PR #49, opened 2015-08-05
Send PageViews together with client perf And refactored unit tests

MaxShehovtsov:
PageView used to be sent synchronously when user does trackPageView(). Now we're changing it to send together with client perf (when page is loaded), i.e. asynchronously. This required changing/refactoring lots of tests.

### DIFF
--- a/JavaScript/JavaScriptSDK/appInsights.ts
+++ b/JavaScript/JavaScriptSDK/appInsights.ts
@@ -82,13 +82,17 @@ module Microsoft.ApplicationInsights {
 
             // initialize page view timing
             this._pageTracking = new Timing("trackPageView");
-            this._pageTracking.action = (name?: string, url?: string, duration?: number, properties?: Object, measurements?: Object) => {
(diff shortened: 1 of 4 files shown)


JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -149,51 +153,51 @@ module Microsoft.ApplicationInsights {
...
+        }
+
+        private trackPageViewInternal(name?: string, url?: string, properties?: Object, measurements?: Object) {
+            var durationMs = 0;
+            // check if timing data is available
+            if (Telemetry.PageViewPerformance.isPerformanceTimingSupported()) {
+                // compute current duration (navigation start to now) for the pageViewTelemetry
+                var startTime = window.performance.timing.navigationStart;
+                durationMs = Telemetry.PageViewPerformance.getDuration(startTime, +new Date);
+
```
AlexBulankou: I think for accurate measurements we should be collecting interval from navigatioStart to pageLoad event start.

PR #50, opened 2015-08-06
synchronizing page view duration with client perf total

MaxShehovtsov:
Page view duration now should be equal to client perf.total


AlexBulankou:
LGTM :shipit:

PR #51, opened 2015-08-08
Telemetry initializer

MaxShehovtsov:
Implementation of Telemtry initializer concept - allows to attach properties and measurements to all telemetry items.

Suggested use:
1) with snippet config

       <SNIPPET>
        }({
            instrumentationKey: "b9d56dad-4148-458d-8dd1-147c61a17b07",
            properties: { prop1: "val1", prop2: "val2" },
            measurements: { m1: 100, m2: 200 }
        });

        window.appInsights = appInsights;        
        appInsights.trackPageView();
2) with queue push (if can't use option 1)

       appInsights.queue.push(function() { 
         appInsights.context.properties = {prop1: "val1", prop2: "val2"}
         appInsights.context.measurements = {m1: 100, m2: 200}
       });

Next steps - stop needing queue.push (though reading appInsights.properties in initialization)

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
@@ -20,13 +20,7 @@ class TelemetryContextTests extends TestClass {
             emitLineDelimitedJson: () => false,
             maxBatchSizeInBytes: () => 1000000,
             maxBatchInterval: () => 1,
-            disableTelemetry: () => false,
-            properties: () => {
-                return { prop1: "val1", prop2: "val2" };
-            },
-            measurements: () => {
-                return { m1: 100, m2: 200 };
-            }
+            disableTelemetry: () => false
         }
 
         this._telemetryContext = new Microsoft.ApplicationInsights.TelemetryContext(this._config);
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
@@ -20,7 +20,8 @@ class TelemetryContextTests extends TestClass {
             emitLineDelimitedJson: () => false,
             maxBatchSizeInBytes: () => 1000000,
             maxBatchInterval: () => 1,
-            disableTelemetry: () => false
+            disableTelemetry: () => false,
+            properties: () => { return { prop1: "val1", prop2: "val2" }; }
         }
 
         this._telemetryContext = new Microsoft.ApplicationInsights.TelemetryContext(this._config);
@@ -66,6 +67,17 @@ class TelemetryContextTests extends TestClass {
                 var contextKeys = new AI.ContextTagKeys();
                 Assert.equal("101", env.tags[contextKeys.sessionId], "session.id");
                 Assert.equal(true, env.tags[contextKeys.sessionIsFirst], "session.isFirst");
+            }
+        });
+
+        this.testCase({
+            name: "TelemetryContext: properties initialized correctly",
+            test: () => {
+                // act
+                var telemetryContext = new Microsoft.ApplicationInsights.TelemetryContext(this._config);
+
+                // verify
+                Assert.deepEqual(this._config.properties(), telemetryContext.properties);
             }
         });
 
@@ -126,10 +138,10 @@ class TelemetryContextTests extends TestClass {
                     // teardown
                     this._telemetryContext.properties = null;
                 }
-            });        
+            });
     }
 
-    private getTestEventEnvelope(properties?:Object) {
+    private getTestEventEnvelope(properties?: Object) {
         var event = new Microsoft.ApplicationInsights.Telemetry.Event('Test Event', properties);
         var eventData = new Microsoft.ApplicationInsights.Telemetry.Common.Data<Microsoft.ApplicationInsights.Telemetry.Event>(Microsoft.ApplicationInsights.Telemetry.Event.dataType, event);
         var eventEnvelope = new Microsoft.ApplicationInsights.Telemetry.Common.Envelope(eventData, Microsoft.ApplicationInsights.Telemetry.Event.envelopeType);
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
@@ -76,155 +76,15 @@ class TelemetryContextTests extends TestClass {
         });
 
         this.testCase({
-            name: "TelemetryContext: properties initialized correctly",
-            test: () => {
-                // act
-                var telemetryContext = new Microsoft.ApplicationInsights.TelemetryContext(this._config);
-
-                // verify
-                Assert.deepEqual(this._config.properties(), telemetryContext.properties);
-            }
-        });
-
-        this.testCase(
-            {
-                name: "TelemetryContext: custom properties applied if telemetry item's properties are undefined",
-                test: () => {
-                    // setup
-                    this._telemetryContext.properties = { prop1: "val1" };
-                    var eventEnvelope = this.getTestEventEnvelope(undefined);
-                    // act
-                    (<any>this._telemetryContext)._applyCustomProperties(eventEnvelope);
-
-                    // verify
-                    var resEvent = (<any>eventEnvelope.data).baseData;
-                    Assert.equal("val1", resEvent.properties.prop1);
-
-                    // teardown
-                    this._telemetryContext.properties = null;
-                }
-            });
-
-
-        this.testCase(
-            {
-                name: "TelemetryContext: telemetry item-specific custom properties are not touched",
-                test: () => {
-                    // setup
-                    this._telemetryContext.properties = { prop1: "val1" };
-                    var eventEnvelope = this.getTestEventEnvelope({ testProp: "testVal" });
-                    // act
-                    (<any>this._telemetryContext)._applyCustomProperties(eventEnvelope);
-
-                    // verify
-                    var resEvent = (<any>eventEnvelope.data).baseData;
-                    Assert.equal("testVal", resEvent.properties.testProp);
-
-                    // teardown
-                    this._telemetryContext.properties = null;
-                }
-            });
-
-        this.testCase(
-            {
-                name: "TelemetryContext: telemetry item-specific properties are not overriden",
-                test: () => {
-                    // setup
-                    this._telemetryContext.properties = { prop1: "context wide value" };
-                    var eventEnvelope = this.getTestEventEnvelope({ prop1: "item specific value" });
-                    
-                    // act
-                    (<any>this._telemetryContext)._applyCustomProperties(eventEnvelope);
-
-                    // verify
-                    var resEvent = (<any>eventEnvelope.data).baseData;
-                    Assert.equal("item specific value", resEvent.properties.prop1);
-
-                    // teardown
-                    this._telemetryContext.properties = null;
-                }
-            });
-
-        this.testCase({
-            name: "TelemetryContext: measurements initialized correctly",
-            test: () => {
-                // act
-                var telemetryContext = new Microsoft.ApplicationInsights.TelemetryContext(this._config);
-
-                // verify
-                Assert.deepEqual(this._config.measurements(), telemetryContext.measurements);
-            }
-        });
-
-        this.testCase(
-            {
-                name: "TelemetryContext: custom measurements applied if telemetry item's measurements are undefined",
-                test: () => {
-                    // setup
-                    this._telemetryContext.measurements = { m1: 100 };
-                    var eventEnvelope = this.getTestEventEnvelope(undefined, undefined);
-                    // act
-                    (<any>this._telemetryContext)._applyCustomMeasurements(eventEnvelope);
-
-                    // verify
-                    var resEvent = (<any>eventEnvelope.data).baseData;
-                    Assert.equal(100, resEvent.measurements.m1);
-
-                    // teardown
-                    this._telemetryContext.measurements = null;
-                }
-            });
-
-
-        this.testCase(
-            {
-                name: "TelemetryContext: telemetry item-specific measurements are not touched",
-                test: () => {
-                    // setup
-                    this._telemetryContext.measurements = { m1: 100 };
-                    var eventEnvelope = this.getTestEventEnvelope(undefined, { m2: 200 });
-                    // act
-                    (<any>this._telemetryContext)._applyCustomMeasurements(eventEnvelope);
-
-                    // verify
-                    var resEvent = (<any>eventEnvelope.data).baseData;
-                    Assert.equal(200, resEvent.measurements.m2);
-
-                    // teardown
-                    this._telemetryContext.measurements = null;
-                }
-            });
-
-        this.testCase(
-            {
-                name: "TelemetryContext: telemetry item-specific custom measurements are not overriden",
-                test: () => {
-                    // setup
-                    this._telemetryContext.measurements = { m1: 100 };
-                    var eventEnvelope = this.getTestEventEnvelope(undefined, { m1: 200 });
-                    
-                    // act
-                    (<any>this._telemetryContext)._applyCustomMeasurements(eventEnvelope);
-
-                    // verify
-                    var resEvent = (<any>eventEnvelope.data).baseData;
-                    Assert.equal(200, resEvent.measurements.m1);
-
-                    // teardown
-                    this._telemetryContext.measurements = null;
-                }
-            });
-
-        this.testCase({
-            name: "TelemetryContext: onInitializeTelemetry is called within track() and gets the envelope as an argument",
+            name: "TelemetryContext: onBeforeSendTelemetry is called within track() and gets the envelope as an argument",
             test: () => {
                 var eventEnvelope = this.getTestEventEnvelope();
 
                 var telemetryInitializer = {
-                    onInitializeTelemetry: (envelope) => { }
+                    initializer: (envelope) => { }
                 }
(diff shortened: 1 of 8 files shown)


AlexBulankou:
Telemetry initializer for JS SDK is one of the top asked features. Great to see a proposed PR!
I think with the telemetry initializer we want to accomplish the following: Allow users to augment **or update** their telemetry items before they are being sent with the **information that only is known at the time of sending**.
I do think the highlighted is important - users would like to update the properties at the time of sending, not just add to the list. Also in many cases users don't know what properties they would like to add to the properties and measurements list before beforehand.
The following proposed signature would illustrate what I mean - this would allow the user to accomplish this:

    }({
        instrumentationKey: "b9d56dad-4148-458d-8dd1-147c61a17b07",
        onBeforeSendTelemetry: myTelemetryInitializer
    });

where the signature of onBeforeSendTelemetry is as follows:

     var myTelemetryInitializer = function(telemetryType, context, properties, measurements){..}

This function would be called before telemetry item is sent - this would allow the user to check what telemetry type is being sent, and augment or update its context, properties or measurements dictionaries with the information that is known at the time of sending.

MaxShehovtsov:
Thanks for the feedback. I like the idea of onBeforeSendTelemetry - can you please give me more complete sample of how you'd use it.

SergeyKanzhelev:
Can you please elaborate on why you'd need to use telemetry initializers? Is it for the default pageView events that being sent? For every other event you can just set properties in the AppInsights.context, can you?

MaxShehovtsov:
@SergeyKanzhelev no, I can't set properties via AppInsights.context. It doesn't have 'properteis' member and there's no code to copy arbitrary properties to telemetry items.
This's been a confusion for some time as our documentation stated that the use case I described is valid - we've fixed our docs at some point.

SergeyKanzhelev:
Would it be better to have arbitrary properties in the context? I'm not against telemetry initializers - thinking about them I can make sense of it. It is just more convenient to set properties in the page context than implementing the function that may be more error prone...

SergeyKanzhelev:
@AlexBulankou, the name ```onBeforeSendTelemetry``` is aligned with the current implementation of telemetry initializers. However while implementing dependency and operations tracking it makes more sense to call ```Initialize``` on telemetry items when you start operation, not when you send it. For instance, you can get the current thread when item was tracked, not when it was attempted to be sent. I'd suggest to change name to something closer to ```OnInitialize```

olegsych:
It's great to see us solving the problem of telemetry initialization :+1:  I have several questions I would like to ask and a few nit-picky comments :blush:. Please see my comments in the code.

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -17,6 +17,8 @@ module Microsoft.ApplicationInsights {
         accountId: () => string;
         sessionRenewalMs: () => number;
         sessionExpirationMs: () => number;
+        properties: () => Object;
+        measurements: () => Object;
```
olegsych: Could you elaborate why you've chosen this particular design and its tradeoffs compared to the available alternatives? I've listed some of the alternative designs that have been suggested or implemented previously.

- This is a direct port of the .NET interface, similar to the 'onInitialize'/'onBeforeSendTelemetry` suggestion made by @SergeyKanzhelev and @AlexBulankou.
```TypeScript
export interface ITelemetryInitializer {
    initialize(envelope: Telemetry.Common.Envelope): void;
}
```

- A single-value initializer, where user-defined function returns value of a single property and is registered like this:
```TypeScript
config.properties["foo_property"] = () => "bar_value";
```

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -145,12 +161,35 @@ module Microsoft.ApplicationInsights {
 
             sessionStateEnvelope.time = Util.toISOStringForIE8(new Date(timestamp));
 
-            tc._track(sessionStateEnvelope); 
+            tc._track(sessionStateEnvelope);
+        }
+
+        private _applyCustomProperties(envelope: Microsoft.Telemetry.Envelope) {
+            if (this.properties && envelope && envelope.data && (<any>envelope.data).baseData) {
```
olegsych: Do we expect all of these conditions? Would it be more appropriate to [fail fast](http://martinfowler.com/ieeeSoftware/failFast.pdf) instead?

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -145,12 +161,35 @@ module Microsoft.ApplicationInsights {
...
+                var telemetryItem = (<any>envelope.data).baseData;
+                telemetryItem.properties = telemetryItem.properties || {};
+                Object.getOwnPropertyNames(this.properties).forEach((val, index, array) => {
```
olegsych: Would a [for...in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) be simpler here?

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -65,10 +67,20 @@ module Microsoft.ApplicationInsights {
...
 
         /**
+        * Custom properties added to all telemetry items (implementation of the TelemetryInitializer concept).
+        */
+        public properties;
+
+        /**
+        * Custom measurements added to all telemetry items (implementation of the TelemetryInitializer concept).
+        */
+        public measurements;
```
olegsych: Do we want to define measurements in `TelemetryContext` knowing that they are supported only by a small subset of telemetry types?

JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
```diff
@@ -68,6 +74,155 @@ class TelemetryContextTests extends TestClass {
...
+            test: () => {
+                // act
+                var telemetryContext = new Microsoft.ApplicationInsights.TelemetryContext(this._config);
+
+                // verify
+                Assert.deepEqual(this._config.properties(), telemetryContext.properties);
+            }
+        });
+
+        this.testCase(
```
olegsych: Can we use bracing style consistent with the existing code?

JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
```diff
@@ -68,6 +74,155 @@ class TelemetryContextTests extends TestClass {
...
+            {
+                name: "TelemetryContext: custom properties applied if telemetry item's properties are undefined",
+                test: () => {
+                    // setup
+                    this._telemetryContext.properties = { prop1: "val1" };
+                    var eventEnvelope = this.getTestEventEnvelope(undefined);
+                    // act
+                    (<any>this._telemetryContext)._applyCustomProperties(eventEnvelope);
```
olegsych: Can we test at the public interface level and call `track` instead of the private implementation level (`_applyCustomProperties`)?

JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
```diff
@@ -68,6 +74,155 @@ class TelemetryContextTests extends TestClass {
...
+
+                    // teardown
+                    this._telemetryContext.measurements = null;
+                }
+            });
+
+
+        this.testCase(
+            {
+                name: "TelemetryContext: telemetry item-specific measurements are not touched",
```
olegsych: Does this test fail independently from the ones above and below?

JavaScript/JavaScriptSDK/TelemetryContext.ts
olegsych: Are we intentionally choosing the term `measurements`, used by the data collection schema, as opposed to the term `metrics` used by the .NET API?

MaxShehovtsov:
@AlexBulankou @SergeyKanzhelev @olegsych @ameyagholkar 
Implemented 2nd approach - onInitializeTelemetry function which invokes on each track().
Usage sample (censores page url):

        var appInsights = window.appInsights || f<.........snippet goes on and on.....>
        }({
            instrumentationKey: "c218014f-b257-4c2f-885e-e20848e49b85"
        });
        window.appInsights = appInsights;

```javascript
        appInsights.queue.push(function () {
            appInsights.context.onInitializeTelemetry = function (envelope) {
                if (envelope.name == 
                 Microsoft.ApplicationInsights.Telemetry.PageView.envelopeType) {
                    var telemetryItem = envelope.data.baseData;
                    telemetryItem.url = "URL CENSORED";
                }

                return true;
            }
        });
```
        appInsights.trackPageView();

olegsych:
Do we want to allow multiple `InitializeTelemetry` callbacks? This may be beneficial if we plan to use them in the SDK code and for large applications that consist of multiple modules with different telemetry requirements.

MaxShehovtsov:
@olegsych I think we need the smallest thing possible (and I'd be ok to remove my 1st implementation with properties/measurements). I'd go incrementally and data/feedback driven. (And in theory you *can* achieve what you want wrappping multiple initializer into just one, but this will be user's responsibility)

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -145,12 +176,35 @@ module Microsoft.ApplicationInsights {
...
+                Object.getOwnPropertyNames(this.properties).forEach((val, index, array) => {
+                    if (telemetryItem.properties[val] === undefined) {
+                        telemetryItem.properties[val] = this.properties[val];
+                    }
+                });
+            }
+        }
+
+        private _applyCustomMeasurements(envelope: Microsoft.Telemetry.Envelope) {
+            if (this.measurements && envelope && envelope.data && (<any>envelope.data).baseData) {
```
AlexBulankou: _applyCustomProperties and _applyCustomMeasurements are identical functions with the exception of source and target object you're populating. Perhaps create a single function that takes in these two objects as input parameters and call this function from both _applyCustomProperties and _applyCustomMeasurements

JavaScript/JavaScriptSDK/TelemetryContext.ts
AlexBulankou: I still think (yeah I mentioned this already many times :)) that since you're adding onInitializeTelemetry, this other mechanism of adding properties and measurements is redundant - since you can do the same with onInitializeTelemetry

JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
```diff
@@ -68,6 +74,253 @@ class TelemetryContextTests extends TestClass {
...
+        });
+
+    }
+
+
+    private getTestEventEnvelope(properties?: Object, measurements?: Object) {
+        var event = new Microsoft.ApplicationInsights.Telemetry.Event('Test Event', properties, measurements);
+        var eventData = new Microsoft.ApplicationInsights.Telemetry.Common.Data<Microsoft.ApplicationInsights.Telemetry.Event>(Microsoft.ApplicationInsights.Telemetry.Event.dataType, event);
+        var eventEnvelope = new Microsoft.ApplicationInsights.Telemetry.Common.Envelope(eventData, Microsoft.ApplicationInsights.Telemetry.Event.envelopeType);
+        return eventEnvelope;
```
AlexBulankou: I think we also want test to verify user can add custom properties and measurements to the telemetry item with onInitializeTelemetry

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -65,10 +67,22 @@ module Microsoft.ApplicationInsights {
...
+        */
+        public properties;
+
+        /**
+        * Custom measurements added to all telemetry items (implementation of the TelemetryInitializer concept).
+        */
+        public measurements;
+        
+        public onInitializeTelemetry: (envelope: Telemetry.Common.Envelope) => boolean;
+
```
AlexBulankou: Hmm - what can user do to the envelope? Can they see what telemetry type this is? Can they change and add properties and measurements? Would it be better to have telemetry type string passed into onInitializeTelemetry as separate argument? Do we need to expose telemetry type constants for the user they can do this check?

AlexBulankou: Since onInitialzeTelemetry not just initializes but also makes decisions about whether to include/exclude telemetry item, should the verb be "initialize"? Perhaps "process" or "transform" would be better verbs. When I think "initialize" I "remove" is not usually in the realm of possible things to happen during "initialization", but when I think "process" this may include "remove". onBeforeSendTelemetry would also be an interesting option.

JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
MaxShehovtsov: will do

JavaScript/JavaScriptSDK/TelemetryContext.ts
MaxShehovtsov: I like onBeforeSendTelemetry, will switch to it

MaxShehovtsov: There are several options what to pass to the callback - all of them narrow the possibilities comparing to passing the whole envelope. And even worse - they make me guessing which one is more useful/needed more. I don't want to guess, so passing the whole thing and waiting for customers' feedback. Does that sound good?

JavaScript/JavaScriptSDK/TelemetryContext.ts
MaxShehovtsov: Agree, but when I start making it one function it's becoming hard/ugly to make it abstract from concrete member - properties or measurements. So I like it the way it's now better (still open for concrete suggestions/code examples)

JavaScript/JavaScriptSDK/TelemetryContext.ts
MaxShehovtsov: @AlexBulankou I'm ok with removing config.properties approach that i've implemented in my 1st iteration
@olegsych I like onBeforeSendTelemetry, even though i'm not sure we need an interface here. The 2nd one (config.properties[xxx] = () => {return sometihng}) looks very good. However its limitation is that if you need to set several properties you end up writing unnecessary code (comparing to onBeforeSendTelemetry).

olegsych:
Changing cardinality is difficult without introducing additional APIs or breaking backward compatibility. You could say that design that requires it is breaking the [open/closed principle](http://www.objectmentor.com/resources/articles/ocp.pdf). What would be the downside of changing the API so that its usage looks like this:
```JavaScript
        appInsights.queue.push(function () {
            ///
            /// This is how you add your telemetry initializer
            ///
            appInsights.context.addTelemetryInitializer(function (envelope) {
                if (envelope.name == 
                 Microsoft.ApplicationInsights.Telemetry.PageView.envelopeType) {
                    var telemetryItem = envelope.data.baseData;
                    telemetryItem.url = "URL CENSORED";
                }

                return true;
            });
        });
```

JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
```diff
@@ -68,6 +74,253 @@ class TelemetryContextTests extends TestClass {
...
+            }
+        });
+
+        this.testCase({
+            name: "TelemetryContext: onInitializeTelemetry returns false means item is not sent",
+            test: () => {
+                var eventEnvelope = this.getTestEventEnvelope();
+                var telemetryInitializer = {
+                    onInitializeTelemetry: (envelope) => {
+                        return false;
```
olegsych: Are we sure that implementing [multiple responsibilities](http://www.objectmentor.com/resources/articles/srp.pdf), telemetry initialization and telemetry processing in a single `onInitializeTelemetry` function is a good idea? In this specific scenario, a `boolean` return value is not obvious and requires documentation to understand. It also makes (potential) future support for multiple telemetry initializers more difficult. 

@upendras, how does this design align with the telemetry processor design you are implementing in the .NET sdk?

MaxShehovtsov: RE multiple responsibility - I think that returning true/false is unnecessary at all, noone asked for the ability to stop sending items. I'll remove this functionality (and your concern for multiple responsibility will be resolved)

RE chaining telemetry initializers - I don't think we need it. I didn't see people asking for it, and therefore no need to invest into it. So answering to your question - the downside is extra testing/maintaining something that noone needs. Let me know why you believe it is indeed needed.

upendras: As @olegsych mentioned, we are implementing chaining in .Net sdk. The reason it would be needed is when you have things like Filters where there can be more than one.

MaxShehovtsov:
@olegsych @AlexBulankou

Another try -
1) removed 1st implemetation via properties/measurements.
2) now exposing telemetry initializers via context.addTelemetryInitializer((envelope)=> {})

MaxShehovtsov:
Updated usage:

```javascript
    <script type="text/javascript">
        <SNIPPET>({
            instrumentationKey: "c218014f-b257-4c2f-885e-e20848e49b85"
        });

        window.appInsights = appInsights;

        // Adding telemetry initializer
        appInsights.queue.push(function () {
            appInsights.context.addTelemetryInitializer(function (envelope) {
                var telemetryItem = envelope.data.baseData;
                if (envelope.name == Microsoft.ApplicationInsights.Telemetry.PageView.envelopeType) {
                    telemetryItem.url = "URL CENSORED";
                }

                telemetryItem.properties = telemetryItem.properties || {};
                telemetryItem.properties["globalProperty"] = "boo";

                telemetryItem.measurements = telemetryItem.measurements || {};
                telemetryItem.measurements["globalMetric"] = 100;
            });
        });

        appInsights.trackPageView();
    </script>
```

olegsych:
:shipit:

PR #52, opened 2015-08-12
Move local storage logging severity from CRITICAL to WARNING. 

OsvaldoRosado:
Our telemetry has shown local storage support to be quite good. These logs can be quieted down a bit.

Also make the recently added session storage logging distinct from local storage logging.


MaxShehovtsov:
:sheep: it

PR #55, opened 2015-08-20
Bug in handling response when using XDomainRequest

searus:
The current version requires the response text to be "" or "200" for it
to be a successful call.
But, the web service actually returns a json string like this:
{"itemsRecieved":2,"itemsAccepted":2,"errors":[]}
This change attempts to parse the response text and check for a zero
length array of errors.

### DIFF
--- a/JavaScript/JavaScriptSDK/Sender.ts
+++ b/JavaScript/JavaScriptSDK/Sender.ts
@@ -221,11 +221,21 @@ module Microsoft.ApplicationInsights {
          * xdr state changes
          */
         public static _xdrOnLoad(xdr: XDomainRequest, payload: string) {
-            if (xdr && (xdr.responseText + "" === "200" || xdr.responseText === "")) {
+            
+            var validResponse: boolean = false;
+            
+            try {
+                // try and parse the response and check for errors
+                validResponse = (JSON.parse(xdr.responseText).errors.length === 0);
+            }
+            catch(e){ }
+            
+            if (validResponse) {
                 Sender._onSuccess(payload);
             } else {
                 Sender._onError(payload, xdr && xdr.responseText || "");
             }
+            
         }
 
         /**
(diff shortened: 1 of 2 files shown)


MaxShehovtsov:
Hi @searus , thank you for your contribution. Can you please describe why you need this change - how does it impact you (your app), what is it for? 
Currently everything 200 means success from our client-server communication perspective. Existence of errors means bad data not sending error.

searus:
Hi @MaxShehovtsov,
In IE9, the XDomainRequest object is used instead of XMLHttpRequest.  This does not allow you to check for the HTTP response code, so instead you look at the response text.  Currently, the App Insights service returns a JSON string, but Sender.ts expects either "" or "200" in the responseText to indicate success.
This means that in IE9 all submissions result in a failure rather than success.

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -221,11 +221,21 @@ module Microsoft.ApplicationInsights {
          * xdr state changes
          */
         public static _xdrOnLoad(xdr: XDomainRequest, payload: string) {
-            if (xdr && (xdr.responseText + "" === "200" || xdr.responseText === "")) {
+
```
searus: This is what https://dc.services.visualstudio.com/v2/track returns:
    {"itemsRecieved":2,"itemsAccepted":2,"errors":[]}
So, the response is never evaluated as successful

MaxShehovtsov: Got it, can you make a simple unit test please to check your parsing?

searus:
I've added a test, but I can't get the projects to build locally so I haven't been able to run it to see it if works.
I'll have another go tomorrow.

MaxShehovtsov:
Sure, thanks a lot! BTW check out wiki for general tests info, also let me know what problems you hit during building/testing

AlexBulankou:
Testing on the latest version of the changes, I see the following test failure:
SenderTests: XDomain sender can be invoked and handles errors (10, 7, 17)Rerun

The other tests are passing.

kamilszostak:
I updated the Sinon library to a new version and I fixed the XDomain test. Please check if the test is still failing for you.

AlexBulankou:
Closing stale PR

PR #61, opened 2015-09-01
Fix "Cannot set property 'assembly' of undefined" error in CreateSimpleException 

olegsych:
The following unhandled exception was reported 671 times by internal AI JS SDK telemetry during the week of 8/16, affecting 24 applications:

```
AI (Internal): _onerror threw [object Error]{ 
stack: 'TypeError: Cannot set property 'assembly' of undefined    
    at <error: TypeError: undefined is not a function>     
    at <error: TypeError: undefined is not a function>     
    at <error: TypeError: undefined is not a function>     
    at <error: TypeError: undefined is not a function>', 
message: 'Cannot set property 'assembly' of undefined', 
name: 'TypeError' while logging error, error will not be collected: null
```

I believe it is caused by a problem in the `Exception.CreateSimpleException` method, which is called by `SendCORSException` called from `_onerror`. The current code in `CreateSimpleException` assumes that a `Telemetry.Exception` instance always has at least one stack frame object in the `parsedStack`. However, when the array is empty, `parsedStack[0]` returns an undefined object, triggering the `TypeError` in the code trying to set the `assembly` property.

This pull request fixes the problem by creating a `Telemetry.Exception`-compatible object without relying on browser-specific exception throwing and handling functionality.

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Telemetry/Exception.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Telemetry/Exception.tests.ts
@@ -156,6 +156,34 @@ Error: testmessage2\n\
                 test(ie, firefox);
             }
         });
+
+        this.testCase({
+            name: "CreateSimpleException returns Exception instance with specified properties",
+            test: () => {
+                var expectedMessage = "Test Message";
+                var expectedTypeName = "Test Type Name";
+                var expectedDetails = "Test Details";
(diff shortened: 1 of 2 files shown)


abaranch:
:shipit:

SergeyKanzhelev:
I like compact code! :shipit:

olegsych:
@MaxShehovtsov, @AlexBulankou, I think this pull request is ready to merge, but I wanted to let you guys do it.

PR #62, opened 2015-09-02
Changing console references to use Util console warn function

ameyagholkar:
With reference to error - AI (Internal): trackPageView failed on page load calculation: [object Error]{ stack: 'undefined', message: ''console' is undefined'.

We are using console.warn directly in two places and console.error in one which will throw errors in case that the "console" object is undefined (as the error above suggests.) In this pull request, we are switching to use the Util console warn func which checks for the presence of the console object and its funcs before using them.


SergeyKanzhelev:
Cool :shipit:

PR #72, opened 2015-11-08
By default page view duration equals Page load time 

MaxShehovtsov:
1) Default scenario: forcing PageView duration being page load time (equal to PageViewPerformance.total).
This will be the default behavior. 

2) If new config parameter ('relativePageViewDuration') is set to true than page view duration will equal to the time from window.performance.timing.navigationStart to the moment of calling the trackPageView(). This is useful when page view duration has a special meaning and is not bound to "physical" page load time. For example, if a page has many ajax calls which execute after DOM and all resources are loaded, and we consider the last ajax call as the end of page load.

### DIFF
--- a/JavaScript/JavaScriptSDK/appInsights.ts
+++ b/JavaScript/JavaScriptSDK/appInsights.ts
@@ -31,7 +31,7 @@ module Microsoft.ApplicationInsights {
         samplingPercentage: number;
         autoTrackPageVisitTime: boolean;
         autoTrackAjax: boolean;
-        relativePageViewDuration: boolean;
+        overridePageViewDuration: boolean;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -185,56 +186,56 @@ module Microsoft.ApplicationInsights {
...
                         }
-                    } catch (e) {
-                        _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.CRITICAL, "trackPageView failed on page load calculation: " + Util.dump(e));
+
+                        this.flush();
                     }
-                }, 100);
-            }
+                    else if (Telemetry.PageViewPerformance.getDuration(start, +new Date) > maxDurationLimit) {
+                        clearInterval(handle);
```
AlexBulankou: why aren't we sending page view performance data in this case (timeout exceeded), only pageview data?

MaxShehovtsov: Because maxDurationLimit reached and performance timing data is not available yet (see previous "if"). I.e. we don't have perf data to report and we don't want to wait any longer, and we still want to report page view.

PR #77, opened 2015-11-13
Setting operation name to window.location.pathname

MaxShehovtsov:
This change will allow to glue page view and page view performance documents in the UI. So that we can show a blade filtered by operation name and showing data of both doc types (raw grid with page views and performance stacked area chart based on page view performance)

### DIFF
--- a/JavaScript/JavaScriptSDK/Context/Operation.ts
+++ b/JavaScript/JavaScriptSDK/Context/Operation.ts
@@ -12,7 +12,10 @@ module Microsoft.ApplicationInsights.Context {
         public syntheticSource: string;
 
         constructor() {
-            this.id = Util.newGuid();
+            this.id = Util.newGuid();            
+            if (window && window.location && window.location.pathname) {
+                this.name = window.location.pathname;
+            }
         }
     }
 } 


JavaScript/JavaScriptSDK/Context/Operation.ts
```diff
@@ -12,7 +12,10 @@ module Microsoft.ApplicationInsights.Context {
         public syntheticSource: string;
 
         constructor() {
-            this.id = Util.newGuid();
+            this.id = Util.newGuid();            
+            if (window && window.location && window.location.pathname) {
+                this.name = window.location.pathname;
```
SergeyKanzhelev: should it be taken from the pageView.name?

MaxShehovtsov: Good question. Currently in UI i see page URL and page name being treated separately - when you look at diagnostics search result for page view. Page name is usually a text description, more usage than diagnostics oriented. Where the url (and derived from it baseUrl) will arguably be more useful to classify the page view from dev perspective.
By default I think url will work better/will be more predictable than window.document.title used for default page name. This obviously can be overridden with a telemetry initializer.

AlexBulankou:
:shipit:

JavaScript/JavaScriptSDK/Context/Operation.ts
SergeyKanzhelev: My point was that you may use whatever customer passed in the initial trackPage() as a name, not read it from document.url. As I mentioned [here](http://apmtips.com/blog/2015/08/15/disable-browser-telemetry-from-the-page/) customer may choose not to send URL by changing 

``` javascript
appInsights.trackPageView();
```

which is ultimately just

``` javascript
appInsights.trackPageView(window.location.pathname, window.location.href);
```

to: 

``` javascript
appInsights.trackPageView("something else", window.location.href);
```
And customer would expect to see the same ```something else``` as operation name for ajax.

MaxShehovtsov: I understand it. However 
1) by default page name equals to document.title so if I set operation name being page name it won't be usable;
 and 
2) there's no explicit connection between this and operation name. So I would not expect change in page name impacting operation name.
I don't see the value in hiding URL, but if this is the case then users will need to override it manually.

Issue #74, opened 2015-11-11
Do not send sampleRate when it is 100


SergeyKanzhelev:

kamilszostak:
Fixed in #78

PR #78, opened 2015-11-18
Do not send sampleRate when it is 100 #74

kamilszostak:
* Refactor aiDataContract object to use enum instead of bool 
* User can use functions in aiDataContract (see Envelope.ts)
* User can mark a fields as hidden an it will not be serialized

### DIFF
--- a/JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
+++ b/JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
@@ -20,14 +20,16 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
             this.name = name;
             this.data = data;
             this.time = Util.toISOStringForIE8(new Date());
-            
+
             this.aiDataContract = {
-                time: true,
-                iKey: true,
-                name: true,
-                sampleRate: true,
-                tags: true,
-                data: true
+                time: FieldType.Required,
+                iKey: FieldType.Required,
+                name: FieldType.Required,
+                sampleRate: () => {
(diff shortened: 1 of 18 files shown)


JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
```diff
@@ -20,14 +20,16 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
...
-                iKey: true,
-                name: true,
-                sampleRate: true,
-                tags: true,
-                data: true
+                time: FieldType.Required,
+                iKey: FieldType.Required,
+                name: FieldType.Required,
+                sampleRate: () => {
+                    return (this.sampleRate == 100) ? FieldType.Hidden : FieldType.Required;
```
SergeyKanzhelev: maybe you need to compare with precision. Not sure how JavaScript handles doubles

MaxShehovtsov:
:shipit:

SergeyKanzhelev:
@MaxShehovtsov so in javascript comparing double (or float) with integer will work just fine? Do we need to compare with precision?

MaxShehovtsov:
@SergeyKanzhelev there's only "number" type in javascript (http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.20) . See this example: http://jsfiddle.net/zoyt3tpn/2/

kamilszostak:
It looks like JS handles float numbers quite well and we don't need to put any special code.

See: http://jsfiddle.net/zoyt3tpn/11/

MaxShehovtsov:
@kamilszostak your example actually disproves it (given the 2nd 'false') :) But I still think it's fine :) 
:sheep: :it:

PR #79, opened 2015-11-20
Update Sinon lib from 1.7.3 to 1.17.3

kamilszostak:


JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -38,7 +38,6 @@ class AppInsightsTests extends TestClass {
     }
 
     public testCleanup() {
-        this.clock.reset();
```
MaxShehovtsov: what's the problem with this?

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -373,7 +373,9 @@ module Microsoft.ApplicationInsights {
 
                     // do not intercept onreadystatechange if it is defined and not a function, because we are not able to call original function in this case, which happends on Firefox 13 and lower
                     if (extensions.IsNullOrUndefined(this.onreadystatechange) || (typeof (this.onreadystatechange) === "function")) {
-                        this.ajaxData.originalOnreadystatechage = this.onreadystatechange;
+                        if ((typeof (this.onreadystatechange) === "function") && (this.onreadystatechange.name !== "onreadystatechangeWrapper")) {
```
MaxShehovtsov: Why do you do this?

kamilszostak: If this method is called twice it will set this.ajaxData.originalOnreadystatechage to onreadystatechangeWrapper and create an infinite loop in onreadystatechangeWrapper function (line 411 and 418 in ajax.ts file).

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
kamilszostak: this.clock is undefined at this point. I think it's being cleared automatically by the sandbox. See: https://github.com/Microsoft/ApplicationInsights-JS/blob/b056eae7cc92c21424e41f90724bcc593621833a/JavaScript/JavaScriptSDK.Tests/TestFramework/TestClass.ts#L163

MaxShehovtsov:
:shipit: Thanks!

PR #80, opened 2015-11-21
Don't pollute the global namespace

gregjhogan:


MaxShehovtsov:
@gregjhogan Thank you!

PR #81, opened 2015-11-22
Single-page application page tracking

gregjhogan:
For multi-page applications the operation ID is different for each page request.  Support for equivalent behavior is needed for single page applications.  Perhaps a new method ```appInsights.trackPageRequest()``` which changes the operation ID is a reasonable way to support this?

### DIFF
--- a/JavaScript/JavaScriptSDK/appInsights.ts
+++ b/JavaScript/JavaScriptSDK/appInsights.ts
@@ -1,6 +1,7 @@
 ﻿/// <reference path="telemetrycontext.ts" />
 /// <reference path="./Telemetry/Common/Data.ts"/>
 /// <reference path="./Util.ts"/>
+/// <reference path="./Context/Operation.ts"/>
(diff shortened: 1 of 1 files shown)


AlexBulankou:
Closing stale PR

gregjhogan:
Does my request make sense? Should I update and resubmit?  If not how do you recommend changing the operation ID when navigating around a single page application?  Thanks!

AlexBulankou:
@gregjhogan apologies for closing this PR without explanation! I only just realised we never properly responded to it. You are absolutely right that for SPAs the user may want to create new operationID and new operation name for all or some of page transitions, however we feel that it doesn't warrant a separate  root level API. The name trackpagerequest can be easily mixed up with trackpageview and gives no indication that operation ID is changing. This API would be redundant, we feel the user should implement this on their side as we are trying to approach introduction of each new API extremely carefully. Going forward we may decide to invest in SPA support and implement some sort of page auto tracking for popular SPA frameworks.

PR #83, opened 2015-11-24
Refactoring ajax monitoring code

MaxShehovtsov:


JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -291,211 +58,284 @@ module Microsoft.ApplicationInsights {
...
 
                         var successStatuses = [200, 201, 202, 203, 204, 301, 302, 303, 304];
 
-                        appInsights.trackAjax(
-                            this.ajaxData.getAbsoluteUrl(),
-                            this.ajaxData.async,
-                            this.ajaxData.ajaxTotalDuration,
-                            successStatuses.indexOf(+this.ajaxData.status) != -1
+                this.appInsights.trackAjax(
+                    (<any>xhr).ajaxData.getAbsoluteUrl(),
```
SergeyKanzhelev: it is not related to this PR - but you need to make dependency.Name to be low cardinality. You cannot/should not just use absolute URL as a name - ideally use host.

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -363,5 +363,16 @@ module Microsoft.ApplicationInsights {
             
             return true;
         }
+
+        public static getAbsoluteUrl = (function () {
+            var a;
+
+            return function (url) {
+                if (!a) a = document.createElement('a');
```
SergeyKanzhelev: are there any drawbacks creating dom elements? Also use `Util.document` to minimize the code better. Is there a way to enforce using proper document object

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -1,288 +1,55 @@
...
-                        interceptSend();
-                        interceptAbort();
-                        initiated = true;
+        private Init = function () {
+            if (this.supportMonitoring()) {
+                this.instrumentOpen();
+                this.instrumentSetRequestHeader();
+                this.instrumentSend();
+                this.instrumentAbort();
+                this.initiated = true;
```
SergeyKanzhelev: should we catch exception and set initiate to some error state so there will be no more attempts to initiate it?

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -291,211 +58,284 @@ module Microsoft.ApplicationInsights {
...
                     var originalSend = XMLHttpRequest.prototype.send;
+            var ajaxMonitorInstance = this;
                     XMLHttpRequest.prototype.send = function (content) {
-                        sendPrefixInterceptor.call(this, content);
+                try {
+                    ajaxMonitorInstance.sendPrefixInstrumentor(this, content);
+                } catch (e) {
+                    _InternalLogging.throwInternalNonUserActionable(
+                        LoggingSeverity.CRITICAL,
+                        "Failed to monitor XMLHttpRequest.send, monitoring data for this ajax call may be incorrect: "
```
SergeyKanzhelev: Is it guaranteed that the original ajax call will still be made? Is there is no guarantee can we do emergency shut off of all further ajax monitoring calls? Is there is guarantee - let's say about it in the error message.

Also - report url or some other identifier of the original ajax call

JavaScript/JavaScriptSDK/ajax/ajax.ts
MaxShehovtsov: the flag sets to true as a last thing after initialization, in other words it's true only if initialization was successful. If it fails we'll get an exception and the flag remains false.
Also ajaxMonitor itself is initialized only once (from appInsights constructor), so there should be no consequent attempts to initialize it again...

JavaScript/JavaScriptSDK/ajax/ajax.ts
MaxShehovtsov: What will happen if I keep using absolute url?

JavaScript/JavaScriptSDK/Util.ts
MaxShehovtsov: To be clear it's wouldn't affect minimization, but it does make code cleaner. Thanks!

JavaScript/JavaScriptSDK/ajax/ajax.ts
SergeyKanzhelev: You'll start getting `OtherValueReservedforAIUser` instead of the name in charts and sausages

JavaScript/JavaScriptSDK/Util.ts
SergeyKanzhelev: can you pls explain why it doesn't affect minimization? Does minifier creates global `document` object to use everywhere?

JavaScript/JavaScriptSDK/ajax/ajax.ts
SergeyKanzhelev: Do we have protection from `AppInsights` being initialized more than once? Will it break anything? 

Also is there anything that could fail in initialization at all?

JavaScript/JavaScriptSDK/ajax/ajax.ts
AlexBulankou: Agreed on the name=host. Looks like we need to change in RDD constructor to continue using absolute url as command name, but use just pathname as RDD name?
What's interesting though is that dependency property that used for grouping and for displaying in properties is already just host as I checked. Looks like backend is adjusting it?:

image

JavaScript/JavaScriptSDK/ajax/ajax.ts
MaxShehovtsov: try/catch guarantees that the original callback will be invoked. 

Adding url to the console output.

JavaScript/JavaScriptSDK/ajax/ajax.ts
MaxShehovtsov: We don't have this protection, however I'm not sure if we need one (although I can imagine whole bunch of problems related to multiple instances of AppInsights). I'd rather stay in reactive mode until we get real use cases.

JavaScript/JavaScriptSDK/Util.ts
MaxShehovtsov: I think you're right - it does affect minification (as a result we'll have 'document' instead of some sort of minified string out of 'util.document').

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -1,528 +1,382 @@
...
-            var extensions = {
-                IsNullOrUndefined: function (obj) {
-                    return typeof (obj) === constants.udf || obj === null;
-                },
+        }
 
-            };
+        ///<summary>Determines whether ajax monitoring can be enabled on this document</summary>
+        ///<returns>True if Ajax monitoring is supported on this page, otherwise false</returns>
+        private supportMonitoring() {
```
AlexBulankou: change TS signature to return true?

JavaScript/JavaScriptSDK/Telemetry/RemoteDependencyData.ts
```diff
@@ -31,16 +31,17 @@ module Microsoft.ApplicationInsights.Telemetry {
         /**
          * Constructs a new instance of the PageEventTelemetry object
```
AlexBulankou: incorrect comment?

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -1,528 +1,382 @@
...
-
-                ///<summary>Function that returns property name which will identify that monitoring for given instance of XmlHttpRequest is disabled</summary>
-                this.GetDisabledPropertyName = function () {
-                    return getDisabledPropertyName();
-                };
-
-
-                function getDisabledPropertyName() {
-                    return "Microsoft_ApplicationInsights_BypassAjaxInstrumentation";
+                } catch (e) {
```
AlexBulankou: Shouldn't this try..catch block be around originSetRequestHeader.apply? Otherwise it is only monitoring for errors in ajaxData.requestSize line.  Actually if we do this this way, then we'll catch all errors including the ones from original xhr request. So what are we really hoping to check with this try..catch?

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -1,528 +1,382 @@
...
+                xhr.ajaxData.CalculateMetrics();
+
+                var successStatuses = [200, 201, 202, 203, 204, 301, 302, 303, 304];
+
+                this.appInsights.trackAjax(
+                    xhr.ajaxData.getHostName(),
+                    xhr.ajaxData.getAbsoluteUrl(),
+                    xhr.ajaxData.async,
+                    xhr.ajaxData.ajaxTotalDuration,
+                    successStatuses.indexOf(+(xhr.ajaxData.status)) != -1
```
AlexBulankou: why not anything < 400?

JavaScript/JavaScriptSDK/ajax/ajax.ts
MaxShehovtsov: It's intentional - I don't want to handle/swallow exceptions from user's code. Also the code I do wrap with try/catch can potentially throw - however it's more a safety pattern than a real need to catch errors here.

JavaScript/JavaScriptSDK/ajax/ajaxUtils.ts
```diff
@@ -0,0 +1,86 @@
+﻿/// <reference path="../logging.ts" />
+/// <reference path="../util.ts" />
+
+module Microsoft.ApplicationInsights {
+    "use strict";
+
+    export class extensions {
```
AlexBulankou: do we have unit tests for these reusable extension classes?

JavaScript/JavaScriptSDK/ajax/ajax.ts
MaxShehovtsov: good idea, will do

JavaScript/JavaScriptSDK/ajax/ajaxUtils.ts
MaxShehovtsov: nope

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -1,528 +1,382 @@
...
-                    XMLHttpRequest.prototype.abort = function () {
-                        if (isMonitoredInstance.call(this)) {
-                            this.ajaxData.aborted = 1;
+                            else {
+                                ajaxMonitorInstance.instrumentOnReadyStateChange(xhr);
+                            }
+                        } catch (e) {
+                            _InternalLogging.throwInternalNonUserActionable(
+                                LoggingSeverity.CRITICAL,
+                                "Failed to instrument XMLHttpRequest.onreadystatechange"
```
SergeyKanzhelev: there should be some sort of protection that will ensure we do not send error too often from the same page. I think that all these "failed to instrument Foo" messages means that something is very broken with the browser/page and we need to stop processing ajaxes on this page completely.

MaxShehovtsov: We have a protection against too many internal logs - https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/Logging.ts#L111

As for "something is very broken" - this is an assumption, perhaps valid, but still an assumption. Let's get errors first and then act on them.

PR #85, opened 2015-12-02
PageView: sending 0 duration when duration is unknown

MaxShehovtsov:
We don't want to lose page views when navigation timing is not supported by browsers (3% of cases).
However since duration property is mandatory we have no choice but setting it to 0.

### DIFF
--- a/JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
+++ b/JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
@@ -75,7 +75,7 @@ module Microsoft.ApplicationInsights.Telemetry {
 
                 if (total == 0) {
                     _InternalLogging.throwInternalNonUserActionable(
-                        LoggingSeverity.WARNING,
+                        LoggingSeverity.CRITICAL,
                         "error calculating page view performance: total='" +
                         total + "', network='" + network + "', request='" + request + "', response='" +
                         response + "', dom='" + dom + "'");
@@ -83,7 +83,7 @@ module Microsoft.ApplicationInsights.Telemetry {
                     // some browsers may report individual components incorrectly so that the sum of the parts will be bigger than total PLT
                     // in this case, don't report client performance from this page                    
                     _InternalLogging.throwInternalNonUserActionable(
-                        LoggingSeverity.WARNING,
+                        LoggingSeverity.CRITICAL,
                         "client performance math error:" + total + " < " + network + " + " + request + " + " + response + " + " + dom);
                 } else {
                     this.durationMs = total;
--- a/JavaScript/JavaScriptSDK/appInsights.ts
+++ b/JavaScript/JavaScriptSDK/appInsights.ts
@@ -190,8 +190,10 @@ module Microsoft.ApplicationInsights {
             if (!Telemetry.PageViewPerformance.isPerformanceTimingSupported()) {
                 // no navigation timing (IE 8, iOS Safari 8.4, Opera Mini 8 - see http://caniuse.com/#feat=nav-timing)
                 _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.CRITICAL,
-                    "trackPageView failed: navigation timing API used for calculation of page duration is not supported in this browser.");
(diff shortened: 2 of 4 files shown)


JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -190,16 +190,18 @@ module Microsoft.ApplicationInsights {
             if (!Telemetry.PageViewPerformance.isPerformanceTimingSupported()) {
                 // no navigation timing (IE 8, iOS Safari 8.4, Opera Mini 8 - see http://caniuse.com/#feat=nav-timing)
                 _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.CRITICAL,
```
SergeyKanzhelev: Do we still need to report it to the customer? It is a known issue that neither customer nor we can address. So perhaps we can omit it. Also zero will always mean "navigation API not supported" so you can query this information without having this error message

JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
```diff
@@ -83,7 +83,7 @@ module Microsoft.ApplicationInsights.Telemetry {
                     // some browsers may report individual components incorrectly so that the sum of the parts will be bigger than total PLT
                     // in this case, don't report client performance from this page                    
                     _InternalLogging.throwInternalNonUserActionable(
-                        LoggingSeverity.WARNING,
+                        LoggingSeverity.CRITICAL,
```
SergeyKanzhelev: do we need page name for easier troubleshooting by customer? Also you need to report the outcome of this problem. Not "error calculating", but something like "page view will report zero due to error in calculating"...

Why do you want customer to see this errors in the portal?

JavaScript/JavaScriptSDK/appInsights.ts
MaxShehovtsov: I think this is about data quality (first assuring customer that page view will be sent, second letting customer know that no duration will be collected) - so I'd keep customers informed.
0 does not only mean that API is not supported - it could be API bug (it's supported but returns 0) or manual override, also different versions of SDK use API differently so it makes it harder to perform investigations about data quality issues. And specific errors simplify it.

JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
MaxShehovtsov: In case of these errors we won't send client perf. However absence of client perf event may also be caused by lack of API support altogether (safari specific case).

I agree that these messages are not actionable. Also I think we should not show AI (Internal) messages in the console - I'll create a bug.

JavaScript/JavaScriptSDK/appInsights.ts
SergeyKanzhelev: I think at some stage everybody who cares about data quality will know that 5% of the data is unreliable due to the error in navigation API. If one needs more information - can customer detect this case and start reporting it himself? Maybe using telemetry initializer? Can we initialize the hidden field with the error descitption so it can be read by telemetry initializer and sent as custom property. My main concern is that we should not pollute telemetry with known errors.

For instance we do not report zero duration for requests telemetry but customer can increase error severity level and detect all such cases.

MaxShehovtsov: Even though it's a known issue - it's still unknown what the percentage will be in, say, one month. Currently it's affecting 1% of customers, so we dicided to send 0 duration. If it becomes 10% (more safari oriented applications, or browser demography changes) we'll have to do something about it. Without this telemetry we'll go blind

MaxShehovtsov:
@SergeyKanzhelev I'm going to merge the pull request - if you still think we need to follow up on our tracing approach, let's discuss it separately.

JavaScript/JavaScriptSDK/appInsights.ts
SergeyKanzhelev: My point is - we have telemetry - the number of zeroes. And it is a known problem that doesn't affect many customers. If it go wildly up we can return to discussion what kind of telemetry we need.

MaxShehovtsov: Right, and my point is - without this telemetry we have no idea if we go wild or not.

Issue #84, opened 2015-11-25
Support ajax monitoring


MaxShehovtsov:
**Ajax calls** are instrumented by overriding prototype.open and prototype.send functions, and onreadystatechange event handler of XMLHttpRequest. 

This is an entrypoint for ajax auto collection: https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/ajax/ajax.ts

Below is the ajax monitoring sequence chart:

* **XHR** – an instance of XMLHttpRequest
* **User-defined** callbacks – while appInsights subscribes (‘attaches’) to XMLHttpRequest.onreadystatechange event there might be other event handlers before or after (if user subscribed before or after initialization of appInsights object).
* **AjaxMonitor** and **AjaxRecord** are the new appInsights entities added for supporting ajax monitoring.
* **AppInsights** is the object exposing our public API (trackXXX methods) and initializing javascript monitoring.

![](https://cloud.githubusercontent.com/assets/10049593/11410408/fe9c0b0e-937b-11e5-9047-19435bfdfa8a.png)

SergeyKanzhelev:
Questions: 
- How monitoring will behave if `send` will be sent again before the callback or result returned?
- Is it a valid scenario to do multiple `send` for one `open`?
- Will we detect long-poll ajax calls and adjust their processing so they do not make average ajax execution time enormously big?
- Do we instrument `abort` method?
- is `Callbacks` only stored in `XHR` object itself? If not - will it keep it from being garbage collected as it keeps delegates on it?
- Will the `ajaxRecord` stay alive after `trackAjax` call? Will Callbacks unsubscribe of `onreadystatechange` after that?
- Do we plan to stop and send in-flight ajax calls on page `onbeforeunload` callback?

As a side note I'd advice to set a lock when you call `original Open` and `original Send` so that "originl" callback wouldn't call you back and enter recursion in case when customer application has some logic that also does override of `XHR` and do not expect somebody in the middle to change callbacks.

MaxShehovtsov:
Thank you for the feedback:

- 2nd should fail (and does so in chrome), however we shouldn't execute send handler 2nd time. I'm adding validations for this.
- long ajax - I think they will be there, spoiling averages, and it's by design
- We do instrument abort however we currently don't send any data about abortion status. So depending on timing when abort was called - if request succeeds it'll be treated as any other ajax call. It makes sense to send the custom property with abortion status though.
[...]

PR #88, opened 2015-12-02
Ajax feedback

MaxShehovtsov:
Addressed feedback from here - https://github.com/Microsoft/ApplicationInsights-JS/issues/84

### DIFF
--- a/JavaScript/JavaScriptSDK/ajax/ajax.ts
+++ b/JavaScript/JavaScriptSDK/ajax/ajax.ts
@@ -92,9 +92,7 @@ module Microsoft.ApplicationInsights {
             var ajaxData = new ajaxRecord();
             ajaxData.method = method;
             ajaxData.requestUrl = url;
-            ajaxData.requestSize += url.length;
-            // If not set async defaults to true 
-            ajaxData.async = extensions.IsNullOrUndefined(async) ? true : async;
+            ajaxData.requestSize += url.length;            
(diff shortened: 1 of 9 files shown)


JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -92,12 +88,25 @@ module Microsoft.ApplicationInsights {
             };
         }
 
+        private openHandler(xhr: XMLHttpRequestInstrumented, method, url, async) {
+            var ajaxData = new ajaxRecord();
+            ajaxData.method = method;
+            ajaxData.requestUrl = url;
+            ajaxData.requestSize += url.length;
+            // If not set async defaults to true 
+            ajaxData.async = extensions.IsNullOrUndefined(async) ? true : async;
```
SergeyKanzhelev: As I mentioned we are making work to remove async attribute from schema. Also it's nullable - you should not default it to true.

SergeyKanzhelev:
Looks good

JavaScript/JavaScriptSDK/ajax/ajax.ts
MaxShehovtsov: I'll remove async

MaxShehovtsov:
@SergeyKanzhelev I removed async property from ajax (rdd) telemetry item

Issue #73, opened 2015-11-10
Do not collect locale and screen resolution by default


SergeyKanzhelev:

MaxShehovtsov:
@SergeyKanzhelev what do you mean 'by default'? Should there be a config parameter turning on this collection?

SergeyKanzhelev:
customer can create a telemetry initializer or insert initialization in config snippet, right? We may document it

AlexBulankou:
This is not a bug in existing product, that was intended by design, so I removed bug label and added enhancement label.

PR #89, opened 2015-12-03
some minimization

SergeyKanzhelev:
Make identifiers shorter and remove some data we collect by default 
[#73](https://github.com/Microsoft/ApplicationInsights-JS/issues/73)

### DIFF
--- a/JavaScript/JavaScriptSDK/Context/Session.ts
+++ b/JavaScript/JavaScriptSDK/Context/Session.ts
@@ -15,10 +15,10 @@ module Microsoft.ApplicationInsights.Context {
          */
         public id: string;
 
-        /**
-         * The true if this is the first session
+        /**  
+         * The true if this is the first session  
          */
-        public isFirst: boolean;
+        public isFirst: boolean;  
 
         /**
          * The date at which this guid was genereated.
@@ -79,7 +79,7 @@ module Microsoft.ApplicationInsights.Context {
                 if (typeof this._sessionHandler === "function") {
                     this._sessionHandler(AI.SessionState.End, this.automaticSession.renewalDate);
                 }
-                this.automaticSession.isFirst = undefined;
+                this.automaticSession.isFirst = undefined; 
                 this.renew();
             } else {
                 this.automaticSession.renewalDate = +new Date;
@@ -116,7 +116,7 @@ module Microsoft.ApplicationInsights.Context {
             }
 
             if (!this.automaticSession.id) {
-                this.automaticSession.isFirst = true;
+                this.automaticSession.isFirst = true; 
                 this.renew();
             }
         }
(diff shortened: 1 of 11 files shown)


JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -227,19 +227,18 @@ module Microsoft.ApplicationInsights {
...
-            var hexValues = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "A", "B", "C", "D", "E", "F"];
-
-            // c.f. rfc4122 (UUID version 4 = xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx)
-            var oct = "", tmp;
-            for (var a = 0; a < 4; a++) {
-                tmp = (4294967296 * Math.random()) | 0;
-                oct += hexValues[tmp & 0xF] + hexValues[tmp >> 4 & 0xF] + hexValues[tmp >> 8 & 0xF] + hexValues[tmp >> 12 & 0xF] + hexValues[tmp >> 16 & 0xF] + hexValues[tmp >> 20 & 0xF] + hexValues[tmp >> 24 & 0xF] + hexValues[tmp >> 28 & 0xF];
-            }
+        public static newId(): string {
+            var base64chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/';
```
MaxShehovtsov: Is it still gonna be a valid GUID?

JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -16,11 +16,6 @@ module Microsoft.ApplicationInsights.Context {
         public id: string;
 
         /**
-         * The true if this is the first session
-         */
-        public isFirst: boolean;
```
MaxShehovtsov: I thought that our UI in Settings->Usage->Users->New users chart is based on isFirst flag. Can you double check pls

JavaScript/JavaScriptSDK/Util.ts
SergeyKanzhelev: Nope. Do you know any places we are using it as a GUID?

JavaScript/JavaScriptSDK/Context/Session.ts
SergeyKanzhelev: Yes, you are correct. We are still using it to detect "new users" vs. "returning users". I really wonder why it is not a field on SessionState event. But nevertheless I returned it back

JavaScript/JavaScriptSDK/Util.ts
MaxShehovtsov: Does server side SDK (including Java and Node) treat is as a string? If yes then you should be fine.

SergeyKanzhelev: Validated. It is string everywhere

SergeyKanzhelev:
@MaxShehovtsov  is it looking ok now? Note, I removed `accountAcquisitionDate` if compare to the Friday and returned isFirst for the session context back

MaxShehovtsov:
Looks good. Merging. Thanks!

PR #90, opened 2015-12-07
Log duration on stopTrackEvent

kamilszostak:
Send the duration between startTrackEvent and stopTrackEvent to the server.

### DIFF
--- a/JavaScript/JavaScriptSDK/appInsights.ts
+++ b/JavaScript/JavaScriptSDK/appInsights.ts
@@ -84,6 +84,15 @@ module Microsoft.ApplicationInsights {
             // initialize event timing
             this._eventTracking = new Timing("trackEvent");
             this._eventTracking.action = (name?: string, url?: string, duration?: number, properties?: Object, measurements?: Object) => {
+                if (properties === undefined) {
+                    properties = { duration: duration };
+                }
+                else {
+                    // do not override existing duration value
+                    if (isNaN(properties["duration"])) {
+                        properties["duration"] = duration;
+                    }
+                }
                 var event = new Telemetry.Event(name, properties, measurements);
                 var data = new ApplicationInsights.Telemetry.Common.Data<ApplicationInsights.Telemetry.Event>(Telemetry.Event.dataType, event);
                 var envelope = new Telemetry.Common.Envelope(data, Telemetry.Event.envelopeType);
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -84,6 +84,15 @@ module Microsoft.ApplicationInsights {
             // initialize event timing
             this._eventTracking = new Timing("trackEvent");
             this._eventTracking.action = (name?: string, url?: string, duration?: number, properties?: Object, measurements?: Object) => {
+                if (properties === undefined) {
```
SergeyKanzhelev: what if properties is null?
Why properties, not measurements?

kamilszostak: Agree. I'll move duration to 'measurements' and check for null as well.

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -84,6 +84,15 @@ module Microsoft.ApplicationInsights {
             // initialize event timing
             this._eventTracking = new Timing("trackEvent");
             this._eventTracking.action = (name?: string, url?: string, duration?: number, properties?: Object, measurements?: Object) => {
+                if (!measurements) {
+                    measurements = { duration: duration };
```
MaxShehovtsov: I'm actually not sure about measurements - where will we see it in the portal? (I know that custom dimensions show up below TOP N properties on event details blade). I understand it may be even better than properties if you could do charts based on it (and if dimensions don't support charting). Have you thought about it?

JavaScript/JavaScriptSDK/appInsights.ts
```diff
@@ -237,7 +246,7 @@ module Microsoft.ApplicationInsights {
         }
         
         /**
-         * Start timing an extended event. Call {@link stopTrackEvent} to log the event when it ends.
+         * Start timing an extended event. Call {@link stopTrackE-vent} to log the event when it ends.
```
MaxShehovtsov: You introduced a breaking change into this comment. :)

kamilszostak: Where did that came from?! Fixed.

JavaScript/JavaScriptSDK/appInsights.ts
SergeyKanzhelev: You can build a chart based on measurements. Worth checking anyway

MaxShehovtsov:
:shipit:

Issue #59, opened 2015-08-28
Api Reference: remove autoCollectErrors, add disableExceptionTracking


MaxShehovtsov:
autoCollectErrors seems to not exist in the SDK, however the docs say that it controlls automatic exception tracking (which actually does not).
On other hand the disableExceptionTracking setting is the one that controls automatic exception collection, and it's not mentioned in the api reference.

MaxShehovtsov:
The proposal is to replace autoCollectErrors with disableExceptionTracking in the API-reference

SergeyKanzhelev:
Make sure to make it consistent with `autoTrackAjax`. Maybe it also requires renaming

MaxShehovtsov:
RE 'autoTrackAjax', i think it's better to rename it to 'disableAjaxTracking' - @AlexBulankou @SergeyKanzhelev what do you guys think? That highlights the intention to collect ajaxes by default.

RE 'disableExceptionTracking' - i'd keep it as it because:
1) It's used in snippet, and we don't want to change the snippet
2) it's consistent with 'disableTelemetry'

AlexBulankou:
I agree best to keep consistent with existing disableX. Thanks!

PR #91, opened 2015-12-08
rename config params

MaxShehovtsov:
…cking instead of autoCollectAjax

Issue #59

### DIFF
--- a/JavaScript/JavaScriptSDK/AppInsights.ts
+++ b/JavaScript/JavaScriptSDK/AppInsights.ts
@@ -24,13 +24,13 @@ module Microsoft.ApplicationInsights {
         maxBatchSizeInBytes: number;
         maxBatchInterval: number;
         enableDebug: boolean;
-        autoCollectErrors: boolean;
+        disableExceptionTracking: boolean;
         disableTelemetry: boolean;
         verboseLogging: boolean;
         diagnosticLogInterval: number;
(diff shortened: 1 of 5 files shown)


JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -109,7 +109,7 @@ module Microsoft.ApplicationInsights {
             this._pageVisitTimeManager = new ApplicationInsights.Telemetry.PageVisitTimeManager(
                 (pageName, pageUrl, pageVisitTime) => this.trackPageVisitTime(pageName, pageUrl, pageVisitTime));
 
-            if (this.config.autoTrackAjax) { new Microsoft.ApplicationInsights.AjaxMonitor(this); }
+            if (!this.config.disableAjaxTracking) { new Microsoft.ApplicationInsights.AjaxMonitor(this); }
```
Dmitry-Matveev: I would expect the same change (from "this..." to "!this...") for this.config.disableExceptionTracking value somewhere in the code.

MaxShehovtsov: It lives here (https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/snippet.js#L41) and doesn't need a change. Thanks!

PR #97, opened 2015-12-15
adding WebApplication.targets as nuget package, and changing references

bengreenier:
choosing to reference WebApplication from nuget package, as this is more likely to be stable across different vs versions, and different vs installations (apparently not everyone has WebApplication.targets by default, my vs2015 ultimate installation did not).


bengreenier:
for reference, i'm full-time ms, so no CLA is needed.

AlexBulankou:
Closing stale PR

Issue #99, opened 2015-12-17
Internal trace: AI (Internal): client performance math error:1 < 0 + 0 + 2579 + 0


AlexBulankou:
Seeing high occurrence of this message

AlexBulankou:
Issue started appearing on 12/3
![image](https://cloud.githubusercontent.com/assets/11968377/11860006/44e0a05c-a425-11e5-9ef7-e29fe3c3768c.png)

AlexBulankou:
For the record these are the percentage of page views affected by each of the warnings that I'm making non-reporting with this change:

![image](https://cloud.githubusercontent.com/assets/11968377/11878722/57ba2108-a4ac-11e5-9bf7-f75409937599.png)

Ruud-cb:
Hi
Is there anything to do to fix this issue other then that it is a warning? Or can it be safely ignored?
I am currently not sure if, when this exception occurs on a logging event, it prevents futher execution of the process. Should every AI logging call be try-catched for this? Kind of an annoying one..

PR #100, opened 2015-12-17
Making messages reporting incorrect page performance warning-only

AlexBulankou:
This means they won't be sent to customer ikey as trace messages and instead will just appear in browser console. These are non-actionable errors for the ikey owner as we know that certain browser either don't support or incorrectly implement page performance API. This is to address #99


madeleinefletcher:
Approved

PR #102, opened 2015-12-17
Fix for bug 98: Object doesn't support property or method 'substring'

madeleinefletcher:


JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
```diff
@@ -77,7 +77,7 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
             if (value) {
                 value = Util.trim(value);
                 if (value.toString().length > DataSanitizer.MAX_STRING_LENGTH) {
-                    value = value.substring(0, DataSanitizer.MAX_STRING_LENGTH);
+                    value = value.toString().substring(0, DataSanitizer.MAX_STRING_LENGTH);
```
AlexBulankou: substring function takes start and *not-exclusive* end, so in this case it should actually be substring(0, DataSanitizer.MAX_STRING_LENGTH+1).
Overall I would rather switch it here and in other places to use substr, which behaves as you would expect - takes start_index and length.

madeleinefletcher:
Looks like there is no difference between substr and substring, so I am going to leave it

JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
madeleinefletcher: Looks like there's no difference between substr and substring in the way they behave here, so I'm going to leave it.

AlexBulankou: sorry my bad, I cannot count

AlexBulankou:
:shipit:

PR #103, opened 2015-12-17
Made ajax dependencies send url as name

madeleinefletcher:
Link to bug 484887 in VSO

### DIFF
--- a/JavaScript/JavaScriptSDK/Util.ts
+++ b/JavaScript/JavaScriptSDK/Util.ts
@@ -397,5 +397,15 @@ module Microsoft.ApplicationInsights {
 
             return result;
         }
+
+        public static getPathName(url): string {
+            var result: string;
+            var a = UrlHelper.parseUrl(url);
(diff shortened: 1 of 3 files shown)


AlexBulankou:
Looks :shipit: to me, but let's have @MaxShehovtsov double check

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -397,5 +397,15 @@ module Microsoft.ApplicationInsights {
 
             return result;
         }
+
```
MaxShehovtsov: Should getHostName be removed (or is it still used)?

madeleinefletcher: I think it can be removed. do you want me to submit a pull request for that?

MaxShehovtsov: that would be great, let's try to avoid the dead code :)

PR #107, opened 2016-01-05
Fixing ajaxes with negative duration

MaxShehovtsov:


JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -177,27 +178,24 @@ module Microsoft.ApplicationInsights {
...
-            }
-        }
-
         private onAjaxComplete(xhr: XMLHttpRequestInstrumented) {
-            try {
+            xhr.ajaxData.responseFinishedTime = dateTime.Now();
+            xhr.ajaxData.status = xhr.status;
                 xhr.ajaxData.CalculateMetrics();
 
+            if (xhr.ajaxData.ajaxTotalDuration < 0) {
```
SergeyKanzhelev: what's the root cause for this? Is it something close to zero and it's a precision problem or something more critical? Could we just report zero or one if it's a precision problem?

MaxShehovtsov: We don't know the root cause yet. That's why I'm adding extra telemetry (to see if it's dates problem). Just to get one step closer to understanding. 
Also preventing negative durations as it messes up overall picture.

SergeyKanzhelev: Tracing will give you requestURL, start time and stop time. Is it correct that start time and request URL are already sent with the ajax event itself? Am I missing something?

MaxShehovtsov: No - we won't send ajax at all in case of negative duration. In this case we will send this new trace.

SergeyKanzhelev: Investigation started from the fact that we saw dependencies with negative duration on some applications, correct? Or we used to send negative duration before and do not send it anymore?

MaxShehovtsov: Some time we send negative durations (based on investigation). This fix prevents it (by not sending negative ajaxes). This fix also adds extra telemetry (trace) which should help us understand why we get negative ajaxes.

MaxShehovtsov:
I'll address remaining feedback in future commits - feel free to post any questions/suggestions.

PR #108, opened 2016-01-07
switching to performance.now() for ajax timing

MaxShehovtsov:
Per http://stackoverflow.com/questions/313893/how-to-measure-time-taken-by-a-function-to-execute 

Our telemetry indicates that ajax start time sometimes is bigger than finish time.


MaxShehovtsov:
Feel free to comment after PR is closed!

PR #109, opened 2016-01-11
Cors fix

MaxShehovtsov:
Simplifying the criteria for CORS exception


AlexBulankou:
Best to use ===  comparisons as good practice even when not strictly required :)
:shipit:

PR #111, opened 2016-01-12
Ajax - client side throttling per view

MaxShehovtsov:
75th percentile of our users have 15 ajaxes per page view. To prevent quota cap setting limit of ajaxes per view to 20.

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Initialization.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Initialization.tests.ts
@@ -28,7 +28,8 @@ class InitializationTests extends TestClass {
             autoTrackPageVisitTime: false,
             samplingPercentage: 33,
             disableAjaxTracking: true,
-            overridePageViewDuration: false
+            overridePageViewDuration: false,
+            maxAjaxCallsPerView: 44
         };
 
         // set default values
@@ -69,7 +70,8 @@ class InitializationTests extends TestClass {
                     disableTelemetry: undefined,
                     verboseLogging: undefined,
                     diagnosticLogInterval: undefined,
-                    samplingPercentage: undefined
+                    samplingPercentage: undefined,
+                    maxAjaxCallsPerView: undefined
                 };
 
                 var snippet = <Microsoft.ApplicationInsights.Snippet> {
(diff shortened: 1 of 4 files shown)


SergeyKanzhelev:
How will it affect SPA?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Initialization.tests.ts
```diff
@@ -28,7 +28,8 @@ class InitializationTests extends TestClass {
             autoTrackPageVisitTime: false,
             samplingPercentage: 33,
             disableAjaxTracking: true,
-            overridePageViewDuration: false
+            overridePageViewDuration: false,
+            maxAjaxCallsPerView: 44
```
SergeyKanzhelev: It should be 42

JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -250,11 +257,18 @@ module Microsoft.ApplicationInsights {
...
+            if (this.config.maxAjaxCallsPerView === -1 ||
+                this._ajaxCallsSent < this.config.maxAjaxCallsPerView) {
+                var dependency = new Telemetry.RemoteDependencyData(absoluteUrl, pathName, totalTime, success, resultCode);
+                var dependencyData = new ApplicationInsights.Telemetry.Common.Data<ApplicationInsights.Telemetry.RemoteDependencyData>(
+                    Telemetry.RemoteDependencyData.dataType, dependency);
+                var envelope = new Telemetry.Common.Envelope(dependencyData, "Microsoft.ApplicationInsights." + this.config.instrumentationKey.replace(/-/g, "") + ".RemoteDependency");
+                this.context.track(envelope);
+                ++this._ajaxCallsSent;
+            } else {
+                _InternalLogging.throwInternalUserActionable(LoggingSeverity.CRITICAL,
```
SergeyKanzhelev: Include page name to the message. Also including pageView.id would be great if we can somehow show this trace message to be related to that pageview. Maybe even show it on the pageview details blade.

Also - would it make sense to only send it once per pageview? Otherwise we send as many traces as ajaxes were on the page.

MaxShehovtsov: Great comments, thanks!

JavaScript/JavaScriptSDK.Tests/CheckinTests/Initialization.tests.ts
MaxShehovtsov: Not a ship blocker though :)

MaxShehovtsov:
@SergeyKanzhelev I think SPAs will still have track page view. Also this logic is configurable, and our main concern is to not break anyone's monitoring with quota cap

PR #112, opened 2016-01-15
supported filtering out items with telemetry initializers

MaxShehovtsov:
Don't want an item to be sent? Return false from telemetry initializer and you're all set!

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
@@ -29,6 +29,7 @@ class TelemetryContextTests extends TestClass {
 
     /** Method called after each test method has completed */
     public testCleanup() {
+        (<any>this._telemetryContext).telemetryInitializers = undefined;
     }
 
     public registerTests() {
@@ -242,7 +243,7 @@ class TelemetryContextTests extends TestClass {
 
                 // teardown
                 
-                (<any>this._telemetryContext).telemetryInitializers = undefined;
+                
             }
         });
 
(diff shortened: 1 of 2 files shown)


SergeyKanzhelev:
From design perspective:
- Would it be more natural to return telemetry item that will be passed to the next telemetry initializer? It will be more like telemetry processors in dotnet.
- If not - would it be more natural for javascript developer to have a behavior of onerror handler. `true` means prevent passing to the default handler, `false` - otherwise. I'm not sure what is a typical pattern being used in other cases in JavaScript, but this pattern seems to be well-known.

MaxShehovtsov:
1) Before this fix telemetry initializers did not return modified item. I don't want to change signature and break someone's code.
2) For me returning 'false' looks perfectly fine (comparing to return 'true' or null or undefined or -1). False means 'no' (while true means 'yes') :-)

SergeyKanzhelev:
Even if you think that the design of `onerror` handler is bad - consistency is better than non-consistency. I don't know what is the standard pattern in JavaScript so if you can give other examples when false means do not proceed - it will be fine with me. I just want to make sure it looks/feels OK for JavaScript developer.

MaxShehovtsov:
I don't see consistency in browser's api, but I do see that people rely on docs heavily. So for me it's the quality of our docs that matters.
Specifically for onerror - I don't see much value in making our initializers consistent with it. And my subjective feeling (see above - true vs false) is that false better describes 'disable sending'.

AlexBulankou:
@SergeyKanzhelev if you're looking for consistency example, returning ```false``` from jQuery event handler has the effect of calling ```stopPropagation()``` (stops the event from bubbling up) and ```preventDefault()``` - prevents default action from occurring. For non-jQuery event handlers returning false prevents default event from occurring. So I think @MaxShehovtsov 's implementation is consistent at least with this pattern.

SergeyKanzhelev:
makes sense than. It means that `onerror` is not standard, not all other event handlers =)

PR #113, opened 2016-01-15
Reliability - added message about data loss

MaxShehovtsov:


JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -27,6 +27,65 @@ declare var XDomainRequest: {
...
+            }
+        }
+
+        static reportLostItems() {
+            try {
+                if (DataLossAnalyzer.enabled && DataLossAnalyzer.appInsights && DataLossAnalyzer.getNumberOfLostItems() > 0) {
+                    DataLossAnalyzer.appInsights.trackTrace(
+                        "AI (Internal): Not all telemetry items were sent from the previous page. Please set maxBatchSizeInBytes and/or maxBatchInterval to increase sending frequency. Count of missing items: "
+                        + DataLossAnalyzer.getNumberOfLostItems()
+                        , null);
```
AlexBulankou: I don't think we should be making this an action item for the customer to change maxBatchSizeInBytes and maxBatchInterval - if I saw this I would be frightened and slightly annoyed I should be tweaking it and it doesn't just work out of the box :) I think we should send it as internal info for our own analysis only, without the message.

MaxShehovtsov:
@AlexBulankou @SergeyKanzhelev - Added split testing mechanism to enable data loss analysis on N% of ikeys.
Also fixed the guid creation algorithm because of negative effect it makes on sampling logic.

JavaScript/JavaScriptSDK/SplitTest.ts
```diff
@@ -0,0 +1,22 @@
...
+    //   if (new SplitTest.isEnabled(<user id>, <percent of users to enable feature for>)){
+    //     newShinyFeature.enabled = true;
+    //   }
+    //
+    export class SplitTest {
+        private hashCodeGeneragor: HashCodeScoreGenerator = new HashCodeScoreGenerator();
+
+        public isEnabled(key: string, percentEnabled: number) {
+            return this.hashCodeGeneragor.getHashCodeScore(key) < percentEnabled;
+        }
```
AlexBulankou: so getHashCodeScore() returns numbers from 0 to 99 inclusive? (so percentEnabled can be from 0 (disabled) to 100 (enabled for everyone))

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -227,18 +227,19 @@ module Microsoft.ApplicationInsights {
...
-            var base64chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/';
-
-            var result = "";
-            var random = Math.random() * 1073741824; //5 symbols in base64, almost maxint
-
-            while (random > 0) {
-                var char = base64chars.charAt(random % 64);
-                result += char;
-                random = Math.floor(random / 64);
+        public static newId() {
```
AlexBulankou: do we need a test to ensure guid format is correct (so we don't regress?)

JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -130,6 +133,8 @@ module Microsoft.ApplicationInsights {
                 (pageName, pageUrl, pageVisitTime) => this.trackPageVisitTime(pageName, pageUrl, pageVisitTime));
 
             if (!this.config.disableAjaxTracking) { new Microsoft.ApplicationInsights.AjaxMonitor(this); }
+            
+            DataLossAnalyzer.reportLostItems();
```
AlexBulankou: for itemQueued function we are checking in DA is enabled in the caller code and for reportLostItems we are checking inside the function. Do we care if this consistent?

AlexBulankou:
:shipit: with suggestions :)

JavaScript/JavaScriptSDK/Util.ts
MaxShehovtsov: we could, although all we care is that the id is a relatively long string

JavaScript/JavaScriptSDK/SplitTest.ts
MaxShehovtsov: yes, plus it's a floating point (so 0 to 99 inclusive, with all the precision)

MaxShehovtsov:
@AlexBulankou @SergeyKanzhelev Added an extra data volume limitation - no more than 10 data loss messages per browser session. This should be enough to make a data loss analysis.

PR #114, opened 2016-01-20
Fix allows user to submit 0 custom duration

madeleinefletcher:
Overriding duration with a duration of 0 will send page view with duration 0 now instead of the value of customDuration.


madeleinefletcher:
@MaxShehovtsov

MaxShehovtsov:
:+1:

PR #117, opened 2016-01-22
Hash code algorithm - min 8 chars input string

MaxShehovtsov:
Existing algorithm for creating userid/operationid produces 5 char strings (much shorter than it used to be with guids). This seems to screw up sampling algorithm (DJB hash) which seems to not work properly on short strings.

This affects both client AND server sampling (as user ids get created on client and passed to server SDK)

Proposed solution is to guarantee minimum 8 chars input string. Based on experiments it produces an acceptable random distribution.


SergeyKanzhelev:
:shipit:

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -227,19 +227,18 @@ module Microsoft.ApplicationInsights {
         /**
          * generate GUID
```
AlexBulankou: Update comment, no longer a guid format?

PR #122, opened 2016-01-26
Removed ajax trace message as per bug 505238

madeleinefletcher:


MaxShehovtsov:
I think we still should trace with lower severity.

PR #123, opened 2016-01-26
Changed severity of ajax message as per bug 505238

madeleinefletcher:


madeleinefletcher:
I will incorporate this change in my previous PR, since I was already changing the logging messages anyway

PR #124, opened 2016-01-26
Leap year testing

kamilszostak:


JavaScript/JavaScriptSDK/Context/User.ts
```diff
@@ -95,7 +95,7 @@ module Microsoft.ApplicationInsights.Context {
                 var acqStr = Util.toISOStringForIE8(date);
                 this.accountAcquisitionDate = acqStr;
                 // without expiration, cookies expire at the end of the session
-                // set it to a year from now
+                // set it to 365 days from now
```
MaxShehovtsov: Why is this change necessary?

kamilszostak: It's not necessary and it's really low priority. On leap year we should add 366 days to make a full year. 
I've decided to update the comment, so when I'll be doing another leap year scan in 4 years there is not confusion what the code is supposed to do.

Issue #125, opened 2016-01-28
Maximum ajax per page view limit reached message is sent on every page view


AlexBulankou:
We are currently sending this trace message once per view
```
else if (this._trackAjaxAttempts === this.config.maxAjaxCallsPerView) {                
 -                _InternalLogging.throwInternalUserActionable(LoggingSeverity.CRITICAL,		 +                _InternalLogging.throwInternalUserActionable(LoggingSeverity.CRITICAL, new _InternalLogMessage(
 -                    "Maximum ajax per page view limit reached, ajax monitoring is paused until the next trackPageView(). In order to increase the limit set the maxAjaxCallsPerView configuration parameter.");		 +                    "Maximum ajax per page view limit reached, ajax monitoring is paused until the next trackPageView(). In order to increase the limit set the maxAjaxCallsPerView configuration parameter."));
```
I think we should send it once per session. @MaxShehovtsov's notes that this will make it harder to diagnose, however since the message is intended for application author (the one who is looking at telemetry in the portal) and not for end-user, once per session will be enough for application author to diagnose the issue.
So I would like to add a flag to sessionStorage once we send this trace message.

PR #126, opened 2016-01-28
Albulank/limit trace messages

AlexBulankou:
Fix for #125


MaxShehovtsov:
Actually - I thought about it. I'm not sure we want to make this change. Right now users will get 1 trace per page view (so overhead is low). In return they'll get clear diagnostics about why there's no ajaxes reported. I'm not sure it's easy to discover this 1-per-session trace in a huge pile of other traces users might have.
A better long term fix would be to align all internal messages to be 1 per session. This will make them have equal weight in customer's eyes, therefore easier to understand.

Since short term gain is minimal and it makes it harder to troubleshoot, I would not make this change.

AlexBulankou:
I agree with making all diagnostic messages to be one per session. I'll just make this fix instead.

MaxShehovtsov:
I think just making this fix will make things inconsistent.

AlexBulankou:
I meant I will make the fix that you suggested. I should have said "that fix", not "this fix" :)

MaxShehovtsov:
:+1:

Issue #127, opened 2016-02-02
WireDep Issues


jasteph:
I'm using WireDep to automatically grab all of my bower packages and inject them into my application. https://github.com/taptapship/wiredep

It is not finding this library. They require that an application define its dependencies and have a main property in it's bower.json file. I tried hardcoding these, but have been unsuccessfuly in getting it to recognize.

Has anyone else came across this issue before?

laughin1:
Was able to get it working with overriding it in Wiredep's config.

jasteph:
Was able to get it working with overriding it in Wiredep's config.

MaxShehovtsov:
I'm glad you figured this out! Would you mind sharing the fix?

jasteph:
In my own bower.json I added an 'overrides' object:

```
  "overrides": {
    "applicationinsights-js": {
      "main": "dist/ai.js"
    }
  }
```

MaxShehovtsov:
Thank you!

PR #128, opened 2016-02-02
adding main to bower.json to help bower tools

jasteph:
It's common practice for third party tools to use the main object in bower.json to detect which are the distribution files. 

I encountered an issue where WireDep was not recognizing AppInsights #127, this fixes the problem.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "applicationinsights-js",
-  "main": "dist/ai.js",
+  "main": "dist/ai.0.js",
   "version": "0.21.5",
   "homepage": "https://github.com/Microsoft/ApplicationInsights-JS",
   "authors": [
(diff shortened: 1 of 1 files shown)


MaxShehovtsov:
@jasteph  I wonder if ai.0.js (the minified version of the script) is more appropriate here - what do you think?

jasteph:
That makes sense as it is 'distribution'. Will change.

jasteph:
Is there a reason why the minified version is named ai.0.js instead of ai.min.js?

MaxShehovtsov:
I believe the idea was to version scripts (like, future versions would have ai.1.js, etc). Right now we're still on 0.xxx, and when we switch to 1.0 we'll most probably rename it to ai.1.js.

MaxShehovtsov:
Thank you @jasteph

PR #129, opened 2016-02-06
Client/server correlation

MaxShehovtsov:
In this PR - setting x-ms-request-id and x-ms-root-id headers to each ajax call. 
Applies on to ajax calls to the same host to prevent CORS issues.
Making sure ajax telemetry gets correct ctx.id, ctx.parentid and rdd.id.

Assumptions: 
- in current implementation a page is considered a root (we might implement request to page relationship later). 
- Therefore ctx.id (root id) equals ctx.parentid  
- Dependency.ctx.parentid equals ctx.id (root id) (and equals ctx.parentid)

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Initialization.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Initialization.tests.ts
@@ -30,7 +30,8 @@ class InitializationTests extends TestClass {
             disableAjaxTracking: true,
             overridePageViewDuration: false,
             maxAjaxCallsPerView: 44,
-            disableDataLossAnalysis: true
+            disableDataLossAnalysis: true,
+            disableCorrelationHeaders: false
         };
 
         // set default values
--- a/JavaScript/JavaScriptSDK/ajax/ajax.ts
+++ b/JavaScript/JavaScriptSDK/ajax/ajax.ts
@@ -14,8 +14,10 @@ module Microsoft.ApplicationInsights {
         private appInsights: AppInsights;
         private initialized: boolean;
         private static instrumentedByAppInsightsName = "InstrumentedByAppInsights";
+        private currentWindowHost;
 
         constructor(appInsights: Microsoft.ApplicationInsights.AppInsights) {
+            this.currentWindowHost = window.location.host;
             this.appInsights = appInsights;
             this.initialized = false;
             this.Init();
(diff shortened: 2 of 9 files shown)


SergeyKanzhelev:
Looks good for me.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Initialization.tests.ts
```diff
@@ -30,7 +30,8 @@ class InitializationTests extends TestClass {
             disableAjaxTracking: true,
             overridePageViewDuration: false,
             maxAjaxCallsPerView: 44,
-            disableDataLossAnalysis: true
+            disableDataLossAnalysis: true,
+            disableCorrelationHeaders: false
```
AlexBulankou: Why do we need this switch: disableCorrelationHeaders?

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -132,7 +134,11 @@ module Microsoft.ApplicationInsights {
         }
 
         private sendHandler(xhr: XMLHttpRequestInstrumented, content) {
-            xhr.ajaxData.requestSentTime = dateTime.Now();            
+            xhr.ajaxData.requestSentTime = dateTime.Now();
+            if (!this.appInsights.config.disableCorrelationHeaders && (UrlHelper.parseUrl(xhr.ajaxData.getAbsoluteUrl()).host == this.currentWindowHost)) {
+                xhr.setRequestHeader("x-ms-request-root-id", this.appInsights.context.operation.id);
+                xhr.setRequestHeader("x-ms-request-id", xhr.ajaxData.id);
+            }
```
AlexBulankou: Should we have try...catch...log around the whole setting headers part to ensure we don't break ajax with this functionality?

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -132,7 +134,11 @@ module Microsoft.ApplicationInsights {
         }
 
         private sendHandler(xhr: XMLHttpRequestInstrumented, content) {
-            xhr.ajaxData.requestSentTime = dateTime.Now();            
+            xhr.ajaxData.requestSentTime = dateTime.Now();
+            if (!this.appInsights.config.disableCorrelationHeaders && (UrlHelper.parseUrl(xhr.ajaxData.getAbsoluteUrl()).host == this.currentWindowHost)) {
```
AlexBulankou: why have currentWindowHost restriction? Host name may be different even for the same component and going forward we'll look into supporting it across components.

JavaScript/JavaScriptSDK/ajax/ajax.ts
MaxShehovtsov: CORS doesn't allow adding headers, unless configured

JavaScript/JavaScriptSDK.Tests/CheckinTests/Initialization.tests.ts
MaxShehovtsov: Because, like CORS in general, there might be less common cases where adding our headers will break customer's server side (e.g. if they whitelist headers for security reason).
My plan is once we roll it out (and support in UI) we'll validate it first and then turn it ON by default.

JavaScript/JavaScriptSDK/ajax/ajax.ts
MaxShehovtsov: CORS doesn't allow adding headers, unless configured

AlexBulankou:
Thanks for clarifying.

PR #140, opened 2016-03-08
Add iKey to telemetry name

abaranch:
Microsoft.ApplicationInsights.Event -> Microsoft.ApplicationInsights.iKey.Event
And same for all other types


SergeyKanzhelev:
:shipit: thanks!

PR #141, opened 2016-03-10
Disabling diagnostic message reporting for a couple of known scenarios

AlexBulankou:


MaxShehovtsov:
Did you check in .js/.js.map files by mistake?

MaxShehovtsov:
:Shipit: after removing js/d.ts/jsmaps

Issue #144, opened 2016-03-14
Page crash iOS Safari in Private mode: infinite loop SessionStorage read warning


bgever:
As of a few weeks ago, user reported issue with **iOS Safari in Private mode**. Normal mode works fine.

When loading the page with the AI script, the browser reports eventually in a message bar:
`A problem occured with this webpage so it was reloaded.`
and the page content has only the message:
`A problem repeatedly occurred on "<URL>".`.
    

Reproduced with Browserstack real browser on various devices and on my own iPhone.

The console log has repeatedly this message (up to 6400 times):

`ai.0.js:1 AI (Internal): NONUSRACT_BrowserCannotWriteSessionStorage message:"Browser failed write to session storage. " props:"{exception:[object DOMException]{\code\:22,\name\:\QuotaExceededError\,\message\:\QuotaExceededError: DOM Exception 22\,\line\:1,\column\:16686,\sourceURL\:\https://az416426.vo.msecnd.net/scripts/a/ai.0.js\}}"`

[When Safari is in Private mode it doesn't allow to read session storage.](http://stackoverflow.com/questions/21159301/quotaexceedederror-dom-exception-22-an-attempt-was-made-to-add-something-to-st)

[The AI code has a try/catch around that](https://github.com/Microsoft/ApplicationInsights-JS/blob/31aacddd1cb8852af0bc3b2ddcd3b4ae71f15c94/JavaScript/JavaScriptSDK/Util.ts#L186), but somehow the attempt is repeated.

![ai safari private infinite loop](https://cloud.githubusercontent.com/assets/1618765/13740127/f4f566c4-ea09-11e5-8b3a-3788b5320b31.PNG)

tsj0603:
I also have the issue for my web site.

For Safari in private browsing the browser hands out the window.localStorage objet but throws an exception when using the SetItem member.

The problem is in Util.ts. The _getStorageObject and _getSessionStorageObject functions return the storage object but when the write functions use the storage object it throws an exception causing the write functions to return false. This causes the caller loop.

The _getStorage*Object functions should use a pattern where they validate the storage object by performing a test write operation before handing out the object. This is a common pattern - e.g. see https://mathiasbynens.be/notes/localstorage-pattern.

bgever:
Hope the AI team can review the PR which should fix this issue. I can't run the unit tests locally, the PS script opens an HTML page with empty unit tests. I can test the results via Browserstack.

bgever:
[...]

PR #145, opened 2016-03-14
Test getting sessionStorage in case private browsing throws an error

bgever:
Issue #144 details how Private browsing in iOS Safari causes the page to crash due to an infinite loop of console warnings due to session storage being inaccessible.

This PR fixes the issue by adding an additional check to the code that validates whether the `sessionStorage` object is available. 

Private browsing in iOS Safari does provide a sessionStorage object, but it throws an error when getting or setting an item. Therefore the current availability check isn't sufficient.

By getting the empty string key during the validation, it would cause an error to throw, which is caught by the existing try/catch that detects whether session storage is available. The getter would otherwise return null or the value for the empty string key (shortest), which is unlikely to be defined.

I was unable to run the unit tests locally. No tests are detected when I run the PS script.

### DIFF
--- a/JavaScript/JavaScriptSDK/Util.ts
+++ b/JavaScript/JavaScriptSDK/Util.ts
@@ -110,6 +110,8 @@ module Microsoft.ApplicationInsights {
         private static _getSessionStorageObject(): Storage {
             try {
                 if (window.sessionStorage) {
+                    // Test getting items in case private browsing throws an error.
+                    window.sessionStorage.getItem('');
                     return window.sessionStorage;
                 } else {
                     return null;


MaxShehovtsov:
Thanks for the PR! (Just to make sure) Did you check that it fixes the issue?
Also here's a summary on how to run tests - https://github.com/Microsoft/ApplicationInsights-JS/wiki/Run-tests-from-Visual-Studio

bgever:
@MaxShehovtsov I have not been able to verify. Do you think I can run the tests on my Mac standalone?

bgever:
Selenium doesn't have a driver to write E2E tests that would work in Private mode. I can only test it manually. In that case I'd need to build the dist package. I can do that tomorrow morning HKT (almost midnight here).

AlexBulankou:
Thanks @bgever! We can run the tests in private mode to verify. I also wanted to suggest that we use storage checks that are suggested in the post that you linked: 

```
// Feature detect + local reference
var storage;
var fail;
var uid;
try {
  uid = new Date;
  (storage = window.localStorage).setItem(uid, uid);
  fail = storage.getItem(uid) != uid;
  storage.removeItem(uid);
  fail && (storage = false);
} catch (exception) {}
```
so we can do these additional changes and run tests in private mode before submitting the fix

AlexBulankou:
These are the changes including sessionStorage/localStorage checks refactored. Ran tests in private mode #146

bgever:
To note, the issue happens with private browsing in *Safari only*. iOS shows the message bar as in the screen shot. Mac OS X does not show the message bar, but a spinning beachball (hourglass) and sometimes hangs. Chrome's Incognito on both Windows and Mac works fine. IE & Edge InPrivate both work fine too.

bgever:
Great to see a fix has been implemented in #146, @AlexBulankou. Thanks a lot for the quick follow-up! Closing this PR in favor of #146.

PR #146, opened 2016-03-14
Improving localStorage/sessionStorage checks as suggested by @bgever

AlexBulankou:
Ran tests in private:
image


bgever:
Please note, Chrome's Incognito mode behaves different than Safari's Private mode. If you have an Apple device, you could use [Passport](https://sourceforge.net/projects/pjs-passport/) (works with Win10) to forward the local port to the network and test with an Apple device on the same network.

AlexBulankou:
Thanks! testing it in Safari private mode

AlexBulankou:
I could reproduce the issue using browser stack and then with my fix I can no longer repro. Compare these with Safari in Private mode:
HAS issue: http://albulank-test20160314-2.azurewebsites.net
FIXED: http://albulank-test20160314-3.azurewebsites.net

kamilszostak:
:shipit:

bgever:
Awesome work @AlexBulankou, tested it as well, and fix works as expected.

What would be the ETA for this to land on the CDN, now it's in master?

AlexBulankou:
Tentative ETA is in about 4 hours, will update if there are any changes to this. Thanks!

On March 14, 2016 6:44:00 PM Bart Verkoeijen <notifications@github.com> wrote:

Awesome work @AlexBulankou<https://github.com/AlexBulankou>, tested it as well, and fix works as expected.

What would be the ETA for this to land on the CDN, now it's in master?

-

AlexBulankou:
New JS SDK version was just deployed. Please verify on your end. 
Thank you! - really appreciate your help in tracking this issue down.

bgever:
Confirmed that the issue has been fixed in our production environment using the CDN. Thanks a ton for the super fast resolution of the issue @AlexBulankou!

PR #147, opened 2016-03-14
Set cookie domain

kamilszostak:
Users can now set a custom cookie domain in the JS script snippet. 
For instance: `cookieDomain: ".example.com"`


kamilszostak:
Once this is merged, I will update the documentation.

JavaScript/JavaScriptSDK/Initialization.ts
```diff
@@ -157,6 +157,7 @@ module Microsoft.ApplicationInsights {
             config.emitLineDelimitedJson = Util.stringToBoolOrDefault(config.emitLineDelimitedJson);
             config.diagnosticLogInterval = config.diagnosticLogInterval || 10000;
             config.autoTrackPageVisitTime = ApplicationInsights.Util.stringToBoolOrDefault(config.autoTrackPageVisitTime);
+            config.cookieDomain = config.cookieDomain;
```
AlexBulankou: Probably didn't mean this?

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -220,8 +220,14 @@ module Microsoft.ApplicationInsights {
         /**
          * helper method to set userId and sessionId cookie
          */
-        public static setCookie(name, value) {
-            Util.document.cookie = name + "=" + value + ";path=/";
+        public static setCookie(name, value, domain?) {
+            var domainAttrib = "";
+
+            if (domain) {
+                domainAttrib = ";domain=" + domain;
```
AlexBulankou: In case domain is not specified should we default to top level domain of the place where the code is executing. So, do something like this: 

`new RegExp(/[-\w]+\.(?:[-\w]+\.xn--[-\w]+|[-\w]{3,}|[-\w]+\.[-\w]{2})$/i).exec(window.location.host)`
I got the regex from here (http://stackoverflow.com/a/6449501). 
Thinking more about it, this would be a bit too risky - we might break a lot of user counting if we get this wrong.

AlexBulankou:
:shipit:

JavaScript/JavaScriptSDK/Initialization.ts
kamilszostak: I'll remove this. I'm also wondering why we have a similar code on line 146.

JavaScript/JavaScriptSDK/Util.ts
SergeyKanzhelev: Not sure how it will be configurable, but we should allow do `if` logic by `location.host` so the same script will work in dev/test and production domain. And also will work when you have two domains like `abc.foo.com` and `cde.foo.ru`

kamilszostak: If the `cookieDomain` is not specified we will use a default domain, the same way as it is currently done and it will work in most cases. 

I wouldn't add any code here to guess the domain.

AlexBulankou: @SergeyKanzhelev with this PR, the customer can specify foo.com in the setting and it will apply to both site1.foo.com and site2.foo.com. This is the use case - allowing cookie to work across subdomains.
Allowing cookie to work across two arbitrary domains without common top level parent (like in your example abc.foo.com and cde.foo.ru) is not the use case for this fix and I don't think it will be technically feasible if it was a use case.
Configuration of different environment be done by adding to the snippet (I image it will be a mix of server and client code) something like this:
`cookieDomain = @Config.CookieDomain;`
or you can have if..else logic here too

SergeyKanzhelev: that's what I meant. Just make sure we can allow if statement in the place where config is defined. Like 
```
if (host.location.endWith("foo.ru")) cookieDomain="foo.ru"
```

Based on your reply it should be totally possible today

kamilszostak: Yes, user can use an inline function to dynamically set the cookie domain. Sample code: 

 ```
           instrumentationKey: "<your-key>",
            cookieDomain: (function () { return 1 === 2 ? "check your math!" : "we're good" })()
```

SergeyKanzhelev:
:shipit:

PR #153, opened 2016-03-15
Only calling window.localStorage and window.sessionStorage after safe checks

AlexBulankou:
Fix for #152


AlexBulankou:
Ran tests successfully under service account

kamilszostak:
:shipit:

Issue #135, opened 2016-02-22
IE10 - Telemetry failing to send


forktrucka:
No page view or event data is being returned by my site when viewed from IE10. Requests are still being reported.

I have noticed that when using IE10, telemetry is failing to send. You get similar issues to below reported. Tested in IE8, IE9,IE11, latest Chrome and firefox.

Custom lib location:

Telemetry transmission failed, some telemetry will be lost: [object Error]{ stack: 'Error: Access is denied.

   at _xhrSender (https://[blah].azurewebsites.net/scripts/lib/ai.0.15.0-build58334.min.js:1:22545)
   at triggerSend (https://[blah].azurewebsites.net/scripts/lib/ai.0.15.0-build58334.min.js:1:22215)
   at Anonymous function (https://[blah].azurewebsites.net/scripts/lib/ai.0.15.0-build58334.min.js:1:21643)', message: 'Access is denied.
', name: 'Error' 

No lib specified:

AI (Internal): Telemetry transmission failed, some telemetry will be lost: Error properties: {"exception":"[object Error]{ stack: 'Error: Access is denied.\r\n\n   at open (https://az416426.vo.msecnd.net/scripts/a/ai.0.js:1:11217)\n   at _xhrSender (https://az416426.vo.msecnd.net/scripts/a/ai.0.js:1:34476)\n   at triggerSend (https://az416426.vo.msecnd.net/scripts/a/ai.0.js:1:33992)\n   at flush (https://az416426.vo.msecnd.net/scripts/a/ai.0.js:1:69470)\n   at Anonymous function (https://az416426.vo.msecnd.net/scripts/a/ai.0.js:1:58106)', message: 'Access is denied.\r\n', name: 'Error'"} 

I'm using a plain VS 2015 - ASP.NET MVC template site and hosting in Azure.

SergeyKanzhelev:
Dup for https://github.com/Microsoft/ApplicationInsights-JS/issues/76 Please check out response there

forktrucka:
Have tried the solution outlined in the other issue. flush doesn't work anywhere it seems.

var appInsights=window.appInsights||function(config){
[...]

PR #154, opened 2016-03-15
Switch to disable 'house keeping'

MaxShehovtsov:
Disables housekeeping in window.onBeforeUnload. In some cases IE throws an error when you try to send ajax request on beforeunload. The flag 'disableFlushOnBeforeUnload' is intended to help users cope with errors they get in this case.

### DIFF
--- a/JavaScript/JavaScriptSDK/Initialization.ts
+++ b/JavaScript/JavaScriptSDK/Initialization.ts
@@ -114,7 +114,7 @@ module Microsoft.ApplicationInsights {
         public addHousekeepingBeforeUnload(appInsightsInstance: AppInsights): void {
             // Add callback to push events when the user navigates away
 
-            if ('onbeforeunload' in window) {
+            if (!appInsightsInstance.config.disableFlushOnBeforeUnload && ('onbeforeunload' in window)) {
                 var performHousekeeping = function () {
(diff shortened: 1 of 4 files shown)


MaxShehovtsov:
#135

JavaScript/JavaScriptSDK/Initialization.ts
```diff
@@ -169,6 +169,10 @@ module Microsoft.ApplicationInsights {
             config.disableCorrelationHeaders = (config.disableCorrelationHeaders !== undefined && config.disableCorrelationHeaders !== null) ?
                 Util.stringToBoolOrDefault(config.disableCorrelationHeaders) :
                 true;
+
+            config.disableFlushOnBeforeUnload = (config.disableFlushOnBeforeUnload !== undefined && config.disableFlushOnBeforeUnload !== null) ?
+                Util.stringToBoolOrDefault(config.disableFlushOnBeforeUnload) :
```
AlexBulankou: Seems like Util.stringToBoolOrDefault could handle null and undefined cases - would have saved us some typing. Or if you are trying to check for null and undefined why not just check if (config.disableFlushOnBeforeUnload){...}

AlexBulankou:
:shipit: with suggestions

Issue #155, opened 2016-03-15
Can't send exceptions with empty method


MaxShehovtsov:
Currently we may end up with an empty method for a stack frame in exception telemetry item. This may happen if error is thrown inside a global scope.
The suggestion is to report "Global scope" or "Unknown" or something like that as a method name in this case

PR #156, opened 2016-03-15
Set exception method to "<no_method>" if it's empty

MaxShehovtsov:
Our backend rejects empty methods on exception stack.


AlexBulankou:
:shipit:

MaxShehovtsov:
#155

Issue #160, opened 2016-03-21
Should sanitize CommandName for URL and truncate at 2048 symbols


SergeyKanzhelev:
See how .NET doing it https://github.com/Microsoft/ApplicationInsights-dotnet/blob/8051d296c39ab446ddd6d7911cc6124111224c65/src/Core/Managed/Shared/Extensibility/Implementation/Property.cs#L87

PR #163, opened 2016-03-22
Truncating CommandName to 2048 characters

madeleinefletcher:
Issue #160

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Telemetry/RemoteDependency.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Telemetry/RemoteDependency.tests.ts
@@ -38,6 +38,22 @@ class RemoteDependencyTests extends ContractTestHelper {
         });
 
         this.testCase({
+            name: name + "Command name is truncated if too long",
+            test: () => {
+                var urlLength = 2049;
+                var longUrl = "";
+                for (var i = 0; i < urlLength; i++) {
+                    longUrl += "A";
+                }
+
+                var telemetry = new Microsoft.ApplicationInsights.Telemetry.RemoteDependencyData(
+                    RemoteDependencyTests.id, RemoteDependencyTests.name, longUrl, RemoteDependencyTests.totalTime, RemoteDependencyTests.success, RemoteDependencyTests.resultCode);
+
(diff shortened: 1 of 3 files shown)


SergeyKanzhelev:
:shipit: thanks! It will help with our backend processing.
@lukekim - you should enforce this limit.

AlexBulankou:
We should do trim() and truncate() to be consistent with Web SDK and also because it is a good thing to do, currently this is doing just truncate()

madeleinefletcher:
Added trimming to sanitization

JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
```diff
@@ -89,6 +89,7 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
 
         public static sanitizeUrl(url) {
             if (url) {
+                url = Util.trim(url);
                 if (url.length > DataSanitizer.MAX_URL_LENGTH) {
                     url = url.substring(0, DataSanitizer.MAX_URL_LENGTH);
```
MaxShehovtsov: this line of code also looks to be responsible for trimming

madeleinefletcher: Yes, I added that after Alex's comment :)

JavaScript/JavaScriptSDK.Tests/CheckinTests/Telemetry/RemoteDependency.tests.ts
```diff
@@ -38,6 +38,22 @@ class RemoteDependencyTests extends ContractTestHelper {
         });
 
         this.testCase({
+            name: name + "Command name is truncated if too long",
+            test: () => {
+                var urlLength = 2049;
+                var longUrl = "";
+                for (var i = 0; i < urlLength; i++) {
+                    longUrl += "A";
```
dnduffy: Just for fun I wanted to share these alternatives:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat
http://stackoverflow.com/questions/1877475/repeat-character-n-times

AlexBulankou:
:shipit:

dnduffy:
:+1:

PR #169, opened 2016-04-19
Fix a crash in UrlHelper if document isn't defined

berickson1:
Calling createElement() on an empty object causes an exception, instead check if document.createElement is defined

### DIFF
--- a/JavaScript/JavaScriptSDK/Util.ts
+++ b/JavaScript/JavaScriptSDK/Util.ts
@@ -453,7 +453,7 @@ module Microsoft.ApplicationInsights {
 
         public static parseUrl(url): HTMLAnchorElement {
             if (!UrlHelper.htmlAnchorElement) {
-                UrlHelper.htmlAnchorElement = UrlHelper.document.createElement('a');
+                UrlHelper.htmlAnchorElement = !!UrlHelper.document.createElement ? UrlHelper.document.createElement('a'): {};
             }
 
             UrlHelper.htmlAnchorElement.href = url;


JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -453,7 +453,7 @@ module Microsoft.ApplicationInsights {
 
         public static parseUrl(url): HTMLAnchorElement {
             if (!UrlHelper.htmlAnchorElement) {
-                UrlHelper.htmlAnchorElement = UrlHelper.document.createElement('a');
+                UrlHelper.htmlAnchorElement = !!UrlHelper.document.createElement ? UrlHelper.document.createElement('a'): {};
```
MaxShehovtsov: We rely on ability to create <a> to parse url. With this fix we're fixing the exception but as far as I understand we won't be able to parse URL, which means the end-to-end scenario is still broken.
Can you describe how you reproduced it? Did you run the sdk in browser?

MaxShehovtsov:
Thanks for the contribution! Please describe how you repro'd it a bit more, was it a non-browser environment?

berickson1:
Yeah this was repro'd in a non-browser environment.  Everything else accessing document within this SDK seems to be fine due to null checks similar to the one I added

JavaScript/JavaScriptSDK/Util.ts
berickson1: This will result in a url being spat out since it's inspecting the .src property, which in a browser would be parsed and transform.  In a non-browser environment this will just be a no-op and I believe this url should already be a fully-qualified url

MaxShehovtsov: So you're saying if it's non-browser then noop is fine because we don't guarantee it working there/sdk user is responsible for making sure it works.

berickson1: Yes, in a non-browser environment url should be absolute anyways since there would be no relative base address

MaxShehovtsov: I see, I think this fix makes sense then as it helps non-browser and doesn't impact in-browser scenarios. Thanks!

MaxShehovtsov:
#171

PR #175, opened 2016-04-25
SPA page view tracking

gregjhogan:
Suggestion to add new SPA mode (through new config option) which supports changing the operation ID on each page track event (perhaps better than #81)


AlexBulankou:
Thanks for PR! The concern I have with this approach is that often by the time trackPageView is called the new virtual page is in fact already loaded and several other telemetry items (e.g. events) were logged in the context of new page view. In this case you would like to renew operation ID first, possibly track some other telemetry and then call trackPageView (passing duration) when virtual page is fully loaded. 
Here is a question, why would you like to auto-renew operation ID as part of this call for SPAs, why not just let the developer do it themselves at time convenient for them?

gregjhogan:
Just throwing out a non-API change idea with the hope that it would cover at least 80% of users (with autoTrackPageVisitTime enabled).  Sounds like this is a poor solution; I am just throwing out ideas hoping one sticks because I really want a way to support this that isn't so fragile.

The bottom line is right now you have to dig into the AI code to figure out how to make single page applications track page views properly.  I initially did just that (using things not publicly documented) and later found what I had implemented no longer worked because the way an operation ID was generated had changed.  It seems like an easy problem to solve once the best solution has been identified.

PR #180, opened 2016-05-02
version update 0.22.9 -> 0.22.10

kamilszostak:
- unify line endings and set them to CRLF
- bump up version number to 0.22.10


MaxShehovtsov:
:shipit:

PR #181, opened 2016-05-02
Fix encoding and update versionDate field in global.props

kamilszostak:


changeSdkVersion.ps1
```diff
@@ -24,13 +24,18 @@ if (-Not ($version -match "\d+\.\d+\.\d+")) {
 
 # update package.json 
 $packagesJson.version = $version
-$packagesJson | ConvertTo-Json | Out-File $packageJsonPath
+$packagesJson = $packagesJson | ConvertTo-Json 
+$packagesJson = $packagesJson -replace "\\u0026", "&"
```
MaxShehovtsov: why is that needed?

MaxShehovtsov: :shipit:

kamilszostak: Json converter is encoding those characters. I need to revert them the original state.

Issue #69, opened 2015-10-16
Improve transmission reliability


SergeyKanzhelev:
See https://w3c.github.io/beacon/

MaxShehovtsov:
This is an interesting API however it seems to have poor browser support. We could try using it for FF and Chrome only.

AlexBulankou:
Changing the title of this issue, so we track the generic way to address this problem - today we are losing about 10% of all telemetry documents that customers are sending, mostly due to navigating away. One way to address it is to store unsent telemetry items in session or local storage and attempt resending on page transition.

AlexBulankou:
We discussed that this item should include supporting
persisting in session or local storage 
P1. retrying in case user navigates away before sending and comes back
P2. retrying in case of error from DC endpoint
P3. retrying in case of lost connection

oising:
Beacon definitely looks promising. 

http://status.modern.ie/beacon

kamilszostak:
Thanks @oising for mentioning the Beacon API. We will consider using it, but it may not be possible because Beacon API does not provide a response callback.

alippai:
This shouldn't be closed. Beacon API is living standard and it was designed for this behavior.

SergeyKanzhelev:
@AlexBulankou repurposed this issue for another investigation. @AlexBulankou , @kamilszostak - should we open another one specifically for beacon API?

PR #189, opened 2016-05-10
Session storage send buffer

kamilszostak:
Task #69 - Improve transmission reliability

+ Add session storage buffer (disabled by default).


JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
```diff
@@ -0,0 +1,420 @@
...
+/// <reference path="../../JavaScriptSDK/sender.ts" />
+/// <reference path="../../javascriptsdk/appinsights.ts" />
+
+class SendBufferTests extends TestClass {
+
+    private getArraySendBuffer: (emitLineDelimitedJson?: boolean) => Microsoft.ApplicationInsights.ArraySendBuffer;
+    private getSessionStorageSendBuffer: (emitLineDelimitedJson?: boolean) => Microsoft.ApplicationInsights.SessionStorageSendBuffer;
+
+    public testCleanup() {
+        Microsoft.ApplicationInsights.DataLossAnalyzer.enabled = false;
```
dnduffy: You were removing usage of this in a prior pull request but I presume the code is staying for potential future re-use?

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
```diff
@@ -0,0 +1,420 @@
...
+            storeSendBufferInSessionStorage: () => false,
+            endpointUrl: () => null,
+            maxBatchSizeInBytes: () => null,
+            maxBatchInterval: () => null,
+            disableTelemetry: () => null
+        };
+
+        this.getArraySendBuffer = (emitLineDelimitedJson?: boolean) => {
+            if (emitLineDelimitedJson) {
+                config.emitLineDelimitedJson = () => emitLineDelimitedJson;
```
dnduffy: I'm just curious but why are functions being used for simple values, are there instances where there is more logic applied to these kinds of "properties"?

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
```diff
@@ -0,0 +1,420 @@
...
+            name: "ArraySendBuffer: can clear empty buffer",
+            test: () => {
+                // setup
+                var buffer = this.getArraySendBuffer();
+
+                // act
+                buffer.clear();
+
+                // verify
+                Assert.equal(0, buffer.count(), "buffer should be empty");
```
dnduffy: It might be good to also assert that it was empty to start with since that is a starting assumption.

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
```diff
@@ -0,0 +1,420 @@
...
+                // act
+                var payload1 = "{ test: test }";
+                var payload2 = "{ }";
+
+                buffer.enqueue(payload1);
+                buffer.enqueue(payload2);
+                var batch = buffer.batchPayloads();
+
+                // verify
+                Assert.equal("[" + payload1 + "," + payload2 + "]", batch, "invalid batch");
```
dnduffy: I know it is just a test but do traditional JavaScript tricks like ['[', payload1, ',', payload2, ']'].join('') apply in TypeScript or will the "compiler" write out JS code which executes optimally automatically?

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
```diff
@@ -0,0 +1,420 @@
...
+        // SessionStorageSendBuffer tests
+
+        this.testCase({
+            name: "SessionStorageSendBuffer: initialize",
+            test: () => {
+                // setup
+                var buffer = this.getSessionStorageSendBuffer();
+
+                // verify
+                Assert.equal(0, buffer.count(), "new buffer should be empty");
```
dnduffy: I guess this covers my earlier comment about asserting it starts empty.

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
```diff
@@ -0,0 +1,420 @@
...
+            }
+        });
+
+        this.testCase({
+            name: "SessionStorageSendBuffer: can enqueue and clear the buffer",
+            test: () => {
+                // setup
+                var buffer = this.getSessionStorageSendBuffer();
```
dnduffy: If the only difference between this test and an earlier one is this.getArraySendBuffer() vs. this.getSessionStorageSendBuffer() can the rest of the logic be put into a helper method and reused?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -38,12 +38,19 @@ class SenderTests extends TestClass {
...
         };
 
-        this.getSender = () => new Microsoft.ApplicationInsights.Sender(config);
-        this.errorSpy = this.sandbox.spy(Microsoft.ApplicationInsights.Sender, "_onError");
-        this.successSpy = this.sandbox.spy(Microsoft.ApplicationInsights.Sender, "_onSuccess");
+        this.getSender = () => {
+            var sender = new Microsoft.ApplicationInsights.Sender(config);
+
+            this.errorSpy = this.sandbox.spy(sender, "_onError");
+            this.successSpy = this.sandbox.spy(sender, "_onSuccess");
```
dnduffy: Just to make sure I understand, errorSpy and successSpy won't actually be defined until getSender is called?
It appears you want a new sender for each call to getSender in which case it makes sense to define errorSpy and successSpy relative to the new sender.
However, it is strange that errorSpy and successSpy are properties on the calling instance since each call to getSender will change the this. assignments and orphan previous instances.

dnduffy: P.S. If the old way was making global handlers that ran for any Sender instance then that seems safer than redefining the properties on each call for a specific instance.  What happens when two calls come in to getSender and are actively used in parallel?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -327,7 +334,7 @@ class SenderTests extends TestClass {
                 var sender: Microsoft.ApplicationInsights.Sender = this.getSender();
                 sender._sender = () => null;
                 var senderSpy = this.sandbox.spy(sender, "_sender");
-                this.maxBatchSizeInBytes = Microsoft.ApplicationInsights.Serializer.serialize(this.testTelemetry).length * 2;
+                this.maxBatchSizeInBytes = Microsoft.ApplicationInsights.Serializer.serialize(this.testTelemetry).length * 2 + 2; // +2 for "[]"
```
dnduffy: Are [] the only extra characters?  Why aren't they included in the result from calling serialize?

JavaScript/JavaScriptSDK.Tests/Selenium/checkinTests.ts
```diff
@@ -16,6 +16,7 @@
 /// <reference path="../checkintests/PageVisitTimeManager.tests.ts" />
 /// <reference path="../checkintests/logging.tests.ts" />
 /// <reference path="../checkintests/sender.tests.ts" />
+/// <reference path="../checkintests/SendBuffer.tests.ts" />
```
dnduffy: Nice, more tests == more good. :)

JavaScript/JavaScriptSDK/SendBuffer.ts
```diff
@@ -0,0 +1,205 @@
...
+         * Returns the number of elements in the buffer
+         */
+        count: () => number;
+
+        /**
+         * Clears the buffer
+         */
+        clear: () => void;
+
+        getItems: () => string[];
```
dnduffy: Comment?

JavaScript/JavaScriptSDK/SendBuffer.ts
```diff
@@ -0,0 +1,205 @@
...
+                    "[" + this._buffer.join(",") + "]";
+
+                return batch;
+            }
+
+            return null;
+        }
+
+        public markAsSent(payload: string[]) {
+            this.clear();
```
dnduffy: I think I can see a semantic readability value to this method but in practice it isn't really doing what the name implies.

JavaScript/JavaScriptSDK/SendBuffer.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        public clearSent(payload: string[]) {
+            // already cleared
```
dnduffy: What value is being added by this method?

JavaScript/JavaScriptSDK/SendBuffer.ts
```diff
@@ -0,0 +1,205 @@
...
+        public clear() {
+            this._buffer.length = 0;
+            this.setBuffer(SessionStorageSendBuffer.BUFFER_KEY, []);
+        }
+
+        public getItems(): string[] {
+            return this._buffer;
+        }
+
+        public batchPayloads(): string {
```
dnduffy: This method appears to be the same between the two buffer classes, should they share a common abstract base?

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -168,22 +180,17 @@ module Microsoft.ApplicationInsights {
                 // Send data only if disableTelemetry is false
                 if (!this._config.disableTelemetry()) {
```
dnduffy: If it has been disabled why store it in the buffer in the first place?

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -200,14 +207,19 @@ module Microsoft.ApplicationInsights {
...
+
             var xhr = new XMLHttpRequest();
             xhr[AjaxMonitor.DisabledPropertyName] = true;
             xhr.open("POST", this._config.endpointUrl(), isAsync);
             xhr.setRequestHeader("Content-type", "application/json");
-            xhr.onreadystatechange = () => Sender._xhrReadyStateChange(xhr, payload, countOfItemsInPayload);
-            xhr.onerror = (event: ErrorEvent) => Sender._onError(payload, xhr.responseText || xhr.response || "", event);
-            xhr.send(payload);
+            xhr.onreadystatechange = () => this._xhrReadyStateChange(xhr, payload, countOfItemsInPayload);
+            xhr.onerror = (event: ErrorEvent) => this._onError(payload, xhr.responseText || xhr.response || "", event);
```
dnduffy: I just want to confirm that in the above two lines it is intentional to use "payload" directly instead of "batch"?

JavaScript/JavaScriptSDK/SendBuffer.ts
kamilszostak: `ArrayBuffer `and `SessionStorageBuffer `implement the same interface and `markAsSent` and `clearSent` are part of that interface. I know, it doesn't make much sense for the `ArrayBuffer `to implement those methods, but I had to add it to keep the `Sender `code insistent.

JavaScript/JavaScriptSDK/SendBuffer.ts
kamilszostak: This might me an option, but since the code is so simple and it's just a couple lines, I didn't want to make it overly complex.

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -168,22 +180,17 @@ module Microsoft.ApplicationInsights {
...
-                    if (this._buffer.length) {
-                        // compose an array of payloads
-                        var batch = this._config.emitLineDelimitedJson() ?
-                            this._buffer.join("\n") :
-                            "[" + this._buffer.join(",") + "]";
-
+                    if (this._buffer.count() > 0) {
                         // invoke send
-                        this._sender(batch, isAsync, this._buffer.length);
+                        this._sender(this._buffer.getItems(), isAsync, this._buffer.count());
```
MaxShehovtsov: This looks confusing - you pass items in but sender actually calls .batchPayloads itself and sends it as. My advice is that you either pass the whole buffer and let _sender operate with it or pass everything as arguments.

dnduffy:
Some tweaks and :shipit:

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
```diff
@@ -496,7 +496,8 @@ class SessionContextTests extends TestClass {
             emitLineDelimitedJson: () => null,
             maxBatchSizeInBytes: () => null,
             maxBatchInterval: () => null,
-            disableTelemetry: () => null
+            disableTelemetry: () => null,
+            storeSendBufferInSessionStorage: () => null
```
AlexBulankou: enableSessionStorageBuffer
enableStorageBuffer
sessionStorageBuffer
storageBuffer
enableItemStorage

JavaScript/JavaScriptSDK/SendBuffer.ts
```diff
@@ -0,0 +1,205 @@
...
+        public count(): number {
+            return this._buffer.length;
+        }
+
+        public clear() {
+            this._buffer.length = 0;
+        }
+
+        public getItems(): string[] {
+            return this._buffer;
```
AlexBulankou: do you want to give them the actual thing that they can modify or a copy, i.e. this._buffer.slice(0) ?

JavaScript/JavaScriptSDK/SendBuffer.ts
```diff
@@ -0,0 +1,205 @@
...
+                    return buffer;
+                }
+            }
+
+            return [];
+        }
+
+        private setBuffer(key: string, buffer: string[]) {
+            var bufferJson = JSON.stringify(buffer);
+            Util.setSessionStorage(key, bufferJson);
```
AlexBulankou: This will fail for various reasons. Catch and log internal message?

AlexBulankou:
:shipit:

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
kamilszostak: Thanks for spotting this. This should not be there, I'll remove it.

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
kamilszostak: Good question. I think it's because `ISenderConfig `and `ITelemetryConfig` depend on `IConfig `which can be changed during the run time.

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
kamilszostak: Done.

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
kamilszostak: What are the benefits of using this trick? String concatenation is expensive and table join will work similar to `StringBuilder `from C#? 

TS compiler doesn't optimize this code. Both versions, before and after the compilation look exactly the same.

JavaScript/JavaScriptSDK.Tests/CheckinTests/SendBuffer.tests.ts
kamilszostak: Yeah.. I thought about it and for some reason I've decided not to do that. On second thought, you're right. It will be better to reuse the test logic. 

Done.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
kamilszostak: `Serializer` returns an object, not an array. I missed the coma, I fixed it to +3 :)

JavaScript/JavaScriptSDK/SendBuffer.ts
kamilszostak: Added.

JavaScript/JavaScriptSDK/SendBuffer.ts
kamilszostak: Yes.

JavaScript/JavaScriptSDK/Sender.ts
kamilszostak: I'm not sure if I understand. We step into the `if` statement if telemetry is enabled (not disabled).

JavaScript/JavaScriptSDK/Sender.ts
kamilszostak: I'll move the `var batch = this._buffer.batchPayloads();` right before the send method, to make it less confusing.

JavaScript/JavaScriptSDK/Sender.ts
kamilszostak: Good point. Fixed.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Context/Session.tests.ts
kamilszostak: `enableSessionStorageBuffer ` is the winner :)

JavaScript/JavaScriptSDK/SendBuffer.ts
kamilszostak: A copy might be a better option. We don't want anybody to mess with the buffer :)

JavaScript/JavaScriptSDK/SendBuffer.ts
kamilszostak: try/catch added.

JavaScript/JavaScriptSDK/Sender.ts
dnduffy: I think now I might have read the condition backwards, maybe I missed the ! character or something.  Sorry. :)

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
kamilszostak: Good point. I'll change the code to not store the `errorSpy `and `successSpy` on the Sender.test class.

JavaScript/JavaScriptSDK/SendBuffer.ts
kamilszostak: I agree. Do you have any suggestions on how to fix it? Change the `markAsSent` name?

Issue #86, opened 2015-12-02
Do not show AI (internal) traces in console by default


MaxShehovtsov:
It's useless to show them in the browser console.

PR #190, opened 2016-05-10
Show AI internal traces in console only when verboseLogging is enabled

kamilszostak:
Fix for #86 .

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
@@ -156,6 +156,7 @@ class LoggingTests extends TestClass {
                     // act
                     var message = new InternalLoggingMessage(1, "error!");
                     InternalLogging.enableDebugExceptions = () => false;
+                    InternalLogging.verboseLogging = () => true;
                     InternalLogging.throwInternalNonUserActionable(Microsoft.ApplicationInsights.LoggingSeverity.CRITICAL, message);
 
                     // verify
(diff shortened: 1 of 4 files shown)


JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
```diff
@@ -165,7 +166,36 @@ class LoggingTests extends TestClass {
                     Assert.equal("AI (Internal): " + "NONUSRACT_BrowserCannotReadLocalStorage message:\"error!\"", InternalLogging.queue[0].message);
 
                     // cleanup
-                    
+                    InternalLogging.verboseLogging = () => false;
```
dnduffy: Would it be safer to read out the original value before setting it to true and then set it back to the original value here rather than explicitly false?

dnduffy: Also, isn't there already a cleanup method that does this?

dnduffy:
:shipit:

JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
kamilszostak: It's a class with a bunch of static methods and properties. It's hard to unit tests it :angry: 

...anyway, I changed it slightly to make the code more clear and easier to read.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
```diff
@@ -5,35 +5,45 @@
...
 
     /**
      * Clears the internal logging queue
      */
     private clearInternalLoggingQueue() {
-        var length = Microsoft.ApplicationInsights._InternalLogging.queue.length;
+        var length = this.InternalLogging.queue.length;
         for (var i = 0; i < length; i++) {
-            Microsoft.ApplicationInsights._InternalLogging.queue.shift();
+            this.InternalLogging.queue.shift();
```
MaxShehovtsov: Just wondering if you could do queue=[] instead of going though the loop.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
```diff
@@ -5,35 +5,45 @@
 
 class LoggingTests extends TestClass {
 
+    InternalLogging = Microsoft.ApplicationInsights._InternalLogging;
```
MaxShehovtsov: is it just a refactoring to simplify future renaming?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -46,6 +46,8 @@ class SenderTests extends TestClass {
         this.successSpy = this.sandbox.spy(Microsoft.ApplicationInsights.Sender, "_onSuccess");
         this.loggingSpy = this.sandbox.stub(Microsoft.ApplicationInsights._InternalLogging, "warnToConsole");
         this.testTelemetry = { aiDataContract: true };
+
+        Microsoft.ApplicationInsights._InternalLogging.verboseLogging = () => true;
```
MaxShehovtsov: Looks like this test overrides defaults and hence must reset it back too.

AlexBulankou:
:shipit:

PR #192, opened 2016-05-13
Log verb as part of dependency name

kamilszostak:
The `trackAjax` function is public so I added `method` (verb) as an optional parameter. This way it won't affect existing customer. If the value is null only the absolute URL will be used. 

The verb isn't exposed in the UI yet, but it is accessible through Application Analytic.


AlexBulankou:
:shipit:

Issue #184, opened 2016-05-04
NONUSRACT_MissingRequiredFieldSpecification


AlexBulankou:
Unique weekly ikeys: 20
Message: AI (Internal): NONUSRACT_MissingRequiredFieldSpecification message:"Missing required field specification. The field is required but not present on source" props:"{field:name,name:baseData}"

Looks like this one is related:
Unique weekly ikeys: 30
Message: AI (Internal): NONUSRACT_TrackPVFailedCalc message:"trackPageView failed on page load calculation: " props:"{exception:[object String]\out of memory\}"

kamilszostak:
I have a fix - #194.

PR #194, opened 2016-05-16
Required field is set to not_specified if value is missing

kamilszostak:
#184. If name is required set the value to 'not_specified' if value is null or empty. 

Common.DataSanitizer.sanitizeString(exception.name || Util.NotSpecified) didn't handle " " names correctly. It will return 'not_specified' with the fix.

### DIFF
--- a/JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
+++ b/JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
@@ -17,7 +17,7 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
         constructor(data: Microsoft.Telemetry.Base, name: string) {
             super();
 
-            this.name = name || Util.NotSpecified;
+            this.name = Common.DataSanitizer.sanitizeString(name) || Util.NotSpecified;
             this.data = data;
             this.time = Util.toISOStringForIE8(new Date());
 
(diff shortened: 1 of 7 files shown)


JavaScript/JavaScriptSDK/Telemetry/Exception.ts
```diff
@@ -77,8 +77,8 @@ module Microsoft.ApplicationInsights.Telemetry {
 
         constructor(exception: Error) {
             super();
-            this.typeName = Common.DataSanitizer.sanitizeString(exception.name || Util.NotSpecified);
-            this.message = Common.DataSanitizer.sanitizeMessage(exception.message || Util.NotSpecified);
+            this.typeName = Common.DataSanitizer.sanitizeString(exception.name) || Util.NotSpecified;
```
dnduffy: I trust that Common.DataSanitizer.sanitizeString handles being passed undefined or null?

dnduffy:
:shipit: if my assumption is true.

JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
```diff
@@ -17,7 +17,7 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
         constructor(data: Microsoft.Telemetry.Base, name: string) {
             super();
 
-            this.name = name;
+            this.name = name || Util.NotSpecified;
```
MaxShehovtsov: So how do you handle whitespaces here?

JavaScript/JavaScriptSDK/Telemetry/Exception.ts
kamilszostak: Yes, sanitizeString() handles null and undefined. I'll add unit tests.

JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
kamilszostak: I added sanitizeString.

PR #201, opened 2016-05-20
Allow passing severityLevel in trackException

kamilszostak:
#170

### DIFF
--- a/API-reference.md
+++ b/API-reference.md
@@ -115,7 +115,7 @@ To send a single measurement, use just the first two parameters. If you take mea
 
 ### trackException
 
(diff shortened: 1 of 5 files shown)


abaranch:
Do you have autocollected exceptions? If yes, do you want to set SeverityLevel for them?..

:shipit:

dnduffy:
:shipit:

kamilszostak:
@abaranch Good point, but I don't want to change the current behavior.

abaranch:
@kamilszostak but that would be consistent with Web SDK where we set it...

PR #203, opened 2016-05-22
Reference TS compiler in tests, removing hardcoded VS version

AlexBulankou:
* JavaScriptSDK.Tests project is referencing tsc in Exec tasks, assuming it is available in PATH (which it is not, unless installed on the box). I'm referencing TS compiler package and referencing it from packages path
* Removing hardcoded VS version, so that solution can be opened and built with VS2015


dnduffy:
:shipit:

MaxShehovtsov:
:shipit: Thanks!

PR #204, opened 2016-05-23
Enable data loss analyzer for session storage buffer

kamilszostak:


JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -104,10 +104,15 @@ module Microsoft.ApplicationInsights {
             }
 
             // enable session storage buffer experiment
-            this.config.enableSessionStorageBuffer = new SplitTest().isEnabled(this.config.instrumentationKey, 10); ;
+            var enableExperiment = new SplitTest().isEnabled(this.config.instrumentationKey, 10);
+            this.config.enableSessionStorageBuffer = enableExperiment;
 
             this.context = new ApplicationInsights.TelemetryContext(configGetters);
 
+            DataLossAnalyzer.appInsights = this;
```
dnduffy: Where is DataLossAnalyzer defined?

dnduffy:
:shipit:

JavaScript/JavaScriptSDK/AppInsights.ts
kamilszostak: DataLossAnalyzer.ts :unamused: 

The class was created a while ago and we used it before to measure the data loss rate. I need to enable it for a couple days to measure if session storage buffer is working as expected.

AlexBulankou: Why does it need to be static? It makes sense for Utils to be static, because it is reasonable that everyone has access to it, but in this case, only sender and appInsights need access to DataLossAnalyzer.

Why do we need to pass the whole appInsights object to DataLossAnalyzer. It needs to be able to trackTrace and flush, would a better approach be (from appInsights):
```
var dataLossAnalyzer = new DataLossAnalyzer();
dataLossAnalyzer.onLossNotify(()=>
{
 trackTrace(...);
 flush();
});
```
This second part of the comment is a suggestion, but I feel more strongly about the first one.

kamilszostak: I agree. I'll create an issue and refactor DataLossAnalyzer in a separate pull request. Thanks!

Issue #196, opened 2016-05-17
Document and clean up trackAjax method


kamilszostak:
* Rename `trackAjax `to `trackDependency`
* Update [API-reference](https://github.com/Microsoft/ApplicationInsights-JS/blob/master/API-reference.md) page
* Make `method: string` parameter required and change the order of parameters

PR #212, opened 2016-05-26
Rename trackAjax to trackDependency

kamilszostak:
#196


MaxShehovtsov:
Did you change the snippet? I also wonder if we can do this change as snippet has trackAjax and potentially someone could already start using it explicitely. If we change the snippet and/or the implementation they'll be broken.

API-reference.md
```diff
@@ -155,6 +155,23 @@ In the portal, you can search on message content and [display individual trackTr
 (Unlike `trackEvent`, you can't filter on the message content in the portal.)
 
 
+### trackDependency
+
+    trackDependency(id: string, method: string, absoluteUrl: string, pathName: string, totalTime: number, success: boolean, resultCode: number) {
+    
+Log an ajax dependency
```
Dmitry-Matveev: Should we consider changing this from "Log an ajax dependency" to just "dependency" then.

dnduffy:
:shipit:

kamilszostak:
@MaxShehovtsov Yes, I updated the snippet. I left `trackAjax` method so existing costumers won't be impacted. But the method is not exposed in `IAppInsights` interfaces or *.d.ts files. This should force users to switch to `trackDependency` after they update the nuget package and try to build their solution. 

Once we release this change we will have to update the snippet code in the JS SDK on-boarding page.

MaxShehovtsov:
:shipit:

PR #216, opened 2016-06-01
Disable Session Storage buffer experiment

kamilszostak:


kamilszostak:
I've decided to disable SessionStorage buffer as it might be a root cause of #217 . 
I'll investigate and re-enable the new buffer when we have a fix.

package.json
```diff
@@ -1,15 +1,15 @@
...
-        "microsoft",
-        "azure"
-    ],
+    "keywords":  [
+                     "browser performance monitoring",
+                     "script errors",
+                     "application insights",
+                     "microsoft",
+                     "azure"
+                 ],
```
dnduffy: nitpick: is the formatting change intentional or did VS auto-format it?

dnduffy:
:shipit:

package.json
kamilszostak: Not intentional. Powershell script to update the version number did that. 
I'll check if I can fix it later today and submit a fix in a separate pull request.

AlexBulankou:
:shipit:

MaxShehovtsov:
:shipit:

PR #218, opened 2016-06-01
Copy 0.22.17 to dist

kamilszostak:


dnduffy:
This is just updating a copy of a build generated aggregate file right?
:shipit:

Issue #221, opened 2016-06-03
npm package is a bit of a mess


ArturDorochowicz:
```
$ npm install applicationinsights-js
└── applicationinsights-js@0.22.17 
```

1. It contains everything, including NuGet.exe, PowerShell files etc.

   ```
$ tree node_modules/applicationinsights-js/ -L 2 
node_modules/applicationinsights-js/
├── API-reference.md
├── bower.json
├── changeSdkVersion.ps1
├── Common.props
├── dist
│   ├── ai.0.js
│   └── ai.js
├── EnlistmentRoot.marker
├── Global.props
├── JavaScript
│   ├── JavaScriptSDK
│   ├── JavaScriptSDK.Tests
│   ├── Microsoft.ApplicationInsights.JavaScript.sln
│   ├── NuGet.JavaScript
│   ├── NuGet.TypeScript
│   └── PostTestScript.cmd
├── LICENSE
├── NuGet.config
├── NuGet.exe
├── NuGet.targets
├── package.json
├── Package.targets
├── PostTestScript.cmd
├── PushNugetPackages.cmd
├── README.md
├── runAllTests.cmd
├── RunTestsInBrowser.ps1
├── ThirdPartyNotices.txt
└── uploadFilesToAzureStorage.ps1

6 directories, 23 files
   ```

2. `main` field in `package.json` points to `JavaScriptSDK/AppInsightsModule`. I doubt it's supposed to point to a TypeScript file, but if so it skips the `JavaScript` directory. Preferably it should point to `dist/ai.js` (i.e. unminified file), I think.

3. The one usable thing in here which are the javascript files in `dist` are actually different version (0.22.15).

4. The minified file is `ai.0.js`. That's a bit unconventional, I would expect `ai.min.js`.

Besides, what's the story here? Is it supposed to be an npm package with TypeScript? Or the point of the package is the JavaScript in `dist` directory?

The readme file in "Import as a module and initialize dynamically (download full Application Insights script from CDN)" talks about "modern JS application", however the code example is TypeScript and the `downloadAndSetup` method can only be found in the package in TypeScript files. That was a bit confusing to me. Please, make this scenario available in JS (that is without me compiling the package).

AlexBulankou:
Thanks @ArturDorochowicz! We'll work on improving this early next week, will have another update Monday on this.

AlexBulankou:
[...]

PR #222, opened 2016-06-05
Module fixes (compiling AppInsightsModule, fixing ``main`` reference)

AlexBulankou:
* Adding separate project for AppInsightsModule (as dicussed offline with @kamilszostak)
* Adding separate interface-only TS project and referencing it from JavaScriptSDK and JavaScriptSDK.Module projects
* Fixing ``main`` reference in package.json
* Updating dist folder and bumping up version
This should address the blocking issues mentioned in #221


MaxShehovtsov:
LGTM, but will give it another look once you merge and it's available for trying out locally.

Global.props
```diff
@@ -90,7 +90,7 @@
     -->
     <SemanticVersionMajor>0</SemanticVersionMajor>
     <SemanticVersionMinor>22</SemanticVersionMinor>
-    <SemanticVersionPatch>17</SemanticVersionPatch>
+    <SemanticVersionPatch>18</SemanticVersionPatch>
     <!-- 
       Date when Semantic Version was changed. 
       Update for every public release.
```
kamilszostak: Please also update the version data on line 98.

dist/ai.js
```diff
@@ -1,3 +1,8 @@
+var __extends = (this && this.__extends) || function (d, b) {
```
kamilszostak: Did you take this file from build drop dir? I made this mistake before and Max suggested to always release from the official build drop directory (\\vscsstor\CIDrops\AIC\DC_JavaScript_master).

kamilszostak:
:shipit:

dist/ai.js
AlexBulankou: you're right, I probably did. We should use your script to deploy.

PR #224, opened 2016-06-07
Disable DataLossAnalyzer for XDomainRequest

kamilszostak:
* Disable DataLossAnalyzer for XDomainRequest 
* Sender calls buffer.GetItems only once


MaxShehovtsov:
:shipit:

Issue #233, opened 2016-06-10
AppInsightsModule hardcodes http path to the script


ArturDorochowicz:
See:
https://github.com/Microsoft/ApplicationInsights-JS/blob/e739caed7bd9d9f051a54889bf2483e1e43b6db7/JavaScript/JavaScriptSDK.Module/AppInsightsModule.ts#L33

The original snippet uses protocol relative path.

MaxShehovtsov:
Thanks! Looks like a bug.

dlebedynskyi:
any change this fix will be released soon to npm?

PR #234, opened 2016-06-13
Updating default setting to aiConfig.url to not hardcode http endpoint

AlexBulankou:
Fix for #233


kamilszostak:
:shipit:

fermaem:
Sorry for asking, but why rather not always posting to https?

AlexBulankou:
@fermaem this is for consistency how our snippet works, but we are actually making the change to always use https soon (next week or so) and when this change is done we will also update static initialization
@kamilszostak FYI

fermaem:
>we are actually making the change to always use https soon

YES :bangbang: :tada: :heart:

Thx for this

PR #238, opened 2016-06-23
Test for d.ts files hosted in DefinitelyTyped repo

kamilszostak:
The file won't compile if we make breaking changes to interfaces or public signatures.


dnduffy:
:shipit:

MaxShehovtsov:
:shipit: nice!

PR #245, opened 2016-06-27
Disable experiment

kamilszostak:
* Disable session storage buffer experiment
* Fix Data loss analyzer bug

### DIFF
--- a/JavaScript/JavaScriptSDK/DataLossAnalyzer.ts
+++ b/JavaScript/JavaScriptSDK/DataLossAnalyzer.ts
@@ -13,6 +13,7 @@
         static reset() {
             if (DataLossAnalyzer.isEnabled()) {
                 Util.setSessionStorage(DataLossAnalyzer.ITEMS_QUEUED_KEY, "0");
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScriptSDK/DataLossAnalyzer.ts
```diff
@@ -13,6 +13,7 @@
         static reset() {
             if (DataLossAnalyzer.isEnabled()) {
                 Util.setSessionStorage(DataLossAnalyzer.ITEMS_QUEUED_KEY, "0");
+                Util.setSessionStorage(DataLossAnalyzer.ISSUES_REPORTED_KEY, "0");
```
MaxShehovtsov: Not sure why is this needed if you're disabling the analyzer

MaxShehovtsov:
:shipit:

JavaScript/JavaScriptSDK/DataLossAnalyzer.ts
kamilszostak: It will be useful if we want to enable the DataLossAnalyzer in the future.

PR #250, opened 2016-06-30
Check if document.cookie is available

### LINKED ISSUE #249: JS SDK fails to initialize due to error: "Uncaught SecurityError: Failed to read the 'cookie' property from 'Document': The document is sandboxed and lacks the 'allow-same-origin' flag"

kamilszostak:
This fixes #249 

I still need to do some testing to validate the SDK works as expected in the `document.cookie` is not accessible. Ideally it should still send all telemetry, but without session and user id.

### DIFF
--- a/JavaScript/JavaScriptSDK/Util.ts
+++ b/JavaScript/JavaScriptSDK/Util.ts
@@ -1,9 +1,9 @@
 ﻿/// <reference path="./logging.ts" />
 module Microsoft.ApplicationInsights {
 
-         /**
-         * Type of storage to differentiate between local storage and session storage
-         */
+    /**
+    * Type of storage to differentiate between local storage and session storage
+    */
     enum StorageType {
         LocalStorage,
         SessionStorage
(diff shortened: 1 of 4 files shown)


JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -258,6 +277,10 @@ module Microsoft.ApplicationInsights {
          * helper method to access userId and sessionId cookie
          */
         public static getCookie(name) {
+            if (!Util.canUseCookies()) {
```
dnduffy: Out of curiosity, can a flag be set at startup rather than making the method call each time?
So "this.canUseCookies" instead of "Util.canUseCookies()"?

dnduffy:
:shipit:

JavaScript/JavaScriptSDK/Util.ts
kamilszostak: Good idea. I kept the canUseCookie() method, but I cache the result.

JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -12,7 +12,9 @@
     localDocument.getElementsByTagName(scriptText)[0].parentNode.appendChild(scriptElement);
 
     // capture initial cookie
-    appInsights.cookie = localDocument.cookie;
+    try {
+        appInsights.cookie = localDocument.cookie;
```
MaxShehovtsov: Just wondering if there's a corresponding code that expects/does not expect appInsights.cookie to be unset

MaxShehovtsov:
:shipit:

PR #251, opened 2016-06-30
Limit session storage buffer size to 100 elements

kamilszostak:


JavaScript/JavaScriptSDK/SendBuffer.ts
```diff
@@ -132,6 +136,17 @@ module Microsoft.ApplicationInsights {
         }
 
         public enqueue(payload: string) {
+            if (this._buffer.length >= SessionStorageSendBuffer.MAX_BUFFER_SIZE) {
+                // sent internal log only once per page view
+                if (!this._bufferFullMessageSent) {
+                    _InternalLogging.throwInternalUserActionable(LoggingSeverity.WARNING,
+                        new _InternalLogMessage(_InternalMessageId.USRACT_SessionStorageBufferFull,
+                            "Maximum buffer size reached: " + this._buffer.length));
+                    this._bufferFullMessageSent = true;
```
dnduffy: When/where does it get set back to false?
I presume that once the buffer has space again that this should no longer be true.

MaxShehovtsov:
why are we doing it?

kamilszostak:
IMHO it's good to have some safety switch to prevent a situation where the buffer becomes too big. Old buffer was cleared on every page load, the new buffer will restore all unsent element from the session storage. If, for instance, client logs a lot of events and the backend becomes very slow the buffer will grow bigger and bigger. 

From what I see an average payload size is ~600bytes. The maxBatchSizeInBytes = 1,000,000bytes, so it will store around 1,666 elements. I'll increase the max buffer size to **2,000**, so it should always trigger send before the buffer limit is reached.

JavaScript/JavaScriptSDK/SendBuffer.ts
kamilszostak: Good point. It will be cleared when the script is loaded again (when user navigates to a next page, etc). I didn't want to log this message too often... but you're right, it might be a good idea to clear it when the buffer has space again.

MaxShehovtsov:
:shipit:

Issue #93, opened 2015-12-11
should prefix.js fall back to https if loaded from file:// uri?


gardnerjr:
from this stack overflow:  http://stackoverflow.com/questions/33952511/microsoft-application-insights-on-cordova-requests-goes-to-file/34230709#34230709

the default prefix/snippet doesn't specify a protocol, falling back to whatever the page got loaded from (presumably so http gets the ai script also via http, and https sites get the script via https so there's no "unsecure content" errors or warnings.

however, if you use the example from a file:// based source, like a cordova app in the stack overflow question above, it also tries to load / submit data to:

file:// dc.services.visualstudio.com/v2/track 

instead of using https.  should there be code in the snippet to fall back to https when sending data to DC even if the source was loaded from a file:// uri?

MaxShehovtsov:
@gardnerjr I don't see any issues of top of my head. 
@scsouthw @olegsych - do you guys remember why we change DC url protocol based on page protocol?

olegsych:
The only reason I can think of is to ensure that an https page talks to AI over https (browsers can block https page talking to http resources). An http page could be more efficient in its bandwidth use if it used http to talk to AI as well, but it wouldn't be terrible if AI scripts could always used HTTPS.

gardnerjr:
i think right now you can repro this in a desktop webpage.  create a "page.html" on your windows desktop.  paste a valid snippet with a valid ikey into it.  open it with the browser via `file:\\c:\users\yourusername\desktop\page.html`  (or whatever you name it, etc).  You won't see any outbound traffic to dc.

first, because the browser will fail to load the ai javascript from a file://uri.  

Then, modify the snippet to use https:// to get the ai javascript.
refresh the browser.  now the AI sdk will download just fine (as it is https instead of file, which doesn't exist), but you still won't see any outbound traffic to dc.

to fix it, you also have to add:

               endpointUrl: "https://dc.services.visualstudio.com/v2/track"

to your config, to force the sdk to send data with a proper protocol, as the ai sdk doesn't have one, it is just `//dc.services...` without a protocol in the downloaded ai.0.js.  once you do the 2 steps above to explicitly specify a protocol, if you refresh the page, you'll see outbound network requests to dc.

PR #258, opened 2016-07-07
Send telemetry over HTTPS

kamilszostak:
Extra code for XDomainRequests is required because IE8,9 requires the same schema [see point 7 here](https://blogs.msdn.microsoft.com/ieinternals/2010/05/13/xdomainrequest-restrictions-limitations-and-workarounds/)

Fixes: #93


dnduffy:
:shipit:

PR #261, opened 2016-07-11
Sent buffer overflow

kamilszostak:
If the back-end is not responding or is responding very slow, the SENT_BUFFER can grow above the max buffer size limit. This can cause a situation when the restored buffer has more than MAX_BUFFER_SIZE elements.

### DIFF
--- a/JavaScript/JavaScriptSDK/SendBuffer.ts
+++ b/JavaScript/JavaScriptSDK/SendBuffer.ts
@@ -187,6 +187,16 @@ module Microsoft.ApplicationInsights {
             var sentElements = this.getBuffer(SessionStorageSendBuffer.SENT_BUFFER_KEY);
             sentElements = sentElements.concat(payload);
 
+            if (sentElements.length > SessionStorageSendBuffer.MAX_BUFFER_SIZE) {
+                // We send telemetry normally. If the SENT_BUFFER is too big we don't add new elements
+                // until we receive a response from the backend and the buffer has free space again (see clearSent method)
+                _InternalLogging.throwInternalUserActionable(LoggingSeverity.CRITICAL,
+                    new _InternalLogMessage(_InternalMessageId.USRACT_SessionStorageBufferFull,
+                        "Sent buffer reached its maximum size: " + sentElements.length));
+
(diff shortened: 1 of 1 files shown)


JavaScript/JavaScriptSDK/SendBuffer.ts
```diff
@@ -128,6 +128,10 @@ module Microsoft.ApplicationInsights {
 
             this._buffer = bufferItems.concat(notDeliveredItems);
 
+            if (this._buffer.length > SessionStorageSendBuffer.MAX_BUFFER_SIZE) {
```
dnduffy: You might want to comment that items on the end will be lost.

JavaScript/JavaScriptSDK/SendBuffer.ts
```diff
@@ -183,6 +187,16 @@ module Microsoft.ApplicationInsights {
             var sentElements = this.getBuffer(SessionStorageSendBuffer.SENT_BUFFER_KEY);
             sentElements = sentElements.concat(payload);
 
+            if (sentElements.length > SessionStorageSendBuffer.MAX_BUFFER_SIZE) {
```
dnduffy: At the point in time when markAsSent is called isn't it too late to throw away data?

kamilszostak: Ideally we should check the total buffer size (BUFFER + SENT_BUFFER) in `enqueue` method, but I try to minimize the number of times we read from session storage. This is why `enqueue` method only checks the BUFFER size. IMHO it's OK to check the size in `markAsSent` method because this is one of two places where we add new items to the SENT_BUFFER (the other one is in the constructor).

dnduffy: I understand not accepting new items to be sent when over quota but something feels wrong about losing track of data that really was already sent, what am I missing?

dnduffy: Per discussion we'll leave it like this for code efficiency and trust that the limits really shouldn't be hit in normal circumstances.

dnduffy:
:shipit:

PR #268, opened 2016-07-14
Create E2ETests as a directory.

mateoatr:
Create E2ETests as a directory instead of a file by appending '\' to the route.


MaxShehovtsov:
A commit description would be helpful :)

mateoatr:
@MaxShehovtsov Just did (y). I was having some build errors with the PowerShell script, it seems that if there isn't any '\' at the end of a route, PowerShell will create a blob instead of a directory.

dnduffy:
:shipit:

kamilszostak:
:shipit:

PR #271, opened 2016-07-18
Fixing npm package

AlexBulankou:
Adding .npmignore file under JavaScript, because we have .gitgnore there
Removing AppInsightsModule.js from source control


kartang:
:shipit:

PR #273, opened 2016-07-18
Handle partial success (response code 206)

kamilszostak:
* Handle partial success
* Add retry logic and exponential back-off


kamilszostak:
dang... I've just noticed David's first commit message. I'll squash and update the RP.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -655,12 +666,497 @@ class SenderTests extends TestClass {
...
+                sender.send(this.testTelemetry);
+
+                Assert.equal(2, sender._buffer.count(), "Buffer has two items");
+
+                // trigger send
+                this.clock.tick(sender._config.maxBatchInterval());
+
+                // retry after 777s. 
+                var now = 1468864738000;
+                this.clock.setSystemTime(now);
```
dnduffy: I think I'm missing something, why change the clock instead of taking two readings?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -655,12 +666,497 @@ class SenderTests extends TestClass {
...
+
+                var delay = 77; 
+                var delayDate = new Date();
+                delayDate.setSeconds(delayDate.getSeconds() + delay);
+
+                this.requestAsserts();
+                this.fakeServer.requests.pop().respond(
+                    206,
+                    { "Content-Type": "application/json", "Retry-After": delayDate.toUTCString() },
```
dnduffy: So for this test with a non-retryable 408 shouldn't the retry-after header not exist?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
kamilszostak: For some reason Sinon is setting the fake clock to some day in 1969. Which is a negative number when we convert it to a number. This doesn't work well with the `if` condition in `_setRetryTime`. 
The easiest way to fix it was to set Sinon clock to a date after 1/1/1970.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -655,12 +666,497 @@ class SenderTests extends TestClass {
...
+                });
+
+                var sender = this.getSender();
+                Assert.ok(sender, "sender was constructed");
+
+                // too many items accepted
+                var response = '{ "itemsReceived": 1, "itemsAccepted": 2, "errors": [] }';
+                var result = <Microsoft.ApplicationInsights.IBackendResponse>(<any>sender)._parseResponse(response);
+
+                Assert.ok(!result, "Parse should fail - there are too itemsAccepted > itemsReceived");
```
dnduffy: Some of these tests seem to be testing Breeze rather than the SDK. ;)

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -655,12 +666,497 @@ class SenderTests extends TestClass {
...
+                var sender = this.getSender();
+                Assert.ok(sender, "sender was constructed");
+
+                var response = '{}';
+                var result = <Microsoft.ApplicationInsights.IBackendResponse>(<any>sender)._parseResponse(response);
+                Assert.ok(!result, "Parse should fail for an empty response");
+
+                response = '{ "itemsReceived": 1, "itemsAccepted": 2, "errors": [] }';
+                result = <Microsoft.ApplicationInsights.IBackendResponse>(<any>sender)._parseResponse(response);
+                Assert.ok(!result, "Parse should fail - itemsAccepted field missing");
```
dnduffy: But the field is not missing, it just has a bad value.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -655,12 +666,497 @@ class SenderTests extends TestClass {
...
+        sender.successSpy.reset();
+        sender.errorSpy.reset();
+    }
+
+    private validatePartialSuccess_Retryable() {
+        var sender = this.getSender();
+        Assert.ok(sender, "sender was constructed");
+
+        // send six items
+        this.fakeServer.requests.pop();
```
dnduffy: What request is being popped off the stack here?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -655,12 +666,497 @@ class SenderTests extends TestClass {
...
+        );
+
+        // verify
+        Assert.ok(sender.successSpy.called, "success was invoked");
+
+        // partial response warning
+        this.logAsserts(1);
+
+        // the buffer has 5 items - payloads 1-5, payload 0 was accepted by the backend and should not be re-send
+        Assert.equal(5, sender._buffer.count(), "Buffer has 5 items to re");
```
dnduffy: "...items to retry."

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
kamilszostak: I removed this test as we can't support `retry-after` right now.

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -199,13 +264,14 @@ module Microsoft.ApplicationInsights {
                     }
 
                     // update lastSend time to enable throttling
-                    this._lastSend = +new Date;
+                    this._lastSend = +new Date; // TODO: Does anything use _lastSend?
```
dnduffy: Can we remove the _lastSend stuff (or the TODO if _lastSend is used)?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
kamilszostak: 1) Yes and no. In theory it shouldn't happen, but I wanted to test if our code doesn't throw when backend returns inconsistent response. 

2) "too many". I improved the comment to explain that there were two errors instead of one.

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -215,6 +281,57 @@ module Microsoft.ApplicationInsights {
...
+                    var backOffSlot = (Math.pow(2, this._consecutiveErrors) - 1) / 2;
+                    var backOffDelay = Math.floor(Math.random() * backOffSlot * SlotDelayInSeconds) + 1;
+                    delayInSeconds = Math.max(Math.min(backOffDelay, 3600), SlotDelayInSeconds);
+                }
+
+                // TODO: Log the backoff time like the C# version does.
+                retryAfterTimeSpan = Date.now() + (delayInSeconds * 1000);
+            }
+
+            // TODO: Log the retry at time like the C# version does.
```
dnduffy: I suggest we either remove the TODO's or do them.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
kamilszostak: Copy/Paste error. I fixed the error message.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
kamilszostak: Sinon library puts an empty request on the stack. I was trying to figure out why, but I gave up.

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -215,6 +281,57 @@ module Microsoft.ApplicationInsights {
...
+         * Parses the response from the backend. 
+         * @param response - XMLHttpRequest or XDomainRequest response
+         */
+        private _parseResponse(response: any, retryAfter?: string): IBackendResponse {
+            try {
+                var result = JSON.parse(response);
+
+                if (result && result.itemsReceived && result.itemsReceived >= result.itemsAccepted &&
+                    result.itemsReceived - result.itemsAccepted == result.errors.length) {
+                    return {
```
dnduffy: I'm curious what the advantage to constructing a new object is over using the parsed one you already have?

dnduffy:
:shipit:

JavaScript/JavaScriptSDK/Sender.ts
kamilszostak: Removed.

JavaScript/JavaScriptSDK/Sender.ts
kamilszostak: Maybe the naming is a bit confusing - it's parsing, basic validation and try/catch. I extracted it to a method to share it between XDomain and XmlHttpRequest handlers.

dnduffy: What I mean is that you call JSON.parse(response) which either fails or gives you an object with those fields and then you validate that those expected fields exist.  So why create a new object in the return statement which is basically a copy of the object you already have in the result variable from the original JSON.parse call?

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -420,9 +416,9 @@ module Microsoft.ApplicationInsights {
             var errors = results.errors.reverse();
             for (var error of errors) {
                 var extracted = payload.splice(error.index, 1)[0];
-                if (error.statusCode == 408 // Timeout
+                if (error.statusCode == 402 // Too many requests over extended time.
+                    || error.statusCode == 408 // Timeout
                     || error.statusCode == 429 // Too many requests.
-                    || error.statusCode == 439 // Too many requests over extended time.
                     || error.statusCode == 500 // Internal server error.
```
dnduffy: I think 439 can still happen in their new SDK but it's just not retryable; however, it sounded like they expected as to dump our cache of pending items when we receive a 439 now.

kamilszostak: Yes, I'll add this functionality in next pull request.

kamilszostak:
I cleared commit messages history and sent a new PR #276.

PR #279, opened 2016-07-22
Do not flood the console!

kamilszostak:
* Limit console logging to only one message of a given type (if !verboseLogging)
* Limit to only one internal message log per page view when the session storage is not available

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
@@ -302,7 +302,6 @@ class LoggingTests extends TestClass {
                 var maxAllowedInternalMessages = 2;
                 var message1 = new this.InternalLoggingMessage(1, "1");
                 var message2 = new this.InternalLoggingMessage(2, "2");
-               
 
                 // setup
                 this.InternalLogging.enableDebugExceptions = () => false;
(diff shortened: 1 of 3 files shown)


JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
```diff
@@ -323,6 +380,42 @@ class LoggingTests extends TestClass {
...
+                this.InternalLogging.throwInternalNonUserActionable(Microsoft.ApplicationInsights.LoggingSeverity.CRITICAL, message2);
+
+                // verify
+                // only two messages should be in the queue, because we have to distinct types
+                Assert.equal(2, this.InternalLogging.queue.length);
+                Assert.equal(this.InternalLogging.queue[0], message1);
+                Assert.equal(this.InternalLogging.queue[1], message2);
+
+                // clean up - reset session storage
+                Microsoft.ApplicationInsights.Util.canUseSessionStorage = utilCanUserSession;
```
dnduffy: Should "utilCanUserSession" be "utilCanUseSession"? (i.e. "Use" vs. "User")

JavaScript/JavaScriptSDK/Logging.ts
```diff
@@ -182,7 +186,21 @@
                 if (typeof (message) !== "undefined" && !!message) {
                     if (typeof (message.message) !== "undefined") {
                         message.message = this.AiUserActionablePrefix + message.message;
-                        this.warnToConsole(message.message);
+
+                        // check if this message type was already logged to console for this page view and if so, don't log it again
+                        var logToConsole = true;
```
dnduffy: I would remove the variable and invert the if statement below to have the key = true setting, remove the else, and just put the console call inside the conditional for shorter/simpler/faster code.

dnduffy:
:shipit:

JavaScript/JavaScriptSDK/Logging.ts
kamilszostak: Of course. I don't know why I made it so complected :)

JavaScript/JavaScriptSDK.Tests/CheckinTests/Logging.tests.ts
kamilszostak: :smile: I copied it from Sender.test.ts. I fixed the typo in both files. Thanks for spotting this!

Issue #287, opened 2016-08-09
Upload all ai.0.js versions to /dist folder


kamilszostak:
Keep all shipped versions of JS SDK in /dist folder, not only the latest one.

PR #288, opened 2016-08-10
Upload all past releases to /Dist folder

kamilszostak:


kamilszostak:
Fixes #287

dnduffy:
:shipit: 
My only question is do we have any logs or telemetry about which ones are being requested?  It may be that we really only needed a few of the old versions but there should be no problem with having them all, more or a curiosity about what is still being requested.

PR #291, opened 2016-08-22
Enable retry handler

kamilszostak:
Enable retry handler for 10% of users.

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
@@ -1170,7 +1170,7 @@ class SenderTests extends TestClass {
         var config = this.getDefaultConfig();
 
         // disable partial response handling
-        config.disableRetry = () => true;
+        config.isRetryDisabled = () => true;
 
         var sender = <SenderWrapper>new Microsoft.ApplicationInsights.Sender(config);
 
@@ -1223,7 +1223,7 @@ class SenderTests extends TestClass {
             maxBatchInterval: () => this.maxBatchInterval,
             disableTelemetry: () => this.disableTelemetry,
             enableSessionStorageBuffer: () => true,
-            disableRetry: () => false
+            isRetryDisabled: () => false
         };
     }
 }
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
@@ -16,7 +16,7 @@ class SenderTests extends TestClass {
     private xdr;
     private fakeServer: SinonFakeServer;
     private getSender: () => SenderWrapper;
-    private loggingSpy;
+    private loggingSpy: SinonStub;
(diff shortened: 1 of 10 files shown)


JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -759,7 +827,7 @@ class SenderTests extends TestClass {
                 // setup
                 XMLHttpRequest = <any>(() => {
                     var xhr = new this.xhr;
-                    xhr.withCredentials = false;
+                    delete xhr.withCredentials;
```
dnduffy: I'm curious why the property is deleted?

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -368,12 +401,20 @@ module Microsoft.ApplicationInsights {
         public _xhrReadyStateChange(xhr: XMLHttpRequest, payload: string[], countOfItemsInPayload: number) {
             if (xhr.readyState === 4) {
                 if ((xhr.status < 200 || xhr.status >= 300) && xhr.status !== 0) {
-                    this._onError(payload, xhr.responseText || xhr.response || "");
+                    if (!this._config.disableRetry() && this._isRetriable(xhr.status)) {
```
dnduffy: When using config.disableRetry() it reads like a command which will disable retrying instead of a getter for a flag.
I think it would be cleared if it were renamed to something like "isRetryDisabled" so that it is clear it is a check and not a command.

dnduffy:
Minor rename and then :shipit:

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
kamilszostak: This is to pretend that the browser is IE - see [Sender](https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/Sender.ts#L148)

JavaScript/JavaScriptSDK/Sender.ts
kamilszostak: I agree, but I was trying to keep it consistent with other configuration switches (.e.g disableFlushOnBeforeUnload, disableAjaxTracking or  enableSessionStorageBuffer).

dnduffy: I'm guessing we can't rename them now that it is 1.0.0?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
dnduffy: I understand now, thanks. :)

JavaScript/JavaScriptSDK/Sender.ts
kamilszostak: Yeah, we can't rename existing switches (somebody can be using them in a snippet). `disableRetry` is new, I'll rename it to `isRetryDisabled`. Thanks!

PR #294, opened 2016-08-24
Snippet versioning

kamilszostak:
Add version number to JS SDK snippet.


JavaScript/JavaScriptSDK.Tests/E2ETests/snippet.tests.ts
```diff
@@ -106,6 +106,35 @@ class SnippetTests extends TestClass {
...
+                    this.loadSnippet(path70);
+                },
+                () => {
+                    sendSpy = this.setAppInsights();
+
+                    window["appInsights"].trackPageView("test", { property: "p1" }, { measurement: 5 });
+                },
+                () => {
+                    Assert.equal(this.queueCallCount, this.queueSpy.callCount, "queue is emptied");
+                    Assert.equal(1, sendSpy.sender.callCount, "v2 send called");
```
dnduffy: Aren't the messages in asserts shown when the assert fails?  They read like they are success messages so maybe I misunderstand the expected behavior.

dnduffy:
👍

JavaScript/JavaScriptSDK.Tests/E2ETests/snippet.tests.ts
kamilszostak: I was trying to keep them consistent with existing tests in this file. I'll change them.

dnduffy: Oh, I thought it was like C# where it would just write "queue is emptied" or "v2 send called" which seems to be opposite of a test failure message.

Issue #293, opened 2016-08-24
startTrackEvent is not stubbed/queued


SergeyKanzhelev:
Moved from: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/252 (reported by @dfederm)

>It looks like in the JavaScriptSnippet, these events are stubbed out with an implementation that basically queues up the invocation until the real appInsights impl is on the page:
["Event","Exception","Metric","PageView","Trace","Ajax"]

>However, this does not encompass all functions on the appInsights object. For example startTrackEvent.

>This can cause lost instrumentation, as well as JS errors when the AI js fails to load (or is blocked by an adblocker in my case).

>Here is an example of the JS error as well as an example of the "trackEvent" function being stubbed out which avoids the JS error (although still loses the instrumentation due to the adblocker):
![image](https://cloud.githubusercontent.com/assets/6445614/17939188/3c807f70-69de-11e6-8a8d-b9e732605369.png)

dfederm:
FYI, I was able to mitigate this by adding this just under the `@Html.ApplicationInsightsJavaScript(TelemetryConfiguration)` call:

```
    <script>
        (function(t){function r(f){if(!t[f])t[f]=function(){var i=arguments;t.queue.push(function(){t[f].apply(t,i)})}};r("startTrackEvent");r("stopTrackEvent")})(appInsights)
    </script>

```

PR #295, opened 2016-08-25
Add missing lazy methods

dfederm:
Fixes issue #293.

I ran `powershell "& .\RunTestsInBrowser.ps1"` and it's all green. I'd like to add a test for this but the current snippet tests seem to hardcode the config so I couldn't think of a great way to simulate the ai.js script not properly loading to test the lazy methods are properly set up. I'm open to suggestion :)

### DIFF
--- a/JavaScript/JavaScriptSDK/snippet.js
+++ b/JavaScript/JavaScriptSDK/snippet.js
@@ -7,6 +7,12 @@
     var localDocument = document;
     var localWindow = window;
     var scriptText = "script";
+    var userContext = "AuthenticatedUserContext";
+    var start = "start";
+    var stop = "stop";
+    var track = "Track";
+    var trackEvent = track + "Event";
+    var trackPage = track + "Page";
     var scriptElement = localDocument.createElement(scriptText);
     scriptElement.src = aiConfig.url || "CDN_PATH";
     localDocument.getElementsByTagName(scriptText)[0].parentNode.appendChild(scriptElement);
@@ -36,14 +42,14 @@
         createLazyMethod("track" + method.pop());
     }
 
-    createLazyMethod("setAuthenticatedUserContext");
-    createLazyMethod("clearAuthenticatedUserContext");
+    createLazyMethod("set" + userContext);
+    createLazyMethod("clear" + userContext);
 
-    createLazyMethod("startTrackEvent");
-    createLazyMethod("stopTrackEvent");
+    createLazyMethod(start + trackEvent);
+    createLazyMethod(stop + trackEvent);
 
-    createLazyMethod("startTrackPage");
-    createLazyMethod("stopTrackPage");
+    createLazyMethod(start + trackPage);
+    createLazyMethod(stop + trackPage);
 
     createLazyMethod("flush");
 
--- a/JavaScript/JavaScriptSDK/snippet.js
+++ b/JavaScript/JavaScriptSDK/snippet.js
@@ -39,6 +39,14 @@
     createLazyMethod("setAuthenticatedUserContext");
     createLazyMethod("clearAuthenticatedUserContext");
 
+    createLazyMethod("startTrackEvent");
+    createLazyMethod("stopTrackEvent");
+
+    createLazyMethod("startTrackPage");
+    createLazyMethod("stopTrackPage");
+
+    createLazyMethod("flush");
+
     // collect global errors
     if (!aiConfig.disableExceptionTracking) {
         method = "onerror";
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -39,6 +39,14 @@
     createLazyMethod("setAuthenticatedUserContext");
     createLazyMethod("clearAuthenticatedUserContext");
 
+    createLazyMethod("startTrackEvent");
+    createLazyMethod("stopTrackEvent");
+
+    createLazyMethod("startTrackPage");
+    createLazyMethod("stopTrackPage");
+
+    createLazyMethod("flush");
```
kamilszostak: Assigning duplicated strings to local variables allows them to be minified to safe space. 
This will reduce the snippet size by 13 bytes.

    const userContext = "AuthenticatedUserContext";
    const trackEvent = "TrackEvent";
    const trackPage = "TrackPage";
    const start = "start";
    const stop = "stop";

    createLazyMethod("set" + userContext);
    createLazyMethod("clear" + userContext);

    createLazyMethod(start + trackEvent);
    createLazyMethod(stop + trackEvent);
    
    createLazyMethod(start + trackPage);
    createLazyMethod(stop + trackPage);

kamilszostak:
Thank you for fixing this issue. 

We have end-to-end tests in [snippet.tests.ts](https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK.Tests/E2ETests/snippet.tests.ts). I started refactoring and cleaning up those tests couple days ago, when I worked on #294. 

I'll add unit tests for your scenario after I finish with the cleanup work.

README.md
```diff
@@ -22,7 +22,7 @@ from CDN or you can override the script hosting location by specifying `url` par
...
+        function r(config){t[config]=function(){var i=arguments;t.queue.push(function(){t[config].apply(t,i)})}}var t={config:config},u=document,e=window,o="script",s=u.createElement(o),i,f;s.src=config.url||"https://az416426.vo.msecnd.net/scripts/a/ai.0.js";u.getElementsByTagName(o)[0].parentNode.appendChild(s);try{t.cookie=u.cookie}catch(h){}for(t.queue=[],i=["Event","Exception","Metric","PageView","Trace","Dependency"];i.length;)r("track"+i.pop());return r("setAuthenticatedUserContext"),r("clearAuthenticatedUserContext"),r("startTrackEvent"),r("stopTrackEvent"),r("startTrackPage"),r("stopTrackPage"),r("flush"),config.disableExceptionTracking||(i="onerror",r("_"+i),f=e[i],e[i]=function(config,r,u,e,o){var s=f&&f(config,r,u,e,o);return s!==!0&&t["_"+i](config,r,u,e,o),s}),t
     }({
         instrumentationKey: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxx"
     });
```
dnduffy: This snippet also needs to get updated in our .NET Core SDK repo: [Resources.resx](https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/master/src/Microsoft.ApplicationInsights.AspNetCore/Resources.resx)

dnduffy:
👍

JavaScript/JavaScriptSDK/snippet.js
dfederm: Great, thanks for the suggestion!

dfederm:
FYI, as measured by the snippet in the readme (which is what I assume most users will use if they use the snippet at all), this change actually reduces the bytes from 1009 to 904. I believe that was mainly due to some mistaken find/replaces in the existing readme snippet though :).

As an aside, after this goes in what's the process for updating the aspnetcore copy of the snippet? Is there an automated process for it or should I just update it in that repo as well?

kamilszostak:
:shipit:

dnduffy:
Just submit a PR for updating the .NET Core SDK repo separately and we'll review it as soon as we see it.

README.md
dfederm: Sure, that's actually wehre I started before the bug got moved to this repo :), I can do that as well. Not sure if y'all prefer this to get merged in first or have both PRs out at the same time. I'll go ahead and prep the change there now.

---
In reply to: [76287186](https://github.com/Microsoft/ApplicationInsights-JS/pull/295#discussion_r76287186) [](ancestors = 76287186)

dnduffy: Doesn't really matter but I've merged this one now anyway.

PR #316, opened 2016-09-16
SetCookie performance improvements

kamilszostak:


JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -76,11 +79,14 @@ module Microsoft.ApplicationInsights.Context {
...
                 // update automaticSession so session state has correct id                
-                this.automaticSession.isFirst = undefined; 
+                this.automaticSession.isFirst = undefined;
                 this.renew();
             } else {
-                this.automaticSession.renewalDate = +new Date;
-                this.setCookie(this.automaticSession.id, this.automaticSession.acquisitionDate, this.automaticSession.renewalDate);
+                // do not update the cookie more often than cookieUpdateInterval
+                if (!this.cookieUpdatedTimestamp || now - this.cookieUpdatedTimestamp > _SessionManager.cookieUpdateInterval) {
+                    this.automaticSession.renewalDate = +new Date;
```
toddreifsteck: I'm a member of the Microsoft Edge team and co-chair of the W3C Web Performance Working Group. Was forwarded this so pardon if we've not met! I recommend using performance.now rather than Date for things such as this. performance.now is monotonically increasing where Date can have skew.

JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -190,6 +196,8 @@ module Microsoft.ApplicationInsights.Context {
             var cookieDomnain = this.config.cookieDomain ? this.config.cookieDomain() : null;
 
             Util.setCookie('ai_session', cookie.join('|') + ';expires=' + cookieExpiry.toUTCString(), cookieDomnain);
+
+            this.cookieUpdatedTimestamp = +new Date;
```
toddreifsteck: I recommend using performance.now in place of Date. See above.

JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -76,11 +79,14 @@ module Microsoft.ApplicationInsights.Context {
             // renew if acquisitionSpan or renewalSpan has ellapsed
             if (acquisitionExpired || renewalExpired) {
                 // update automaticSession so session state has correct id                
-                this.automaticSession.isFirst = undefined; 
+                this.automaticSession.isFirst = undefined;
                 this.renew();
```
nolanlawson: `renew()` is also calling `setCookie()` ultimately, and doesn't seem to be guarded by the same "only update cookie every X milliseconds logic."

Rather than manual `new Date()` checking, would it be possible to wrap the entire `update` function in a throttled function? For instance you could use Lodash `_.throttle` or just `setTimeout`/`setInterval` to achieve this. E.g.:

```js
this.update = _.throttle(originalUpdateFunction.bind(this), 60000);
```

JavaScript/JavaScriptSDK/Context/Session.ts
kamilszostak: Done. We have a [dataTime.Now()](https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/ajax/ajaxUtils.ts#L34) function which uses performance.now internally.

JavaScript/JavaScriptSDK/Context/Session.ts
kamilszostak: Done.

JavaScript/JavaScriptSDK/Context/Session.ts
kamilszostak: Thanks for spotting this. Currently `renew()` is only called from `update()` (which has a throttling check) and `initializeAutomaticSession()` (which is only called once per page view). I agree that a throttled function is a right solution, but since it is not currently required I stick with the current implementation.

kamilszostak:
I'm resolving remaining comments in this pull request #339

PR #318, opened 2016-09-19
XDomain sender should drop the telemetry if the Endpoint protocol and the hosting page protocol don't match

### LINKED ISSUE #314: JS SDK should not send telemetry over http
HTTPs is already used by default. Due to XDomain limitations, old IE browsers are still using HTTP if a hosting website is served over HTTP. In such case, the SDK should not send telemetry data.

kamilszostak:
XDomainRequest requires the same protocol as the hosting page. If the protocol doesn't match, we can't send the telemetry :(

Fixes #314

### DIFF
--- a/JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
+++ b/JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
@@ -29,7 +29,7 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
                     return (this.sampleRate == 100) ? FieldType.Hidden : FieldType.Required;
                 },
                 tags: FieldType.Required,
-                data: FieldType.Required,
+                data: FieldType.Required
             };
         }
     }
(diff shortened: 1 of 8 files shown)


JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
```diff
@@ -29,7 +29,7 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
                     return (this.sampleRate == 100) ? FieldType.Hidden : FieldType.Required;
                 },
                 tags: FieldType.Required,
-                data: FieldType.Required,
```
dnduffy: Technically the trailing commas are okay on the last row.  I've always liked not having to add them when adding new rows but it doesn't really matter.

dnduffy:
:+1:

JavaScript/JavaScriptSDK/Telemetry/Common/Envelope.ts
kamilszostak: They are not ok in IE7 and IE8 :cry:. Minimizer will remove them, but if you want to debug an unminified script, older IE's will throw an exception.

Fun fact: `[1,].length` returns 2 in IE7 :)

dnduffy: Wow that is really messed up for old IE to not allow them!

PR #320, opened 2016-09-26
Move snippet version to sdkVersion

kamilszostak:


SergeyKanzhelev:
Ideally you need to provide a way for extensions to add this postfix as well.

PR #325, opened 2016-10-17
Add support for tracking properties and measurements in trackDependency

### LINKED ISSUE #322: Pass properties object in `trackDependency`
Is there a plan to introduce `properties` in the `trackDependency` API ([ref](https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/AppInsights.ts#L273)) like `trackException` or other API's?

In my case, I want to track some custom properties with my Ajax call which signifies the source component of the call (for eg.)

DennisPan:
Fix #322


kamilszostak:
:shipit:

kamilszostak:
I also created an issue #327 to update all schemas based on the latest [bonds](https://github.com/Microsoft/ApplicationInsights-Home/tree/master/EndpointSpecs/Schemas/Bond).

PR #333, opened 2016-10-26
Enable correlation headers

### LINKED ISSUE #331: Why is disableCorrelationHeaders setting true by default?
Shouldn't it be enabled by default?
Another question is why do we need a check to only limit setting them for the same hostname only. In case my server-side application (instrumented with the same ikey) happens to be hosted on a sever with different name, I still will benefit from the feature. And if they are different applications, still no harm done, I just won't see correlated telemetry.

kamilszostak:
Fixes #331.


nicosabena:
This is a breaking change, because now external sites will be getting those additional headers and might reject them.

PR #336, opened 2016-10-27
Expose an option to disable cookies

### LINKED ISSUE #335: option to disable cookies completely
In order to use AI with a customer, we need to remove all cookies. There is already `Util.canUseCookies()`, could this function take an global option to disable it? Same for localstorage.

kamilszostak:
Fixes #335

### DIFF
--- a/API-reference.md
+++ b/API-reference.md
@@ -263,6 +263,10 @@ Values that control how the telemetry data is sent.
         // Default 500 - controls how many ajax calls will be monitored per page view.
         // Set to -1 to monitor all ajax calls on the page.
         maxAjaxCallsPerView: number;
+
+        // If true, the SDK will not store or read any data from cookies.
+        // Default: false
+        isCookieUseDisabled: boolean;
         
         // Custom cookie domain. This is helpful if you want to share Application Insights cookies across subdomains.
         cookieDomain: string;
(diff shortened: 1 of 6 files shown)


API-reference.md
```diff
@@ -281,6 +281,10 @@ Values that control how the telemetry data is sent.
         // The url from where the JS SDK will be downloaded. 
         // Default 'https://az416426.vo.msecnd.net/scripts/a/ai.0.js'
         url: string;
+
+        // If true, the SDK will not store or read any data from cookies.
+        // Default: false
+        disableCookies: boolean;
```
SergeyKanzhelev: can you move it closer to `cookieDomain` setting so they will be sitting together.

pelikhan:
Can there be a flag to disable local storage as well?

API-reference.md
kamilszostak: Done

kamilszostak:
I updated the pull request. 

* Exposed and option to disable session and local storage
* CanUseLocal and CanUseSession storage methods now cache the results, so it will be slightly faster
* Renamed configuration fields : `isCookieUseDisabled` and `isStorageUseDisabled`

pelikhan:
Looks great!

Issue #337, opened 2016-10-31
Address feedback from #316


kamilszostak:
Address feedback from @toddreifsteck and @nolanlawson in https://github.com/Microsoft/ApplicationInsights-JS/pull/316

PR #339, opened 2016-11-01
Use performance.now in session class

kamilszostak:
Fixing #337

### DIFF
--- a/JavaScript/JavaScriptSDK/Context/Session.ts
+++ b/JavaScript/JavaScriptSDK/Context/Session.ts
@@ -71,7 +71,7 @@ module Microsoft.ApplicationInsights.Context {
                 this.initializeAutomaticSession();
             }
 
-            var now = +new Date;
+            var now = dateTime.Now();
 
             var acquisitionExpired = now - this.automaticSession.acquisitionDate > this.config.sessionExpirationMs();
             var renewalExpired = now - this.automaticSession.renewalDate > this.config.sessionRenewalMs();
@@ -84,7 +84,7 @@ module Microsoft.ApplicationInsights.Context {
             } else {
                 // do not update the cookie more often than cookieUpdateInterval
                 if (!this.cookieUpdatedTimestamp || now - this.cookieUpdatedTimestamp > _SessionManager.cookieUpdateInterval) {
-                    this.automaticSession.renewalDate = +new Date;
+                    this.automaticSession.renewalDate = +new Date(now);
                     this.setCookie(this.automaticSession.id, this.automaticSession.acquisitionDate, this.automaticSession.renewalDate);
                 }
             }
@@ -197,7 +197,7 @@ module Microsoft.ApplicationInsights.Context {
 
             Util.setCookie('ai_session', cookie.join('|') + ';expires=' + cookieExpiry.toUTCString(), cookieDomnain);
 
-            this.cookieUpdatedTimestamp = +new Date;
+            this.cookieUpdatedTimestamp = dateTime.Now();
         }
 
         private setStorage(guid: string, acq: number, renewal: number) {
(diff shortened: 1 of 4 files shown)


JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -71,7 +71,7 @@ module Microsoft.ApplicationInsights.Context {
                 this.initializeAutomaticSession();
             }
 
-            var now = +new Date;
+            var now = dateTime.Now();
```
toddreifsteck: So dateTime.Now() is actually calling performance.now()?

JavaScript/JavaScriptSDK/Context/Session.ts
```diff
@@ -84,7 +84,7 @@ module Microsoft.ApplicationInsights.Context {
             } else {
                 // do not update the cookie more often than cookieUpdateInterval
                 if (!this.cookieUpdatedTimestamp || now - this.cookieUpdatedTimestamp > _SessionManager.cookieUpdateInterval) {
-                    this.automaticSession.renewalDate = +new Date;
+                    this.automaticSession.renewalDate = +new Date(now);
```
toddreifsteck: Passing performance.now to construct a Date is not an accurate way to use it. Instead, you would remove the use of the Date type and replace it and all Date comparisons on this field with performance.now lookup's/comparisons.

JavaScript/JavaScriptSDK/Context/Session.ts
kamilszostak: Yes, it's a wrapper for performance.now, which also checks if the function is supported. See [ajaxUtils.ts](https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/ajax/ajaxUtils.ts#L32)

toddreifsteck: K, the comment on that method isn't accurate. You may want to correct that. performance.now is the time since the context was created and not the time since 1970.

kamilszostak: Actually, we want to use it instead of Date, so it needs to be a number of milliseconds since 1970. 
The correct value is `window.performance.now() + window.performance.timing.navigationStart`.

toddreifsteck: If you really MUST have a date, that is accurate.

toddreifsteck: You will also have code that only runs in the window context. In the worker context, you won't have a navigationStart so you'll need another technique for determining when the timeOrigin (aka timing.navigationStart) is.

kamilszostak: We use it to set a timestamp of each event and sent it to AI backend, so it has to be a Date. The code will fall back to `Date().getTime()` if `window.performance` is not available.

JavaScript/JavaScriptSDK/Context/Session.ts
kamilszostak: Fixed.

JavaScript/JavaScriptSDK/ajax/ajaxUtils.ts
```diff
@@ -31,17 +31,17 @@ module Microsoft.ApplicationInsights {
 
     export class dateTime {
         ///<summary>Return the number of milliseconds since 1970/01/01 in local timezon</summary>
-        public static Now = (window.performance && window.performance.now) ?
+        public static Now = (window.performance && window.performance.now && window.performance.timing) ?
             function () {
-                return performance.now();
+                return window.performance.now() + window.performance.timing.navigationStart;
```
toddreifsteck: This is probably fine. :-) Be aware that mixing this and Date.now() can cause variance of up to ~16 ms as they are not guaranteed to use the same internals for calculating their times. Again.. probably fine.

PR #340, opened 2016-11-02
Resource timing

KamilSzostak:
* Remove performance analyzer
* Add ResourceTiming manager (disabled by default)


kamilszostak:
I'm working on unit tests and I'll add them before I merge to master.

Issue #327, opened 2016-10-17
Update bond schemas


KamilSzostak:
Update all schemas based on the latest version from https://github.com/Microsoft/ApplicationInsights-Home/tree/master/EndpointSpecs/Schemas/Bond 

This will make changes like [this](https://github.com/Microsoft/ApplicationInsights-JS/pull/325) easier.

PR #372, opened 2016-12-30
Add target to rdd data, update command name and name for RDD

kartang:


JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -1454,8 +1454,10 @@ class AppInsightsTests extends TestClass {
             test: () => {
                 var appInsights = new Microsoft.ApplicationInsights.AppInsights(this.getAppInsightsSnippet());
                 var trackStub = this.sandbox.stub(appInsights.context, "track");
-                var name = "test";
-                var url = "http://myurl.com";
+                var pathName = "https://tempurl.net/api/temp/ABCD";
+                var commandName = "GET https://tempurl.net/api/temp/ABCD?param1=test&param2=test";
+                var url = "https://tempurl.net/api/temp/ABCD?param1=test&param2=test";
```
dnduffy: It doesn't really matter but I would probably have built url as pathName + the param string and then build commandName as "GET " + url so there would only be one copy of the string to type/maintain.

kartang: Makes sense. Will update that. Thanks!

JavaScript/JavaScriptSDK/Telemetry/RemoteDependencyData.ts
```diff
@@ -39,8 +40,10 @@ module Microsoft.ApplicationInsights.Telemetry {
             super();
 
             this.id = id;
-            this.name = this.formatDependencyName(method, absoluteUrl);
-            this.commandName = Common.DataSanitizer.sanitizeUrl(commandName);
+            this.target = UrlHelper.parseUrl(absoluteUrl).hostname;
+            this.commandName = this.formatDependencyName(method, absoluteUrl);
```
OsvaldoRosado: What about the `data` field that replaces `commandName`?

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -1607,8 +1611,9 @@ class AppInsightsTests extends TestClass {
                 // Assert
                 Assert.ok(trackStub.called, "Track should be called");
                 var rdd = <Microsoft.ApplicationInsights.Telemetry.RemoteDependencyData>(<any>trackStub.args[0][0]).data.baseData;
-                Assert.equal(url, rdd.name);
-                Assert.equal(name, rdd.commandName);
+                Assert.equal(pathName, rdd.name);
+                Assert.equal(url, rdd.commandName);
```
OsvaldoRosado: Is this right? The url variable holds the plain url, shouldn't the commandName object have "GET "+url?

kartang: The method here is null, and in that case, url will be the command name.

JavaScript/JavaScriptSDK/Telemetry/RemoteDependencyData.ts
kartang: I believe the transformation is handled in breeze? I though breeze today expects commandName  as opposed to data?

kamilszostak:
@SergeyKanzhelev Updating schema might be a breaking change. I would suggest updating in in next minor release (1.1) - #327.

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -318,5 +320,67 @@ module Microsoft.ApplicationInsights {
...
+            try {
+                var data = (<any>envelope).data;
+                if (data && data.baseType === Microsoft.ApplicationInsights.Telemetry.RemoteDependencyData.dataType) {
+                    var rddData = <Microsoft.ApplicationInsights.Telemetry.RemoteDependencyData>(<any>data.baseData);
+                    if (rddData) {
+                        this._fixRDDDepricatedValues(rddData);
+                    }
+                }
+            } catch (e) {
+            }
```
kamilszostak: Consider logging a warning message. It won't be reported unless a user is in the debug mode or has enabled verboseLogging. 

sample code: 
                _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.WARNING,
                    new _InternalLogMessage(_InternalMessageId.<newEnumName>, "error message" + Util.getExceptionName(e),
                        { exception: Util.dump(e) }));

kartang: Thanks! Updated the logging!

JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -331,6 +331,9 @@ module Microsoft.ApplicationInsights {
                     }
                 }
             } catch (e) {
+                _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.WARNING,
+                    new _InternalLogMessage(_InternalMessageId.NONUSRACT_FailedToFixDepricatedValues, "Failed to parse the base data object, to fix the depricated values " + Util.getExceptionName(e),
```
OsvaldoRosado: Typo: Depricated -> Deprecated

JavaScript/JavaScriptSDK.Tests/JavaScriptSDK.Tests.csproj
```diff
@@ -140,6 +140,9 @@
     <Content Include="E2ETests\standalone\sprint69.html" />
     <Content Include="E2ETests\standalone\test.html" />
     <Content Include="E2ETests\standalone\tests.js" />
+    <Content Include="geckodriver.exe">
```
dnduffy: Why is it "Content"?

Issue #367, opened 2016-12-12
Page views issued by GoogleBot have very long duration


kamilszostak:
JS SDK uses navigation timing API to get the page view duration. When a web page is visited by a GoogleBot navigation data provided by the API is invalid. E.g. networkDuration=81,328,397ms (~23h).

PR #378, opened 2017-01-10
Don't report duration if it's >= 5min

kamilszostak:


kamilszostak:
Fixing #367.

JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
```diff
@@ -137,12 +146,22 @@ module Microsoft.ApplicationInsights.Telemetry {
...
+            var duration = undefined;
             if (!(isNaN(start) || isNaN(end))) {
                 duration = Math.max(end - start, 0);
             }
 
             return duration;
         }
+
+        public static maxDurationExceeded(...durations: number[]): boolean {
+            for (var i = 0; i < durations.length; i++) {
```
AlexBulankou: why is it public? Doesn't seem like it is called anywhere outside PageViewPerformance class.

kamilszostak: PageViewManager is also using this method.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Telemetry/PageViewPerformance.tests.ts
```diff
@@ -124,6 +150,61 @@ class PageViewPerformanceTelemetryTests extends ContractTestHelper {
...
+                    Assert.equal(undefined, data.networkConnect);
+                    Assert.equal(undefined, data.sentRequest);
+                    Assert.equal(undefined, data.receivedResponse);
+                    Assert.equal(undefined, data.domProcessing);
+
+                    if (i === 0) {
+                        // check props only for the first timingModifier
+                        Assert.equal("AI (Internal): NONUSRACT_MaxDurationExceeded message:\"exceeded maximum duration value (5min). Browser perf data won't be sent.\" props:\"{total:300000,network:1,request:27,response:12,dom:299959}\"", actualLoggedMessage);
+                    } else {
+                        Assert.ok(actualLoggedMessage.lastIndexOf("AI (Internal): NONUSRACT_MaxDurationExceeded message:\"exceeded maximum duration value (5min)", 0) === 0);
```
dnduffy: Why is the cutoff 5 minutes?
Can we detect bad values without an explicit time cutoff point?

kamilszostak: Durations reported by bots are >1h (usually in a 2-5h range). I can change the code to exclude durations longer than 1h. I can also check a user agent to see if it's a GoogleBot. 

I don't want to remove a valid page load duration (event if it's 5min+), so I'll change it to use more strict rules first (duration >1h, user agent == GoogleBog). If our users will still see invalid values reported by navigation.timing API, we will make rules less strict.

JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts
```diff
@@ -137,12 +146,31 @@ module Microsoft.ApplicationInsights.Telemetry {
...
+
+        /**
+         * GoogleBot is returning invalid values in performance.timing API.
+         * This method tells if given durations should be excluded from collection.
+         */
+        public static shouldCollectDuration(...durations: number[]): boolean {
+            let userAgent = navigator.userAgent;
+            let isGoogleBot = userAgent ? userAgent.toLowerCase().indexOf("googlebot") !== -1 : false;
+
+            if (isGoogleBot) {
```
dnduffy: Shouldn't this be an or condition?  It is GoogleBot, or it is > MAX_DURATION_ALLOWED?
Maybe GoogleBot pretends to be IOS and still reports 3 hours.

kamilszostak: I still want to report page duration from GoogleBot if it gives accurate numbers. Let's keep the rule more strict for now. We can revisit it after a week if it's not enough.

PR #394, opened 2017-02-02
Beacon API support

KamilSzostak:
Adds experimental support for a [Beacon API](https://www.w3.org/TR/beacon/). The feature is disabled by default but can be enabled by setting `isBeaconApiDisabled` config to `false`.


SergeyKanzhelev:
Will we know that the data came via `sendBeacon`? 

`sendBeacon` is designed to send data that otherwise can be lost on page unload. So it doesn't provide re-try. Should SDK implementation use regular http when page is not being unloaded yet for more chances to send data?

JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -407,6 +419,29 @@ module Microsoft.ApplicationInsights {
...
+            var queued = (<any>navigator).sendBeacon(url, batch);
+
+            if (queued) {
+                this._buffer.markAsSent(payload);
+            } else {
+                _InternalLogging.throwInternalNonUserActionable(LoggingSeverity.WARNING,
+                    new _InternalLogMessage(_InternalMessageId.NONUSRACT_TransmissionFailed, ". " +
+                        "Failed to send telemetry with Beacon API."));
+
+                // consider switching back to XHR
```
SergeyKanzhelev: would warnings be visible for customer so they know to switch `sendBeacon` off? I'd say it is important to implement the fallback before releasing a version of SDK even for internal testing.

kamilszostak: Yes, warnings are visible in the console. I agree I'll add a fallback logic.

kamilszostak:
We won't know if a browser sent data using Beacon API, the same way as we are unable to tell if a browser sent it using XDomain or XmlHttp request. 

Beacon API is designed to let the browser decide when is the best time to send telemetry - it can even happen when a user had already left the page. Yes, the API doesn't allow to handle response codes, so we are unable to handle errors (retry on server error, partial success, exponential backoff, etc.). 
The goal is to give flexibility to our users to choose which data transmission method works best for them (AJAX or Beacon API) - both of them have their own pros and cons.

SergeyKanzhelev:
@KamilSzostak - instead of being flexible and require user to decide on communication API - we may use beacon when page is unloading and regular ajax when it's not. Complete switch to beacon API will require the change in endpoint behavior so we will accept telemetry more often.

JavaScript/JavaScriptSDK/Sender.ts
SergeyKanzhelev: My point was that we need to give visibility for a customer that after the switch to beacon so much telemetry got lost... Or something along this lines. Maybe even make it "error" instead of "warning"

KamilSzostak: I changed it to error. I'll write a wiki page to describe pros and cons of both senders (AJAX-based vs Beacon).

KamilSzostak:
@SergeyKanzhelev we shouldn't rely on any code which is running when a page is unloading. A browser may terminate the javascript execution and not record a Beacon API call. 
I'm not planning on making the Beacon API a default solution to send telemetry, I see it as a feature which our customers can use in case a default solution doesn't fully work for them. For instance: user doesn't care if 100% of his telemetry is recorded, but he's trying to optimize page performance and PLT on mobile devices. In this case, a Beacon API might be a better choice, because a browser will decide what is the is the best time to send data. I'll prepare a Wiki page which explains pros and cons of both solutions. 

This doesn't require any change on the backend side to work. The AppInsights backend will return a response and JS SDK simply won't be able to process it. We may in the future send events to a fast-track endpoint (if we ever decide to implement it).

I agree, it is a good idea to call `sendBeacon` more often and let the browser decide when is the best time to send the actual request to the server. This can be achieved by changing the `maxBatchInterval` value to let's say 5s (instead of 15s which is a default value) or a `maxBatchSize` which I already changed to `64kb`.

PR #400, opened 2017-02-09
msToTimeSpan handles float numbers

kamilszostak:


JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -525,15 +527,6 @@ module Microsoft.ApplicationInsights {
 
             return true;
         }
-
-        public static getDurationString(valueInMS: number): string {
-            if (valueInMS >= 0) {
-                var date = new Date(valueInMS);
-                return (date.getUTCDate() - 1) + "." + date.getUTCHours() + ":" + date.getUTCMinutes() + ":" + date.getUTCSeconds() + "." + date.getUTCMilliseconds();
-            } else {
-                return "0.0:0:0.0";
```
OsvaldoRosado: If we're removing this, should msToTimeSpan gain the ability to handle days? getDurationString could handle it

kamilszostak: Done.

PR #402, opened 2017-02-09
Powershell script to check the sdk size

KamilSzostak:


scripts/CheckSdkSize.ps1
```diff
@@ -0,0 +1,56 @@
...
+
+if(-not (test-path($aiMinZip))) {
+    set-content $aiMinZip ("PK" + [char]5 + [char]6 + ("$([char]0)" * 18))
+    (dir $aiMinZip).IsReadOnly = $false  
+}
+
+$shellApplication = new-object -com shell.application
+$zipPackage = $shellApplication.NameSpace($aiMinZip)
+$zipPackage.CopyHere($aiMin)
+Start-sleep -milliseconds 500
```
dnduffy: What is the sleep for?

scripts/CheckSdkSize.ps1
```diff
@@ -0,0 +1,56 @@
...
+
+$metrics = New-Object 'system.collections.generic.dictionary[[string],[double]]'
+$metrics.Add('ai', $aiSize);
+$metrics.Add('ai.min', $aiMinSize);
+$metrics.Add('a.min.zip', $aiMinZipSize);
+
+$telemetry.TrackEvent("JsSdkSize", $null, $metrics);
+$telemetry.Flush();
+
+$MAX_SDK_SIZE = 25 * 1025; # size in bytes
```
dnduffy: 1024 vs. 1025?

scripts/CheckSdkSize.ps1
```diff
@@ -0,0 +1,56 @@
...
+$telemetry = New-Object "Microsoft.ApplicationInsights.TelemetryClient"
+$telemetry.InstrumentationKey = $ikey;
+
+$metrics = New-Object 'system.collections.generic.dictionary[[string],[double]]'
+$metrics.Add('ai', $aiSize);
+$metrics.Add('ai.min', $aiMinSize);
+$metrics.Add('a.min.zip', $aiMinZipSize);
+
+$telemetry.TrackEvent("JsSdkSize", $null, $metrics);
+$telemetry.Flush();
```
dnduffy: Should the telemetry calls be wrapped in a try catch so the script won't fail if AI can't send (or does the SDK swallow errors)?

scripts/CheckSdkSize.ps1
kamilszostak: To give some time to copy the file to a zip. Otherwise, it's showing that the zip files has only 24b.

scripts/CheckSdkSize.ps1
kamilszostak: How did that happen? I need more coffee :)

scripts/CheckSdkSize.ps1
dnduffy: How do you know it finished in 500 milliseconds?

scripts/CheckSdkSize.ps1
dnduffy: ;)

scripts/CheckSdkSize.ps1
OsvaldoRosado: I don't believe the SDK is supposed to ever throw, so that we don't crash production web servers

dnduffy: True, I'm "paranoid".

scripts/CheckSdkSize.ps1
kamilszostak: I knew you're gonna ask about that... I'll check what I can do :)

KamilSzostak: OK. I'm gonna use the GzipStream from C# which is synchronous.

Issue #289, opened 2016-08-16
Add a callback to notify that the SDK failed to send telemetry data


kamilszostak:

sorgloomer:
I can see at least two ways how this could be implemented:

1: Adding an extra parameter to the `track*` calls, like `trackEvent("MyEvent", myProps, { onSuccess: handleSuccess });`. This way the user could be notified about every single telemetry item separately.

2: Adding global event handlers to the `appInsights` object: `appInsights.addEventListener("sendSucceeded", handleSuccess);`. This way the user gets to know whether telemetry sending works in general, and the onSuccess callback might even imply that the sending queue exhausted.

I personally prefer the second one, it would solve my current problem in #313.

Issue #208, opened 2016-05-24
Refactor DataLossAnalyzer


KamilSzostak:
See- #204.

Change DataLossAnaluzer to hook to onSuccess/onError callbacks from #289.

PR #404, opened 2017-02-10
Remove DataLossAnalyzer and PerfAnalyzer

KamilSzostak:
Comment out DataLossAnalyzer and remove PerfAnalyzer classes. Both functionalities are turned off and were designed for SDK internal use only.


JavaScript/JavaScriptSDK.Tests/CheckinTests/DataLossAnalyzer.tests.ts
```diff
@@ -89,4 +89,5 @@ class DataLossAnalyzerTests extends TestClass {
     }
 }
 
-new DataLossAnalyzerTests().registerTests(); 
\ No newline at end of file
+// Uncomment if you want to use DataLossanalyzer
+// new DataLossAnalyzerTests().registerTests();
```
Dmitry-Matveev: Is there a better way to disable DataLossAnalyzer rather then commenting it in 4 places? E.g. some configuration setting to only change one bit back and worth and avoid the case when someone forgets to uncomment/comment one of those.

KamilSzostak: Yes, this requires a refactoring of DataLossAnalyzer (#208). I'll do that once we have this task done #289.

PR #405, opened 2017-02-11
Optimize internal logging

KamilSzostak:
This change will reduce the SDK size by ~1kb (minimized+gzipped).


SergeyKanzhelev:
Can you pls summarize changes and main ideas of this PR?

KamilSzostak:
Sorry, I know that the change is quite big, but it's mostly find/replace type of changes. The main idea is to reduce the SDK size. The change includes the following parts:
* combine `throwInternalNonUserActionable` and `throwInternalUserActionable` into one `throwInternal(isUserAct=false, ...)` method 
* I create an `_InternalLogMessage` object inside the `throwInternal` method, instead of passing it as a parameter
* remove `USRACT_` and `NONUSRACT_` prefixes from `_InternalMessageId` enum
* cleanup `Initialization.getDefaultConfig()` method

SergeyKanzhelev:
You can also regenerate the schema from public types only like here: https://github.com/Microsoft/ApplicationInsights-node.js/tree/develop/Schema It will remove some internal and legacy types and fields and ultimately make script smaller

KamilSzostak:
Yes, I'm planning to regenerate schema - #327.

PR #417, opened 2017-03-03
Add correlation headers for request within the same domain

### LINKED ISSUE #414: ApplicationInsights doesn't set x-ms-request-root-id & x-ms-request-id headers in Internet Explorer 11
I was testing Application Insights with IE and noticed that the request headers weren't getting set in the following code:

```
                if (!this.appInsights.config.disableCorrelationHeaders && (ApplicationInsights.UrlHelper.parseUrl(xhr.ajaxData.getAbsoluteUrl()).host == this.currentWindowHost)) {
                    var rootId = this.appInsights.context.operation.id;
                    xhr.setRequestHeader("x-ms-request-root-id", rootId);
                    xhr.setRequestHeader("x-ms-request-id", xhr.ajaxData.id);
                }
```

ApplicationInsights.UrlHelper.parseUrl(xhr.ajaxData.getAbsoluteUrl()).host == this.currentWindowHost) evaluates to false. Because ApplicationInsights.UrlHelper.parseUrl(xhr.ajaxData.getAbsoluteUrl()).host returns a port number at the end, like:

subdomain.domain.com:443 while this.currentWindowHost is subdomain.domain.com

KamilSzostak:
Fixes #414.

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
@@ -1715,6 +1715,41 @@ class AppInsightsTests extends TestClass {
         });
 
         this.testCase({
+            name: "Ajax - root/parent id are set only for dependency calls within the same domain",
+            test: () => {
+                var snippet = this.getAppInsightsSnippet();
+                snippet.disableAjaxTracking = false;
+                snippet.disableCorrelationHeaders = false;
+                snippet.maxBatchInterval = 0;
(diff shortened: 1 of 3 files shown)


jbrantly:
It seems like this is the wrong fix to the issue. Instead of ignoring the port number, the port number should always be included since same-origin includes the port number. The way this is currently means that it's possible to add those headers to CORS requests (which is how I ran into this issue).

KamilSzostak:
Thanks, @jbrantly. I opened an issue to fix it #459.

jbrantly:
👍

PR #424, opened 2017-03-16
Provide help on CORS script errors

KamilSzostak:


JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -470,7 +470,7 @@ module Microsoft.ApplicationInsights {
         * happens in a script from other domain (cross origin, CORS).		
         */
         public static isCrossOriginError(message: string, url: string, lineNumber: number, columnNumber: number, error: Error): boolean {
-            return (message === "Script error." || message === "Script error") && error === null;
+            return (message === "Script error." || message === "Script error") && error == null;
```
OsvaldoRosado: Why is this no longer triple-equals?

KamilSzostak: IE is using `undefined`, all other browsers `null`. This will match both values.

OsvaldoRosado: Would `!error` be better so that linters don't pick this up in the future?

OsvaldoRosado: Also curious. Does this string change in different locales?

KamilSzostak: I changed it to `!error`. I checked in polish version of Firefox and the error message is not localized.

PR #427, opened 2017-03-22
Capture errors thrown from the snippet script block

### LINKED ISSUE #282: Unhandled browser exceptions not being tracked.
I've included the applicationinsights-js module and initialized it with my instrumentation key. I also tried tracking page views which do show up in the azure portal.

I've been trying to get unhandled browser exceptions to show up in the portal with no success. I've tried manually throwing the exceptions or faking invalid operations but neither of them are tracked by app insights. Is there something I'm missing or any way to get logging to show if app insights is ignoring my exceptions?

I encountered this issue with [this ](https://github.com/Microsoft/ApplicationInsights-JS#include-ai-js-sdk-script-and-initialize-statically) javascript example and [this ](https://github.com/Microsoft/ApplicationInsights-JS#import-as-a-module-and-initialize-dynamically-download-full-application-insights-script-from-cdn) example with typescript.

KamilSzostak:
Fixes #282. 

* capture errors when they are thrown from the snippet script block
* wrap the correct `onerror` method in AppInsightsModule.ts
* add end-to-end unit test

### DIFF
--- a/JavaScript/JavaScriptSDK.Module/AppInsightsModule.ts
+++ b/JavaScript/JavaScriptSDK.Module/AppInsightsModule.ts
@@ -68,18 +68,21 @@ class AppInsightsModule {
             aiObject.queue = [];
         }
 
-        var scriptElement = document.createElement("script");
-        scriptElement.src = aiConfig.url || "https://az416426.vo.msecnd.net/scripts/a/ai.0.js";
-        document.head.appendChild(scriptElement);
+        setTimeout(function () {
+            var scriptElement = document.createElement("script");
+            scriptElement.src = aiConfig.url || "https://az416426.vo.msecnd.net/scripts/a/ai.0.js";
+            document.head.appendChild(scriptElement);
+        });
 
-        // collect global errors
+        // collect global errors by wrapping the window.onerror method
         if (!aiConfig.disableExceptionTracking) {
-            AppInsightsModule._createLazyMethod("_onerror");
-            var originalOnError = window["_onerror"];
-            window["_onerror"] = function (message, url, lineNumber, columnNumber, error) {
+            let method = "onerror";
+            AppInsightsModule._createLazyMethod("_" + method);
+            var originalOnError = window[method];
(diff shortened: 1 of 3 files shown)


JavaScript/JavaScriptSDK.Module/AppInsightsModule.ts
```diff
@@ -68,18 +68,21 @@ class AppInsightsModule {
...
 
-        // collect global errors
+        // collect global errors by wrapping the window.onerror method
         if (!aiConfig.disableExceptionTracking) {
-            AppInsightsModule._createLazyMethod("_onerror");
-            var originalOnError = window["_onerror"];
-            window["_onerror"] = function (message, url, lineNumber, columnNumber, error) {
+            let method = "onerror";
+            AppInsightsModule._createLazyMethod("_" + method);
+            var originalOnError = window[method];
```
OsvaldoRosado: Was there any reason this was capturing `_onerror` vs `onerror` before?

JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -13,9 +13,11 @@
     var track = "Track";
     var trackEvent = track + "Event";
     var trackPage = track + "Page";
-    var scriptElement = localDocument.createElement(scriptText);
-    scriptElement.src = aiConfig.url || "CDN_PATH";
-    localDocument.getElementsByTagName(scriptText)[0].parentNode.appendChild(scriptElement);
+    setTimeout(function () {
```
OsvaldoRosado: Does setTimeout end up actually end up solving the throw problem?

KamilSzostak: Yes. I tested it in IE 11, Edge, FF and Chrome.

JavaScript/JavaScriptSDK.Module/AppInsightsModule.ts
KamilSzostak: I was a bug, `originalOnError` was always null. We need to wrap `window.onerror` method.

PR #441, opened 2017-04-11
Add source maps to un-minified library (ai.js)

KamilSzostak:
This PR fixes source maps in the un-minified `ai.js` file. I'll add source maps to minified library in a next pull request.

Changes: 
* fix import path
* point to typescript source files in source maps (TypeScriptSourceRoot)

### DIFF
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
@@ -1,9 +1,9 @@
 ﻿/// <reference path="..\TestFramework\Common.ts" />
 /// <reference path="../../JavaScriptSDK/Util.ts"/>
-/// <reference path="../../JavaScriptSDK/sender.ts" />
+/// <reference path="../../JavaScriptSDK/Sender.ts" />
 /// <reference path="../../JavaScriptSDK/SendBuffer.ts"/>
-/// <reference path="../../javascriptsdk/appinsights.ts" />
-/// <reference path="../../JavaScriptSDK/util.ts" />
+/// <reference path="../../JavaScriptSDK/Appinsights.ts" />
+/// <reference path="../../JavaScriptSDK/Util.ts" />
 
 class SenderWrapper extends Microsoft.ApplicationInsights.Sender {
     errorSpy: SinonSpy;
--- a/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
+++ b/JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
@@ -1,6 +1,6 @@
 ﻿/// <reference path="../testframework/common.ts" />
-/// <reference path="../../JavaScriptSDK/telemetrycontext.ts" />
-/// <reference path="../../javascriptsdk/appinsights.ts" />
+/// <reference path="../../JavaScriptSDK/TelemetryContext.ts" />
+/// <reference path="../../JavaScriptSDK/AppInsights.ts" />
 /// <reference path="../../JavaScriptSDK/Telemetry/Common/Envelope.ts"/>
 
 class TelemetryContextTests extends TestClass {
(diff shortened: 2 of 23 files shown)


JavaScript/JavaScriptSDK/JavaScriptSDK.csproj
```diff
@@ -56,6 +56,15 @@
...
     <TypeScriptModuleKind>AMD</TypeScriptModuleKind>
+    <TypeScriptJSXEmit>None</TypeScriptJSXEmit>
+    <TypeScriptCompileOnSaveEnabled>True</TypeScriptCompileOnSaveEnabled>
+    <TypeScriptNoImplicitAny>False</TypeScriptNoImplicitAny>
+    <TypeScriptOutFile>min\$(OutFileBaseName).js</TypeScriptOutFile>
+    <TypeScriptOutDir />
+    <TypeScriptNoEmitOnError>True</TypeScriptNoEmitOnError>
+    <TypeScriptMapRoot>
+    </TypeScriptMapRoot>
+    <TypeScriptSourceRoot>https://raw.githubusercontent.com/Microsoft/ApplicationInsights-JS/master/JavaScript</TypeScriptSourceRoot>
```
AlexBulankou: If my understanding is correct, this will cause source map to point at the current state of master, which may be out of date? Should we instead be pointing at sources with release tag? Would this require making this variable set at build time?

KamilSzostak: You're right. I thought about it and I was planning to make two branches: develop and master (the same as we have for other SDKs), but this will still not work if somebody is using an older version of the SDK. Adding a release tag to the URL is a better solution. 

I'll add the release tag to the URL and update the release script to do that automatically - #443.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -1,9 +1,9 @@
 ﻿/// <reference path="..\TestFramework\Common.ts" />
 /// <reference path="../../JavaScriptSDK/Util.ts"/>
-/// <reference path="../../JavaScriptSDK/sender.ts" />
+/// <reference path="../../JavaScriptSDK/Sender.ts" />
 /// <reference path="../../JavaScriptSDK/SendBuffer.ts"/>
-/// <reference path="../../javascriptsdk/appinsights.ts" />
-/// <reference path="../../JavaScriptSDK/util.ts" />
+/// <reference path="../../javascriptsdk/Appinsights.ts" />
```
OsvaldoRosado: What's going on with the capitalization here. Sometimes the folder name is all lowercase, sometimes it is pascalcase. Same for filename. Some are all lowercase, some have the first letter uppercase, some are pascalcase

JavaScript/JavaScriptSDK.Tests/CheckinTests/TelemetryContext.tests.ts
```diff
@@ -1,6 +1,6 @@
 ﻿/// <reference path="../testframework/common.ts" />
-/// <reference path="../../JavaScriptSDK/telemetrycontext.ts" />
-/// <reference path="../../javascriptsdk/appinsights.ts" />
+/// <reference path="../../JavaScriptSDK/TelemetryContext.ts" />
+/// <reference path="../../javascriptsdk/AppInsights.ts" />
```
OsvaldoRosado: Nit: folder name casing

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
KamilSzostak: All paths in source maps are case sensitive (at least in Edge). I'm trying to fix all casing issues, but I missed a couple spots.

JavaScript/JavaScriptSDK/JavaScriptSDK.csproj
OsvaldoRosado: I found there's a TSC option you can use to bundle it into the sourcemaps and avoid this problem. https://www.typescriptlang.org/docs/handbook/compiler-options.html look for `--inlineSources`

KamilSzostak: I saw it but for some reason, it didn't work for me when I tried it. I guess it was the issue with the path capitalization because now it is working without any issues.

PR #448, opened 2017-04-21
Fix Markdown table formatting for parameters

stewartadam:


stewartadam:
Hold on, looks like this didn't quite fully fix it. Previews correctly in VS Code but not in GitHub itself.

stewartadam:
OK should preview correctly in GitHub now. GitHub doesn't appear to permit empty table headers, so I've added Parameter/Description and Property/Description as appropriate.

Issue #300, opened 2016-08-29
Stubbed startTrackPage/stopTrackPage should have correct timestamp


kamilszostak:
The SDK should report correct timestamps for stubbed/queued `startTrackPage` or `stopTrackPage` events. Even if they are called before the full library (`ai.0.js`) file is downloaded.

PR #454, opened 2017-05-04
#300 Stubbed startTrackPage/stopTrackPage should have correct timestamp

GeovannyLopez:
Stubbed startTrackPage/stopTrackPage should have correct timestamp #300

Team:
@juliocarpio4
@GeovannyLopez
@JoseSic
@LuisSic
@mariare19
@pnadalini
@ricardoher12

Engineers in charge:
@oscargarciacolon
@gzepeda

### DIFF
--- a/JavaScript/JavaScriptSDK/AppInsights.ts
+++ b/JavaScript/JavaScriptSDK/AppInsights.ts
@@ -155,7 +155,7 @@ module Microsoft.ApplicationInsights {
          * Starts timing how long the user views a page or other item. Call this when the page opens. 
          * This method doesn't send any telemetry. Call {@link stopTrackTelemetry} to log the page when it closes.
          * @param   name  A string that idenfities this item, unique within this HTML document. Defaults to the document title.
-         * @param StartDate The date that identifies the original start date that will override the current date. Defaults to the current date.
+         * @param StartDate A date that identifies the original start date that will override the current date. Defaults to the current date.
          */
         public startTrackPage(name?: string, StartDate?: Date) {
             try {
@@ -180,7 +180,7 @@ module Microsoft.ApplicationInsights {
          * @param   url   String - a relative or absolute URL that identifies the page or other item. Defaults to the window location.
          * @param   properties  map[string, string] - additional data used to filter pages and metrics in the portal. Defaults to empty.
          * @param   measurements    map[string, number] - metrics associated with this page, displayed in Metrics Explorer on the portal. Defaults to empty.
-         * @param   EndDate The date that identifies the original end date that will override the current date. Defaults to the current date.
+         * @param   EndDate A date that identifies the original end date that will override the current date. Defaults to the current date.
          */
         public stopTrackPage(name?: string, url?: string, properties?: Object, measurements?: Object, EndDate?: Date) {
             try {
--- a/JavaScript/JavaScriptSDK/AppInsights.ts
+++ b/JavaScript/JavaScriptSDK/AppInsights.ts
@@ -236,7 +236,7 @@ module Microsoft.ApplicationInsights {
          * @param   measurements    map[string, number] - metrics associated with this page, displayed in Metrics Explorer on the portal. Defaults to empty.
          * @param   EndDate A date that identifies the original end date that will override the current date. Defaults to the current date.
          */
-        public stopTrackPageInternal(name?: string, url?: string, properties?: Object, measurements?: Object, endTime?: number) {
+        private stopTrackPageInternal(name?: string, url?: string, properties?: Object, measurements?: Object, endTime?: number) {
             try {
                 if (typeof name !== "string") {
                     name = window.document && window.document.title || "";
--- a/JavaScript/JavaScriptSDK/AppInsights.ts
+++ b/JavaScript/JavaScriptSDK/AppInsights.ts
@@ -155,40 +155,41 @@ module Microsoft.ApplicationInsights {
          * Starts timing how long the user views a page or other item. Call this when the page opens. 
          * This method doesn't send any telemetry. Call {@link stopTrackTelemetry} to log the page when it closes.
          * @param   name  A string that idenfities this item, unique within this HTML document. Defaults to the document title.
-         * @param startDate A date that identifies the original start date that will override the current date. Defaults to the current date.
          */
-        public startTrackPage(startDate?: Date); //Method firm with date
-        public startTrackPage(name?: string, startDate?: Date); //Method firm with name and date
-        public startTrackPage(param1?: (string | Date), param2?: Date) {
-
-            var name: string;
-            var startDate: Date;
-
+        public startTrackPage(name?: string) {
             try {
-
-                if (param2 != null) {
-                    name = <string>param1;
-                    startDate = param2;
-                }
-
-                else if (param1 != null) {
-                    if (typeof param1 === "string") {
-                        name = param1;
-                        startDate = null;
-                    }
-
-                    else {
-                        name = window.document && window.document.title || "";
-                        startDate = param1;
-                    }
-                }
-
-                else if (typeof param1 !== "string") {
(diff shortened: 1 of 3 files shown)


JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -155,14 +155,16 @@ module Microsoft.ApplicationInsights {
          * Starts timing how long the user views a page or other item. Call this when the page opens. 
          * This method doesn't send any telemetry. Call {@link stopTrackTelemetry} to log the page when it closes.
          * @param   name  A string that idenfities this item, unique within this HTML document. Defaults to the document title.
+         * @param StartDate A date that identifies the original start date that will override the current date. Defaults to the current date.
          */
-        public startTrackPage(name?: string) {
+        public startTrackPage(name?: string, StartDate?: Date) {
```
KamilSzostak: `name` argument is optional. Please make sure this will work if a user didn't pass it and we automatically added startDate. 

Also, please change argument name to lower case `startDate`.

JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -32,6 +32,7 @@
             var originalArguments = arguments;
             // Queue-up a call to the real method
             appInsights.queue.push(function() {
+
```
KamilSzostak: Please also record the original call time and add it to argument list of `startTrackPage` and `stopTrackPage` methods.

JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -49,8 +50,9 @@
     createLazyMethod(start + trackEvent);
     createLazyMethod(stop + trackEvent);
 
-    createLazyMethod(start + trackPage);
-    createLazyMethod(stop + trackPage);
+    //Record the current time for the methods startTrackPage and stopTrackPage
+    createLazyMethod(start + trackPage, new Date());
```
SergeyKanzhelev: Can the `new Date()` be calculated inside the method itself if no value was passed to it?

GeovannyLopez:
I have made the suggested corrections.

JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -30,8 +30,12 @@
         appInsights[name] = function() {
             // Capture the original arguments passed to the method
             var originalArguments = arguments;
+            if (name.indexOf(trackPage) >= 0) {
+                [].push.call(originalArguments, new Date().getTime());
```
KamilSzostak: Does it work when a user calls `trackPageView` without `name` parameter?  Is new Date used as the name?

GeovannyLopez: I change startTrackPage method to work if the user calls it without _name_ parameter

gzepeda: waiting for @KamilSzostak feedback.

JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -155,14 +155,40 @@ module Microsoft.ApplicationInsights {
          * Starts timing how long the user views a page or other item. Call this when the page opens. 
          * This method doesn't send any telemetry. Call {@link stopTrackTelemetry} to log the page when it closes.
          * @param   name  A string that idenfities this item, unique within this HTML document. Defaults to the document title.
+         * @param startDate A date that identifies the original start date that will override the current date. Defaults to the current date.
          */
-        public startTrackPage(name?: string) {
+        public startTrackPage(startDate?: Date); //Method firm with date
+        public startTrackPage(name?: string, startDate?: Date); //Method firm with name and date
+        public startTrackPage(param1?: (string | Date), param2?: Date) {
```
KamilSzostak: This looks very hacky :) especially that the method with `param1` and `param2 is public and will be very confusing for SDK users. 

IMHO it would be better to only have one public method `public startTrackPage(name?: string)` which will internally call `startTrackPageInternal(name: string, startDate?: Date)`.  It will be an internal private method of `AppInsights` class. The snippet.js code can call `startTrackPageInternal` and pass a right set of parameters. The code may look like this: 

```
if (name.indexOf(trackPage) >= 0) {
   // change the function name from `startTrackPage` to `startTrackPageInternal` (the same for stopTrack)
   // check if `originalArguments` contains name, if not set it to `null`
   // [].push a call to startTrackPageInternal(name, new Date().getTime())
```

This way you will always have a name parameter passed to `startTrackPageInternal`. 

Thanks in advance for implementing this change!

JulioCarpio4: How can we check if 'originalArguments' contains name, @KamilSzostack? We're inplementing the change you suggest, but we haven't been able to check if the name is within the 'originalArguments'.

GeovannyLopez: Hi @KamilSzostak We followed your advice and implement the Internal methods. 
Thanks for your feedback.

gzepeda:
Hello @KamilSzostak, have you had a chance to check the latest commit?

JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -162,7 +162,34 @@ module Microsoft.ApplicationInsights {
                     name = window.document && window.document.title || "";
                 }
```
KamilSzostak: This is not required here. `name` will be set in `startTrackPageInternal`.

JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -178,8 +205,9 @@ module Microsoft.ApplicationInsights {
          * @param   url   String - a relative or absolute URL that identifies the page or other item. Defaults to the window location.
          * @param   properties  map[string, string] - additional data used to filter pages and metrics in the portal. Defaults to empty.
          * @param   measurements    map[string, number] - metrics associated with this page, displayed in Metrics Explorer on the portal. Defaults to empty.
+         * @param   EndDate A date that identifies the original end date that will override the current date. Defaults to the current date.
          */
-        public stopTrackPage(name?: string, url?: string, properties?: Object, measurements?: Object) {
+        public stopTrackPage(name?: string, url?: string, properties?: Object, measurements?: Object, endTime?: number) {
             try {
                 if (typeof name !== "string") {
```
KamilSzostak: The same here, both variables will be set in `stopTrackPageInternal`. No need to duplicate the code.

JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -30,8 +31,15 @@
         appInsights[name] = function() {
             // Capture the original arguments passed to the method
             var originalArguments = arguments;
+            if (name.indexOf(trackPage) > 0) {
```
KamilSzostak: This will also catch `trackPageView` method, which we should not touch.

JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -49,8 +57,8 @@
     createLazyMethod(start + trackEvent);
     createLazyMethod(stop + trackEvent);
 
-    createLazyMethod(start + trackPage);
-    createLazyMethod(stop + trackPage);
+    createLazyMethod(start + trackPage + internal);
```
KamilSzostak: `startTrackPageInternal` is an internal method, there is no need to expose it to users. You need to create a lazy method for the `startTrackPage`. In line 35 you need to modify the originalArguments to change `startTrackPage` to `startTrackPageInternal` (the same for stop).

sylvanaar:
The age of this PR doesn't exactly reflect well on the health of this project.

gzepeda:
Yes, we probably need to redo it. I will ask the team if they can close the PR.

GeovannyLopez:
According to the commets we decided to close this pull request.

PR #469, opened 2017-06-07
Add a default telemetry initializer to exclude browserLink requests

### LINKED ISSUE #466: BrowserLink telemetry being sent to Azure
Moved from https://github.com/Microsoft/ApplicationInsights-Home/issues/122

Reported by @ppumkin

Browserlink shows up as a dependency AJAX and is flooding my charts while debugging.

I had a look at the 

`<Add Type="Microsoft.ApplicationInsights.Web.SyntheticUserAgentTelemetryInitializer, Microsoft.AI.Web">`

and 

`<Add Type="Microsoft.ApplicationInsights.Web.RequestTrackingTelemetryModule, Microsoft.AI.Web">`

but it is not entirely clear to me how to filter out browserLink from Visual Studio

I am currently evaluating on the free tier and I know I probably need to setup separate Azure Resources  but I really do not need to be sending and using data for browserlink any way. (especially when you have a large dev team) 

image

KamilSzostak:
Fixes #466.

### DIFF
--- a/JavaScript/JavaScriptSDK/TelemetryContext.ts
+++ b/JavaScript/JavaScriptSDK/TelemetryContext.ts
@@ -79,6 +79,7 @@ module Microsoft.ApplicationInsights {
         constructor(config: ITelemetryConfig) {
             this._config = config;
             this._sender = new Sender(config);
+            this.telemetryInitializers = [];
 
             // window will be undefined in node.js where we do not want to initialize contexts
             if (typeof window !== 'undefined') {
@@ -92,6 +93,8 @@ module Microsoft.ApplicationInsights {
                 this.session = new Context.Session();
                 this.sample = new Context.Sample(config.sampleRate());
             }
+
+            this._addDefaultTelemetryInitializers();
         }
 
         /**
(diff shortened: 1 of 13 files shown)


JavaScript/JavaScriptSDK/TelemetryContext.ts
```diff
@@ -92,14 +93,15 @@ module Microsoft.ApplicationInsights {
                 this.session = new Context.Session();
                 this.sample = new Context.Sample(config.sampleRate());
             }
+
+            this._addDefaultTelemetryInitializers();
```
Dmitry-Matveev: Do we need to provide a way to disable this initializer for someone who wants to get browser link telemetry or someone who has similarly routes in the app?

KamilSzostak: Yeah, I thought about it and decided that instrumenting browser link is not useful and I didn't want to add another config field (we already have plenty).
But you're right, somebody with similar routes may want to disable it. I'll add it.

SergeyKanzhelev: Can you remove telemetry initializer? Or without strong types in JS you'd never know which one to remove?

KamilSzostak: You can, but it will hacky ;/ telmetryInitializers array is private and is not exposed in d.ts, functions are usually hard to compare and it will be difficult to find it among other initializers.

KamilSzostak:
I added a config to enable browserLink tracking.

PR #471, opened 2017-06-07
CORS check should include a port number

### LINKED ISSUE #459: The port number should be included in the CORS check
According to the [Same Origin Policy](https://www.w3.org/Security/wiki/Same_Origin_Policy), the origin is the same if the protocol, host and port is the same. 

The port number was removed in #417, this has to be fixed.

KamilSzostak:
Fixes #459.

### DIFF
--- a/JavaScript/JavaScriptSDK/ajax/ajax.ts
+++ b/JavaScript/JavaScriptSDK/ajax/ajax.ts
@@ -17,7 +17,7 @@ module Microsoft.ApplicationInsights {
         private currentWindowHost;
 
         constructor(appInsights: Microsoft.ApplicationInsights.AppInsights) {
-            this.currentWindowHost = window.location.hostname;
+            this.currentWindowHost = window.location.host;
             this.appInsights = appInsights;
             this.initialized = false;
             this.Init();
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -140,9 +140,9 @@ module Microsoft.ApplicationInsights {
         private sendHandler(xhr: XMLHttpRequestInstrumented, content) {
             xhr.ajaxData.requestSentTime = dateTime.Now();
 
-            // Add correlation headers only for requests within the same domain (ignore the port number)
+            // Add correlation headers only for requests within the same host and port number
```
Dmitry-Matveev: I do not see that port is added? Either description is misleading or the fix was not fully implemented. Can you please double check?

Dmitry-Matveev: Maybe, UrlHelper returns host with port included? Then, pls. disregard the comment.

KamilSzostak: It is :) `host` (in opposite to `hostname`) contains the port number. It's also covered by the unit test.

PR #472, opened 2017-06-08
Prevent duplicate data if snippet is inserted twice

### LINKED ISSUE #464: Prevent duplicate data from being sent in case of AI snippet is inserted twice
In Azure Web Apps scenario we cannot reliably determine if AI injection already happened or not. 

Enabling URL re-write feature is now a manual step for customer and it's not turned on by the default, so it would be an edge case if customer enabled it and has client-side injection from VS at the same time. But there could be scenarios when customer tried client-side collection from the extension first and enabled it from VS and re-deployed the app later on. At that point of time he starts receiving duplicate data.

So it's not high important now as I've mentioned this feature is disabled by the default, but it would be really nice to have it. And the priority will rise if we'll want to turn it on by the default.

KamilSzostak:
Fixes #464.

### DIFF
--- a/JavaScript/JavaScriptSDK/snippet.js
+++ b/JavaScript/JavaScriptSDK/snippet.js
@@ -78,4 +78,8 @@
 
 // global instance must be set in this order to mitigate issues in ie8 and lower
 window.appInsights = appInsights;
-appInsights.trackPageView();
+
+// if somebody calls the snippet twice, don't report page view again
+if (appInsights.queue && appInsights.queue.length === 0) {
+    appInsights.trackPageView();
+}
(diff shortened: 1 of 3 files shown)


JavaScript/JavaScriptSDK/snippet.js
```diff
@@ -78,4 +78,8 @@
 
 // global instance must be set in this order to mitigate issues in ie8 and lower
 window.appInsights = appInsights;
-appInsights.trackPageView();
+
+// if somebody calls the snippet twice, don't report page view again
+if (appInsights.queue && appInsights.queue.length === 0) {
```
Dmitry-Matveev: Just curious, may we have some race condition in case of a single snippet, so that queue.length is 0 at this time and not even the first TrackPageView() is called, e.g. for the first Page View?

KamilSzostak: Just to give some context. I was trying to find a solution which doesn't introduce a global var and is as small as possible. This code will be executed before the full library is downloaded and the snippet is only initializing an empty queue. 
Of course, you can break it if you put `appInsights.queue.push('haha')` before this code, but we have to assume that nobody will change the snippet code and assume it will still work.

PR #473, opened 2017-06-08
Log the response code (status) in request onError handler

KamilSzostak:
With this change, the SDK will log the response code (status) in request `onError` handler.

### DIFF
--- a/JavaScript/JavaScriptSDK/Sender.ts
+++ b/JavaScript/JavaScriptSDK/Sender.ts
@@ -370,6 +370,18 @@ module Microsoft.ApplicationInsights {
             this._setupTimer();
         }
 
+        private _formatErrorMessage(xhr: XMLHttpRequest, xdr: XDomainRequest, message?: string): string {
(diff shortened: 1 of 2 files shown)


JavaScript/JavaScriptSDK/Sender.ts
```diff
@@ -370,6 +370,18 @@ module Microsoft.ApplicationInsights {
             this._setupTimer();
         }
 
+        private _formatErrorMessage(xhr: XMLHttpRequest, xdr: XDomainRequest, message?: string): string {
+            if (xhr) {
```
Dmitry-Matveev: Minor: would it be preferable to have to methods, so that usage is a bit easier to read, e.g. formatErrorMessageXdr(), formatErrorMessageXhr()? Now it sumps around a bit like (null, xdr) vs. (xhr, null).

KamilSzostak: Good point, fixed.

PR #474, opened 2017-06-08
Fix telemetry context function type

### LINKED ISSUE #434: Bad/non working sample for setting default properties.
The codesnippet on https://docs.microsoft.com/en-us/azure/application-insights/app-insights-api-filtering-sampling#add-properties-itelemetryinitializer  dosent work with SDK 107 

Compiling it with typescript gives the following errors.
Argument of type '(envelope: IEnvelope) => void' is not assignable to parameter of type '(envelope: IEnvelope) => boolean'
Property 'data' does not exist on type 'IEnvelope'

KamilSzostak:
Fixes #434.

### DIFF
--- a/API-reference.md
+++ b/API-reference.md
@@ -414,7 +414,7 @@ Sends telemetry to the endpoint.
 
 ### addTelemetryInitializer
 
-        public addTelemetryInitializer(telemetryInitializer: (envelope: Telemetry.Common.Envelope) => boolean)
+        public addTelemetryInitializer(telemetryInitializer: (envelope: Telemetry.Common.Envelope) => boolean | void)
 
 Adds a telemetry initializer to the collection. Telemetry initializers will be called one by one, in the order they were added,
 before the telemetry item is pushed for sending.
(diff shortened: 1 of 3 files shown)


API-reference.md
```diff
@@ -414,7 +414,7 @@ Sends telemetry to the endpoint.
 
 ### addTelemetryInitializer
 
-        public addTelemetryInitializer(telemetryInitializer: (envelope: Telemetry.Common.Envelope) => boolean)
+        public addTelemetryInitializer(telemetryInitializer: (envelope: Telemetry.Common.Envelope) => boolean | void)
```
Dmitry-Matveev: Is it a back compat change (e.g. we now require a Boolean but previously were OK with void)? If not - we may simply want to update instructions in the [docs](https://docs.microsoft.com/azure/application-insights/app-insights-api-filtering-sampling#add-properties-itelemetryinitializer).

KamilSzostak: Yes, it's a back compatibility change. We were always accepting bool|void, but definition types only allowed boolean.

mattias-skold: @KamilSzostak - docs still not working with definition types,  Microsoft.ApplicationInsights.IEnvelope doesn't have member data...

Issue #445, opened 2017-04-20
Save events tracked in beforeUnload


ptrstpp950:
I'm trying to save custom metric - how long user was on mu page - using Application Insights. I wrote following code (it is a part):
```
 var start = new Date();
            
var recordMetrics = function () {
    var end = new Date();
    var timeSpentTotal = (end.getTime() - start.getTime()) / 1000;
    if (window.appInsights) {
        window.appInsights.trackMetric("timeSpentTotal", timeSpentTotal);
        window.appInsights.flush();
        console.log("Sent data with timeSpent");
    }
}
window.addEventListener("beforeunload",
    function (e) {
        recordMetrics();
    });
```
But every request after the `flush` function is canceled. I'm almost sure that this is because App Insights create asynchronous AJAX request. Can I somehow change this to synchronous?

OsvaldoRosado:
`beforeunload` is a tricky event to work with, mainly because of the asynchronous problem you've identified. We have some code to try to send events for you during beforeunload, as a best effort. (See https://github.com/Microsoft/ApplicationInsights-JS/blob/cac7b363a7f59147adff510f698bc3f4943f2ab0/JavaScript/JavaScriptSDK/Initialization.ts#L120 for reference)

Using flush like this should be synchronous however. Are you finding that is not the case? Perhaps try `appInsights.context._sender.triggerSend(false);` instead.

ptrstpp950:
Can I create a pull request with a boolean parameter in `flush` method?
Using "private" method is like a naughty hack

OsvaldoRosado:
Can you first verify that flush is sending asynchronously? The documentation for that method says it should be synchronous. The workaround I suggested is of course not a permanent fix if this is the case, but something to unblock between SDK releases

KamilSzostak:
Adding a bool parameter to `flush` method to force it to be synchronous is a good idea. Please feel free to send a PR. 

By default, the SDK will not block the user from navigating to a new page (e.g. by calling sync flush) as this may impact the navigation experience. But of course, there is nothing stopping the developer to block the navigation. Please keep in mind that this is not guaranteed to work because the browser may close any pending requests and terminate any script execution. 

[...]

PR #477, opened 2017-06-23
Allow flush to be sync

nacjacks:
Allows the flush method exposed via the SDK to be called with an optional parameter to perform the flush synchronously.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.suo
 *.user
 *.sln.docstates
+.idea/
 
 # Build results
 [Dd]ebug/
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 *.suo
 *.user
 *.sln.docstates
-.idea/
 
 # Build results
 [Dd]ebug/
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@
 *.suo
 *.user
 *.sln.docstates
+.idea/
```
KamilSzostak: is this required?

JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -407,10 +407,19 @@ module Microsoft.ApplicationInsights {
 
         /**
          * Immediately send all queued telemetry.
+         * @param {boolean} async - If flush should be call asynchronously
          */
-        public flush() {
+        public flush(async?: boolean) {
```
KamilSzostak: Can you use `public flush(async = true)`? Then you don't have to check if async is boolean. 

The same improvement can be done in the [sender.triggerSend()](https://github.com/Microsoft/ApplicationInsights-JS/blob/2745ffadaa834ad7532182e5aa4417b1cb7907c8/JavaScript/JavaScriptSDK/Sender.ts#L256) method. 

Thanks!

KamilSzostak:
Thanks for contributing. This PR is addressing an issue mentioned in #445.

.gitignore
nacjacks: Not specifically, but for some reason Webstorm decided to add that folder my project.

jerms0518:
@KamilSzostak Any idea when this change will be released?

PR #498, opened 2017-08-04
Migrate to grunt build

KamilSzostak:
Done:
* `grunt` to build ai.min and snippet scripts
* `grunt module` to build the module script
* `grunt tests` to run unit and e2e tests
* migrated to TS 2.4.1

TODO:
* remove unused files (nuget files, JavaScriptSDK, Interfaces, Module and Tests  csproj)
* compile with `noImplicitAny: true`


JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -667,7 +667,12 @@ class AppInsightsTests extends TestClass {
 
                 // mock user agent
                 let originalUserAgent = navigator.userAgent;
-                this.setUserAgent("Googlebot/2.1");
+                try {
+                    this.setUserAgent("Googlebot/2.1");
+                } catch (ex) {
+                    Assert.ok(true, 'cannot run this test in the current setup - try Chrome');
+                    return;
```
AlexBulankou: why do we need to catch and ok-return, instead of proceeding as if Beacon API is not supported?

JavaScript/JavaScriptSDK.Tests/phantomJS.debug.js
```diff
@@ -0,0 +1,7 @@
+page = require('webpage').create();
+page.open('file:///E:/Dev/ApplicationInsights-JS/JavaScript/JavaScriptSDK.Tests/E2ETests/E2E.DisableTelemetryTests.htm?hidepassed', function () {
```
AlexBulankou: probably not intended to check this in?

JavaScript/Microsoft.ApplicationInsights.JavaScript.sln
```diff
@@ -1,13 +1,9 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
```
AlexBulankou: Why do we still need SLN and *PROJ files?

AlexBulankou: Never mind, I saw the comment that you put in saying this will be removed as part of subsequent changes.

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
KamilSzostak: I need at least one `Assert` otherwise, JQuery fails with an error - "No assertions executed...".

AlexBulankou: I meant to catch this but proceed with remainder of the test as if Beacon API is not enabled

KamilSzostak: This test validates if googlebot is detected correctly by the SDK. To do that I need to change the user agent which is not supported in phantomJs. There is nothing to validate if I cannot override the user agent. 

The long term plan is to switch from phantomJS to headless Chrome, once it's available which supports user agent overriding.

JavaScript/JavaScriptSDK.Tests/phantomJS.debug.js
KamilSzostak: Yes, I used this file to debug tests in PhantomJS.

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    // Use Typescript from node_modules instead of builtin to VSCode
```
OsvaldoRosado: Comments aren't usually valid JSON. Is vscode okay with this?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
```diff
@@ -1111,6 +1111,10 @@ class SenderTests extends TestClass {
         this.testCase({
             name: "SenderTests: send() is using BeaconAPI sender if the BeaconAPI is enabled",
             test: () => {
+                if (!(<any>navigator).sendBeacon) {
```
OsvaldoRosado: Are there types we can update rather than cast to any?

JavaScript/JavaScriptSDK.Tests/CheckinTests/Telemetry/PageViewPerformance.tests.ts
```diff
@@ -90,12 +90,12 @@ class PageViewPerformanceTelemetryTests extends ContractTestHelper {
             test: () => {
 
                 var timing = <PerformanceTiming>{};
-                timing.navigationStart = 1;
-                timing.connectEnd = 10;
-                timing.requestStart = 11;
-                timing.responseStart = 30;
-                timing.responseEnd = 42;
-                timing.loadEventEnd = 60;
+                (<any>timing).navigationStart = 1;
```
OsvaldoRosado: Why is this cast necessary?

tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
```
OsvaldoRosado: Any way to have grunt read from here rather than defining these options in multiple places?

.vscode/settings.json
KamilSzostak: Removed.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Sender.tests.ts
KamilSzostak: I think I added that when I was fixing unit tests and forgot to remove it. It's gone now.

JavaScript/JavaScriptSDK.Tests/CheckinTests/Telemetry/PageViewPerformance.tests.ts
KamilSzostak: PerformanceTiming properties are read-only, to change them I need to cast them to _any_. 
One way to fix this is to remove PerformanceTiming type and create a simple object with required properties. I like strongly typed objects, but in this case casting to _any_ contradicts using a type, doesn't it? 

```
                var timing = {
                    navigationStart: 1,
                    connectEnd: 10,
                    requestStart: 11,
                    responseStart: 30,
                    responseEnd: 42,
                    loadEventEnd: 60,
                };
```
instead of 
```
                var timing = <PerformanceTiming>{};
                (<any>timing).navigationStart = 1;
                (<any>timing).connectEnd = 10;
                (<any>timing).requestStart = 11;
                (<any>timing).responseStart = 30;
                (<any>timing).responseEnd = 42;
                (<any>timing).loadEventEnd = 60;
```

OsvaldoRosado: How was this working before when it was strongly typed and not casted?

tsconfig.json
KamilSzostak: Yes, I tried it at the beginning when I started the migration to grunt and I couldn't get it to work. But now when everything is building and test are passing it was much easier :)

PR #556, opened 2017-12-06
update schemas

KamilSzostak:
#327


JavaScript/JavaScriptSDK.Interfaces/Contracts/Generated/RemoteDependencyData.ts
```diff
@@ -1,51 +1,99 @@
...
         public min: number;
         public max: number;
         public stdDev: number;
         public dependencyKind: AI.DependencyKind;
+        
+        /**
+         * Indication of successfull or unsuccessfull call.
+         */
+        public success: boolean;
         public async: boolean;
```
SergeyKanzhelev: isn't async is gone in public schema?

JavaScript/JavaScriptSDK.Interfaces/Contracts/Generated/RemoteDependencyData.ts
```diff
@@ -1,51 +1,99 @@
...
-
-        /* Following fields should be marked as depricated in the next release, and must be removed in the subsequent breaking release */
         public kind: AI.DataPointType;
         public value: number;
+        
+        /**
+         * Request duration in format: DD.HH:MM:SS.MMMMMM. Must be less than 1000 days.
+         */
+        public duration: string;
         public count: number;
```
SergeyKanzhelev: these fields gone in public schema

KamilSzostak: did we replace duration with 'value'? Or it's just gone and the JS SDK should not report any duration?

JavaScript/JavaScriptSDK.Interfaces/Contracts/Generated/RemoteDependencyData.ts
KamilSzostak: Ok, I'll remove it.

JavaScript/JavaScriptSDK.Interfaces/Contracts/Generated/RemoteDependencyData.ts
KamilSzostak: it's there: https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/RemoteDependencyData.bond

SergeyKanzhelev: I meant `count`. It's not there

KamilSzostak: Yes, I reviewed all bond files in AppInsights-Home and remove all fields that are not present there. 

After removing those fields I've noticed some no longer used code, which I'm removing right now.

PR #565, opened 2018-01-31
Correlation headers updates and refactoring

iusafaro:
1) combine legacy headers into Request-Id
2) change logic for dependency id calculation
3) extend interface with traceDependencyData
4) read Request-Context response header to get appId and attach it to the dependency target
5) add excludedDomains list for the correlation


JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -1710,14 +1710,14 @@ class AppInsightsTests extends TestClass {
...
         this.testCase({
-            name: "Ajax - root/parent id are set and passed correctly",
+            name: "Ajax - Request-Id is set and passed correctly",
             test: () => {
                 var snippet = this.getAppInsightsSnippet();
                 snippet.disableAjaxTracking = false;
                 snippet.disableCorrelationHeaders = false;
                 snippet.maxBatchInterval = 0;
                 var appInsights = new Microsoft.ApplicationInsights.AppInsights(snippet);
-                var trackStub = this.sandbox.spy(appInsights, "trackDependency");
```
KamilSzostak: do we have any tests for `trackDependency` or only for `trackDependencyData`?

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -1796,8 +1792,33 @@ class AppInsightsTests extends TestClass {
                 (<any>xhr).respond("200", {}, "");
 
                 // Assert
-                Assert.equal(null, (<any>xhr).requestHeaders['x-ms-request-id'], "x-ms-request-id should not be set");
```
KamilSzostak: Do we need any backward compatibility and still support `x-ms'*`  headers? Breeze will understand `Request-ID`, but how about Vortex?

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
iusafaro: i left unit tests for trackDependency without any changes as they're testing public API, but unit tests for ajax collection now for trackDependencyData - so both are covered now. i can make a duplicate of trackDependency unit tests for trackDependencyData as well.

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
iusafaro: didn't understand this... this format is for all server-side SDKs, but not for Breeze or Vortex... how those are involved here?

PR #573, opened 2018-02-28
Update schemas to add pageview.id field

iusafaro:


JavaScript/JavaScriptSDK/Telemetry/PageView.ts
```diff
@@ -21,9 +22,10 @@ module Microsoft.ApplicationInsights.Telemetry {
         /**
          * Constructs a new instance of the PageEventTelemetry object
          */
-        constructor(name?: string, url?: string, durationMs?: number, properties?: any, measurements?: any) {
+        constructor(id?: string, name?: string, url?: string, durationMs?: number, properties?: any, measurements?: any) {
```
SergeyKanzhelev: this change may break customers who constructing pageView today. especially since every argument here is optional and some people may rely on position of argument, not name. + @OsvaldoRosado who did some API clean up recently. What would be your suggestion?

SergeyKanzhelev: Update API reference doc in this repository as well.

OsvaldoRosado: I would not change the order of any positional arguments, especially because customers get automatically upgraded to the latest version. This new parameter should be added to the end instead. That is, of course, only if this is part of the public API. Do we expose this class to end users or do we only abstract it behind a track method?

In terms of API cleanup, we addressed this upgrade conflict issue by making all of our TelemetryClient functions take a dictionary of arguments instead of positional arguments, but I don't think we can make a similar change here without breaking everyone.

iusafaro: put id as a last parameter

iusafaro: id was already mentioned as a field for telemetry item... no need to update.

jpiyali: optional parameters should be added at the end?

iusafaro: yes, i already updated it - you're looking at the outdated version )

jpiyali: ah, thanks!

PR #575, opened 2018-03-21
Parse appId from Breeze reponse and pass it as Request-Context

iusafaro:


JavaScript/JavaScriptSDK.Tests/Selenium/ai.tests.d.ts
```diff
@@ -2259,6 +2283,10 @@ declare module Microsoft.ApplicationInsights {
          * The object describing a session tracked by this object.
          */
         session: Context.Session;
+        /**
+         * AppId of this component if returned by the backend.
+         */
+        appId: () => string;
```
SergeyKanzhelev: Is this `TelemetryContext` object represented with the single ikey? In .NET we are keeping a hash map as every item may have it's own ikey.

JavaScript/JavaScriptSDK/RequestResponseHeaders.ts
```diff
@@ -5,17 +5,32 @@ module Microsoft.ApplicationInsights {
...
+         */
+        public static requestContextAppIdFormat = "appId=cid-v1:";
+
         /**
          * Request-Id header
          */
         public static requestIdHeader = "Request-Id";
+
+        /**
+         * Sdk-Context header.
```
SergeyKanzhelev: Can you expand this comment explaining where this header is used?

JavaScript/JavaScriptSDK.Tests/Selenium/ai.tests.d.ts
```diff
@@ -1708,12 +1728,16 @@ declare module Microsoft.ApplicationInsights {
...
          * to maintain consistency with the xhrSender's contract
+         * Note: XDomainRequest does not support custom headers and we are not able to get
+         * appId from the backend for the correct correlation.
          */
         private _xdrSender(payload, isAsync);
         /**
          * Send Beacon API request
          * @param payload {string} - The data payload to be sent.
          * @param isAsync {boolean} - not used
+         * Note: Beacon API does not support custom headers and we are not able to get
```
jpiyali: Why was the design to go with AppId and not Ikey? Though Ikey is mutable for a resource, it still has to be valid in order to send data to. Seems beacon api users wont get the benefit of this feature, and going with ikey would have been useful here?

JavaScript/JavaScriptSDK.Tests/Selenium/ai.tests.d.ts
iusafaro: we don't support multi-key scenario in javascript sdk and, yes, we do have only one TelemetryContext for the whole execution.

SergeyKanzhelev: API reference says you can change [config properties dynamically](https://github.com/Microsoft/ApplicationInsights-JS/blob/master/API-reference.md#config). Also `ikey` is part of an [envelope](https://github.com/Microsoft/ApplicationInsights-JS/blob/master/API-reference.md#class-envelope) which is accessible from telemetry processor.

So it is not completely unsupported. 

I'd suggest to keep a hash map instead. It would be more error-prone.

JavaScript/JavaScriptSDK.Tests/Selenium/ai.tests.d.ts
iusafaro: @SergeyKanzhelev, @jpiyali,
Correct me if I'm wrong - are we talking about the following?
1) pass ikey:<ikey> in a Request-Context header 
2) add <ikey> to the source of the server-side request
3) modify UX to work with <ikey> from source

It'll be very difficult to change UX to work with both ikeys and appIds as all Draft queries are bound only to appId and it cannot be implemented without significant refactoring and performance degradation. 

We can proceed in two ways:
1) Ensure that current approach works and then remove custom header requirement from Breeze - actually nothing will change because Breeze responds with appId today only if batch from the same key
2) Support ikey:<ikey> format but extend server SDK, to resolve it to appId and append appId to the request source like it's done today.

JavaScript/JavaScriptSDK/RequestResponseHeaders.ts
iusafaro: Done

JavaScript/JavaScriptSDK.Tests/Selenium/ai.tests.d.ts
SergeyKanzhelev: I wouldn't make it server SDK responsibility. Otherwise it will create another attack vector - sending many different ikeys will force an app to run many queries to Breeze. Plus there will be a delay in obtaining this mapping that will delay app delivery.

Does Kusto supports addressing an application by ikey? I know it supports by name and ARM ID. Is it slower request? I don't understand why performance will suffer - it is basically one indirection in the query preparation. Actual query will be the same.

iusafaro: No, Kusto doesn't support ikey. We solved this problem before once - when we sent hashed ikey - and it resolved in a heavy queries for finding right appId or componentId by ikey. Even if we'll send ikey (instead of hashed one - and that still slightly concerns me from the security point) then we'll need to resolve it to appId anyway by additional call to Breeze or CDS in order to get information from Draft. Our UX queries are complicated and slow at the moment and we keep constantly fighting with the performance overhead and ongoing optimization. 

So my proposal here would be not to support Beacon API now as it's disabled by the default and is used only by couple customers who are dealing with other problems with JS SDK + Beacon API (basically it doesn't work)... And later on, when we'll completely test the new channel and ensure that it's working, remove Sdk-Context header requirement from Breeze and return appId in any case. BTW this header was introduced to limit issues with multi-ikey scenarios, but we found other security issues and now Breeze return appId only if all items from the batch are from the same ikey. In this logic it's safe to remove header.

JavaScript/JavaScriptSDK.Tests/Selenium/ai.tests.d.ts
iusafaro: Fair. It's partially supported. Please create an issue about correct support for multi-ikey behavior in JS SDK.

SergeyKanzhelev: please create an issue or address it in PR now =)

JavaScript/JavaScriptSDK.Tests/Selenium/ai.tests.d.ts
SergeyKanzhelev: Thank you for the explanation. Perhaps the feature that allows to pre-set the `ikey` to `appid` mapping may be useful here. Basically let customer configure it in application deployment time. On certain environments like web apps we can pre-set it during the deployment

SergeyKanzhelev: this feature will address beacon API support concern

iusafaro: @SergeyKanzhelev, please check out my latest commit with exposing appId in the config.

API-reference.md
```diff
@@ -332,6 +333,11 @@ interface IConfig {
     // If true, the SDK will track all [Browser Link](https://docs.microsoft.com/en-us/aspnet/core/client-side/using-browserlink) requests. 
     // Default: false
     isBrowserLinkTrackingEnabled: boolean;
+
+    // AppId is used for the correlation between AJAX dependencies happening on the client-side with the server-side requests. 
+    // When Beacon API is enabled it can not be used automatically, but can be set manually in the configuration.
+    // Default: null
+    appId: string;
```
SergeyKanzhelev: same comment as before. It needs to be made a hash map so mapping between key-appid can be registered, not a single ikey

JavaScript/JavaScriptSDK.Tests/Selenium/ai.tests.d.ts
iusafaro: It's not about one line of change: Improve multi-ikey support #576

API-reference.md
iusafaro: I would disagree here - you can set only one configuration instance and only one ikey in this object, so logically it's correct to have one appId related to that ikey. As a user I would be really confused by having a map here.
I've opened issue and it'll require additional design discussions: Improve multi-ikey support #576

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -159,6 +159,12 @@ module Microsoft.ApplicationInsights {
 
             if (CorrelationIdHelper.canIncludeCorrelationHeader(this.appInsights.config, xhr.ajaxData.getAbsoluteUrl())) {
                 xhr.setRequestHeader(RequestHeaders.requestIdHeader, xhr.ajaxData.id);
+                if (this.appInsights.context) {
+                    var appId = this.appInsights.context.appId();
```
SergeyKanzhelev: same comment as before. Just mention in the corresponding issue. This `appid` should be taken from ajaxData object's instrumentation key. This is why for instance we call telemetry initializers twice in .NET - making sure that ikey is populated on `DependencyTelemetry`

API-reference.md
SergeyKanzhelev: we can have a single `appId` here. However there should be mechanism to provide mapping for more keys somehow. For the multi-key users.

Issue #566, opened 2018-02-08
Windows OS cuts up string variables to 1024 letters


biosis21:
Hi!
When I send data via trackEvent method and attribute **properties** contains array type of data you will see serealized as a string.
`{identifiers: ['1', '2', '3', ... , '5000' ]}` - I sent to the application insight library method
received `{identifiers: '1,2, 3, ... , 5000'}` - Track event sent to server
Windows platform cannot support this and browser tab is broken. I cannot do anything with my browser. Warning message is:
![screenshot_3 jpgss jpg](https://user-images.githubusercontent.com/6310176/35973577-a1874c38-0cde-11e8-83ea-bba8f660fafb.jpg)

SergeyKanzhelev:
Yes. `sanitizeProperties` should sanitize values to [8192](https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/EventData.bond#L19), not 1024 as it does today:
https://github.com/Microsoft/ApplicationInsights-JS/blob/cac7b363a7f59147adff510f698bc3f4943f2ab0/JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts#L70-L81

@jpiyali can you take a look?

jpiyali:
sorry, missed this thread earlier, @SergeyKanzhelev - current max string length is 1024. Where is the 8192 definition coming from?
* Max length allowed for custom values.
*/
        private static MAX_STRING_LENGTH = 1024;

SergeyKanzhelev:
I'm not sure where the current string limit comes from. Every field has it's own limit. For the custom property value - the limit is 8K.

jpiyali:
Thanks @SergeyKanzhelev, I confirmed in collector that this should be 8k as well. @biosis21 - we'll fix this in the next release. I'll post a release date soon.

jpiyali:
We are trying to push this fix out this week.

jpiyali:
Fix released in 1.0.16. Closing issue.

PR #580, opened 2018-03-30
Bug fix for data sanitization error for property value length, should…

jpiyali:
… be 8k

### DIFF
--- a/JavaScript/JavaScriptSDK/Logging.ts
+++ b/JavaScript/JavaScriptSDK/Logging.ts
@@ -86,6 +86,7 @@
         UrlTooLong,
         SessionStorageBufferFull,
         CannotAccessCookie,
+        IdTooLong
     }
 
     export class _InternalLogMessage {
--- a/JavaScript/JavaScriptSDK/Logging.ts
+++ b/JavaScript/JavaScriptSDK/Logging.ts
@@ -86,7 +86,7 @@
         UrlTooLong,
         SessionStorageBufferFull,
         CannotAccessCookie,
-        IdTooLong
+        IdTooLong,
     }
 
     export class _InternalLogMessage {
--- a/JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
+++ b/JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
@@ -12,7 +12,17 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
         private static MAX_NAME_LENGTH = 150;
 
         /**
+         * Max length allowed for Id field in page views.
+         */
+        private static MAX_ID_LENGTH = 128;
+
+        /**
          * Max length allowed for custom values.
+         */
+        private static MAX_PROPERTY_LENGTH = 8192;
+
+        /**
+         * Max length allowed for names
          */
         private static MAX_STRING_LENGTH = 1024;
 
@@ -67,15 +77,15 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
             return name;
         }
 
-        public static sanitizeString(value) {
+        public static sanitizeString(value, maxLength: number = DataSanitizer.MAX_STRING_LENGTH) {
             if (value) {
                 value = Util.trim(value);
-                if (value.toString().length > DataSanitizer.MAX_STRING_LENGTH) {
-                    value = value.toString().substring(0, DataSanitizer.MAX_STRING_LENGTH);
+                if (value.toString().length > maxLength) {
+                    value = value.toString().substring(0, maxLength);
                     _InternalLogging.throwInternal(
                         LoggingSeverity.WARNING,
                         _InternalMessageId.StringValueTooLong,
-                        "string value is too long. It has been truncated to " + DataSanitizer.MAX_STRING_LENGTH + " characters.",
+                        "string value is too long. It has been truncated to " + maxLength + " characters.",
                         { value: value }, true);
                 }
             }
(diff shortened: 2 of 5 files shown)


jpiyali:
Issue Id is https://github.com/Microsoft/ApplicationInsights-JS/issues/566

JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
```diff
@@ -67,15 +77,15 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
...
                 value = Util.trim(value);
-                if (value.toString().length > DataSanitizer.MAX_STRING_LENGTH) {
-                    value = value.toString().substring(0, DataSanitizer.MAX_STRING_LENGTH);
+                if (value.toString().length > maxLength) {
+                    value = value.toString().substring(0, maxLength);
                     _InternalLogging.throwInternal(
                         LoggingSeverity.WARNING,
                         _InternalMessageId.StringValueTooLong,
-                        "string value is too long. It has been truncated to " + DataSanitizer.MAX_STRING_LENGTH + " characters.",
+                        "string value is too long. It has been truncated to " + maxLength + " characters.",
```
gardnerjr: as a future task, can we possibly also get the *key* corresponding of the value that was trimmed in this message?  the message tells you what the *value* was that was truncated, but knowing that it was a custom property with name `x` or the field `url` or whatever would be really useful to track down what the source of that value was.

JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
```diff
@@ -156,6 +166,23 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
             return measurements;
         }
 
+        public static sanitizeId(id: string) {
```
gardnerjr: you changed the other method to take a value for maxlength that has a default, why not do the same here?  (or why is there a different method for ids at all, when it's the same rules but different trim length?

JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
```diff
@@ -156,6 +166,23 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
...
+            if (id) {
+                id = Util.trim(id);
+                if (id.length > DataSanitizer.MAX_ID_LENGTH) {
+                    id = id.substring(0, DataSanitizer.MAX_ID_LENGTH);
+                    _InternalLogging.throwInternal(
+                        LoggingSeverity.WARNING,
+                        _InternalMessageId.UrlTooLong,
+                        "id is too long, it has been truncated to " + DataSanitizer.MAX_URL_LENGTH + " characters.",
```
gardnerjr: the message has a different value (url length) than the checks did (id length)

gardnerjr: (and has messageid url too long)

JavaScript/JavaScriptSDK/Logging.ts
```diff
@@ -86,6 +86,7 @@
         UrlTooLong,
         SessionStorageBufferFull,
         CannotAccessCookie,
+        IdTooLong
```
SergeyKanzhelev: nit: it is a good practice to have comma in the end of the list.

JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
jpiyali: Fixed.

JavaScript/JavaScriptSDK/Logging.ts
jpiyali: Updated.

JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
jpiyali: Combined sanitizeUrl and sanitizeId() to call single sanitizeInput() method.

JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
jpiyali: It is possible to do this now with the associated length (value is different for name vs. custom property). we use this method for these two types only).

JavaScript/JavaScriptSDK/Telemetry/Common/DataSanitizer.ts
```diff
@@ -156,6 +154,27 @@ module Microsoft.ApplicationInsights.Telemetry.Common {
             return measurements;
         }
 
+        public static sanitizeId(id: string): string {
+                return id ? DataSanitizer.sanitizeInput(id, DataSanitizer.MAX_ID_LENGTH, _InternalMessageId.IdTooLong).toString() : id;
+        }
+
+        public static sanitizeInput(input: any, maxLength: number, _msgId: _InternalMessageId) {
```
Dmitry-Matveev: `sanitizeInput()` is indirectly tested with `sanitizeUrl()` and `sanitizeId()` - would you like to explicitly test it as well?
It's in public API and one would not know that it's untested if one is to take dependency on this method later.

jpiyali: Added unit tests.

PR #582, opened 2018-04-11
temporarily remove pageview.id initialization

iusafaro:
Piyali and I started separate conversation with Kusto team about only this particular field and they haven't replied to us for the past two weeks. So, we decided to release without this change then bring it back and immediately deploy to AIMon with it to see how it's going.


SergeyKanzhelev:
Can you provide more reasoning in PR description?

PR #584, opened 2018-04-13
Prevent getting Request-Context header when it's not returned

iusafaro:


JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -260,8 +260,12 @@ module Microsoft.ApplicationInsights {
 
         private getCorrelationContext(xhr: XMLHttpRequestInstrumented) {
             try {
-                var responseHeader = xhr.getResponseHeader(RequestHeaders.requestContextHeader);
-                return CorrelationIdHelper.getCorrelationContext(responseHeader);
+                var responseHeadersString = xhr.getAllResponseHeaders();
+                var index = responseHeadersString.toLowerCase().indexOf(RequestHeaders.requestContextHeader.toLowerCase());
```
jpiyali: add null check

JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -260,8 +260,14 @@ module Microsoft.ApplicationInsights {
 
         private getCorrelationContext(xhr: XMLHttpRequestInstrumented) {
             try {
-                var responseHeader = xhr.getResponseHeader(RequestHeaders.requestContextHeader);
-                return CorrelationIdHelper.getCorrelationContext(responseHeader);
+                var responseHeadersString = xhr.getAllResponseHeaders();
+                if (responseHeadersString !== null) {
+                    var index = responseHeadersString.toLowerCase().indexOf(RequestHeaders.requestContextHeader.toLowerCase());
```
OsvaldoRosado: Would it be better to make the RequestHeaders constant itself lowercase than run toLowerCase on it?

OsvaldoRosado: I'm also curious why this is necessary. Docs say `xhr.getResponseHeader` will return null if not found. Wouldn't it be easier to check for that since we have to run it anyway?

jpiyali: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getAllResponseHeaders

iusafaro: yes, made it a separate constant. 
the problem is that it returns null but also logs an error to the console, even if surrounded by a try/catch if this header is not listed in the access-control-expose-headers.

OsvaldoRosado: I mean, why do we have to run `getAllResponseHeaders` just to check `indexOf` and call `getResponseHeader` right after? `getResponseHeader` is already safe and will return null if the header is not present.

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getResponseHeader

OsvaldoRosado: Didn't see your comment in time @iusafaro. Makes sense!

jpiyali: She is fixing the bug stemming from xhr.getResponseHeader (throws a console error for each dependency to cross domain).

PR #591, opened 2018-04-24
Creating folder for coreSDK.

jpiyali:
Initial copy of all files from Application insights SDK.


SergeyKanzhelev:
In the interest of Open Source development it would be great if you can create an issue outlining high level plan of introduction of core SDK.

Also this branch is behind of `master`. It means that some of copies of files already old. Is it critical for code development?

jpiyali:
I will create a separate issue tracking the explanation behind the coreSDK folder. Thank you for guidance here.
Its not critical for code to stay in sync. M1 goal is to pick up existing SDK, clean up as required to have:

1.	Lightweight core that initializes and sets up registration and initializes plugins (not merging this back to Application insights JS)
2.	Works with a AI channel to send data (Basel is working on this).
3.	Works with 1st party channel to send data.

I do need a build leg for this folder and for main build to exclude this folder altogether. My offline thread about the repo is about this. We will keep stuff here short term and update build definition.

jpiyali:
Created issue https://github.com/Microsoft/ApplicationInsights-JS/issues/595 to track work against.

PR #593, opened 2018-04-26
Parse response & config setting for CORS correlation

iusafaro:


SergeyKanzhelev:
@iusafaro can you please provide a description and issue number for this PR?

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -580,19 +580,24 @@ module Microsoft.ApplicationInsights {
         /**
         * Checks if a request url is not on a excluded domain list and if it is safe to add correlation headers
         */
-        public static canIncludeCorrelationHeader(config: IConfig, requestUrl: string) {
+        public static canIncludeCorrelationHeader(config: IConfig, requestUrl: string, currentHost: string) {
             if (config && config.disableCorrelationHeaders) {
```
jpiyali: is this a global setting for enabling/not enabling correlation. and enableCorsCorrelation is for cross domain specificially?

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -580,19 +580,24 @@ module Microsoft.ApplicationInsights {
...
                 return false;
             }
 
+            let requestHost = UrlHelper.parseUrl(requestUrl).host;
+            if (config && !config.enableCorsCorrelation && requestHost !== currentHost) {
```
jpiyali: probably should ignore case sensitivity.

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
```diff
@@ -1862,6 +1901,7 @@ class AppInsightsTests extends TestClass {
                 var snippet = this.getAppInsightsSnippet();
                 snippet.disableAjaxTracking = false;
                 snippet.disableCorrelationHeaders = false;
+                snippet.enableCorsCorrelation = true;
```
jpiyali: this should be tested with false as default?

JavaScript/JavaScriptSDK/Util.ts
iusafaro: defined everything in the issue as Sergey requested - Create separate configuration setting for enabling CORS correlation #594

JavaScript/JavaScriptSDK.Tests/CheckinTests/appInsights.tests.ts
iusafaro: this is the test for excluded domains that are honored only for cross-domain. i have different tests for  all combinations of those flags and every other test has enableCorsCorrelation  as false by the default.

JavaScript/JavaScriptSDK/Util.ts
iusafaro: yes. thanks!

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -580,19 +580,28 @@ module Microsoft.ApplicationInsights {
...
+        public static canIncludeCorrelationHeader(config: IConfig, requestUrl: string, currentHost: string) {
             if (config && config.disableCorrelationHeaders) {
                 return false;
             }
 
+            if (!requestUrl) {
+                return false;
+            }
+
+            let requestHost = UrlHelper.parseUrl(requestUrl).host.toLowerCase();
```
SergeyKanzhelev: you need to include `port` into comparison. As I recall

SergeyKanzhelev: Ideally paste a link to official documentation on how browsers compare hosts for CORS. There are official rules

iusafaro: Port included in the host - I've double checked for different browsers. This comparison existed since the initial implementation (legacy headers) and here we improved it to be case insensitive.

JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -580,19 +580,28 @@ module Microsoft.ApplicationInsights {
...
+            if (config && !config.enableCorsCorrelation && requestHost !== currentHost) {
```
jpiyali: This should be if((config && !config.enableCorsCorrelation) || requestHost !== currentHost) // either should return false

iusafaro: no, it should be && => we should proceed (not return false) if hosts don't match but enableCorsCorrelation is true.

molochiy:
hi guys, you are doing amazing stuff. how long should we wait for this changes in production?

PR #600, opened 2018-05-03
create NPM package for coreSDK

barustum:
This change allows us to be able to create npm package for core sdk. For now, I'll publish to npm manually and take a dependency on it from the AppInsights Channel plugin. Eventually, I'll setup an automated build just like how we do it for JS SDK.

Note that I am publishing this package on my own npm login, and not the appinsights one. So for testing purposes, and as I am making changes, I will be incrementing the package.json version number. Otherwise, npm wouldn't allow me to publish. Once we are ready to publish to appinsights npm login, we can reset the version to 0.0.1


barustum:
I will abandon this pull request for now as me and @jpiyali synched up last week. We will create a common npm package for core and AI channel to use.

PR #610, opened 2018-05-24
Turn common and coresdk into namespace(less) modules

barustum:
- Turn common and coresdk into external modules
- support building commonjs and amd without making code changes from consumer side.


applicationinsights-common.ts
```diff
@@ -0,0 +1,5 @@
+export { Util, CorrelationIdHelper, UrlHelper } from './AppInsightsCommon/Util';
```
jpiyali: I would recommend you move all files inside applicationinsights common down into a sub dir. That way this file can be part of the root AppInsightsCommon directory.

PR #612, opened 2018-05-29
Update API-reference.md

hofuchi:
data type for samplingPercentage was boolean, changing to number since it's a percentage and default was 100


jpiyali:
@hofuchi - thank you for updating the docs.

PR #616, opened 2018-06-06
Add Validators. Refactor coreSDK folder

barustum:
- Add Validators
- Move files from coreSDK to Common and Channel
- Fix cjs and amd folder structures for Common and Channel to match coreSDK


barustum:
Please note that the files under "Generated" are modified manually and not generated. I will investigate how to run the generation tool and update them.

coreSDK/amd/.npmignore
```diff
@@ -5,6 +5,6 @@
 
 # ... but these files
 !package.json
-!bundle/aicore.js
```
barustum: @jpiyali I changed from aicore to applicationinsights-core-js for consistency. I can change it back if you feel like we should stick with aicore. I can go either way. If we change it to applicationinsights-core-js, we'll need to update the npm-vsts feed

jpiyali: should the file name be just core.js and not so large? I checked a few other existing public packages, names do not need to be reflective of package?

AppInsightsChannel/Sender.ts
```diff
@@ -13,31 +14,43 @@ import { MetricValidator } from './TelemetryValidation/MetricValidator';
...
 declare var XDomainRequest: {
     prototype: IXDomainRequest;
     new(): IXDomainRequest;
 };
 
-export class Sender implements ITelemetryPlugin {
-    priority: number; // todo set priority
-    public identifier: string; // todo set identifier
+export class Sender implements IChannelControls {
+    public priority: number; // todo set priority
```
jpiyali: set any value > 100

AppInsightsChannel/TelemetryValidation/PageViewPerformanceValidator.ts
```diff
@@ -1,10 +1,28 @@
...
+        if (!item.domainProperties ||
+            !item.domainProperties["domProcessing"] ||
+            !item.domainProperties["duration"] ||
+            !item.domainProperties["name"] ||
+            !item.domainProperties["networkConnect"] ||
+            !item.domainProperties["perfTotal"] ||
+            !item.domainProperties["receivedResponse"] ||
+            !item.domainProperties["sentRequest"] ||
+            !item.domainProperties["url"]) {
+            return false;
```
jpiyali: add log when returning false.

AppInsightsChannel/cjs/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "applicationinsights-channel-js",
+    "version": "0.0.1",
+    "description": "Microsoft Application Insights JavaScript SDK Channel",
+    "main": "bundle/aichannel.js",
+    "devDependencies": {
+        "grunt": "1.0.1",
+        "grunt-contrib-qunit": "2.0.0",
```
jpiyali: we should review each of our new package.json for the grunt dependency.

PR #636, opened 2018-07-23
feat(fetch): add fetch monitor

MaximBalaganskiy:


JavaScript/JavaScriptSDK/ajax/fetch.ts
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        private Init(): void {
+            if (this.supportsMonitoring()) {
+                this.instrumentFetch();
+                this.initialized = true;
+            }
+        }
+
+        public static DisabledPropertyName: string = "Microsoft_ApplicationInsights_BypassFetchInstrumentation";
```
jpiyali: Unused, did you want to add if fetch request have this property set so that those are not monitored?

jpiyali:
@MaximBalaganskiy  - thanks for contributing to the SDK. can you please add some unit tests for the monitor?

MaximBalaganskiy:
Can do. Looked at the custom test framework and couldn't figure out in a
timeframe I had. Can you point me to an example on how to use addTestAsync?

jpiyali:
Check out TestCaseAsync usage in AppInsightsModule.tests.ts. Option is minimal but would it work for you?
For the unused variable, possibly remove it since its not being used currently.

Thanks!

MaximBalaganskiy:
I've updated the monitor to check the disable flag. As for the tests, I couldn't make it work - this project's test harness is a bit hard to get around. Promise stub is missing and to add it I would have to include another package source code into the project, plus typings. Sorry, I'm just not that committed to this PR. 
Feel free to close it. I'll be using my monitor for now.

MaximBalaganskiy:
ok, i figured it out for the async in browser `useFakeTimers` sould be false.... but the question still stands for tests not working in CLI

MaximBalaganskiy:
Ok, all done

JavaScript/JavaScriptSDK/AppInsights.ts
```diff
@@ -131,6 +133,7 @@ module Microsoft.ApplicationInsights {
 
             if (!this.config.disableAjaxTracking) {
```
jpiyali: Thanks for nailing the unit tests @	MaximBalaganskiy . I missed adding a comment here, can you add a disableFetchTracking and default it to true for the fetchMonitor. We can take changes in PROD and test. Also, this can potentially impact customer bill, so we cannot start collecting by default.

MaximBalaganskiy:
Done. I also fixed an issue when an async test could not be run with the `setTimeout` being overwritten by a stub.

MaximBalaganskiy:
What was causing the issue in IE10? Could that be fixed instead of reverting the PR?

MaximBalaganskiy:
I've just tried the `ai.1.0.19.js` script in IE10 without a Promise polyfill and had no errors at all.

jpiyali:
@MaximBalaganskiy  - issue is in IE8 and below. We could have tried a promise polyfill, but that would have increased the size of the SDK needlessly.

jpiyali:
we'll make the implementation as an extension in the SDK 2.0.0. We'll be releasing beta at end of the month.

MaximBalaganskiy:
IE10 does not have `Promise` either

PR #637, opened 2018-07-23
clean up and add test framework for applicationinsights-analytics

barustum:
1. Clean up PageViewManager.trackPageView to make it more humanly readable
2. Add test framework for applicationinsights-analytics-js
3. Add 1 test to ensure test can run fine. Will add more tests later.


barustum:
Anandoning this pull request as I am merging into another private branch that already has a pull request sent.

PR #638, opened 2018-07-23
trackPageView cleanup and add test framework

barustum:
1. Clean up PageViewManager.trackPageView to make it more humanly readable
2. Add test framework for applicationinsights-analytics-js
3. Add 1 test to ensure test can run fine. Will add more tests later.


ApplicationInsights/JavaScriptSDK/Telemetry/PageViewManager.ts
```diff
@@ -54,50 +56,52 @@ export class PageViewManager {
...
                 measurements);
-                this._channel.flush();
+            this._channel.flush();
             pageViewSent = true;
         }
 
+        // now try to send the page view performance telemetry
         var maxDurationLimit = 60000;
-
-        if (!PageViewPerformance.isPerformanceTimingSupported()) {
```
jpiyali: you'll be adding this back again right when the return false on isPerformanceTimingSupported switches to looking at browser functionality? If so, can you track this both through a task/

barustum: I did not remove it. I moved it up to line 62. With this pull request, I did not change any functionality to what it was. Just cleaning this method as it was confusing to me. Now it's a little bit more readable.

barustum: Once you're free, we can pair on this to discuss the change as it is easier to walk through it face to face.

PR #640, opened 2018-07-24
generate definition files for all builds but module

barustum:


gruntfile.js
```diff
@@ -54,7 +54,7 @@ module.exports = function (grunt) {
                 out: 'AISKU/amd/bundle/aisdk-js.js'
              },
             module: {
-                tsconfig: './tsconfig.json',
+                tsconfig: './tsconfigmodule.json',
```
jpiyali: add a comment in the new file as to why its own tsconfig for the module

PR #644, opened 2018-08-02
Switch to using webpack, check in bootstrap code

jpiyali:
PR contains:
- Switch to webpack for packing, use umd target for pack.
- Check in dist files for m2.
- Remove r.js configuration for prior packing
- Check in snippet code m2 for bootstrap scenario.


AISKU/webpack.config.js
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }));
+        filename_suffix = filename_suffix + ".min";
+    }
+
+    filename = filename_suffix + ".js";
+
+    return {
+        entry: entry,
+        mode: "development",
```
barustum: >development [](start = 15, length = 11)

did you mean to change this back to production?

barustum:
did you mean to check in the empty file snippet.build.js?

Issue #654, opened 2018-08-24
ai.0.js failing on 8 and older version of IE


kshcon:
Hi,

I have used Application Insights on many modules of one of our web-based application. I have added Application Insights for the JavaScript tracking as well.
Some of the HTML pages are destined for hardware devices which run IE version 7 or older. Since last night, 10:00 PM UTC, the JS are failing due to this Application Insights usage. It is failing even if we are using desktop browsers with emulation of IE 7 or 8. It started working fine when we disabled App Insights in the app (we stopped calling the JS function initializing it).

I'd appreciate if anyone confirms it is an incompatibility in the new version or if there is a way to call it differently in the older versions. If no option works, we will be forced to keep the JS logging disabled.

The function we use to initialize App Insights is as following:

> function ConfigureAppInsight(instrumentationKey) {
>     if (instrumentationKey != null && instrumentationKey.length > 0) {
>         var appInsights = window.appInsights || function (config) {
>             function i(config) { t[config] = function () { var i = arguments; t.queue.push(function () { t[config].apply(t, i) }) } } var t = { config: config }, u = document, e = window, o = "script", s = "AuthenticatedUserContext", h = "start", c = "stop", l = "Track", a = l + "Event", v = l + "Page", y = u.createElement(o), r, f; y.src = config.url || "https://az416426.vo.msecnd.net/scripts/a/ai.0.js"; u.getElementsByTagName(o)[0].parentNode.appendChild(y); try { t.cookie = u.cookie } catch (p) { } for (t.queue = [], t.version = "1.0", r = ["Event", "Exception", "Metric", "PageView", "Trace", "Dependency"]; r.length;) i("track" + r.pop()); return i("set" + s), i("clear" + s), i(h + a), i(c + a), i(h + v), i(c + v), i("flush"), config.disableExceptionTracking || (r = "onerror", i("_" + r), f = e[r], e[r] = function (config, i, u, e, o) { var s = f && f(config, i, u, e, o); return s !== !0 && t["_" + r](config, i, u, e, o), s }), t
>         }({
>             instrumentationKey: instrumentationKey
>         });
> 
>         window.appInsights = appInsights;
>         appInsights.trackPageView();
>         return appInsights;
>     }
>     else
>         return null;
> }

Error seen in the browser console:
![appinsightissuescreenshot2](https://user-images.githubusercontent.com/42668899/44574457-16e0ac00-a7a7-11e8-9536-40b986c343e6.png)

[...]

PR #655, opened 2018-08-24
Remove promise chaining syntax

jpiyali:
The issue with the fetch api instrumentation is that fetch api is based on promise and there is no support for promise in older versions of some browsers. While the code is not run in old versions, script load is failing. This may be due to promise chain syntax that we had. so, taking that out and replacing with alternate syntax.
I still think this won't work, as I'm seeing another script error, much earlier in the script though.
Need to test this out some more. We may need to take out fetch support altogether if this doesnt work.


jpiyali:
#654

jpiyali:
The script error was resolved, am going to try this to validate fix.

PR #664, opened 2018-09-10
Remove Logging from Common and use Core's DiagnosticLogger

markwolff:
This update removes common's logging functionality and uses Core's DiagnosticLogger instead. In addition, Util.ts and DataSanitizer.ts were changed to have all function-type members to require a DiagnosticLogger argument in order to keep these classes as static. Changes are towards next version of SDK.


ApplicationInsights/JavaScriptSDK/Extensions/ajax/ajax.ts
```diff
@@ -1,4 +1,5 @@
-﻿import { RequestHeaders, Util, LoggingSeverity, _InternalMessageId, _InternalLogging, CorrelationIdHelper } from 'applicationinsights-common';
+﻿import { RequestHeaders, Util, CorrelationIdHelper } from 'applicationinsights-common';
```
jpiyali: remove extra line

PR #667, opened 2018-09-13
Add IConfig flag for onerror instrumented

markwolff:
Internal flag which will be used to determine if window.onerror has been already instrumented by Application Analytics


AppInsightsCommon/Interfaces/IConfig.ts
```diff
@@ -31,4 +31,7 @@ export interface IConfig {
     isBrowserLinkTrackingEnabled?: boolean;
     appId?: string;
     enableCorsCorrelation?: boolean;
+
+    // Internal
+    aiInstrumented?: boolean;
 }
```
jpiyali: PerhAps call this exceptionsInstrumented

PR #668, opened 2018-09-13
add track exception

markwolff:
- Add trackException to ApplicationInsightsAnalytics API

- Add option to ApplicationInsightsAnalytics config to enable auto collection of uncaught exceptions (this is opt-out, as it was in old SDK)


ApplicationInsights/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -199,7 +201,79 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
...
+                { exception: Util.dump(e) });
+        }
+    }
+
+    /**
+     * @description Custom error handler for Application Insights Analytics
+     * @param {IAutoExceptionTelemetry} exception
+     * @memberof ApplicationInsights
+     */
+    public _onerror(exception: IAutoExceptionTelemetry): void {
```
jpiyali: did this exist before?

AISKU/snippet/snippet.js
```diff
@@ -31,14 +31,39 @@
...
 
-    var method = ["PageView"];
+    var method = ["PageView", "Exception"];
     while (method.length) {
         createLazyMethod("track" + method.pop());
     }
 
+    // Collect global errors
+    if (aiConfig.extensionsConfig &&
+        aiConfig.extensionsConfig.ApplicationInsightsAnalytics &&
```
jpiyali: use aiConfig.extensionsConfig[this.identifier] as configurations are by plugin identifier

jpiyali: or add comment here

ApplicationInsights/JavaScriptSDK/ApplicationInsights.ts
barustum: Yes. This is the method that will be called by the snippet (snippet.js line 53) in the case of auto collection. It essentially calls trackException. 

---
In reply to: [217542252](https://github.com/Microsoft/ApplicationInsights-JS/pull/668#discussion_r217542252) [](ancestors = 217542252)

AISKU/snippet/snippet.js
markwolff: I've added a comment to clarify 7c915f3

PR #670, opened 2018-09-17
Add startstopTrackEvent

markwolff:
Adds startTrackEvent and stopTrackEvent to the AppInsights (ApplicationInsightsAnalytics) module

### DIFF
--- a/ApplicationInsights/JavaScriptSDK/ApplicationInsights.ts
+++ b/ApplicationInsights/JavaScriptSDK/ApplicationInsights.ts
@@ -5,7 +5,7 @@
 
 import {
     IConfig,
-    Util, PageViewPerformance,
+    Util, PageViewPerformance, Event,
     PageView, IEnvelope, RemoteDependencyData,
     Data, Metric, Exception, SeverityLevel
 } from "applicationinsights-common";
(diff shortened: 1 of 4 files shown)


ApplicationInsights/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -141,14 +138,35 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
...
-        telemetryItem.instrumentationKey = this._globalconfig.instrumentationKey;
-
-        var iKeyNoDashes = this._globalconfig.instrumentationKey.replace(/-/g, "");
-        telemetryItem.name = telemetryItem.name.replace("{0}", iKeyNoDashes);
+        this._setTelemetryNameAndIKey(telemetryItem);
 
         this.core.track(telemetryItem);
     }
 
+    public startTrackEvent(name: string) {
```
SergeyKanzhelev: Do we need those methods? I feel like those are hard to explain and nobody actually uses them

jpiyali: We don't have history but I thought the intent was to allow people to do timing using a custom Event (dfferent than page view perf calculation). I agree it does not provide much value as part C data. Do you have any history in terms of customer asks/features this was added to support initially?

PR #672, opened 2018-09-19
Enable DiagnosticLogger in AISKU

markwolff:
Resolves compile issues from DiagnosticLogger changes in Core


AISKU/Init.ts
```diff
@@ -92,5 +87,5 @@ try {
     //     // core.track(telemetryItem);
     // }
 } catch (e) {
-    _InternalLogging.warnToConsole('Failed to initialize AppInsights JS SDK: ' + e.message);
+    // _InternalLogging.warnToConsole('Failed to initialize AppInsights JS SDK: ' + e.message);
```
jpiyali: Add comment here we need to find proper home to warn to console when SDK initialization fails.

AISKU/Initialization.ts
```diff
@@ -152,7 +147,8 @@ export class Initialization {
 
         config.enableDebug = Util.stringToBoolOrDefault(config.enableDebug);
         config.disableExceptionTracking = Util.stringToBoolOrDefault(config.disableExceptionTracking);
-        config.verboseLogging = Util.stringToBoolOrDefault(config.verboseLogging);
+        config.consoleLoggingLevel = config.consoleLoggingLevel || 0;
```
jpiyali: add comment on what is 0 and 2

AISKU/package.json
```diff
@@ -15,11 +15,11 @@
...
-        "applicationinsights-analytics-js": "0.0.x",
-        "applicationinsights-channel-js": "0.0.x",
-        "applicationinsights-common": "0.0.x",
-        "applicationinsights-core-js": "0.0.x",
-        "applicationinsights-properties-js": "0.0.x",
+        "applicationinsights-analytics-js": "0.0.89",
+        "applicationinsights-channel-js": "0.0.31",
+        "applicationinsights-common": "0.0.77",
+        "applicationinsights-core-js": "0.0.79",
+        "applicationinsights-properties-js": "0.0.4",
```
jpiyali: update to odd and update this package (to use updated core/common) then udpate this one.

PR #687, opened 2018-10-01
Fix ajax tracking; fix analytics exception autocollection

markwolff:
- [x] DONE: Update package versions once ready


ApplicationInsights/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -428,7 +428,7 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
                 }
 
                 return handled;
-            }
+            }.bind(this);
```
jpiyali: is it failing on this._onerror?

jpiyali: What does the comment // handled could be typeof function mean?

markwolff: yes it fails at this._onerror. It might be bad to pollute the `this` of onerror. Do you think it's better to call a window instance? if so, will it be ok for multiple instances of sdk initialized?

comment is just clarifying why it is handled !== true instead of !handled

jpiyali: Yeah, you can potentially just do window[sdkinstancename]._onerror(). We should save the sdkinstance name somewhere in this class (pass it in through configuration).

AISKU/Initialization.ts
```diff
@@ -91,18 +92,16 @@ export class Initialization implements IApplicationInsights {
...
         extensions.push(this.dependencies);
         extensions.push(this.appInsights);
 
         // initialize core
+        window[windowName] = this;
         this.core.initialize(this.config, extensions);
-
-        // initialize extensions
-        this.appInsights.initialize(this.config, this.core, extensions);
-        appInsightsChannel.initialize(this.config, this.core, extensions);
```
markwolff: Can these be deleted since core initializes each extension?

AISKU/Initialization.ts
```diff
@@ -91,18 +92,16 @@ export class Initialization implements IApplicationInsights {
...
         let appInsightsChannel: Sender = new Sender();
+        const windowName: string = this.config.extensionConfig.ApplicationInsightsAnalytics.sdkInstanceName || Initialization.defaultName;
 
         extensions.push(appInsightsChannel);
         extensions.push(this.properties);
         extensions.push(this.dependencies);
         extensions.push(this.appInsights);
 
         // initialize core
+        window[windowName] = this;
```
markwolff: important

AISKU/Initialization.ts
```diff
@@ -192,7 +191,14 @@ export class Initialization implements IApplicationInsights {
...
 
-        let config = configuration.extensionConfig && configuration.extensionConfig[identifier] ? <IConfig>configuration.extensionConfig[identifier] : {};
+        // Undefined checks
+        if (!configuration.extensionConfig) {
+            configuration.extensionConfig = {};
+        }
+        if (!configuration.extensionConfig[identifier]) {
+            configuration.extensionConfig[identifier] = {};
+        }
+        const config: IConfig = configuration.extensionConfig[identifier]; // ref to main config
```
markwolff: Note: `configuration.extensionConfig[identifier]` will get loaded with non-overridden default config

AISKU/Initialization.ts
jpiyali: you should not need to do this at this point. This should have been setup through loadAppInsights() return call.

AISKU/Initialization.ts
```diff
@@ -192,7 +191,14 @@ export class Initialization implements IApplicationInsights {
             identifier = identifier ? identifier : "ApplicationInsightsAnalytics";
```
jpiyali: Lets get rid of the identifier parameter in this method, its making this complicated, and its not as if we can support a different config for a different identifier here.

PR #705, opened 2018-10-30
Add misc files for publishing

markwolff:
- Add main entry point in rollup and package.json
- Add typedoc comments
- Add readme to all sub-repos
- Rename AISKU readme-v2 to README.md
- Rename applicationinsights-sdk-js to applicationinsights-web


AISKU/package.json
```diff
@@ -1,9 +1,10 @@
 {
     "name": "@microsoft/applicationinsights-web",
-    "version": "1.0.0-beta.2",
+    "version": "1.0.0-beta",
     "description": "Microsoft Application Insights Javascript SDK API 1.0 beta",
-    "main": "dist-esm/applicationinsights-sdk-js.js",
-    "types": "types/applicationinsights-sdk-js.d.ts",
+    "main": "dist/applicationinsights-web.min.js",
+    "module": "dist-esm/applicationinsights-web.js",
+    "types": "types/applicationinsights-web.d.ts",
```
jpiyali: totally unrelated to pr: did the api extractor plugin work to roll up interfaces to single file?

AISKU/src/Initialization.ts
```diff
@@ -19,16 +24,20 @@ export interface IApplicationInsights extends IAppInsights, IDependenciesPlugin,
     appInsights: ApplicationInsights;
 };
 
+/**
+ * Application Insights API
+ * @class Initialization
+ * @implements {IApplicationInsights}
+ */
 export class Initialization implements IApplicationInsights {
```
jpiyali: One thing we are missing here is making channels available or just the flush api. its a common requirement.

jpiyali: Internally, it will call getchannelcontrols().foreach(c => c.flush)

markwolff: got it. will add for next beta version

AISKU/package.json
markwolff: It left all of the external types as external instead of trying to resolve them

PR #712, opened 2018-11-09
Add window webworker checks to Common (Basic SKU)

markwolff:


AISKULight/rollup.config.js
```diff
@@ -17,7 +17,7 @@ const browserRollupConfigFactory = (isProduction, libV = '1') => {
       file: `browser/aib.${libV}.js`,
       banner: banner,
       format: "umd",
-      name: "aibasic",
+      name: "Microsoft.Azure",
```
jpiyali: does this create all types in namespace Microsoft.Azure?

markwolff: Only when not using a loader, and just referencing global namespace. For AISKU we manually modify window so it is at both places. It is a required field in rollup for bundling as umd.

jpiyali: but this is the rollup for browser (umd?)

markwolff: yes, it is the fallback for if no cjs or amd
```js
(function (global, factory) {
    typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
    typeof define === 'function' && define.amd ? define(['exports'], factory) :
    (factory((global.Microsoft = global.Microsoft || {}, global.Microsoft.Azure = {})));
}(this, (function (exports) { 'use strict';
```

PR #714, opened 2018-11-17
Change properties plugin to support CS 4.0 for part A extensions

jpiyali:


AppInsightsCommon/src/Interfaces/PartAExtensions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export interface UserExt {
+    id?: string;
+    localId?: string;
+    authId?: string;
+}
+
+export class PartAExtensions {
+    public UserExtensionName:string = "User";
+    public WebxtensionName = "Web";
```
markwolff: WebExtensionName

PR #724, opened 2018-11-28
Fix initialization from cdn

markwolff:
- Add build badges for minified/gzipped sizes
- Fix bug preventing initialization from cdn


AppInsightsCommon/src/Interfaces/PartAExtensions.ts
```diff
@@ -1,4 +1,4 @@
-import { ContextTagKeys } from '../applicationinsights-common';
+import { ContextTagKeys } from '../Interfaces/Contracts/Generated/ContextTagKeys';
```
jpiyali: looks like my mistake!

PR #728, opened 2018-11-30
Remove minified files from package.json, fix ConfigManager bug

markwolff:
- Fix ConfigurationManager bug
- Remove all instances of "browser" in package.json
- Bump dependencies to latest

### DIFF
--- a/AppInsightsCommon/src/Interfaces/IConfig.ts
+++ b/AppInsightsCommon/src/Interfaces/IConfig.ts
@@ -264,7 +264,7 @@ export interface IConfig {
 export class ConfigurationManager {
     public static getConfig(config: IConfiguration & IConfig, field: string, identifier?: string, defaultValue: number | string | boolean = false): number | string | boolean {
         let configValue;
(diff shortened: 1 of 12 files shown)


AppInsightsCommon/src/Interfaces/IConfig.ts
```diff
@@ -264,10 +264,10 @@ export interface IConfig {
 export class ConfigurationManager {
     public static getConfig(config: IConfiguration & IConfig, field: string, identifier?: string, defaultValue: number | string | boolean = false): number | string | boolean {
         let configValue;
-        if (!identifier || !config.extensionConfig || !config.extensionConfig[identifier]) {
-            configValue = config[field];
-        } else {
+        if (identifier && config.extensionConfig && config.extensionConfig[identifier] && !CoreUtils.isNullOrUndefined(config.extensionConfig[identifier][field])) {
```
jpiyali: was this still buggy?

markwolff: yes need to check for not undefined instead of truthy

PR #730, opened 2018-12-06
Add support to allow customers to reference new CDN with config parameter that provides deprecated api support.

jpiyali:
This provides an easy mechanism for customers to try out new SDK bits without changing their instrumentation code.
- Will add more test cases
- need to also add support for context.addTelemetryInitializer

### DIFF
--- a/AISKU/src/ApplicationInsightsDeprecated.ts
+++ b/AISKU/src/ApplicationInsightsDeprecated.ts
@@ -1,6 +1,6 @@
 import { IConfig, PageViewPerformance, SeverityLevel, Util, IPageViewPerformanceTelemetry, 
     IPageViewTelemetry, ITraceTelemetry, IMetricTelemetry, 
-    IAutoExceptionTelemetry, IDependencyTelemetry, IExceptionTelemetry } from "@microsoft/applicationinsights-common";
+    IAutoExceptionTelemetry, IDependencyTelemetry, IExceptionTelemetry, IEventTelemetry } from "@microsoft/applicationinsights-common";
 import { ITelemetryContext } from "@microsoft/applicationinsights-properties-js/types/Interfaces/ITelemetryContext";
 import { Snippet, IApplicationInsights } from "./Initialization";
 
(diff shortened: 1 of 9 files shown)


AISKU/src/ApplicationInsightsDeprecated.ts
```diff
@@ -0,0 +1,317 @@
...
+        this.appInsightsNew.clearAuthenticatedUserContext();
+    }
+    
+    _onerror(message: string, url: string, lineNumber: number, columnNumber: number, error: Error) {
+        this.appInsightsNew._onerror(<IAutoExceptionTelemetry>{ message: message, url: url, lineNumber: lineNumber, columnNumber: columnNumber, error: error });
+    }
+    
+    
+    startTrackEvent(name: string) {
+        throw new Error("Method not implemented.");
```
markwolff: How do we want to handle methods that don't exist in new sdk? If this method is rarely called in someone's code and they dont look for it, it will crash at a potentially bad time. Do we want to remove this so it fails at compile/intellisense time instead? or show a warning?

AISKU/src/ApplicationInsightsDeprecated.ts
```diff
@@ -0,0 +1,317 @@
...
+        config.verboseLogging = Util.stringToBoolOrDefault(config.verboseLogging);
+        config.emitLineDelimitedJson = Util.stringToBoolOrDefault(config.emitLineDelimitedJson);
+        config.diagnosticLogInterval = config.diagnosticLogInterval || 10000;
+        config.autoTrackPageVisitTime = Util.stringToBoolOrDefault(config.autoTrackPageVisitTime);
+
+        if (isNaN(config.samplingPercentage) || config.samplingPercentage <= 0 || config.samplingPercentage >= 100) {
+            config.samplingPercentage = 100;
+        }
+
+        config.disableAjaxTracking = Util.stringToBoolOrDefault(config.disableAjaxTracking);
```
markwolff: also need disableFetchTracking (and enableOldTags?)

AISKU/src/ApplicationInsightsDeprecated.ts
markwolff: also if these are deprecated, should document as such here (and docs, interface)

jpiyali: I added support for startTrackEvent, stopTrackEvent. I think there are valid use case scenario for this api.

AISKU/src/ApplicationInsightsDeprecated.ts
jpiyali: True, let me see whats a good solution for properties in this scenario.

jpiyali: I'll need to essentially change properties package to support all old or all new. I'll do that in next PR along with support for addTelemetryInitializer.

PR #731, opened 2018-12-10
Update for Part B/C schema changes

markwolff:
- Add `properties`, `measurements`, to PageView, PageViewPerformance Telemetry classes.


ApplicationInsights/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -271,12 +276,16 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
         this.core.track(telemetryItem);
     }
 
+    /**
+     * Send browser performance metrics.
+     * @param pageViewPerformance
+     * @param customProperties
```
jpiyali: wondering if we should mark these as unused in the SKUs in comments

markwolff: actually we can just delete from SKU since its external facing

PR #733, opened 2018-12-12
Add loading performance info to README

markwolff:
- Add short blurb of performance info to README


AISKU/README.md
```diff
@@ -184,6 +184,16 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 	npm run test
 	```
 
+## Performance
+At just 22 KB gzipped, and taking only ~15 ms to initialize, Application Insights adds a neglible amount of loadtime to your website. By using the snippet, the essential components of the script are loaded inline synchronously, so it does not require a network roundtrip to begin tracking your page.
```
jpiyali: maybe say "The minimal components of the script are loaded syncrhonously. Full is loaded asynchronously and does not block to begin tracking your page"

Examples/react-application-insights-demo/.gitignore
```diff
@@ -1,23 +0,0 @@
-# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
-
```
jpiyali: You;'ll move eerything under Examples\ to the samples repo right?

markwolff: Yes. only 1 sample was in Examples/. now located [here](https://github.com/Azure-Samples/application-insights-react-demo)

PR #738, opened 2018-12-17
bump versions for 1.0.0-beta.9

markwolff:


AppInsightsCommon/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@microsoft/applicationinsights-common",
-    "version": "1.0.0-beta.12",
+    "version": "1.0.0-beta.13",
```
jpiyali: u possibly need to do this separately, otherwise, build will not pass.

PR #742, opened 2018-12-29
Use bullet points for alternatives

StefanSchoof:
Makes it clearer that not both things need to done.


StefanSchoof:
I was reading at first thought I need step one and two (since this was numbers) I think just change the number to bullets makes it easier to understand

StefanSchoof:
The rest of text had make it clear that only one option is needed.

PR #746, opened 2019-01-07
Release 1.0.0-beta.11

markwolff:
- rename `timestamp` to `time`
- rename `instrumentationKey` to `ikey`
- Add request-context field to autocollected fetch requests when `enableCorsCorrelation` is enabled
- In initialization, make `queue`, `version` optional
- rework ordering of `loadAppInsights` call to ensure preinstrumentation autocollection works
- re-enable disabled ajax plugin tests

### DIFF
--- a/AISKU/package.json
+++ b/AISKU/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@microsoft/applicationinsights-web",
-    "version": "1.0.0-beta.10",
+    "version": "1.0.0-beta.11",
     "description": "Microsoft Application Insights Javascript SDK API 1.0 beta",
     "main": "dist/applicationinsights-web.js",
     "module": "dist-esm/applicationinsights-web.js",
@@ -22,10 +22,10 @@
     },
     "dependencies": {
         "@microsoft/applicationinsights-analytics-js": "^1.0.0-beta.12",
-        "@microsoft/applicationinsights-channel-js": "^1.0.0-beta.11",
-        "@microsoft/applicationinsights-common": "^1.0.0-beta.13",
-        "@microsoft/applicationinsights-core-js": "^1.0.0-beta.3",
-        "@microsoft/applicationinsights-dependencies-js": "^1.0.0-beta.11",
-        "@microsoft/applicationinsights-properties-js": "^1.0.0-beta.8"
+        "@microsoft/applicationinsights-channel-js": "^1.0.0-beta.12",
+        "@microsoft/applicationinsights-common": "^1.0.0-beta.14",
+        "@microsoft/applicationinsights-core-js": "^1.0.0-beta.4",
+        "@microsoft/applicationinsights-dependencies-js": "^1.0.0-beta.12",
+        "@microsoft/applicationinsights-properties-js": "^1.0.0-beta.9"
     }
 }
--- a/AISKU/package.json
+++ b/AISKU/package.json
@@ -24,7 +24,7 @@
         "@microsoft/applicationinsights-analytics-js": "^1.0.0-beta.12",
         "@microsoft/applicationinsights-channel-js": "^1.0.0-beta.12",
         "@microsoft/applicationinsights-common": "^1.0.0-beta.14",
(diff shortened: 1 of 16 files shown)


AISKU/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@microsoft/applicationinsights-web",
-    "version": "1.0.0-beta.10",
+    "version": "1.0.0-beta.11",
```
jpiyali: should be 12 if these changes were already in the release that you did

AISKU/src/ApplicationInsightsContainer.ts
```diff
@@ -3,13 +3,14 @@ import { Initialization as ApplicationInsights, Snippet, IApplicationInsights }
 
 export class ApplicationInsightsContainer {
 
-    public static getAppInsights(snippet: Snippet, version: number) : IApplicationInsights | IAppInsightsDeprecated {
-        let initialization = new ApplicationInsights(snippet);
+    public static getAppInsights(snippet: Snippet, version: number = 2.0, applyAppInsights?: () => void) : IApplicationInsights | IAppInsightsDeprecated {
+        const initialization = new ApplicationInsights(snippet);
+        if (applyAppInsights) { applyAppInsights(); }
```
jpiyali: what is this for?

AppInsightsCommon/src/TelemetryItemCreator.ts
```diff
@@ -39,8 +39,8 @@ export class TelemetryItemCreator {
 
         let telemetryItem: ITelemetryItem = {
             name: envelopeName,
-            timestamp: new Date(),
-            instrumentationKey: "", // this will be set in TelemetryContext
+            time: new Date().toISOString(),
+            ikey: "", // this will be set in TelemetryContext
```
jpiyali: please update to iKey. This is possibly documented wrong in the 1ds page.

AISKU/package.json
```diff
@@ -22,10 +22,10 @@
     },
     "dependencies": {
         "@microsoft/applicationinsights-analytics-js": "^1.0.0-beta.12",
```
jpiyali: can u change this to 1.0.0-beta.11. this should be 11 and not 12.

markwolff: I bumped it to the version that will eventually be published. There was a single change to the analytics plugin since it uses the instrumentation key field

AISKU/src/ApplicationInsightsContainer.ts
jpiyali: wondering if this work can be done inside the constructor of Applicationinsights here (new ApplicationInsights(snippet);). Seems it has the object snippet available to setup the fields prior to doing the required initialization through loadAppInsights() which does the emptyQueue which is impacted.

AISKU/package.json
jpiyali: i see, if you updated this, not sure why I am not seeing a diff here.

jpiyali: i changed to 11 and then i realized u had published 11 already from a different branch but version was behind.

AISKU/src/ApplicationInsightsContainer.ts
markwolff: It could be done this way too. Is it more confusing if snippet is modified in some other constructor though? This is the only reason I just went with a callback

jpiyali: I see, perhaps expose a method called updateSnippetDefs or something like that in Initialization and just call that. If you want to keep as is, thats fine, but update name, I didnt really catch what was going on.

AppInsightsCommon/src/TelemetryItemCreator.ts
markwolff: updated

AISKU/src/ApplicationInsightsContainer.ts
markwolff: updated

PR #752, opened 2019-01-15
Version updates

jpiyali:
- Fix dependency telemetry fields
- update grunt file, build.cmd
- update SDK version number (2.0.2-beta)

### DIFF
--- a/AISKU/src/ApplicationInsightsDeprecated.ts
+++ b/AISKU/src/ApplicationInsightsDeprecated.ts
@@ -76,7 +76,7 @@ export class AppInsightsDeprecated implements IAppInsightsDeprecated {
         this.appInsightsNew.trackDependencyData(
             <IDependencyTelemetry>{
                 id: id,
-                absoluteUrl: absoluteUrl,
+                target: absoluteUrl,
                 type: pathName,
                 duration: totalTime,
                 properties: { HttpMethod: method },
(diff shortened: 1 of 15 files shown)


AISKU/src/ApplicationInsightsDeprecated.ts
```diff
@@ -76,7 +76,7 @@ export class AppInsightsDeprecated implements IAppInsightsDeprecated {
         this.appInsightsNew.trackDependencyData(
             <IDependencyTelemetry>{
                 id: id,
-                absoluteUrl: absoluteUrl,
+                target: absoluteUrl,
```
hectorhdzg: target: target ?

gruntfile.js
```diff
@@ -175,7 +175,7 @@ module.exports = function (grunt) {
             },
             snippetvNext: {
                 files: {
-                    'AISKU/dist-history/snippet.min.js': ['AISKU/snippet/snippet.js']
+                    'AISKU/snippet/snippet.min.js': ['AISKU/snippet/snippet.js']
```
markwolff: Should delete the dist-history snippet minified file. Also consider deleting the rest of the old sdk version in there as well

jpiyali: We want to keep history of shipped versions checked in as well in case someone accidentally deletes from storage account. I'll do another PR to update dist-hist with the bits that we release on Monday.

PR #753, opened 2019-01-17
Feature/supporting react native

shortstevenn:
These are the remaining changes that I needed which started with pull req:
https://github.com/Microsoft/applicationinsights-channel-js/pull/43

I want to use AI with react-native in a create-react-native app that hasn't been ejected.

To make that happen I had to do some odd DI to make all the constructors happy - and deal with failures in friendlier ways.

Ended up getting it all constructed in a function like this:
```
 initialize(config) {
        if(!this.initialized) {
            if(!config) {
                throw "Cannot use an uninitilized object";
            }

            var aiCore = new AppInsightsCore();
            config.logger = config.logger || new DiagnosticLogger();
            config.enableSessionStorageBuffer = false;

            var sender = new Sender(config, aiCore);
            sender.initialize(config, aiCore);
            aiCore.initialize(config, [sender]);
            this.insights = new ApplicationInsightsInternal(config)
            this.insights.appInsights.initialize(config, aiCore, [sender]);
        }
        this.initialized = true;
    }
```

@markwolff - thanks for all your help with this.


AISKU/src/Initialization.ts
```diff
@@ -241,23 +241,24 @@ export class Initialization implements IApplicationInsights {
...
+        let appInsightsChannel: Sender = sender || new Sender();
 
         extensions.push(appInsightsChannel);
         extensions.push(this.properties);
         extensions.push(this.dependencies);
         extensions.push(this.appInsights);
 
         // initialize core
-        this.core.initialize(this.config, extensions);
+        core || this.core.initialize(this.config, extensions);
```
markwolff: `core.initialize` should be idempotent, so I don't think this is needed. Also, is the core you are passing in required to be already initialized?

shortstevenn: Hmm, not sure if it's idempotent or not. Since I was initializing the core already, I skipped it the second time.
But yes, core has to be initialized with the earlier channel sender

AISKU/src/Initialization.ts
```diff
@@ -241,23 +241,24 @@ export class Initialization implements IApplicationInsights {
...
      */
-    public loadAppInsights(legacyMode: boolean = false): IApplicationInsights {
+    public loadAppInsights(legacyMode: boolean = false, core?: AppInsightsCore, sender?:Sender): IApplicationInsights {
 
         // dont allow additional channels/other extensions for legacy mode; legacy mode is only to allow users to switch with no code changes!
         if (legacyMode && this.config.extensions && this.config.extensions.length > 0) {
             throw new Error("Extensions not allowed in legacy mode");
         }
 
+        this.core = core || new AppInsightsCore();
```
jpiyali: what is special about this input core and sender object here?

AISKU/src/Initialization.ts
shortstevenn: Apparently you were right and calling it again didn't matter - updated

AISKU/src/Initialization.ts
jpiyali: There are other changes in this sku that are dom dependent i think. Wouldn't a better fix be to have conditional checks for env and appropriately take care of the dependencies?

shortstevenn: Probably, I just got it working for my purposes - but yes, that would be a much better generic approach.

jpiyali: Are you building core locally with some changes and passing that in to the SKU entry point here? Would appreciate if you share your core changes. just with this change alone in the SKU, are you able to get the E2E working?

shortstevenn: The only changes required were in channel, and @markwolff  merged that pull req.

I'm constructing it locally like this and importing this class as ApplicationInsights - a little convoluted I know...
```
import { Sender }  from "@microsoft/applicationinsights-channel-js";
import { AppInsightsCore, DiagnosticLogger } from '@microsoft/applicationinsights-core-js'
import { ApplicationInsights as ApplicationInsightsInternal } from '@microsoft/applicationinsights-web';

global.window.location = global.window.location || {
    host: ""
}

class ApplicationInsightsWrapper {
    initialize(config) {
        if(!this.initialized) {
            if(!config) {
                throw "Cannot use an uninitialized object";
            }

            var aiCore = new AppInsightsCore();
            config.logger = config.logger || new DiagnosticLogger();
            config.enableSessionStorageBuffer = false;

            var sender = new Sender(config, aiCore);
            sender.initialize(config, aiCore);
            aiCore.initialize(config, [sender]);
            this.insights = new ApplicationInsightsInternal(config)
            this.insights.appInsights.initialize(config, aiCore, [sender]);
        }
        this.initialized = true;
    }

    trackEvent(event, properties) { this.insights.trackEvent(event, properties); }
    trackPageView(event, properties) { this.insights.trackPageView(event, properties); }
    trackException(event, properties) {this.insights.trackException(event, properties); }
    trackTrace(event, properties) { this.insights.trackTrace(event, properties); }
    trackMetric(event, properties) { this.insights.trackMetric(event, properties); }
    startTrackPage(name) { this.insights.startTrackPage(name); }
    stopTrackPage(name, url, properties) { this.insights.stopTrackPage(name, url, properties); }
    startTrackEvent(name) { this.insights.startTrackEVent(name); }
    stopTrackEvent(name, properties, measurements) { this.insights.trackException(name, properties, measurements); }
    trackPageViewPerformance(performance, properties) { this.insights.trackPageViewPerformance(performance, properties); }
    addTelemetryInitializer(telemetry) { this.insights.addTelemetryInitializer(telemetry); }
}

const ApplicationInsights = new ApplicationInsightsWrapper();
export {
    ApplicationInsights
}
```

I'm sure there is some clever javascript trick I can use for the composition rather than passing those functions through like that - but I'm really new to JS and haven't taken time to figure it out yet.

jpiyali: If you add this code 
global.window.location = global.window.location || {
    host: ""
}

prior to initializing using the application insights npm package, it doesnt work? The code does not have anything special in the types (sender, core) for which this PR was started.

shortstevenn: Sorry for the delay. I'll try this out. I don't think it worked when I first did it - thus the pull req. I'll give it another go though so I can be thorough.

shortstevenn:
Looks like this is no longer necessary - ty for the help though.

PR #755, opened 2019-01-17
add requestType to correctly label fetch requests in telemetry

markwolff:
All auto tracked requests are listed as `Ajax` in portal. Now that we support collection of fetch requests, we should label it properly.

- rename `commandName` legacy code
- Add requestType to correctly label fetch requests

TODO:
- [ ] Add tests


extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -541,6 +542,7 @@ export class AjaxMonitor implements ITelemetryPlugin, IDependenciesPlugin, IInst
                     id: ajaxData.id,
                     target: ajaxData.getAbsoluteUrl(),
                     type: ajaxData.getPathName(),
+                    requestType: "Fetch",
```
hectorhdzg: So Type and requestType are different things?, this doesn't look good in Ajax you sue type and in Fetch calls you use both requestType and type, is this new field documented anywhere?

markwolff: type is existing field. Would prefer to name this type. The default value of `requestType` is `"Ajax"`, but it should probably be explicitly set here.

markwolff: Also I think `type` is old `commandName`

markwolff:
Closing for now. Will redo in another PR

PR #762, opened 2019-01-23
Export types for part B in light sku

jpiyali:
When creating part B for track api, it is useful for user to view types that are supported.


AISKULight/index.ts
```diff
@@ -107,13 +107,12 @@ export {
     ITelemetryItem
 } from "@microsoft/applicationinsights-core-js";
 export {
-    SeverityLevel,
```
markwolff: Will still need SeverityLevel exported

PR #764, opened 2019-01-28
Update event telemetry to include properties and measurements

jpiyali:


AppInsightsCommon/src/Interfaces/IEventTelemetry.ts
```diff
@@ -8,4 +8,14 @@ export interface IEventTelemetry {
      * @memberof IEventTelemetry
      */
     name: string;
+
+    /**
+     * Property bag to contain additional custom properties (Part C)
+     */
+    properties?: { [key: string]: string };
```
markwolff: can type be switched to object?

jpiyali: not in this PR. I need to follow up on two things before I can make the switch. Channel will still need to downcast till there is better support for object.

PR #769, opened 2019-01-30
add publish to 'next' cdn script

markwolff:
Refactors the old powershell script for current CDN container


AISKU/scripts/uploadFilesToAzureStorage.ps1
```diff
@@ -0,0 +1,62 @@
+$storageAccountName = "daprodcdn";
+$containerName = "next";
+
+# get js sdk directory
```
jpiyali: Please test with test subscriptions first.

AISKU/snippet/snippet.js
```diff
@@ -11,7 +11,7 @@ var aisdk = window[aiName] || (function (aiConfig) {
     var localDocument = document, localWindow = window, scriptText = "script";
     setTimeout(function () {
         var scriptElement = localDocument.createElement(scriptText);
-        scriptElement.src = aiConfig.url || "https://az416426.vo.msecnd.net/beta/ai.1.min.js";
+        scriptElement.src = aiConfig.url || "https://az416426.vo.msecnd.net/next/ai.1.min.js";
```
hectorhdzg: What is this purpose of next?

markwolff: Using a different container name going forward. Old container was called `beta`, using `next` now

AISKU/scripts/uploadFilesToAzureStorage.ps1
```diff
@@ -0,0 +1,62 @@
...
+$jsSdkDir = Split-Path (Split-Path $MyInvocation.MyCommand.Path) -Parent;
```
jpiyali: also, write wrappers for next and beta. Eventually, you want to be able to do three deployments one by one for flighting. beta, next, scripts/xxx (maybe scripts/b)

AISKU/snippet/snippet.js
hectorhdzg: So next will always have latest, is that correct?

markwolff: I think we will be doing it like this
`scripts/b`: latest stable version
`next`: next stable version (for flighting)
`beta`: additional experimental flighting endpoint

PR #770, opened 2019-01-30
update fields with properies & measurements

### LINKED ISSUE #768: [Beta] Initialization class does not implements IApplicationInsights correctly
In many places in the Initialization class for the new application insights beta SDK, some optional properties from the IApplicationInsights interface are not implemented correctly. I would expect this class to passthrough all track calls down to the `this.appInsights` property but it appears to be skipping some `customProperties`.

```ts
    /**
     * Logs that a page, or similar container was displayed to the user.
     * @param {IPageViewTelemetry} pageView
     * @param {{ [key: string]: any; }} [customProperties]
     * @memberof Initialization
     */
    public trackPageView(pageView: IPageViewTelemetry) {
        this.appInsights.trackPageView(pageView);
    }
```
[Source](https://github.com/Microsoft/ApplicationInsights-JS/blob/0e6d1929dddf0392d30cce4a4fef30d39d0b3084/AISKU/src/Initialization.ts#L87)

As you can see, there are two properties in the doc (an in the interface) but only one in the implementation.

In this case, it prevents us from sending custom properties on page views, even though the underlining class expects them.

markwolff:
Fixes #768 
Removes `customProperties` to API fields where `properties`/`measurements` named argument is present.

### DIFF
--- a/AISKU/src/Initialization.ts
+++ b/AISKU/src/Initialization.ts
@@ -71,17 +71,15 @@ export class Initialization implements IApplicationInsights {
     /**
      * Log a user action or other occurrence.
      * @param {IEventTelemetry} event
-     * @param {{ [key:string]: any }} [customProperties]
      * @memberof Initialization
      */
-    public trackEvent(event: IEventTelemetry, customProperties?: { [key:string]: any }) {
-        this.appInsights.trackEvent(event, customProperties);
+    public trackEvent(event: IEventTelemetry) {
+        this.appInsights.trackEvent(event);
     }
 
     /**
      * Logs that a page, or similar container was displayed to the user.
      * @param {IPageViewTelemetry} pageView
-     * @param {{ [key: string]: any; }} [customProperties]
      * @memberof Initialization
      */
     public trackPageView(pageView: IPageViewTelemetry) {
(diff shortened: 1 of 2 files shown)


AISKU/src/Initialization.ts
```diff
@@ -71,17 +71,15 @@ export class Initialization implements IApplicationInsights {
     /**
      * Log a user action or other occurrence.
      * @param {IEventTelemetry} event
-     * @param {{ [key:string]: any }} [customProperties]
      * @memberof Initialization
      */
-    public trackEvent(event: IEventTelemetry, customProperties?: { [key:string]: any }) {
-        this.appInsights.trackEvent(event, customProperties);
```
jpiyali: customProperties wills stay Mark. It reflects part C.

markwolff: Added this part back in

AISKU/src/Initialization.ts
```diff
@@ -213,12 +210,10 @@ export class Initialization implements IApplicationInsights {
     /**
      * Log a dependency call (e.g. ajax)
      * @param {IDependencyTelemetry} dependency
-     * @param {{[key: string]: any}} [customProperties]
-     * @param {{[key: string]: any}} [systemProperties]
      * @memberof Initialization
      */
-    public trackDependencyData(dependency: IDependencyTelemetry, customProperties?: {[key: string]: any}): void {
-        this.dependencies.trackDependencyData(dependency, customProperties);
+    public trackDependencyData(dependency: IDependencyTelemetry): void {
```
hectorhdzg: So how custom properties will be send now?, dependency.properties?

markwolff: Yes, but this is only the external facing API. No change to any internal dependencies plugin APIs.

PR #792, opened 2019-02-27
bump react-native plugin to 1.0.0-beta

markwolff:
- Bumps version to `1.0.0-beta.0`

### DIFF
--- a/vNext/extensions/applicationinsights-react-native/package.json
+++ b/vNext/extensions/applicationinsights-react-native/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@microsoft/applicationinsights-react-native",
-    "version": "1.0.0-beta.0",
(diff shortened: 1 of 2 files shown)


vNext/extensions/applicationinsights-react-native/package.json
```diff
@@ -35,8 +35,7 @@
         "tslib": "^1.9.3"
     },
     "peerDependencies": {
-        "react-native": "*",
-        "react-native-device-info": "^0.26.4"
```
jpiyali: you dont want this anymore?

markwolff: I think I only want react-native as a peer. For device-info, users will need to link it anyways, so this is just so it shows up on the npmjs dependency tree. So I moved from peer to pure dependency

PR #798, opened 2019-03-06
Bug fix for duplicate telemetry issue in vNext ( 4088796)

jpiyali:
Allow channel configuration to be passed in only through config.channels parameter. Disallow through config.extensions parameter, add unit tests. Fixes bug with multiple items being sent in telemetry in fabrikam.

### DIFF
--- a/vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
+++ b/vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
@@ -46,10 +46,6 @@ export class AppInsightsCore implements IAppInsightsCore {
         this._notificationManager = new NotificationManager();        
         this.config.extensions = CoreUtils.isNullOrUndefined(this.config.extensions) ? [] : this.config.extensions;
 
-        // if (this.config.channels && this.config.channels.length > 0) {
-        //     for (let i = 0; i this.config.extensions
-        // }
-
         // add notification to the extensions in the config so other plugins can access it
         this.config.extensionConfig = CoreUtils.isNullOrUndefined(this.config.extensionConfig) ? {} : this.config.extensionConfig;
         this.config.extensionConfig.NotificationManager = this._notificationManager;
--- a/vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
+++ b/vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
@@ -43,8 +43,12 @@ export class AppInsightsCore implements IAppInsightsCore {
 
         this.config = config;
 
-        this._notificationManager = new NotificationManager();
+        this._notificationManager = new NotificationManager();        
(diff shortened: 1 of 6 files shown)


vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -43,9 +43,13 @@ export class AppInsightsCore implements IAppInsightsCore {
 
         this.config = config;
 
-        this._notificationManager = new NotificationManager();
+        this._notificationManager = new NotificationManager();        
         this.config.extensions = CoreUtils.isNullOrUndefined(this.config.extensions) ? [] : this.config.extensions;
 
+        // if (this.config.channels && this.config.channels.length > 0) {
```
hectorhdzg: Remove commented code

vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -317,14 +330,25 @@ class ChannelController implements ITelemetryPlugin {
...
                     // Initialize each plugin
-                    queue.forEach(queueItem => queueItem.initialize(config, core, extensions));
+                    queue.forEach(queueItem => {
+                        if (queueItem.priority < ChannelControllerPriority) {
+                            invalidChannelIdentifier = queueItem.identifier;
+                        }
+                        queueItem.initialize(config, core, extensions)
+                    });
+
+                    if (invalidChannelIdentifier) {
```
hectorhdzg: How this is related to having duplicated events?

vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -364,5 +388,6 @@ class ChannelController implements ITelemetryPlugin {
 }
 
 const validationError = "Extensions must provide callback to initialize";
+const validationErrorInExt = "Channels must be provided through config.channels only";
 const ChannelControllerPriority = 500;
-const duplicatePriority = "One or more extensions are set at same priority";
```
hectorhdzg: Are you not checking for duplicated priorities anymore?

markwolff: I think the current behavior is for them to execute in order they are added if they have identical priorities. @jpiyali can correct if wrong

jpiyali: yes, we dont check for duplicated priorities anymore. Abhi or you had raised that this was causing issues when I added the check. The string is unused.

vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
jpiyali: The duplicated events was coming from not adding the channel side loaded correctly. We've seen this in the past, when channel is not added correctly in core initialization, we saw duplicate events. I updated fabrikam code to reflect usage there as well.

PR #804, opened 2019-03-12
add time in IDependency to override generated time

### LINKED ISSUE #628: AJAX telemetry has timestamp at end of request rather than start
AJAX telemetry appears to record the event timestamp at the time of collection rather than at the time of the call being made. This can be reproduced by setting up a simple server that responds slowly to a request. The implication of this is that the Portal UI can present misleading timelines of what actually happened during the pageview.

For example, I have an AJAX that takes 5 seconds to complete (as shown by Fiddler):
image

When inspecting the generated telemetry envelope, we see the recorded time as after ClientDoneResponse:
image

From the [schema description of the time field][0], I would expect the time of my AJAX starting to be the time reported in the envelope, though admittedly it's not clear what exactly "event" refers to:

> Event date time when telemetry item was created. This is the wall clock time on the client when the event was generated. 

It is clearer in the context of these two issues however that it refers to the event the telemetry item is created for, not the instantiation time of the telemetry item itself:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/658
https://github.com/Microsoft/ApplicationInsights-Home/issues/133

Interestingly, this problem doesn't seem to be limited to autocollection. From what I can tell by a quick look in the code, there doesn't appear to be any way currently to adjust the timestamp of these events, even if manually collecting a dependency via `RemoteDependencyData`.

[0]: https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Docs/Envelope.md

markwolff:
Add `startTime` number (format = ms since 1/1/1970) to `IDependencyTelemetry`. `TelemetryCreator` does not generate a fresh timestamp if this field is present.

- Fixes #628


vNext/extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -544,14 +546,15 @@ export class AjaxMonitor implements ITelemetryPlugin, IDependenciesPlugin, IInst
...
-                let dependency: IDependencyTelemetry = {
+                let dependency= <IDependencyTelemetry>{
                     id: ajaxData.id,
                     target: ajaxData.getAbsoluteUrl(),
                     name: ajaxData.getPathName(),
                     type: "Fetch",
                     duration: ajaxData.ajaxTotalDuration,
                     success: false,
                     responseCode: 0,
+                    time: ajaxData.requestSentTime,
```
hectorhdzg: Why do you need to override time of the event?

markwolff: We are currently generating the timestamp when the telemetry item is generated, instead of when the request starts. This is inconsistent with how other SDKs handle this. #628 has further discussion on this

hectorhdzg: Ok, so it will be better to actually call this property sentTime or somethign like that so people calling the APIs directly don't get confused of what time means here

vNext/channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
```diff
@@ -57,6 +57,9 @@ export abstract class EnvelopeCreator {
     protected static createEnvelope<T>(logger: IDiagnosticLogger, envelopeType: string, telemetryItem: ITelemetryItem, data: Data<T>): IEnvelope {
         let envelope = new Envelope(logger, data, envelopeType);
         envelope.iKey = telemetryItem.iKey;
+        if (data.baseData && (data.baseData as any).time) {
```
hectorhdzg: This time property only applies to dependency events but you will check for it for every single other event, is there a better place to add this code?

jpiyali: Move this to after line 239. Once u get the envelope back for dependency, update timestamp with baseData.time (or baseData.requestSentTime as you name it)

EnvelopeCreator.createEnvelope<RemoteDependencyData>(logger, RemoteDependencyData.envelopeType, telemetryItem, data);

vNext/extensions/applicationinsights-dependencies-js/src/ajax.ts
markwolff: renamed to `sentTime`

vNext/channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
markwolff: moved to 239

ketaanhshah:
Hey folks, Any update on this code fix being pushed to the latest version and if the issues is removed?

MSNev:
Reopening for tracking the changes for the release notes -- will reclose after release is fully deployed

MSNev:
Duh - this is a pull request not an issue -- reclosing

PR #813, opened 2019-03-19
PageViewPerformance property names updated

hectorhdzg:
PageViewPerformance property names updated
Enable sending system properties for PageViewPerformance
and Exceptions events.
Analytics track event overriding event name fix

### DIFF
--- a/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
+++ b/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
@@ -5,9 +5,10 @@
 
 import {
     IConfig, Util, PageViewPerformance, IAppInsights, PageView, RemoteDependencyData, Event, IEventTelemetry,
-    TelemetryItemCreator, Data, Metric, Exception, SeverityLevel, Trace, IDependencyTelemetry,
+    TelemetryItemCreator, Metric, Exception, SeverityLevel, Trace, IDependencyTelemetry,
     IExceptionTelemetry, ITraceTelemetry, IMetricTelemetry, IAutoExceptionTelemetry,
-    IPageViewTelemetryInternal, IPageViewTelemetry, IPageViewPerformanceTelemetry, ConfigurationManager
+    IPageViewTelemetryInternal, IPageViewTelemetry, IPageViewPerformanceTelemetry, IPageViewPerformanceTelemetryInternal,
+    ConfigurationManager, DateTimeUtils
 } from "@microsoft/applicationinsights-common";
 
 import {
@@ -17,6 +18,7 @@ import {
 } from "@microsoft/applicationinsights-core-js";
 import { PageViewManager, IAppInsightsInternal } from "./Telemetry/PageViewManager";
 import { PageVisitTimeManager } from "./Telemetry/PageVisitTimeManager";
+import { PageViewPerformanceManager } from './Telemetry/PageViewPerformanceManager';
 import { ITelemetryConfig } from "../JavaScriptSDK.Interfaces/ITelemetryConfig";
 
 "use strict";
@@ -39,8 +41,9 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
     private _eventTracking: Timing;
     private _pageTracking: Timing;
     private _telemetryInitializers: { (envelope: ITelemetryItem): boolean | void; }[]; // Internal telemetry initializers.
-    private _pageViewManager: PageViewManager;
-    private _pageVisitTimeManager: PageVisitTimeManager;
+    protected _pageViewManager: PageViewManager;
+    protected _pageViewPerformanceManager: PageViewPerformanceManager;
+    protected _pageVisitTimeManager: PageVisitTimeManager;
 
     // Counts number of trackAjax invokations.
     // By default we only monitor X ajax call per view to avoid too much load.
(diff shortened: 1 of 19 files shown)


vNext/shared/AppInsightsCommon/src/Telemetry/PageViewPerformance.ts
```diff
@@ -30,167 +28,15 @@ export class PageViewPerformance extends PageViewPerfData implements ISerializab
...
-    */
-    public getDurationMs() {
-        return this.durationMs;
-    }
-
     /**
      * Constructs a new instance of the PageEventTelemetry object
      */
-    constructor(logger: IDiagnosticLogger, name: string, url: string, unused: number, properties?: {[key: string]: string}, measurements?: {[key: string]: number}) {
+    constructor(logger: IDiagnosticLogger, name: string, url: string, unused: number, properties?: { [key: string]: string }, measurements?: { [key: string]: number }, id?: string) {
```
markwolff: is `id` being used anywhere?

vNext/shared/AppInsightsCommon/src/Util.ts
```diff
@@ -709,11 +709,11 @@ export class DateTimeUtils {
      * Gets duration between two timestamps
      */
     public static GetDuration = function (start: number, end: number): number {
-        var result = null;
-        if (start !== 0 && end !== 0 && !CoreUtils.isNullOrUndefined(start) && !CoreUtils.isNullOrUndefined(end)) {
-            result = end - start;
+        var duration = undefined;
+        if (!(isNaN(start) || isNaN(end))) {
```
markwolff: no 0 checking is being done anymore

hectorhdzg: 0 is a valid start number right?, this was the reason I was not able to use this Util in pvp scenarios, Math.max will take care of negative numbers if end is 0 or less

markwolff: If it was blocking you, it's probably fine to remove. I think for our case 0 was never a valid start time

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -17,14 +18,15 @@ import {
...
+import { PageViewPerformanceManager } from './Telemetry/PageViewPerformanceManager';
 import { ITelemetryConfig } from "../JavaScriptSDK.Interfaces/ITelemetryConfig";
 
 "use strict";
 
 const durationProperty: string = "duration";
 
 export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IAppInsightsInternal {
-    public static Version = "2.0.1-beta";
+    public static Version = "2.0.0-beta.3";
```
jpiyali: this is unused, can be removed. As per discussion with @markwolff , we should use this to stamp the sdk.libVer in the overall sku code (currently properties plugin is doing the work, but that may not be avl. in all skus).

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -262,12 +265,14 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
...
-        let telemetryItem = TelemetryItemCreator.create<PageViewPerformance>(pageViewPerformance,
+    public sendPageViewPerformanceInternal(pageViewPerformance: IPageViewPerformanceTelemetryInternal, properties?: { [key: string]: any }, systemProperties?: { [key: string]: any }) {
+        let telemetryItem = TelemetryItemCreator.create<IPageViewPerformanceTelemetryInternal>(
+            pageViewPerformance,
             PageViewPerformance.dataType,
             PageViewPerformance.envelopeType,
             this._logger,
-            properties);
+            properties,
+            systemProperties);
```
jpiyali: @hectorhdzg  - do you need part A to be set for this api as well? If there is no requirement, lets keep the extra parameter out.

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -345,6 +353,28 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
         }
     }
 
+    /**
+    * @ignore INTERNAL ONLY
+    * @param exception
+    * @param properties
+    * @param systemProperties
+    */
+    public sendExceptionInternal(exception: IExceptionTelemetry, customProperties?: { [key: string]: any }, systemProperties?: { [key: string]: any }) {
```
jpiyali: same here, does 1P require systemproperties to be passed in for exception telemetry?

vNext/shared/AppInsightsCommon/src/Util.ts
```diff
@@ -709,11 +709,11 @@ export class DateTimeUtils {
...
+            duration = Math.max(end - start, 0);
```
jpiyali: did you get a bug? I get it its safeguard code, but if end does turn out to be less than start, we should get that fbk about the capture, rather than reporting 0.

vNext/shared/AppInsightsCommon/src/Telemetry/PageViewPerformance.ts
hectorhdzg: Not used in Breeze, removing from here

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
hectorhdzg: Yes, is one of the reasons of this PR

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
hectorhdzg: Yes we need to send part A to shared events between 1st and 3rd party analytics plugins, pageView, pageViewPerformance and Exception

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
hectorhdzg: This is being used in core, is being added to the event ext.sdk.libVer property, AI channel doesn't use this and first party properties override it but is actually being used

PR #827, opened 2019-04-08
analytics: initialize autoTrackPageVisitTime deps

### LINKED ISSUE #826: autoTrackPageVisitTime = true causes error
Application Insights JS SDK Version: 2.0.0-rc.1
Module format: UMD/Bundled
Problem files: 
    [node_modules]/@microsoft/applicationinsights-web/browser/ai.2.js
    [node_modules]/@microsoft/applicationinsights-web/dist/applicationinsights-web.js

**Description:** 
 
_We get an error when calling **appInsights.trackPageView()** after setting configuration **autoTrackPageVisitTime** to **true**_ 

**Source line:**
```javascript
ApplicationInsights.prototype.trackPageView = function (pageView, customProperties) {
            try {
                this._pageViewManager.trackPageView(pageView, customProperties);
                if (this.config.autoTrackPageVisitTime) {
                    this._pageVisitTimeManager.trackPreviousPageVisit(pageView.name, pageView.uri);
                    // PS!!! The error occurs on the line above, saying _pageVisitTimeManager is undefined
               }
            […]
```

**Stack trace:**
```text
TypeError: Cannot read property 'trackPreviousPageVisit' of undefined
    at ApplicationInsights.trackPageView (node_modules\@microsoft\applicationinsights-web\dist\applicationinsights-web.js:2697:49)
    at Initialization.trackPageView (node_modules\@microsoft\applicationinsights-web\dist\applicationinsights-web.js:5771:31)
```

**Suspected cause:** 
 
PageVisitTimeManager is missing in the bundling for the files in the folders "dist" and "browser" under "node_modules/@microsoft/applicationinsights-web".

**Details:** 
 
The @microsoft/applicationinsigts-xxx packages are compiled/bundled into three formats in the folders under [node_modules/[packagename]: **"browser"**, **"dist"** and **"dist-esm"**.

The **"dist-esm"** folder contains ESM module format import statements and is referenced by the "module" property in package.json. For applications using WebPack this will become the perferred source files since WebPack prefers this when inspecting package.json.

The files in the **"dist"** and **"browser"** folders seem to be in a UMD module format, but they also seem to be **bundled** to include all the source code from the other depended-upon packages.

I expected the files under **"browser"** to be a no-modules alternative that allows you to simply add the script through a standard script element on an HTML page. (TIP: If this is the intent, this file may be referenced by a "browser" property in package.json, which it is currently not).

I then expected the files under **"dist"** to be in a UMD format - which it is - but I did not expect it to be bundled. In the application where we experience the problem we are still using AMD/RequireJS, and it would work fine to have UMD module format without bundling (it would require some more path configuration for RequireJS though).

Anyway: It seems like the PageVisitTimeManager class/module was left out in this bundling process, so whenever you set autoTrackPageVisitTime to true, an error is thrown when calling trackPageView().

PS! I have tested scripts under both "browser" and "dist", but I experience the same problem in both.

markwolff:
Add missing initialization for `autoTrackPageVisitTime`
- Fixes #826

### DIFF
--- a/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
+++ b/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
@@ -221,12 +221,13 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
      * @param customProperties Additional data used to filter events and metrics. Defaults to empty.
      * If a user wants to provide duration for pageLoad, it'll have to be in pageView.properties.duration
      */
-    public trackPageView(pageView: IPageViewTelemetry, customProperties?: ICustomProperties) {
+    public trackPageView(pageView?: IPageViewTelemetry, customProperties?: ICustomProperties) {
         try {
-            this._pageViewManager.trackPageView(pageView, customProperties);
+            const inPv = pageView || {};
+            this._pageViewManager.trackPageView(inPv, customProperties);
 
             if (this.config.autoTrackPageVisitTime) {
-                this._pageVisitTimeManager.trackPreviousPageVisit(pageView.name, pageView.uri);
+                this._pageVisitTimeManager.trackPreviousPageVisit(inPv.name, inPv.uri);
             }
         } catch (e) {
             this._logger.throwInternal(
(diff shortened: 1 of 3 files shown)


vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -565,6 +567,22 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
...
     }
 
+    /**
+     * Log a page visit time
+     * @param    pageName    Name of page
+     * @param    pageVisitDuration Duration of visit to the page in milleseconds
+     */
+    private trackPageVisitTime(pageName: string, pageUrl: string, pageVisitTime: number) {
+        var properties = { PageName: pageName, PageUrl: pageUrl };
+        this.trackMetric({
```
hectorhdzg: What is this metric?, is this something available before?

markwolff: Yes, it was in previous SDK version
https://github.com/Microsoft/ApplicationInsights-JS/blob/3ea7cd9839277802de0b99c8250062f3b752c1c7/JavaScript/JavaScriptSDK/AppInsights.ts#L425-L434

PR #840, opened 2019-04-17
common: add exception convert to/from interface

markwolff:
Adds `CreateFromInterface` static method and `toInterface` instance method. When converting back and fourth with these, result is equivalent. Interface mode only contains valid Part B.

### DIFF
--- a/vNext/shared/AppInsightsCommon/src/Telemetry/Exception.ts
+++ b/vNext/shared/AppInsightsCommon/src/Telemetry/Exception.ts
@@ -10,6 +10,7 @@ import { FieldType } from '../Enums';
 import { SeverityLevel } from '../Interfaces/Contracts/Generated/SeverityLevel';
 import { Util } from '../Util';
 import { IDiagnosticLogger } from '@microsoft/applicationinsights-core-js';
+import { IExceptionInternal, IExceptionTelemetry, IExceptionDetailsInternal, IExceptionStackFrameInternal } from '../Interfaces/IExceptionTelemetry';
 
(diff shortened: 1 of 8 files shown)


vNext/shared/AppInsightsCommon/src/Telemetry/Exception.ts
```diff
@@ -26,21 +27,47 @@ export class Exception extends ExceptionData implements ISerializable {
...
-        if (severityLevel) {
-            this.severityLevel = severityLevel;
-        }
+    public static CreateFromInterface(logger: IDiagnosticLogger, exception: IExceptionInternal): Exception {
+        const exceptions = exception.exceptions.map((ex: IExceptionDetailsInternal) => _ExceptionDetails.CreateFromInterface(logger, ex));
+        return new Exception(logger, {...exception, exceptions: exceptions});
     }
 
+    public toInterface(): IExceptionInternal {
+        const { exceptions, properties, measurements, severityLevel, ver } = this;
```
jpiyali: anonymous instance!

vNext/shared/AppInsightsCommon/src/Telemetry/Exception.ts
```diff
@@ -26,21 +27,47 @@ export class Exception extends ExceptionData implements ISerializable {
...
+        return {
+            ver: ver,
+            exceptions: exceptions.map((exception: _ExceptionDetails) => exception.toInterface()),
+            severityLevel,
+            properties,
```
jpiyali: why no key assignment needed here?

PR #842, opened 2019-04-17
add isCookieUseDisabled to CoreUtils

markwolff:
- Add `isCookieUseDisabled` to properties plugin for cases where analytics plugin is not available.

### DIFF
--- a/vNext/extensions/applicationinsights-properties-js/src/PropertiesPlugin.ts
+++ b/vNext/extensions/applicationinsights-properties-js/src/PropertiesPlugin.ts
@@ -9,7 +9,7 @@ import {
 } from '@microsoft/applicationinsights-core-js';
 import { TelemetryContext } from './TelemetryContext';
 import { PageView, ConfigurationManager,
-    IConfig, PropertiesPluginIdentifier, IPropertiesPlugin, Extensions, IDevice, Util } from '@microsoft/applicationinsights-common';
+    IConfig, PropertiesPluginIdentifier, IPropertiesPlugin, Extensions } from '@microsoft/applicationinsights-common';
 import { ITelemetryConfig } from './Interfaces/ITelemetryConfig';
 
 export default class PropertiesPlugin implements ITelemetryPlugin, IPropertiesPlugin {
@@ -29,7 +29,6 @@ export default class PropertiesPlugin implements ITelemetryPlugin, IPropertiesPl
             sessionRenewalMs: () => 30 * 60 * 1000,
             samplingPercentage: () => 100,
             sessionExpirationMs: () => 24 * 60 * 60 * 1000,
-            isCookieUseDisabled: () => false,
             cookieDomain: () => null,
             sdkExtension: () => null,
             isBrowserLinkTrackingEnabled: () => false,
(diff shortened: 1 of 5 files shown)


vNext/extensions/applicationinsights-properties-js/src/PropertiesPlugin.ts
```diff
@@ -46,6 +47,11 @@ export default class PropertiesPlugin implements ITelemetryPlugin, IPropertiesPl
         }
 
         this._logger = core.logger;
+
+        if (this._extensionConfig.isCookieUseDisabled()) {
+            Util.disableCookies();
```
hectorhdzg: You are now disabling cookies here and in Analytics extension, if some other extension use cookies in the future you will add the same code over there, it would be better to have this code in Core and disable cookies during Core initialization, in that case I can start using Cookie methods in Common library in 1p code and I don't need to worry about this config ever

markwolff: This is true. I would like to put this stuff in core, but it cannot take a dependency on common.

jpiyali: there is a coreUtil where you could move the isCookieUseDisabled() functionality to. Core has access to config at init, it can set it early on.

markwolff: good call! I will move them to CoreUtils

vNext/shared/AppInsightsCommon/src/Util.ts
```diff
@@ -288,28 +287,24 @@ export class Util {
...
 
     /*
      * helper method to tell if document.cookie object is available
      */
     public static canUseCookies(logger: IDiagnosticLogger): any {
-        if (Util._canUseCookies === undefined) {
-            Util._canUseCookies = false;
-
-            try {
-                Util._canUseCookies = Util.document.cookie !== undefined;
```
jpiyali: this still needs to remain here. Core should be agnostic of concept of document object.

vNext/shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
```diff
@@ -3,6 +3,9 @@
 "use strict";
 
 export class CoreUtils {
+    private static _canUseCookies: boolean;
+    private static document: any = typeof document !== "undefined" ? document : {};
```
jpiyali: move back to common. Core should not take dependency on document object. Keep CoreUtils solely for what is user specified for cookie usage (isCookieUsageDisabledByUser, something like that).

markwolff: moved stuff that uses document back to common. Only `disableCookies` is in CoreUtils now.

vNext/shared/AppInsightsCommon/src/Util.ts
markwolff: Added back

Issue #848, opened 2019-04-25
TypeError: undefined is not an object (evaluating 'this._core.logger')


slolam:
Hi,

I am trying to develop extension so that I could use it with NativeScript app. It's good to see finally all functionality is broken down in the their respective extension in vNext. I still see Storage (local/session) must be separated into into own extension.

There are couple of bugs in the vNext dependency extension. I am providing the line# in the code and what it should be.

https://github.com/Microsoft/ApplicationInsights-JS/blob/master/vNext/extensions/applicationinsights-dependencies-js/src/ajax.ts#L109

https://github.com/Microsoft/ApplicationInsights-JS/blob/master/vNext/extensions/applicationinsights-dependencies-js/src/ajax.ts#L164

https://github.com/Microsoft/ApplicationInsights-JS/blob/master/vNext/extensions/applicationinsights-dependencies-js/src/ajax.ts#L194

https://github.com/Microsoft/ApplicationInsights-JS/blob/master/vNext/extensions/applicationinsights-dependencies-js/src/ajax.ts#L222

it should be **ajaxMonitorInstance._core.logger.throwInternal(**

Thanks

PR #849, opened 2019-04-25
Update ajax.ts

slolam:
Incorrect *this* is referred in the overridden anonymous function handler of XMLHttpRequest prototype. 

Fixes for the issue #848


markwolff:
Good catch and thanks for the PR! Will get this published with rc5

slolam:
Thanks @markwolff

Issue #863, opened 2019-05-09
Correlation not working


LeopoldLerch:
Hi, 
I am using 2.0.0-rc4

I am sending Ajax-requests to my web service. However, I am not able to bring them together. I guess it is because the "request-id" header always consists of just "f1+QH".

I debugged it down to this line in AjaxMonitor.prototype.createFetchRecord

`        if (this._context && this._context.telemetryTrace && this._context.telemetryTrace.traceID) {
            // this format corresponds with activity logic on server-side and is required for the correct correlation
            id = "|" + this._context.telemetryTrace.traceID + "." + Util_Util.newId();
        }
        else {
            id = Util_Util.newId();
        }`

_context is always empty, so it always just generates a plain ID and then is not able to connect it to the operation.

this seems to be due this line in AjaxMonitor.prototype.initialize:

`            if (extensions.length > 0 && extensions) {
                var propExt = extensions[PropertiesPluginIdentifier];
                if (propExt) {
                    this._context = propExt.context; // we could move IPropertiesPlugin to common as well
                }
            }`

Altough there is an Key for an PropertiesPlugin in extensions, it is named "PropertiesPlugin", not "AppInsightsPropertiesPlugin", what is the value of the variable (altough the Object referenced has an "Identifier" with that value").

However, due to that "_context" is null and therefore the requestId cannot be generated to correlate the request with the server-side metrics.

Pls have a look into that

jdudleyie:
Hi,

I was coming here to report an issue - it looks like it might be the same issue as this one. If not please let me know and I can move this into a new Issue.

I have an Angular app sending ajax requests to my .NET Core API. Both my Angular App and my .NET Core API are running in different containers/pods in Kubernetes.

I have added my Application Insights Instrumentation Key to both of them, and added the relevant code.

I can see both on my Application Map in the Azure Portal, but although my client calls my API, the calls do not link up - i.e. it shows the 1) client ajax call target and the 2) API as two separate nodes - even though they are the same thing.

As the target of my client ajax calls is "api.todocalendar.app", and my API was labelled as "todo-calendar-api" - in an effort to link them I tried setting the "telemetry.Context.Cloud.RoleName" to "api.todocalendar.app" - but it still showed them as separate nodes (I have since removed this change.

I also tried adding "Microsoft Application Insights for Kubernetes" - I see the extra telemetry fields in my Azure Analytics, but I still have the same issue.

[...]

PR #865, opened 2019-05-09
Fix dependencies plugin not correctly finding properties plugin

markwolff:
Addresses #863

### DIFF
--- a/vNext/extensions/applicationinsights-dependencies-js/src/ajax.ts
+++ b/vNext/extensions/applicationinsights-dependencies-js/src/ajax.ts
@@ -338,7 +338,7 @@ export class AjaxMonitor implements ITelemetryPlugin, IDependenciesPlugin, IInst
         }
     }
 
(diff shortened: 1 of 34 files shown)


vNext/extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -338,7 +338,7 @@ export class AjaxMonitor implements ITelemetryPlugin, IDependenciesPlugin, IInst
         }
     }
 
-    priority: number = 161;
+    priority: number = 171;
```
hectorhdzg: Why you need dependency extension to be after properties extension?

markwolff: Will this break any of your stuff? Basically this plugin grabs `_context` from the properties plugin, which is not defined until after the properties plugin has initialized.

Issue #857, opened 2019-05-03
appId not auto generating


lworkman:
I'm using `applicationinsights-js` in an Angular 5 app with a node.js backend (running `applicationinsights`). I've been struggling for the past day to get correlation working correctly between the two services, and it seems like the issue has come down to a missing correlation header (`Request-Context`). My analytics class constructor looks like this:
```
    this.appInsightsClient = new ApplicationInsights({
      config: {
        instrumentationKey: "MY KEY HERE",
        enableCorsCorrelation: true,
        disableFetchTracking: false,
        disableCorrelationHeaders: false
      }
    });
    this.appInsightsClient.loadAppInsights();
```
And it appears to track all requests properly _without proper correlation since `Request-Context` is missing_. I could only get `Request-Context` included in the headers by adding a manual `appId`:
```
    this.appInsightsClient = new ApplicationInsights({
      config: {
        instrumentationKey: "MY KEY HERE",
        enableCorsCorrelation: true,
        disableFetchTracking: false,
        disableCorrelationHeaders: false,
        appId: "Editor"
      }
    });
    this.appInsightsClient.loadAppInsights();
```

**Is it necessary to manually set the `appId`**? The typescript definition makes it sound the appId should be automatically set _unless_ you're using the Beacon API.  Although the correlation appears to be working now, I'd like to know if having to set `appId` manually is a bug, intended, or a result of me messing up my setup.

jpiyali:
Manual setup is not required for appId. The setting enableCorsCorrelation and disableCorrelationHeaders should be enough. Are you not seeing these headers setup in your outgoing dependencies from your client?

jpiyali:
Please note only dependency requests sent through with XmlHttpRequest are instrumented.

jpiyali:
Forgot to mention, if you use Fetch, you have to enable tracking explicitly 

disableFetchTracking | true | If true, Fetch requests are not autocollected. Default is true
-- | -- | --

lworkman:
@jpiyali with the below settings:
```
new ApplicationInsights({
  config: {
    instrumentationKey: "MY KEY HERE",
    enableCorsCorrelation: true,
    disableFetchTracking: false,
    disableCorrelationHeaders: false
  }
});
this.appInsightsClient.loadAppInsights();
```

The follow headers are sent:
![image](https://user-images.githubusercontent.com/12467662/57157575-ee796a80-6d95-11e9-93e5-0194cd3b1622.png)

[...]

PR #866, opened 2019-05-09
save appId response from breeze channel in context

markwolff:
Addresses #857 
- Saves appId response from Breeze Channel in TelemetryContext instance. This is then used in the dependencies plugin when it grabs its context ref from properties plugin

### DIFF
--- a/vNext/extensions/applicationinsights-properties-js/src/PropertiesPlugin.ts
+++ b/vNext/extensions/applicationinsights-properties-js/src/PropertiesPlugin.ts
@@ -9,12 +9,13 @@ import {
 } from '@microsoft/applicationinsights-core-js';
 import { TelemetryContext } from './TelemetryContext';
 import { PageView, ConfigurationManager,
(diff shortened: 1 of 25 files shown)


vNext/extensions/applicationinsights-properties-js/src/PropertiesPlugin.ts
```diff
@@ -47,6 +48,8 @@ export default class PropertiesPlugin implements ITelemetryPlugin, IPropertiesPl
 
         this._logger = core.logger;
         this.context = new TelemetryContext(core.logger, this._extensionConfig);
+        this._breezeChannel = Util.getExtension(extensions, BreezeChannelIdentifier);
+        this.context.appId = () => this._breezeChannel ? this._breezeChannel["_appId"] : null;
```
hectorhdzg: Is Breeze channel appId available during initialization?

markwolff: Only if it is provided in config. Otherwise, it is only available after sending telemetry the first time.

PR #869, opened 2019-05-13
PR: Whitelisting of correlation-Domains

LeopoldLerch:
Hi, 
I added a possibility to whitelist the domain, on which the correlation-Headers should be added on Ajax-/Fetch-Requests. 

It first evaluates the correlationHeaders-Array (if given) and then additionaly the correlationHeaderExcludedDomains(if given). 

(well, my Styler, made some addtitional stuff too, obv.).

### DIFF
--- a/vNext/shared/AppInsightsCommon/src/Util.ts
+++ b/vNext/shared/AppInsightsCommon/src/Util.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 import { StorageType } from "./Enums";
-import { CoreUtils, _InternalMessageId, LoggingSeverity, IDiagnosticLogger } from "@microsoft/applicationinsights-core-js";
+import { CoreUtils, _InternalMessageId, LoggingSeverity, IDiagnosticLogger, IPlugin } from "@microsoft/applicationinsights-core-js";
 import { IConfig } from "./Interfaces/IConfig";
 import { RequestHeaders } from "./RequestResponseHeaders";
 import { DataSanitizer } from "./Telemetry/Common/DataSanitizer";
@@ -10,7 +10,6 @@ import { ICorrelationConfig } from "./Interfaces/ICorrelationConfig";
 
 export class Util {
     private static document: any = typeof document !== "undefined" ? document : {};
-    private static _canUseCookies: boolean = undefined;
     private static _canUseLocalStorage: boolean = undefined;
     private static _canUseSessionStorage: boolean = undefined;
     // listing only non-geo specific locations
@@ -309,7 +308,7 @@ export class Util {
             };
         }
 
-        return Util._canUseCookies;
+        return CoreUtils._canUseCookies;
     }
 
     /**
@@ -547,6 +546,20 @@ export class Util {
      */
     public static IsBeaconApiSupported(): boolean {
         return ('sendBeacon' in navigator && (<any>navigator).sendBeacon);
+    }
+
+    public static getExtension(extensions: IPlugin[], identifier: string) {
+        let extension = null;
+        let extIx = 0;
+
+        while (!extension && extIx < extensions.length) {
+            if (extensions[extIx] && extensions[extIx].identifier === identifier) {
+                extension = extensions[extIx];
+            }
+            extIx++;
+        }
+
+        return extension;
     }
 }
 
--- a/vNext/shared/AppInsightsCommon/src/Util.ts
+++ b/vNext/shared/AppInsightsCommon/src/Util.ts
@@ -625,11 +625,11 @@ export class CorrelationIdHelper {
 
         let includedDomains = config && config.correlationHeaderDomains;
         if (includedDomains) {
-            for (let i = 0; i < includedDomains.length; i++) {
-                let regex = new RegExp(includedDomains[i].toLowerCase().replace(/\./g, "\.").replace(/\*/g, ".*"));
-                if (!regex.test(requestHost)) {
-                    return false;
-                }
+            if (includedDomains.length == 0 || includedDomains.every((domain) => {
+                let regex = new RegExp(domain.toLowerCase().replace(/\./g, "\.").replace(/\*/g, ".*"));
+                return !regex.test(requestHost);
+            })) {
+                return false;
             }
         }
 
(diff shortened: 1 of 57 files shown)


vNext/shared/AppInsightsCommon/src/Util.ts
```diff
@@ -623,6 +623,16 @@ export class CorrelationIdHelper {
             return false;
         }
 
+        let includedDomains = config && config.correlationHeaderDomains;
+        if (includedDomains) {
+            for (let i = 0; i < includedDomains.length; i++) {
+                let regex = new RegExp(includedDomains[i].toLowerCase().replace(/\./g, "\.").replace(/\*/g, ".*"));
```
hectorhdzg: nip: Adding comments about all regex validation makes the code easier to read

vNext/shared/AppInsightsCommon/src/Util.ts
```diff
@@ -623,6 +623,16 @@ export class CorrelationIdHelper {
...
+                if (!regex.test(requestHost)) {
```
goenning: shouldn't it be like this?

```
if (regex.test(requestHost)) {
  return true;
}
```

if my config is `correlationHeaderDomains: ["*.mydomain.com", "mydomain.com"]`, I'd expect this function to return true for https://mydomain.com/api/v1/users, which I believe your code doesn't.

LeopoldLerch: Well, it returns fals if it does not match. Thats a difference.

If I would return true, if it matches, i would exclude the whole "BlackList" check following. 

In the current version, you can whitelist domains, and then define separate areas of the whitelist, which are blacklisted. If it does not match a blacklisted one or there are no blacklisted ones, it returns true by default.

But you can change if this is not required.

goenning: So back to my example, `https://mydomain.com/api/v1/users` doesn't match `"*.mydomain.com"`, so it'll return false and not include the headers, but it should, because I've whitelisted it (second item on array).

LeopoldLerch: Hi
First, the regex for a domain of type *.mydomain.com would be transformed to /.*\.mydomain\.com.*/g.
This of course would not match "https://mydomain.com/api/v1/users", because you are explicitly saying it has to have a "." before "mydomain", and your domain doesn´t. In your criteria the Dot is not optional.

Second, it matches only against the host of the url. The rest is stripped apart, so it matches the regex only against mydomain.com.

I know the matching mechanism isn´t quite powerful, but i did it reuse from the already existing blacklist mechanism. So blacklisting with that type of criteria wouldn´t have worked in the past too.

I discussed with Mark Wolff, which helped me starting here (bit thx to him again) an future possibility with a callback mechanism for the the calculation, if the headers should be appended or not (which would give the developers a much more powerful tool then the current black- and now whitelists).

LeopoldLerch: sorry, had to update my post, as the Stars in the regex went missing

LeopoldLerch: Ah, I see what you are going to.

There is a bug with multiple whitelisted ones, as it will return false withe the first not matching. Of course it should first match against all of them and return only false, when there is no match at all. 

This is of course a bit different to blacklist, as it is ok there, when it returns false for the first match.

How can I fix that? Do you reject the PR and I create a new one after my fix?

goenning: Maybe my example was not the best. But what I’m saying is that if I have whitelisted 5 different domains and the first one doesn’t match, it’ll skip it by returning false on that function without checking for all of the other 4 domains.

Example: if I whitelist `[google.com, mydomain.com]`, it wouldn’t include headers on a request to `mydomain.com/api/v1/users`, but it should as it’s been whitelisted.

Or maybe I’m misunderstood something?

markwolff: You are correct. As it is now, it cannot support multiple domains. It can either return true on the first match as you initially suggested, or return false if all don't match. I can't think of a scenario where we need to look at the blacklist if the whitelist is set, so it is probably fine to go with the more simple `return true` route. 
@LeopoldLerch You can just fix this in the existing PR

goenning: and if my website uses Ajax to many different hosts but I want to support only 2 of them, do I need to do something like:

```
correlationHeaderDomains: ['google.com', 'mydomain.com'],
correlationHeaderExcludedDomains: ['*']
```

Because as far as I understood, if the domain doesn't match either a whitelist or a blacklist, it'll default to true (set the headers), right?

I wonder if the default should change based on the existence of a whitelist. If we have a whitelist it'll only set headers to whitelisted domains. But if we don't, then it'll set headers to all. Otherwise, what's the point of a whitelist without a blacklist?

LeopoldLerch: if there is a whitelist, it will only continue, when the host matches any of the whitelisted domains. If it doesn´t it will return false.

If there is no whitelist, it will continue to check if there is a blacklist. If there is a blacklist and it matches any of the blacklisted domains, it will return false.

otherwise it will continue to return true as the default

LeopoldLerch: The new commit should do the trick. 
I hope it is OK to use Array.some, (as it is supported since IE9)

The new Version does  the following
Check if an array for whitelists is there
Check if an existing array has an entry, that passes the whitelistcheck, otherwise returning false

If no array is there or it passes the whitelistcheck, it continues to the blacklist check. If no blacklist is there or it passes the blacklist-check, "true" will be returned and the headers will be appended

goenning: it looks good to me now! I really hope we can get this merged soon so I can start correlating the requests 👍

markwolff:
Thanks for the contribution! Will have this land in the next release.

goenning:
@LeopoldLerch thanks for implementing the feature 👍 

@markwolff Is this going to be release on both v1 and v2? Do you have an idea when would that happen?

markwolff:
@goenning Likely only v2. Hoping to get it out there by next week or the week after.

PR #873, opened 2019-05-14
Added a classname for the appInsights HOC's div

florianchevallier:
See [here](https://github.com/Azure/react-appinsights/pull/89) for the discussion


flip-it:
I'm using React Native Web in my project. I have also the issue with full height. Adding a class name though is not practical for me. Would it instead be possible to have an option to specify the component to pass in? Eg:

`component={View}`

This is how [react-intl](https://github.com/formatjs/react-intl/wiki/Components#using-react-intl-with-react-native) and [react-transition-group](https://reactcommunity.org/react-transition-group/transition-group#TransitionGroup-prop-component) do.

florianchevallier:
@hectorhdzg Is it good for you like that ?

hectorhdzg:
@flip-it we will add both options in next release, allowing className and component to be provided as well

SanjiKir:
Hey @florianchevallier , thanks for your work on react SDK for app insights. I have one question though, what's the need of empty div wrapper around the user component? I mean if for some reason behind the scenes your are putting your custom components in render why just not to use `Fragment`?

florianchevallier:
> Hey @florianchevallier , thanks for your work on react SDK for app insights. I have one question though, what's the need of empty div wrapper around the user component? I mean if for some reason behind the scenes your are putting your custom components in render why just not to use `Fragment`?

Hi ! In my case, on my app, I had a full height CSS app and without this solution, it broke all the design. The problem here was [this div](https://github.com/florianchevallier/ApplicationInsights-JS/blob/04da5e746f70a4b51da4ce42e871dac25fc991d3/vNext/extensions/applicationinsights-react-js/src/withAITracking.tsx#L78-L88) which couldn't be customized and therefore, couldn't add a `height: 100%` on it

Issue #877, opened 2019-05-16
TrackPageViewPerformance is not logging perf measurements


vishwam:
Version: 2.0.0-rc.5
Browser: Chrome 74

Repro:
1. Call TrackPageViewPerformance (or PageView): `appInsights.trackPageViewPerformance({ name: document.title, uri: location.href })`

Expected: event is emitted with following fields populated: ['durationMs', 'duration', 'perfTotal', 'networkConnect', 'sentRequest', 'receivedResponse', 'domProcessing']

Actual: none of the `baseData.aiDataContract` fields are emitted. You can put a breakpoint at `PageViewPerformanceEnvelopeCreator.Create` (in the channel-js package) to verify that the data is present in the incoming telemetryItem, but not in output PageViewPerformance object: 

```
> telemetryItem.baseData:
domProcessing: "00:00:03.360"
duration: "00:00:03.553"
durationMs: 3553
isValid: true
name: "Navigate"
networkConnect: "00:00:00.001"
perfTotal: "00:00:03.553"
properties: {cv: "5zks3bvx.6", …}
receivedResponse: "00:00:00.002"
sentRequest: "00:00:00.186"
uri: "https://apps.saas.localhost/create"

> baseData
aiDataContract:
    domProcessing: 0
    duration: 0
    measurements: 0
    name: 0
    networkConnect: 0
    perfTotal: 0
    properties: 0
    receivedResponse: 0
    sentRequest: 0
    url: 0
    ver: 1
    __proto__: Object
measurements: undefined
name: "Navigate"

```

vishwam:
Closing this issue since it's been fixed in 2.0.0-rc6

PR #878, opened 2019-05-16
fix missing part b in pageviewperformance

markwolff:
Addresses https://github.com/microsoft/ApplicationInsights-JS/issues/877


vNext/shared/AppInsightsCommon/src/Telemetry/PageViewPerformance.ts
```diff
@@ -31,12 +32,21 @@ export class PageViewPerformance extends PageViewPerfData implements ISerializab
     /**
      * Constructs a new instance of the PageEventTelemetry object
```
jpiyali: why is pageViewTelemetry receiving pageViewPerformanceTelemtry?

PR #899, opened 2019-05-28
common: add tests for canIncludeCorrelationHeader

markwolff:
- Add tests for `canIncludeCorrelationHeader`. 
- minor: Replace `Array.some` logic


markwolff:
@LeopoldLerch

LeopoldLerch:
I might be wrong, but Array.prototype.forEach is also not supported in IE < 9

markwolff:
Yep you are correct. Moving forward, it might not be feasible to plan for <IE8 support at all.

LeopoldLerch:
If really needed down there they should use polyfills. IE 8 is really old

Mark Wolff <notifications@github.com> schrieb am Mi., 29. Mai 2019, 18:57:

PR #903, opened 2019-05-29
fix tag issue

xiao-lix:


vNext/channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
```diff
@@ -207,9 +207,21 @@ export abstract class EnvelopeCreator {
         //   }
 
         let tgs = {};
-        item.tags.forEach(tg => {
-            tgs = { ...tgs, ...tg };
-        });
+        let tagsCopy = item.tags;
```
markwolff: Is it possible to do this without creating a copy?

vNext/extensions/applicationinsights-properties-js/src/TelemetryContext.ts
```diff
@@ -105,14 +105,14 @@ export class TelemetryContext implements ITelemetryContext {
     public applyInternalContext(event: ITelemetryItem) {
```
markwolff: The type of `ITelemetryItem` should be changed from `Tags[]`

vNext/channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
xiao-lix: yes, just removed the copy and make sure all tests passed

Issue #891, opened 2019-05-22
Jest broken by applicationinsights-react-js


csuich2:
In attempting to switch from react-appinsights to application-insights-react-js, many of my Jest tests starting failing with this error:

```
  ● Test suite failed to run

    Cannot find module './cjs/react.development.js' from 'applicationinsights-react-js.js'
```

This is easily reproducible by cloning the [demo app](https://nuget.pkg.github.com/Azure-Samples/application-insights-react-demo), then attempting to run tests. You may want to consider an Azure Pipelines build to the demo app to catch this kind of thing.

lanfear:
confirmed, simple swap from final RC to this release (with the minor breaking config changes) blows up tests while the runtime code seems to work fine.  please assist.  with RC6 i had a full suite of passing tests.

csuich2:
Interesting. I believe I had the same issue with RC5, but thought it would be fixed before the final RC, so I just waited to make the switch.

lanfear:
> Interesting. I believe I had the same issue with RC5, but thought it would be fixed before the final RC, so I just waited to make the switch.

fwiw, the 'final' RC candidate i was using was from this repo `https://github.com/Azure/react-appinsights` because i was unaware of the name change.   possible there were minor variances between the new repo location and the old, i was unaware this new repo existed (if it did indeed exist before the final release).  so it's possible we were using different RCs, although i'm sure my was listed in package.json as RC6.

hectorhdzg:
We are working on fixing this issue, like @csuich2  mentioned we found it while updating the sample app and we believe is specific to [Create React App](https://facebook.github.io/create-react-app/docs/getting-started) Jest configuration, let me know if you see this error in different environments.

Itrulia:
I also encounter this when using `"next": "^8.1.0",` when starting the development server via `next` (note that you need to visit the URL first for it to compile)

lanfear:
> We are working on fixing this issue, like @csuich2 mentioned we found it while updating the sample app and we believe is specific to [Create React App](https://facebook.github.io/create-react-app/docs/getting-started) Jest configuration, let me know if you see this error in different environments.

thanks for the quick follow-up, fwiw, i can confirm i used create-react-app to bootstrap my initial application.

valentinvoilean:
until gets fixed, you can mock it:

`jest.mock('@microsoft/applicationinsights-react-js', () => ({ ReactPlugin: jest.fn() }));`

[...]

PR #905, opened 2019-05-30
React (rollup): add commonjs plugin, preferBuiltins

markwolff:
Addresses #891 
Modify rollup config to prevent react plugin from crashing users' Jest unit tests
- Adds commonjs rollup plugin
- Adds `preferBuiltins: true` to the node-resolve plugin
- Deletes browser rollup config, since the package.json does not reference a `browser` entry point

### DIFF
--- a/vNext/extensions/applicationinsights-react-js/rollup.config.js
+++ b/vNext/extensions/applicationinsights-react-js/rollup.config.js
@@ -1,6 +1,7 @@
 import nodeResolve from "rollup-plugin-node-resolve";
 import {uglify} from "rollup-plugin-uglify";
 import replace from "rollup-plugin-replace";
+import commonjs from "rollup-plugin-commonjs";
 
 const version = require("./package.json").version;
 const outputName = "applicationinsights-react-js";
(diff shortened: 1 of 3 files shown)


vNext/extensions/applicationinsights-react-js/rollup.config.js
```diff
@@ -87,8 +55,6 @@ const nodeUmdRollupConfigFactory = (isProduction) => {
 };
 
 export default [
-  browserRollupConfigFactory(true),
-  browserRollupConfigFactory(false),
-  nodeUmdRollupConfigFactory(true),
+  nodeUmdRollupConfigFactory(false),
```
hectorhdzg: isProduction = true is missing

vNext/extensions/applicationinsights-react-js/rollup.config.js
```diff
@@ -11,45 +12,6 @@ const banner = [
   " */"
 ].join("\n");
 
-const browserRollupConfigFactory = isProduction => {
```
hectorhdzg: Why remove browser entirely?

markwolff: It was not being referenced in package.json, so to avoid confusion and bloating the package.

markwolff: added it back for consistency with other extensions. Also to enable scenarios where one might want to self host browserified version

PR #908, opened 2019-05-30
React extension removing dependency on Analytics package

hectorhdzg:
Removing dependency on Analytics package


vNext/extensions/applicationinsights-react-js/test/ReactAI.test.ts
```diff
@@ -34,33 +34,46 @@ describe("ReactAI", () => {
     const history = createBrowserHistory();
     jest.useFakeTimers();
     init();
-    let analyticsExtension = new ApplicationInsights();
+    let analyticsExtension = {
```
markwolff: Can this use the analyics plugin type so that it is not difficult to maintain?

vNext/extensions/applicationinsights-react-js/src/ReactPlugin.ts
```diff
@@ -20,7 +19,7 @@ export default class ReactPlugin implements ITelemetryPlugin {
     public priority = 165;
```
markwolff: cab you change this to 180? Properties plugin is 170 and this plugin sends telemetry in its initialization steps.

PR #918, opened 2019-06-06
add operation.name, pageview.id

markwolff:
Adds previously existing behavior to automatically name a newly started operation.
V1 SDK:
https://github.com/microsoft/ApplicationInsights-JS/blob/a96b1a0e77a43e6ba64369cae635c678e009f4f8/JavaScript/JavaScriptSDK/Context/Operation.ts#L19-L23

Also saves the current operation id to each PageView.id as was done in V1 SDK.
https://github.com/microsoft/ApplicationInsights-JS/blob/a96b1a0e77a43e6ba64369cae635c678e009f4f8/JavaScript/JavaScriptSDK/AppInsights.ts#L140-L141

### DIFF
--- a/vNext/channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
+++ b/vNext/channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
@@ -365,9 +365,15 @@ export class PageViewEnvelopeCreator extends EnvelopeCreator {
         }
 
         let bd = telemetryItem.baseData as IPageViewTelemetryInternal;
+
+         // special case: pageview.id is grabbed from current operation id. Analytics plugin is decoupled from properties plugin, so this is done here instead. This can be made a default telemetry intializer instead if needed to be decoupled from channel
+        let currentContextId;
+        if (telemetryItem.ext && telemetryItem.ext.trace && telemetryItem.ext.trace.traceID) {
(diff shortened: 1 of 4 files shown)


vNext/channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
```diff
@@ -365,9 +365,15 @@ export class PageViewEnvelopeCreator extends EnvelopeCreator {
         }
 
         let bd = telemetryItem.baseData as IPageViewTelemetryInternal;
+
+         // special case: pageview.id is grabbed from current operation id. Analytics plugin is decoupled from properties plugin, so this is done here instead. This can be made a default telemetry intializer instead if needed to be decoupled from channel
+        let currentContextId;
+        if (telemetryItem.ext && telemetryItem.ext.trace && telemetryItem.ext.trace.traceID) {
```
hectorhdzg: What about PageViewPerformance is not a problem also in there?

markwolff: Is not a field in V1 SDK, so it relies on operation.id instead.
https://github.com/microsoft/ApplicationInsights-JS/blob/a96b1a0e77a43e6ba64369cae635c678e009f4f8/JavaScript/JavaScriptSDK/Telemetry/PageViewPerformance.ts#L19-L31

PR #928, opened 2019-06-10
analytics: move to last priority of default plugins

markwolff:
Custom telemetry intializers are run in `processTelemetry` in the analytics plugin. It needs to be made as the last priority among the default plugins so that all other telemetry information from other plugins is available to custom telemetry initializers.

Also removes minified entry point from `core`

### DIFF
--- a/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
+++ b/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
@@ -30,7 +30,7 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
     public static Version = "2.0.1"; // Not currently used anywhere
     public initialize: (config: IConfiguration, core: IAppInsightsCore, extensions: IPlugin[]) => void;
     public identifier: string = "ApplicationInsightsAnalytics"; // do not change name or priority
-    public priority: number = 200;// take from reserved priority range 100- 200
+    public priority: number = 180; // take from reserved priority range 100- 200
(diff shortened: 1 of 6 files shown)


vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -30,7 +30,7 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
     public static Version = "2.0.1"; // Not currently used anywhere
     public initialize: (config: IConfiguration, core: IAppInsightsCore, extensions: IPlugin[]) => void;
     public identifier: string = "ApplicationInsightsAnalytics"; // do not change name or priority
-    public priority: number = 160;// take from reserved priority range 100- 200
+    public priority: number = 199;// take from reserved priority range 100- 200
```
hectorhdzg: React Native is 199

vNext/shared/AppInsightsCore/package.json
```diff
@@ -15,7 +15,6 @@
     "main": "dist/applicationinsights-core-js.js",
     "module": "dist-esm/applicationinsights-core-js.js",
     "types": "types/applicationinsights-core-js.d.ts",
-    "browser": "browser/applicationinsights-core-js.min.js",
```
hectorhdzg: Is this not used anymore?

markwolff: forgot to doc this change. In debugger it is using minified version for only core due to this. Other packages' entry point is `"main": dist/<unminified-entry>.js`

PR #936, opened 2019-06-14
update beaconSender, add size limit and clear buffer after sending data

xiao-lix:


vNext/channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -501,18 +501,33 @@ export class Sender implements IChannelControlsAI {
     private _beaconSender(payload: string[], isAsync: boolean) {
         var url = this._config.endpointUrl();
         var batch = this._buffer.batchPayloads(payload);
+        var browserLimit = 63000;
```
markwolff: Make this a static named const

vNext/channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -501,18 +501,33 @@ export class Sender implements IChannelControlsAI {
...
+        var count = 0;
+        var overLimit = false;
+
+        for (var i = 0; i < payload.length; i++){
```
markwolff: I'm wondering if instead of having these checks we can instead just fallback to xhr whenever `sendBeacon` fails to queue the event. This way we don't need to do any size checks, etc

xiao-lix: makes sense

vNext/channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -511,8 +511,11 @@ export class Sender implements IChannelControlsAI {
 
         if (queued) {
             this._buffer.markAsSent(payload);
+            // no response from beaconSender, clear buffer
+            this._onSuccess(payload, payload.length);
         } else {
-            this._logger.throwInternal(LoggingSeverity.CRITICAL, _InternalMessageId.TransmissionFailed, ". " + "Failed to send telemetry with Beacon API.");
+            this._logger.throwInternal(LoggingSeverity.WARNING, _InternalMessageId.TransmissionFailed, ". " + "Failed to send telemetry with Beacon API, retried with xhrSender.");
+            this._xhrSender(payload, true);
```
markwolff: We can now add (e2e) tests to make sure that it will correctly fallback here. Also add a test to verify it will use beacons when that mode is enabled.

PR #939, opened 2019-06-17
Add missing `namePrefix` to `IConfig` interface

andrewconnell:


andrewconnell:
I see the build failed in Travis, but I can't tell what the error is... appears to be a setup config... but not sure... happy to fix if someone can provide some insight.

vNext/shared/AppInsightsCommon/src/Interfaces/IConfig.ts
```diff
@@ -245,6 +245,14 @@ export interface IConfig {
      */
     enableCorsCorrelation?: boolean;
 
+    /**
+     * @description An optional value that will be used as name postfix for localStorage and cookie name.
+     * @type {string}
+     * @memberof IConfig
+     * @defaultValue null
+     */
+    namePrefix
```
markwolff: this should be `namePrefix?: boolean`
This will fix the travis errors

andrewconnell: The datatype seems to be a string. When I set it to a string, I saw the prefix I set (a string) appended to the properties in the browser's local storage...

But the docs on the root README in the repro simply said `undefined` so I wasn't sure what you wanted here.

andrewconnell: Oops... NM... I see my error. Let me fix. Should be `namePrefix?: string;`

markwolff: oops yes a string

PR #947, opened 2019-06-25
Feature: add SPA auto route change tracking

markwolff:
Adds new off-by-default configuration `enableAutoRouteTracking`, which listens for changes in the `window.history` stack to trigger a new pageview/operation. Each listener forwards events to the catch-all `locationchange` event to actually track the new pageview.

All events are prefixed with `config.namePrefix`, so that multiple instances of the SDK do not duplicate pageview telemetry to each instance.

README changes can be uncommented when the feature is shipped

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ Most configuration fields are named such that they can be defaulted to falsey. A
 | appId | null | AppId is used for the correlation between AJAX dependencies happening on the client-side with the server-side requests. When Beacon API is enabled, it cannot be used automatically, but can be set manually in the configuration. Default is null |
 | enableCorsCorrelation | false | If true, the SDK will add two headers ('Request-Id' and 'Request-Context') to all CORS requests tocorrelate outgoing AJAX dependencies with corresponding requests on the server side. Default is false |
 | namePrefix | undefined | An optional value that will be used as name postfix for localStorage and cookie name.
+<!-- | enableAutoRouteTracking | false | Automatically track route changes in Single Page Applications (SPA). If true, each route change will send a new Pageview to Application Insights. Hash route changes changes (`example.com/foo#bar`) are also recorded as new page views. -->
 
 ## Single Page Applications
 
--- a/README.md
+++ b/README.md
@@ -158,13 +158,9 @@ Most configuration fields are named such that they can be defaulted to falsey. A
 
 ## Single Page Applications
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -154,6 +154,7 @@ Most configuration fields are named such that they can be defaulted to falsey. A
 | appId | null | AppId is used for the correlation between AJAX dependencies happening on the client-side with the server-side requests. When Beacon API is enabled, it cannot be used automatically, but can be set manually in the configuration. Default is null |
 | enableCorsCorrelation | false | If true, the SDK will add two headers ('Request-Id' and 'Request-Context') to all CORS requests tocorrelate outgoing AJAX dependencies with corresponding requests on the server side. Default is false |
 | namePrefix | undefined | An optional value that will be used as name postfix for localStorage and cookie name.
+<!-- | enableAutoRouteTracking | false | Automatically track route changes in Single Page Applications (SPA). If true, each route change will send a new Pageview to Application Insights. Hash route changes changes (`example.com/foo#bar`) are also recorded as new page views. -->
```
hectorhdzg: Un-comment text

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -563,6 +567,41 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
...
+                window.dispatchEvent(new Event(this.config.namePrefix + "locationchange"));
+                return ret;
+            })(history.replaceState);
+
+            window.addEventListener(this.config.namePrefix + "popstate",()=>{
+                window.dispatchEvent(new Event(this.config.namePrefix + "locationchange"));
+            });
+
+            window.addEventListener(this.config.namePrefix + "locationchange", () => {
+                if (this._properties) this._properties.context.telemetryTrace.traceID = Util.newId();
```
hectorhdzg: Checking for nulls in context and context.telemetryTrace would be better here

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -563,6 +567,41 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
...
+         */
+        if (this.config.enableAutoRouteTracking === true && typeof history === "object" && typeof window === "object") {
+            // Find the properties plugin
+            extensions.forEach(extension => {
+                if (extension.identifier === PropertiesPluginIdentifier) {
+                    this._properties = extension;
+                }
+            });
+
+            history.pushState = ( f => function pushState() {
```
hectorhdzg: We need to test this in IE 9, pushState was added in IE10 would be good to understand how this code behaves in IE9

README.md
markwolff: Going wait until it is released so that it is not ambiguous whether or not it is available yet

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
markwolff: Agreed. I will add nullchecks on the history functions. afaik, SPAs made to work with IE9 are including polyfills, so I should probably just make sure we are not introducing any error logging for those.

PR #951, opened 2019-06-27
move old validation code to tests

xiao-lix:


vNext/AISKU/Tests/applicationinsights.e2e.tests.ts
```diff
@@ -658,7 +665,33 @@ export class ApplicationInsightsTests extends TestClass {
...
+                    case PageViewPerformance.dataType:
+                        return PageViewPerformanceValidator.PageViewPerformanceValidator.Validate(payload, baseType);
+                    case RemoteDependencyData.dataType:
+                        return RemoteDepdencyValidator.RemoteDepdencyValidator.Validate(payload, baseType);
+        
+                    default:
+                        return EventValidator.EventValidator.Validate(payload, baseType);
+                }
+            }
+            return currentCount === expectedCount && isValid && !!this._ai.context.appId();
```
markwolff: It does not look like this `return` will ever get reached if `currentCount === expectedCount`

xiao-lix: yes you are right.. fixed it! 👍

PR #956, opened 2019-07-02
SDK must work even if there are no extensions

hectorhdzg:
Allowing core to work even if extensions are not present, most basic scenario of SDK core+ channel


vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -53,7 +53,7 @@ export class AppInsightsCore implements IAppInsightsCore {
         this.logger = new DiagnosticLogger(config);
 
         // Initial validation
-        extensions.forEach((extension: ITelemetryPlugin) => {
+        this._extensions.forEach((extension: ITelemetryPlugin) => {
```
markwolff: I believe `this._extensions` is an empty array at this point

vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -145,11 +145,6 @@ export class AppInsightsCore implements IAppInsightsCore {
             }
         });
 
-        // Remove sender channels from main list
-        if (c < this._extensions.length) {
-            this._extensions.splice(c);
-        }
```
markwolff: What was this code meant to do?

vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
hectorhdzg: I see, so is actually checking the input, my mistake reverting this change

vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
hectorhdzg: This code is basically removing ChannelController from the array, I don't see any use of this, _extensions is only used in track method after this and having Channel Controller in there doesn't look to affect any scenario, _extensions is also a public property but is only used in SKU to identify if Properties extension is available

vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
hectorhdzg: Merged both extensions and config.extension so validation applies to everything

vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
hectorhdzg: Adding this logic back and including check for empty extensions during track

vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -238,114 +235,23 @@ export class AppInsightsCore implements IAppInsightsCore {
...
 
         if (CoreUtils.isNullOrUndefined(telemetryItem.iKey)) {
-            this._notifiyInvalidEvent(telemetryItem);
+            this._notififyInvalidEvent(telemetryItem);
             throw Error("telemetry instrumentationKey required");
         }
     }
 
-    private _notifiyInvalidEvent(telemetryItem: ITelemetryItem): void {
+    private _notififyInvalidEvent(telemetryItem: ITelemetryItem): void {
```
markwolff: typo

PR #965, opened 2019-07-10
Fix rendering issue in readme

andrewconnell:
README.md in root of repo had extra code fencing notes that caused the "Source Map Support" section to get treated as code and incorrectly rendered.


markwolff:
Thanks for the fix!

Issue #974, opened 2019-07-19
New Snippet Setup pollutes the global namespace (window)


PeterOrneholm:
Using the new [Snippet Setup](https://github.com/microsoft/ApplicationInsights-JS#snippet-setup-ignore-if-using-npm-setup) introduces 5 global variables (on the window object). I can see the following being added to window.

- `aisdk`
- `ainame`
- `appInsightsSDK`
- `appInsights`
- `sdkInstance`

Is this really necessary? Is there any usecase for when I should use / change these variables? I'd like a minimal global footprint (`appInsights`) and scope the rest to a local function. IF, they need to be global for some reason, a consistent prefix should be in place IMHO; `sdkInstance` feels very generic.

MSNev:
An update snippet which address this issue has been checked in with #1258 
This also updated the Readme, so it's now available on the readme page, will close this issue once we publish a full release so that the snippet.js and snippet.min.js are included in the NPM packages.
The snippet change is a lot larger because it's closing a gap that existed by sending an exception event to the server when it detects that the main SDK script could not be downloaded from the CDN, it also has some additional snippet only configuration options.

MSNev:
Included in v 2.5.5 which is now fully deployed

PR #978, opened 2019-07-22
update snippet

xiao-lix:
Related issue: https://github.com/microsoft/ApplicationInsights-JS/issues/974

How did I solve the issue: 
Wrap the snippet in a self-execute function that includes variable `aiName` inside the function scope. Grab appInsights instance if exists, or to create one and write to window object. After running the snippet, there will be only one global variable left in the window: 
- `appInsights` SDK instance

Users can create multiple appInsights instances by giving different names and adding a unique namePrefix for each instance config.


vNext/AISKU/snippet/snippet.js
```diff
@@ -1,96 +1,98 @@
...
 
-// if somebody calls the snippet twice, don't report page view again
-if (aisdk.queue && aisdk.queue.length === 0) {
-     aisdk.trackPageView({});
-}
+    // if somebody calls the snippet twice, don't report page view again
+    if (aisdk.queue && aisdk.queue.length === 0) {
+        aisdk.trackPageView({});
+    }
+}) ();
```
markwolff: nit: end on newline

vNext/AISKU/snippet/snippet.js
```diff
@@ -1,96 +1,98 @@
...
-window[sdkInstance] = "appInsights";
-var aiName = window[sdkInstance]; // provide non default instance name through key appInsightsSDK
-var aisdk = window[aiName] || (function (aiConfig) {
-    var appInsights = {
-        config: aiConfig
-    };
-    appInsights.initialize = true; // initialize sdk on download
+﻿(function() {
+    var sdkInstance = "appInsightsSDK";
+    window[sdkInstance] = "appInsights";
```
markwolff: As per our discussions, lets see if we can have this as a `var` instead of another `window` variable.

xiao-lix: yeah it works fine with one window variable, and the name can be updated to user defined.

vNext/AISKU/snippet/snippet.js
```diff
@@ -1,7 +1,5 @@
 ﻿(function() {
-    var sdkInstance = "appInsightsSDK";
-    window[sdkInstance] = "appInsights";
-    var aiName = window[sdkInstance]; // provide non default instance name through key appInsightsSDK
+    var aiName = "appInsights";
```
markwolff: With this test, can you test that using 2 snippets, each with different `namePrefix`, on the same page works correctly?

xiao-lix: Yes everything works correctly

vNext/AISKU/snippet/snippet.js
```diff
@@ -1,96 +1,96 @@
...
+        var trackEvent = track + method[0];
+        createLazyMethod("start" + trackEvent);
+        createLazyMethod("stop" + trackEvent);
 
-    createLazyMethod("setAuthenticatedUserContext");
-    createLazyMethod("clearAuthenticatedUserContext");
-    createLazyMethod("flush");
+        createLazyMethod("setAuthenticatedUserContext");
+        createLazyMethod("clearAuthenticatedUserContext");
+        createLazyMethod("flush");
```
hectorhdzg: Can you create a lazy method for addTelemetryInitializer as well here? #985

xiao-lix: yeah I can add this in a separate pr.

vNext/AISKU/snippet/snippet.js
```diff
@@ -1,96 +1,96 @@
...
-    // Note: ApplicationInsightsAnalytics is the extension string identifier for
-    //  AppAnalytics. It is defined in ApplicationInsights.ts:ApplicationInsights.identifer
-    if (!(aiConfig.disableExceptionTracking === true ||
-        (aiConfig.extensionConfig &&
-        aiConfig.extensionConfig.ApplicationInsightsAnalytics &&
-        aiConfig.extensionConfig.ApplicationInsightsAnalytics.disableExceptionTracking === true))) {
+        // Collect global errors
+        // Note: ApplicationInsightsAnalytics is the extension string identifier for
+        //  AppAnalytics. It is defined in ApplicationInsights.ts:ApplicationInsights.identifer
+        if (!(aiConfig.disableExceptionTracking === true ||
```
MSNev: Also to get better compression
var extCfg = aiConfig.extensionConfig || {};
var aiAnalytics = extCfg.ApplicationInsightsAnalytics || {};
if (!(aiConfig.disableExceptionTracking === true || aiAnalytics.disableExceptionTracking === true)) {

vNext/AISKU/snippet/snippet.js
```diff
@@ -1,96 +1,96 @@
...
-        config: aiConfig
-    };
-    appInsights.initialize = true; // initialize sdk on download
+﻿(function() {
+    var aiName = "appInsights";
+    var aisdk = window[aiName] || (function (aiConfig) {
+        var appInsights = {
+            config: aiConfig
+        };
+        appInsights.initialize = true; // initialize sdk on download
```
MSNev: Just add this to the declaration above.
var appInsights = {
            initialize: true,
            config: aiConfig
        };

vNext/AISKU/snippet/snippet.js
```diff
@@ -1,96 +1,96 @@
...
-        var scriptElement = localDocument.createElement(scriptText);
-        scriptElement.src = aiConfig.url || "https://az416426.vo.msecnd.net/next/ai.2.min.js";
-        localDocument.getElementsByTagName(scriptText)[0].parentNode.appendChild(scriptElement);
-    });
+        // Assigning these to local variables allows them to be minified to save space:
+        var localDocument = document, localWindow = window, scriptText = "script";
+        setTimeout(function () {
+            var scriptElement = localDocument.createElement(scriptText);
+            scriptElement.src = aiConfig.url || "https://az416426.vo.msecnd.net/next/ai.2.min.js";
+            localDocument.getElementsByTagName(scriptText)[0].parentNode.appendChild(scriptElement);
```
MSNev: Can we change this as I don't believe it really doing what the intent around having the snippet is meant to do.

What I mean by this, is that this line will cause a new script tag to be dropped into the same place in the document as the first script (which should be this snippet AND should also be in the <head> section), this will cause this new script tag to become a blocking script under the following condition.
- If there is at least 1 more inline script after the execution of this snippet that is also in the <head> (i.e. a blocking script -- i.e. the critical path scripts)

We should really be adding this at the end of the document or at the very least (the easiest) within the <body> and not the <head>

markwolff: I think we want to block at least until the lazy methods are set up and the `onerror` handler is patched. If it is moved to the end document and becomes more "async", would we miss a lot of the initial telemetry from the pageload that our current blocking accomplishes? From my experience, users extremely sensitive to performance have either tweaked this snippet to their liking (using importScripts, etc) or made the containing `<script>` tag itself `async`.

MSNev: Spoke with Mark offline, this comment is about how we load the main script and not the snippet, so we wouldn't miss any telemetry etc as the snippet takes care of caching and replaying any reported telemetry between the snippet initialization and the loading of the main script.

MSNev:
I'm dealing with this as part of the snippet updates for reporting script load failures.

PR #980, opened 2019-07-24
SPA trigger trackpageview with refUri when route changes

xiao-lix:


xiao-lix:
reopened another pull request for this

PR #995, opened 2019-08-05
update traceheader name

xiao-lix:


vNext/shared/AppInsightsCommon/src/RequestResponseHeaders.ts
```diff
@@ -26,12 +26,12 @@ export class RequestHeaders {
...
      * W3C distributed tracing protocol header
      */
-    public static traceParentHeader = "traceparentHeader";
+    public static traceParentHeader = "traceparent";
 
     /**
      * W3C distributed tracing protocol state header
      */
-    public static traceStateHeader: "tracestateHeader"; // currently not used
+    public static traceStateHeader: "tracestate"; // currently not used
```
hectorhdzg: Is this wrong as well?

xiao-lix: yes I checked the website here: https://www.w3.org/TR/trace-context/#tracestate-field the header name is "tracestate" but this field is not in use for now.

PR #999, opened 2019-08-13
routes: add ie polyfill for Event constructor

### LINKED ISSUE #998: enableAutoRouteTracking breaks angular router in IE11
To reproduce create a new angular app (ng new my-app), create any component and route to it, and then add application insights. Routing works fine without enableAutoRouteTracking but fails with enableAutoRouteTracking: true with the error "Unhandled Navigation Error:"

Routing works fine with route tracking in Edge, Firefox, and Chrome. This is only an issue in IE11.

Versions:
angular 8.2.2
applicationinsights 2.1.1

markwolff:
Fixes #998 

Adds analytics plugin polyfill for `new Event` so that it can be run in Ie

### DIFF
--- a/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
+++ b/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
@@ -92,18 +92,6 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
         return config;
     }
 
-    private static _createDomEvent(eventName: string): Event {
-        let event: Event = null;
-
-        if (typeof Event === "function") { // Use Event constructor when available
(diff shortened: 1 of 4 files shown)


vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -92,6 +92,19 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
         return config;
     }
 
+    private static _createDomEvent(eventName: string): Event {
```
hectorhdzg: Putting this in a Util class instead would be better

markwolff: Moved to common's `Util`

PR #1005, opened 2019-08-16
Users/lxiao/angular plugin

xiao-lix:


xiao-lix:
with version conflicts

PR #1011, opened 2019-08-21
angular plugin track pageview with router change

xiao-lix:


vNext/extensions/applicationinsights-angular-js/src/AngularPlugin.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+import {
+    IConfig, IPageViewTelemetry, IMetricTelemetry, IAppInsights
+} from "@microsoft/applicationinsights-common";
+import {
+    IPlugin, IConfiguration, IAppInsightsCore,
+    ITelemetryPlugin, CoreUtils, ITelemetryItem,
+    IDiagnosticLogger, _InternalMessageId, LoggingSeverity, ICustomProperties
+} from "@microsoft/applicationinsights-core-js";
```
ElanHasson: I was under the impression `@microsoft/applicationinsights-core-js` is legacy. Is this not accurate?

If not, what is the difference between `@microsoft/applicationinsights-core-js` and `@microsoft/applicationinsights-web` ?

hectorhdzg: @microsoft/applicationinsights-web is our SKU, basically a wrapper of all extensions and core we have available, including analytics, properties, dependencies and AI channel, legacy package is called [applicationinsights-js](https://www.npmjs.com/package/applicationinsights-js)

PR #1012, opened 2019-08-21
update housekeeping flush using beaconSender

xiao-lix:


vNext/AISKULight/index.ts
```diff
@@ -75,13 +75,14 @@ export class ApplicationInsights {
     /**
      * Immediately send all batched telemetry
      * @param {boolean} [async=true]
+     * @param {boolean} [isBeaconSender]
      * @memberof ApplicationInsights
      */
-    public flush(async: boolean = true) {
+    public flush(async: boolean = true, isBeaconSender?: boolean) {
```
markwolff: change to `isBeaconSender = false`

vNext/AISKULight/index.ts
```diff
@@ -75,13 +75,14 @@ export class ApplicationInsights {
     /**
      * Immediately send all batched telemetry
      * @param {boolean} [async=true]
+     * @param {boolean} [isBeaconSender]
```
markwolff: `isBeaconSender=false`

vNext/channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -43,9 +43,12 @@ export class Sender implements IChannelControlsAI {
         throw new Error("Method not implemented.");
     }
 
-    public flush() {
+    public flush(async?: boolean, isBeaconSender?: boolean) {
+        if (isBeaconSender) {
+            this._sender = this._beaconSender;
```
markwolff: This will cause each telemetry item after this flush to be sent with beacons, even if we don't intend to

vNext/channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -307,7 +310,8 @@ export class Sender implements IChannelControlsAI {
      * Immediately send buffered data
      * @param async {boolean} - Indicates if the events should be sent asynchronously
      */
-    public triggerSend(async = true) {
+    public triggerSend(async?: boolean) {
+        if (!async) async = true;
```
markwolff: This will result in async always being true. Use [Typescript default parameters instead](https://www.typescriptlang.org/docs/handbook/functions.html#optional-and-default-parameters)

vNext/shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IChannelControls.ts
```diff
@@ -27,9 +27,10 @@ export interface IChannelControls extends ITelemetryPlugin {
     /**
      * Flush to send data immediately; channel should default to sending data asynchronously
      * @param async: send data asynchronously when true
+     * @param isBeaconSender: send data through beacon when true
      * @param callBack: if specified, notify caller when send is complete 
      */
-    flush(async: boolean, callBack?: () => void): void;
+    flush(async: boolean, isBeaconSender?: boolean, callBack?: () => void): void;
```
markwolff: Verify that this interface isn't used anywhere else here or in other SDKs. 
@hectorhdzg Is this used anywhere that you know of?

vNext/channels/applicationinsights-channel-js/src/Sender.ts
markwolff: Also if the browser does not support beacons, we will need to ignore this parameter

vNext/channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -320,7 +317,11 @@ export class Sender implements IChannelControlsAI {
                     var payload = this._buffer.getItems();
 
                     // invoke send
-                    this._sender(payload, async);
+                    if (isBeaconSender && navigator.sendBeacon(this._config.endpointUrl())) {
```
markwolff: this will crash IE

PR #1023, opened 2019-08-30
udpate pollInternalLogs method of core

xiao-lix:
update this function interface so 1P can pass in a eventName to update the name filed of an item of MessageData type.

### DIFF
--- a/vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
+++ b/vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
@@ -208,7 +208,7 @@ export class AppInsightsCore implements IAppInsightsCore {
     /**
      * Periodically check logger.queue for
      */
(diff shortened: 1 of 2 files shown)


vNext/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -208,7 +208,7 @@ export class AppInsightsCore implements IAppInsightsCore {
     /**
      * Periodically check logger.queue for
      */
-    pollInternalLogs(): number {
+    pollInternalLogs(intervalMS?: number, eventName?: string): number {
```
hectorhdzg: Why adding intervalMS?, I believe config.diagnosticLogInterval is used for this purpose already

xiao-lix: removed

PR #1035, opened 2019-09-04
Update to use performanceNavigationTiming interface for supported bro…

xiao-lix:
…wser

### DIFF
--- a/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
+++ b/vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
@@ -38,13 +38,22 @@ export class PageViewPerformanceManager {
          *  |---network---||---request---|---response---|---dom---|
          *  |--------------------------total----------------------|
          */
+        var navigationTiming = this.getPerformanceNavigationTiming();
         var timing = this.getPerformanceTiming();
-        if (timing) {
-            var total = DateTimeUtils.GetDuration(timing.navigationStart, timing.loadEventEnd);
-            var network = DateTimeUtils.GetDuration(timing.navigationStart, timing.connectEnd);
-            var request = DateTimeUtils.GetDuration(timing.requestStart, timing.responseStart);
-            var response = DateTimeUtils.GetDuration(timing.responseStart, timing.responseEnd);
-            var dom = DateTimeUtils.GetDuration(timing.responseEnd, timing.loadEventEnd);
+        if (navigationTiming || timing) {
+            if (navigationTiming) {
+                var total = navigationTiming.duration;
+                var network = DateTimeUtils.GetDuration(navigationTiming.startTime, navigationTiming.connectEnd);
+                var request = DateTimeUtils.GetDuration(navigationTiming.requestStart, navigationTiming.responseStart);
+                var response = DateTimeUtils.GetDuration(navigationTiming.responseStart, navigationTiming.responseEnd);
+                var dom = DateTimeUtils.GetDuration(navigationTiming.responseEnd, navigationTiming.loadEventEnd);
+            } else {
+                var total = DateTimeUtils.GetDuration(timing.navigationStart, timing.loadEventEnd);
+                var network = DateTimeUtils.GetDuration(timing.navigationStart, timing.connectEnd);
+                var request = DateTimeUtils.GetDuration(timing.requestStart, timing.responseStart);
+                var response = DateTimeUtils.GetDuration(timing.responseStart, timing.responseEnd);
+                var dom = DateTimeUtils.GetDuration(timing.responseEnd, timing.loadEventEnd);
+            }
 
             if (total == 0) {
                 this._logger.throwInternal(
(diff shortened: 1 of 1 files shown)


vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
```diff
@@ -82,13 +91,27 @@ export class PageViewPerformanceManager {
...
+    public getPerformanceTiming(): PerformanceTiming | null {
         if (this.isPerformanceTimingSupported()) {
             return window.performance.timing;
         }
 
         return null;
     }
+    public getPerformanceNavigationTiming(): PerformanceNavigationTiming | null {
+        if (this.isPerformanceNavigationTimingSupported()) {
+            return window.performance.getEntriesByType("navigation")[0] as PerformanceNavigationTiming;
```
markwolff: From what I can tell from the spec, this is always an array with a single element. Is this correct?

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
```diff
@@ -38,13 +38,22 @@ export class PageViewPerformanceManager {
...
          */
+        var navigationTiming = this.getPerformanceNavigationTiming();
         var timing = this.getPerformanceTiming();
-        if (timing) {
-            var total = DateTimeUtils.GetDuration(timing.navigationStart, timing.loadEventEnd);
-            var network = DateTimeUtils.GetDuration(timing.navigationStart, timing.connectEnd);
-            var request = DateTimeUtils.GetDuration(timing.requestStart, timing.responseStart);
-            var response = DateTimeUtils.GetDuration(timing.responseStart, timing.responseEnd);
-            var dom = DateTimeUtils.GetDuration(timing.responseEnd, timing.loadEventEnd);
+        if (navigationTiming || timing) {
```
markwolff: can this line be removed? and replace the `} else {` with an `else if`

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
```diff
@@ -38,13 +38,22 @@ export class PageViewPerformanceManager {
...
+            if (navigationTiming) {
```
hectorhdzg: Please verify this is working correctly in IE and Safari

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
```diff
@@ -38,13 +38,22 @@ export class PageViewPerformanceManager {
          *  |---network---||---request---|---response---|---dom---|
          *  |--------------------------total----------------------|
          */
+        var navigationTiming = this.getPerformanceNavigationTiming();
```
hectorhdzg: Is this value more accurate or why is being added now?

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
xiao-lix: There's chance that neither of these two exists, then it will do nothing. So this check needs to be kept?

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
xiao-lix: `PerformanceTiming` interface is deprecated : 
https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming
So add `PerformanceNavigationTiming` interface for supported browsers: 
https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
xiao-lix: Yes, there is only one `PerformanceNavigationTiming` object in the performance timeline: 
(the first green box): https://www.w3.org/TR/navigation-timing-2/#PerformanceResourceTiming
but there're 4 types of NavigationTiming:
https://www.w3.org/TR/navigation-timing-2/#sec-performance-navigation-types

vNext/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
xiao-lix: Verified working correctly in IE

PR #1061, opened 2019-09-27
bug fix - move sampleRate field out from tag to root

xiao-lix:


AISKU/Tests/TelemetryValidation/CommonValidator.ts
```diff
@@ -1,5 +1,5 @@
 import { ITypeValidator } from './ITypeValidator';
-import { IEnvelope } from '@microsoft/applicationinsights-common';
+import { IEnvelope, SampleRate } from '@microsoft/applicationinsights-common';
```
markwolff: Is this being used in this file?

AISKU/Tests/applicationinsights.e2e.tests.ts
```diff
@@ -82,6 +85,8 @@ export class ApplicationInsightsTests extends TestClass {
             this.successSpy = this.sandbox.spy(sender, '_onSuccess');
             this.loggingSpy = this.sandbox.stub(this._ai['core'].logger, 'throwInternal');
             this.trackSpy = this.sandbox.spy(this._ai['dependencies'], 'trackDependencyDataInternal')
+            this.isSampledInSpy = this.sandbox.stub(sender, '_isSampledIn', () => true);
```
markwolff: I don't think this var is used anywhere, so you can just not assign this to a variable

channels/applicationinsights-channel-js/Tests/Sample.tests.ts
```diff
@@ -31,6 +31,7 @@ export class SampleTests extends TestClass {
 
                 Assert.ok(this.sample.isSampledIn(this.item));
                 Assert.ok(scoreStub.notCalled);
+                Assert.ok(!this.item[SampleRate], "sampleRate field is hidden when the sampling percentage is 100");
```
markwolff: Does this test  fail on the other test cases with a <100 sampling rate?

xiao-lix: Thought about it, it actually doesn't make sense to test it here. SampleRate field will be added in the envelope in "processTelemetry" method of Sender, then serialized off if it equals to 100 in the same function. so removed from here.

PR #1067, opened 2019-10-01
add window check

xiao-lix:


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
```diff
@@ -92,14 +92,14 @@ export class PageViewPerformanceManager {
...
-        if (this.isPerformanceTimingSupported()) {
+        if (this.isPerformanceTimingSupported() && typeof window === "object") {
             return window.performance.timing;
         }
 
         return null;
     }
     public getPerformanceNavigationTiming(): PerformanceNavigationTiming | null {
-        if (this.isPerformanceNavigationTimingSupported()) {
+        if (this.isPerformanceNavigationTimingSupported() && typeof window === "object") {
```
hectorhdzg: isPerformanceNavigationTimingSupported already checks for window

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -243,7 +243,7 @@ export class AjaxMonitor implements ITelemetryPlugin, IDependenciesPlugin, IInst
 
     // Fetch Stuff
     protected instrumentFetch(): void {
-        if (!this.supportsFetch() || this._fetchInitialized) {
+        if (!this.supportsFetch() || this._fetchInitialized || !window) {
```
hectorhdzg: window check is already in supportsFetch method, add typeof window check in there instead

shared/AppInsightsCommon/src/Util.ts
```diff
@@ -213,7 +213,7 @@ export class Util {
     public static getSessionStorageKeys(): string[] {
         const keys = [];
 
-        if (Util.canUseSessionStorage()) {
+        if (Util.canUseSessionStorage() && typeof window === "object") {
```
hectorhdzg: Same here add typeof check inside canUseSessionStorage method

xiao-lix: double checked, window is checked in _getVerifiedStorageObject method that is called in canUseSessionStorage method. So no need to check it again.

PR #1068, opened 2019-10-01
enable w3c by default

xiao-lix:


MSNev:
I'm taking this PR over and will republish an update shortly

Issue #1013, opened 2019-08-22
Consider logging of unhandled promises


nphmuller:
Currently ApplicationInsights-JS doesn't log uncaught errors of promises. To make sure all errors are caught and logged, something like the following snippet could be implemented:
``` js
window.addEventListener("unhandledrejection", function(promiseRejectionEvent) { 
    // handle error here, for example log   
});
```

It looks a lot like implementing `window.onerror`, which is already subscribed upon.

MSNev:
This is also referenced in Issue
Unhandled rejected promise tracking #168
I'm linking and going to close #168 and being fixed by this issue

MSNev:
Will be closing this issue once v 2.4.3 (which includes the option to enbale this) is published the the CDN

MSNev:
Version 2.4.3 is now published to CDN and NPM and addresses this enhancement

PR #1069, opened 2019-10-02
Added support for logging unhandled promise rejections

aaronpowell:
- New global handled for unhandledrejection, similar to onerror
- New config property to test for

Addresses #1013

### DIFF
--- a/AISKU/package.json
+++ b/AISKU/package.json
@@ -44,7 +44,7 @@
         "source-map-loader": "^0.2.3",
         "tslint": "^5.19.0",
         "tslint-config-prettier": "^1.18.0",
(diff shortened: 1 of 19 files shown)


aaronpowell:
Well this is a bit annoying. The TypeScript version needs to be bumped to support this but it looks like that bump breaks other parts of the codebase and I'm struggling with getting it to run locally (I get a whole bunch of different errors ☹️)

aaronpowell:
Yay! I got it building locally and looks like all errors are fixed!

aaronpowell:
Closing PR as I'll create a new one that doesn't require a TypeScript upgrade

PR #1070, opened 2019-10-03
Added support for logging unhandled promise rejections

aaronpowell:
- New global handled for unhandledrejection, similar to onerror
- New config property to test for

Addresses #1013

Replaced #1069

### DIFF
--- a/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
+++ b/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
@@ -593,14 +593,14 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
             !this.config.autoUnhandledPromiseInstrumented) {
             // We want to enable exception auto collection and it has not been done so yet
             const onunhandledrejection = "onunhandledrejection";
-            const originalOnUnhandledRejection = window[onunhandledrejection];
-            window.onunhandledrejection = (error: PromiseRejectionEvent) => {
-                const handled = originalOnUnhandledRejection && (originalOnUnhandledRejection.call(window, error) as any);
+            const originalOnUnhandledRejection = _window[onunhandledrejection];
+            _window.onunhandledrejection = (error: PromiseRejectionEvent) => {
+                const handled = originalOnUnhandledRejection && (originalOnUnhandledRejection.call(_window, error) as any);
                 if (handled !== true) { // handled could be typeof function
                     instance._onerror({
                         message: error.reason.toString(),
                         error: error.reason instanceof Error ? error.reason : new Error(error.reason.toString()),
-                        url: window.location.href,
+                        url: _window.location.href,
                         lineNumber: 0,
                         columnNumber: 0
                     });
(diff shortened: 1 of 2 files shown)


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -576,6 +580,28 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
             this.config.autoExceptionInstrumented = true;
         }
 
+        if (this.config.disableExceptionTracking === false &&
+            !this.config.autoUnhandledPromiseInstrumented) {
+            // We want to enable exception auto collection and it has not been done so yet
+            const onunhandledrejection = "onunhandledrejection";
+            const originalOnUnhandledRejection = window[onunhandledrejection];
```
MSNev: As part of my unload PR, I've added a getWindow() helper method which you will need to use as the SDK can be run in an environment without window, document, navigator etc defined. And this causes a ReferenceError exception to be thrown when accessing window in one of these environments.

Hopefully, the react issue and my PR can be checked in shortly so we can pick up your change as well.

MSNev: My Unload and EnvUtils (hasWindow(), getWindow()) is now checked in and I've disabled the failing reactivenative tests

aaronpowell: Shall I merge in that branch and use the `getWindow` function?

MSNev: If you rebase to master you will get the methods and then yes, if you use getWindow() (Checking that it doesn't return null and/or use hasWindow() before hooking the events) then this looks good.

aaronpowell: Updated the PR with a rebase of `master` and using the new method (see https://github.com/microsoft/ApplicationInsights-JS/pull/1070/commits/4788af78d584c90087ab1d6ad910702af2b2d344)

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -585,6 +589,28 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
             this.config.autoExceptionInstrumented = true;
         }
 
+        if (this.config.disableExceptionTracking === false &&
+            !this.config.autoUnhandledPromiseInstrumented) {
+            // We want to enable exception auto collection and it has not been done so yet
+            const onunhandledrejection = "onunhandledrejection";
+            const originalOnUnhandledRejection = _window[onunhandledrejection];
+            _window.onunhandledrejection = (error: PromiseRejectionEvent) => {
```
MSNev: _window may be null (in an electron app), please add a "&& _window" check to the if statement on lines 592-593

aaronpowell: Done, and I also rebased `master`.

PR #1088, opened 2019-10-14
bug fix - Session cookie with expire=0 does not expire after the the …

xiao-lix:
…browser closes


shared/AppInsightsCommon/src/Util.ts
```diff
@@ -328,6 +328,7 @@ export class Util {
      * helper method to set userId and sessionId cookie
      */
     public static setCookie(logger: IDiagnosticLogger, name, value, domain?) {
+        value = value + ";SameSite=none";
```
markwolff:
```suggestion
        value = value + ";SameSite=None";
```

Issue #1094, opened 2019-10-23
[BUG] Server Side Rendering fails when using Gatsby


aaronpowell:
**Description/Screenshot**

I'm using [Gatsby](https://gatsbyjs.org) to create a static website but when I attempted to generate the static HTML files from the React components the Server Side Rendering (SSR) fails in AppInsights:

```
[...]

PR #1096, opened 2019-10-24
Issue 1094

aaronpowell:
This addresses #1094 by doing `typeof` tests against globals to ensure they exist.


extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -322,7 +322,9 @@ export class AjaxMonitor implements ITelemetryPlugin, IDependenciesPlugin, IInst
     /// <returns>True if Ajax monitoring is supported on this page, otherwise false</returns>
     private supportsAjaxMonitoring(): boolean {
         let result = true;
-        if (CoreUtils.isNullOrUndefined(XMLHttpRequest) ||
+
+        if (typeof XMLHttpRequest !== 'function' ||
```
markwolff: The ajax tests are failing because XMLHttpRequest is not a function for some versions of IE
```suggestion
        if (typeof XMLHttpRequest === 'undefined' ||
```

legacy/JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -23,7 +23,7 @@ module Microsoft.ApplicationInsights {
         private currentWindowHost;
 
         constructor(appInsights: Microsoft.ApplicationInsights.AppInsights) {
-            this.currentWindowHost = window.location.host && window.location.host.toLowerCase();
+            this.currentWindowHost = typeof window == 'object' ? window.location.host && window.location.host.toLowerCase() : undefined;
```
markwolff:
```suggestion
            this.currentWindowHost = typeof window === 'object' ? window.location.host && window.location.host.toLowerCase() : undefined;
```

aaronpowell:
Thanks @markwolff, both those changes are actioned. Didn't realise xhr wasn't always detected as `function`, was it `object`? (I'm just curious)

And I can't believed I missed a `===` 🤦‍♂

markwolff:
@aaronpowell Yes I believe it was reported as an `object` for IE9 (and 10?), even though we can patch it in the same way

Issue #1075, opened 2019-10-04
Enable and maintain compatibility for older browsers including IE8


MSNev:
**Is your feature request related to a problem? Please describe.**
As an SDK there are numerous users which cannot control the browsers that customers use. As such we need to ensure that the SDK continues to work and does not break the JS execution when loaded in an older browser.

**Describe the solution you'd like**
As part of this we need to have the generated be **compatible** with ES3 and use either polyfil implementations or striped back emulation of any required ES5/6 supported features.

**Describe alternatives you've considered**
Beyond terminating support for older browsers (which we can't do at this point) -- There are not alternatives

**Additional context**
While it would be ideal to just not support IE8 and older generation (non ES5) browsers there are numerous large customers that continue to require pages to "work". 
This does NOT mean that we will support ALL features, just that we will include changes to not cause any application using this API to fail.

MSNev:
Change commited

PR #1102, opened 2019-10-31
Enable and maintain compatibility for older browsers including IE8 #1075

MSNev:
Is your feature request related to a problem? Please describe.
As an SDK there are numerous users which cannot control the browsers that customers use. As such we need to ensure that the SDK continues to work and does not break the JS execution when loaded in an older browser.

Describe the solution you'd like
As part of this we need to have the generated be compatible with ES3 and use either polyfil implementations or striped back emulation of any required ES5/6 supported features.

### DIFF
--- a/AISKU/tsconfig.json
+++ b/AISKU/tsconfig.json
@@ -4,7 +4,7 @@
     "inlineSources": true,
     "noImplicitAny": false,
     "module": "es6",
-    "target": "es5",
+    "target": "es3",
     "moduleResolution": "Node",
     "alwaysStrict": true,
     "suppressImplicitAnyIndexErrors": true,
--- a/README.md
+++ b/README.md
@@ -231,7 +231,8 @@ This version comes with the bare minimum amount of features and functionalities 
 ## Upgrading from the old Version of Application Insights
 
 Breaking changes in the SDK V2 version:
-- To allow for better API signatures, some of the apis such as trackPageView, trackException have been updated. Running in IE8 or lower versions of the browser is not supported.
+- To allow for better API signatures, some of the apis such as trackPageView, trackException have been updated.
+- ES3 (IE8) compatibility, while running in IE8 or lower versions of the browser is not an officially supported scenario we are working to maintain ES3 level compatibility to ensure that the SDK will not cause any unexpected failures due to Javascript parsing error. See [ES3/IE8 Compatibility](#es3ie8-compatibility) below for further information.
 - Telemetry envelope has field name and structure changes due to data schema updates.
 - Moved `context.operation` to `context.telemetryTrace`. Some fields were also changed (`operation.id` --> `telemetryTrace.traceID`)
   - If you want to maunally refresh the current pageview id (e.g. in SPA apps) this can be done with `appInsights.properties.context.telemetryTrace.traceID = Util.newId()`
@@ -324,7 +325,7 @@ While the script downloads from the CDN, all tracking of your page is queued. On
 
 ![Chrome](https://raw.githubusercontent.com/alrra/browser-logos/master/src/chrome/chrome_48x48.png) | ![Firefox](https://raw.githubusercontent.com/alrra/browser-logos/master/src/firefox/firefox_48x48.png) | ![IE](https://raw.githubusercontent.com/alrra/browser-logos/master/src/edge/edge_48x48.png) | ![Opera](https://raw.githubusercontent.com/alrra/browser-logos/master/src/opera/opera_48x48.png) | ![Safari](https://raw.githubusercontent.com/alrra/browser-logos/master/src/safari/safari_48x48.png)
 --- | --- | --- | --- | --- |
-Latest ✔ | Latest ✔ | 9+ ✔ | Latest ✔ | Latest ✔ |
+Latest ✔ | Latest ✔ | 9+ Full ✔<br>8- Compatible | Latest ✔ | Latest ✔ |
 
 ## Contributing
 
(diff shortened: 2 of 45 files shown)


legacy/JavaScript/JavaScriptSDK.Module/AppInsightsModule.ts
```diff
@@ -64,7 +64,7 @@ class AppInsightsModule {
     }
 
     private static _download(aiConfig: Microsoft.ApplicationInsights.IConfig) {
-        AppInsightsModule.appInsightsInstance.config = aiConfig;
```
MSNev: While the get accessor will exist for an ES5+ environment we can't use it here as we *could* be running in an ES3 version where the get accessor won't exist.

shared/AppInsightsCommon/src/Util.ts
```diff
@@ -542,7 +517,7 @@ export class Util {
      * Returns string representation of an object suitable for diagnostics logging.
      */
     public static dump(object: any): string {
-        const objectTypeDump: string = Object.prototype.toString.call(object);
+        const objectTypeDump: string = Object[prototype].toString.call(object);
```
MSNev: Using index lookup (array) to support better minification of the resulting JS code.

shared/AppInsightsCore/src/JavaScriptSDK/ChannelController.ts
```diff
@@ -91,10 +91,19 @@ export class ChannelController implements ITelemetryPlugin {
...
+            CoreUtils.arrForEach(arr, queueItem => queueItem.initialize(config, core, extensions));
 
             this.channelQueue.push(arr);
         }
     }
+
+    /**
+     * Static constructor, attempt to create accessors
+     */
+    // tslint:disable-next-line
```
MSNev: Disabling the tslint rules which don't like having a private, static, variables after declaring public members as we need for this to be the last definition as we NEED the ChannelController.prototype.getChannelControls has been defined.

shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
```diff
@@ -2,18 +2,27 @@
 // Licensed under the MIT License.
 "use strict";
 
+// Added to help with minfication
+let prototype = "prototype";
```
MSNev: Added to support better code minification by using index lookup to access the XXXX.prototype.

AISKU/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "inlineSources": true,
     "noImplicitAny": false,
     "module": "es6",
-    "target": "es5",
+    "target": "es3",
```
MSNev: NOTE: I've only changed the production code to be generated as ES3. All of the test projects can and should remain as ES5+

shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
```diff
@@ -2,18 +2,27 @@
 // Licensed under the MIT License.
 "use strict";
```
markwolff: Just curious, for all of the helper functions in here, do you think there would be any benefit to adding a compat check so that it calls the native version if it is available? Would this provide any perf benefit?

legacy/JavaScript/JavaScriptSDK.Module/AppInsightsModule.ts
```diff
@@ -64,7 +64,7 @@ class AppInsightsModule {
...
+        AppInsightsModule.getAppInsightsInstance().config = aiConfig;
```
markwolff: We probably won't ever deploy the stuff in the legacy folder again so I don't think we need these changes for the v1 SDK

shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
MSNev: I did consider this and decided against it for 2 reasons
1) It adds extra code to the overall package
2) See the Note on the polyfils: Note: For consistency this will not use the Array.prototype.xxxx implementation if it exists as this would
     * cause a testing requirement to test with and without the implementations

README.md
```diff
@@ -348,3 +349,52 @@ git add <...your changes and rush change file...>
...
+- track()
+- trackException()
+
+Beyond terminating support for older browsers that only support ES3, (which we cannot do at this point in time) we will endeavour to maintain support for the above two (2) use-cases.
+
+### ES3/IE8 Features, Solutions, Workarounds and Polyfil style helper functions
+As part of contributing to the project the following table highlights all of the currently known issues and the available solution/workaround. During PR and reviewing please ensure that you do not use the unsupported feature directly and instead use (or provide) the helper, soultion or workaround.
+
+This table does not attempt to include ALL of the ES3 unsuported features, just the currently known functions that where being used at the time or writing. You are welcome to contribute to provide additional helpers, workarounds or documentation of values that should not be used.
```
xiao-lix: The table below seems not correctly formatted, it appears as a big paragraph and still shows `| --- |` these marks in preview.

MSNev: Looks like I need an extra blank line for GitHub to render this correctly.

legacy/JavaScript/JavaScriptSDK.Module/AppInsightsModule.ts
MSNev: As legacy doesn't have it's own tsconfig.json, I needed to change this.
While I *could* put a tsconfig.json here, it's (I think) safer to fix this class -- just in case -- we need to publish the older version again.

MSNev: As discussed offline added a comment stating that this may not be published and therefore users should check their deployed version

shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
```diff
@@ -26,8 +35,231 @@ export class CoreUtils {
         });
     }
 
+    /**
+     * Convert a date to I.S.O. format in IE8
+     */
+    public static toISOString(date: Date) {
+        if (CoreUtils.isDate(date)) {
+            if (Date[prototype].toISOString) {
```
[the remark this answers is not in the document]
MSNev: Will fix

Issue #1095, opened 2019-10-24
Add an Error Boundary to the React plugin


aaronpowell:
React has introduced [Error Boundaries](https://reactjs.org/docs/error-boundaries.html) as a way to handle component render failure in a clean way, rather than that the component tree crashing.

Part of their guidance of using an Error Boundary is to log errors to a monitoring service, such as AppInsights.

Creating a component that can do this is pretty simple, I've [written a blog post on how to do it](https://www.aaron-powell.com/posts/2019-10-24-using-react-error-boundaries-with-appinsights/), but it'd nice if this was provided as part of the React extension.

AndrewCraswell:
I second this. It was one of the first things I realized that was missing when examining the React Plugin.

aaronpowell:
> I second this. It was one of the first things I realized that was missing when examining the React Plugin.

Did you have a look at the approach in my blog post? Is that how you'd expect to use it? I'm tempted to submit it as a PR, but I am unsure if it's how others would want to use it.

hiraldesai:
Can you please advise when these changes are planned for a release? I'd like to use it in one of my projects. 

BTW - I was one of the contributors on [the old React AI plugin](https://github.com/Azure/react-appinsights) which was deprecated in favor of this one. I am so happy how matured it has become over time! Thank you for the great work @aaronpowell and team!

markwolff:
Closing this since it should be available in the next release

MSNev:
Reopening for release tracking -- will close once the release is fully deployed

MSNev:
Release is now fully deployed

PR #1111, opened 2019-11-08
Creating React error boundary plugin

aaronpowell:
This is to address issue #1095.

Added a new component to the React plugin and exported it from the entry point for the extension.


burkeholland:
ping @markwolff, @hectorhdzg and @MSNev

markwolff:
@MSNev This is ready to be merged in if you think it looks okay

Issue #1080, opened 2019-10-07
[Enhancement] AI is not catching all "Unload" dom events


MSNev:
**Description/Screenshot**
The Beforeunload is not the only navigation away event that should be caught the list of events should include
- beforeunload
- unload
- pagehide

It should also be attached to the document as not all browsers (mostly older ones) support the events on window.

egrendonDev:
Hi MSVen,

Have you made any progress on this issue? We are also experiencing similar issues on Desktop Chrome 78. Users appears to randomly have missing application insight events. The options have been set to 100% sampling. And we are are calling "appInsightsFlush()"  on every browser "beforeunload" event.


`	public initAppInsightsAnalytics(): void {
		if (!this.isAppInsightsInitialized) {
			AppInsights.downloadAndSetup({
				instrumentationKey: this.appInsightsKey,
				samplingPercentage: 100  // 100% of all telemetry will be sent to Application Insights
			});

			this.isAppInsightsInitialized = true;
		}
	}`


`	/**
	 * Method will initialize the session expired subscription monitoring
	 */
	private initBrowserWindowClosedSubscribe() {
		this.windowProvider.browserWindowClosed$
			.pipe(takeUntil(this.subscriptionKiller))
			.subscribe((event: any) => {
					//console.log('Attempt to flush before we close browser window');
					//Attempt to flush AppInsights before we close browser window
					this.appInsights.appInsightsFlush();
				}
			);
	}`


Thanks,
Ernesto

MSNev:
Not yet, as I've been working on the IE8 compatability item, I suspect I should get to this by the end of november (not withstanding the holidays).
The change itself is not that large, I just have a few other high priority items on my work list before I get to this one. The main one that we are missing is the pagehide event (https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event) which is when the user hits the browser back button.

MSNev:
Hi Ernesto,

PR created to address this enhancement

Cheers
Nev

egrendonDev:
@MSNev - This is great news! ETA on a new possible release?

MSNev:
The change should get checked in this week, but I don't have a timeline for when we would push a new release yet. That's probably still the end of Nov unless we end up with some sort of High Priority issue that needs addressing/publishing.

MSNev:
PR checked in and fixed should be included in v2.3.2 or later

PR #1113, opened 2019-11-08
[Enhancement] AI is not catching all "Unload" dom events #1080

MSNev:


AISKU/src/Initialization.ts
```diff
@@ -12,6 +12,29 @@ import { AjaxPlugin as DependenciesPlugin, IDependenciesPlugin } from '@microsof
 
 "use strict";
 
+/**
+ * Trys to add an event handler for the specified event to the window, body and document
+ * @param eventName {string} - The name of the event
+ * @param callback {any} - The callback function that needs to be executed for the given event
+ * @return {boolean} - true if the handler was successfully added
+ */
+function addEventHandler(eventName: string, callback: any): boolean {
```
hectorhdzg: Move this to Core Utils, if we really want to attach to both window and window["body"] everywhere

MSNev: For the normal window events then yes we would, but this is not the case for all events...

hectorhdzg: We do have multiple scenarios for window events, having this in Core instead of SKU code would better so it can be reused

MSNev: Moved to CoreUtils

AISKU/src/Initialization.ts
```diff
@@ -365,17 +366,30 @@ export class Initialization implements IApplicationInsights {
...
-                appInsightsInstance.appInsights.core.logger.throwInternal(
-                    LoggingSeverity.CRITICAL,
-                    _InternalMessageId.FailedToAddHandlerForOnBeforeUnload,
-                    'Could not add handler for beforeunload');
+            if (!appInsightsInstance.appInsights.config.disableFlushOnBeforeUnload) {
+                // Hook the unload event for the document, window and body to ensure that the client events are flushed to the server
+                // As just hooking the window does not always fire (on chrome) for page navigations.
+                let added = CoreUtils.addEventHandler('beforeunload', performHousekeeping);
+                added = CoreUtils.addEventHandler('pagehide', performHousekeeping) || added;
+                if (!added) {
```
markwolff: If pagehide fails and beforeunload succeeds would that be considered an error?

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -84,10 +83,12 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
     private _trackAjaxAttempts: number = 0;
 
     // array with max length of 2 that store current url and previous url for SPA page route change trackPageview use.
-    private _prevUri: string = typeof window === "object" && window.location && window.location.href || "";
+    private _prevUri: string;       // Assigned in the constructor
```
markwolff: nit: whitespace

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -257,8 +258,9 @@ export class ApplicationInsights implements IAppInsights, ITelemetryPlugin, IApp
      * @param systemProperties System level properties (Part A) that a user can add to the telemetry item
      */
     public sendPageViewInternal(pageView: IPageViewTelemetryInternal, properties?: { [key: string]: any }, systemProperties?: { [key: string]: any }) {
-        if (typeof document !== "undefined") {
-            pageView.refUri = pageView.refUri === undefined ? document.referrer : pageView.refUri;
+        let _document = getDocument();
+        if (_document) {
```
markwolff: Will this need an explicit undefined check still? Or does using the utility function prevent crashes for where document doesn't exist?

shared/AppInsightsCommon/src/Telemetry/Common/DataSanitizer.ts
```diff
@@ -134,7 +134,7 @@ export class DataSanitizer {
             const tempProps = {};
             for (let prop in properties) {
                 let value = properties[prop];
-                if (typeof value === "object" && typeof JSON !== "undefined") {
+                if (CoreUtils.isObject(value) && !CoreUtils.isUndefined(JSON)) {
```
markwolff: Will using `JSON` in this way crash where it does not exist? e.g. need some sort of `_getJSON` utility, or just keep the typeof check here.

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -0,0 +1,113 @@
...
+export function hasWindow(): boolean {
+    return Boolean(typeof window === strObject && window);
+}
+
+/**
+ * Returns the global window object if it is present otherwise null.
+ * This helper is used to access the window object without causing an exception
+ * "Uncaught ReferenceError: window is not defined"
+ */
+export function getWindow(): Window|null {
```
markwolff: nit: `Window | null`, same for rest in this file

AISKU/src/Initialization.ts
MSNev: No, we just need to attach to at least one of the events (i.e. whatever the browser supports) so that we ensure that the housekeeping is performed.

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
MSNev: The utility takes care of the undefined check and explicitly returns null if document is undefined or already null

AISKU/src/Initialization.ts
markwolff: Got it. So only 1 of these will be supported in each browser?

shared/AppInsightsCommon/src/Telemetry/Common/DataSanitizer.ts
MSNev: Yes, we should have a helper for this of continue to use the typeof JSON -- nice catch

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
MSNev: I seem to have this for all of the items that return an object?

MSNev: Ah -- spacing...

AISKU/src/Initialization.ts
MSNev: Not exactly, some browsers do support both, but not all browsers.
e.g. 
pagehide == firefox and older IOS/MAC (< v12 - does not support beforeunload)
https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event

beforeunload == all modern and some older browser
https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

Registering both is fine

markwolff: If both are registered then would the event handlers be called twice? It is fine as is since it is only calling flush currently, but if we were send an event in there in the future, would it be duped?

MSNev: Yes, this is something that we would need to deal with, but it's that case that we would always have to deal with it to ensure that we capture all events...
We *probably* should also be listening to the "unload" event -- just in case as well (we do in identity) but looking at the support matrix, all browsers that support unload also support beforeunload -- so it *should* be safe.

Issue #991, opened 2019-08-02
Don't work with React HOOKS


ecatugy:
Please send suport for React Hooks, components class is obsolete.

veerprabakar:
Please support the function components with hooks.

enterpriseSoftware:
Yes I would suggest a hooks implementation ASAP for react since that is the direction it's headed.

khowling:
wrote a custom hook to replace withAITracking, seams to work ok, i have a 'key' paramter because i have a tabbed UI in the component i am tracking, and the key is the tab state

function useAITracking(componentName, key) {
  useEffect( () => {
    let start= new Date()
    return () => {
      reactPlugin.trackMetric({
        average: (new Date() - start)/1000,
        name: 'React Component (seconds)',
        sampleCount: 1
      }, { 'Component Name': componentName, 'Navigation': key });
    };
  }, [componentName, key])

}

aaronpowell:
I've done a write up on how you can create custom hooks to work with AppInsights, you'll find it here: https://dev.to/aaronpowell/combining-react-hooks-with-appinsights-5692

In it there is a Context, allowing you to `useContext` to access the React plugin, a hook for tracking a normal event (using `trackEvent` for AppInsights) and one for doing metric tracking like the HOC does.

Does it look like what people would expect from a programmatic model of AppInsights + Hooks?

MSNev:
Included in v 2.5.5 which is now fully deployed

fab7jj:
@aaronpowell @MSNev is there any official documentation on how to use the new stuff for react hooks?

las3r:
I second this question. Is there any documentation?

aaronpowell:
Docs have been written and are currently being reviewed by our content team. Fingers crossed they should be available soon.

aaronpowell:
Docs are now published - https://docs.microsoft.com/en-us/azure/azure-monitor/app/javascript-react-plugin#using-react-hooks

las3r:
thanks for this!

dps88:
@aaronpowell It would be great to have those docs in the [react Readme.md](https://github.com/microsoft/ApplicationInsights-JS/tree/master/extensions/applicationinsights-react-js) as well

aaronpowell:
Good suggestion @dps88. I've opened a PR to add it.

HJordan35:
[...]

PR #1120, opened 2019-11-19
Introducing React Hooks for AppInsights

aaronpowell:
_Targeting issue #991._

This PR introduces some new features to the React (DOM) package, custom hooks that can be used to track an event or metrics on a component (metric tracking is similar to the HOC `withAITracking`).

It is based off the [blog post I wrote](https://dev.to/aaronpowell/combining-react-hooks-with-appinsights-5692) on the subject of Hooks + AppInsights.


aaronpowell:
Presently this PR will be blocked as the version of TypeScript in use, 2.5.3, does not support [Conditional Types](http://www.typescriptlang.org/docs/handbook/advanced-types.html#conditional-types) (they were added in 2.8).

This was uncovered as I had to update the `@types/react` packages in use because it was using type definitions from pre-hooks React, but the type definitions use Conditional Types.

extensions/applicationinsights-react-js/src/useTrackEvent.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @copyright Microsoft 2019
+ */
+import { useState, useEffect, useRef } from "react";
+import ReactPlugin from "./ReactPlugin";
+
+export default function useCustomEvent<T>(
+  reactPlugin: ReactPlugin,
+  eventName: string,
+  eventData: T,
+  skipFirstRun = true
```
xiao-lix: @aaronpowell do you mind explaining why this `skipFirstRun` flag is needed here? thanks.

aaronpowell: @xiao-lix The reason is to mimic more closely the way the non-hook version works. With `useEffect` hooks the effect is triggered on each value update _including_ the initial setting of the value, meaning we would fire off too early and potentially have unwanted events tracked. I explain it more in the post I wrote: https://www.aaron-powell.com/posts/2019-11-19-combining-react-hooks-with-appinsights/#ignoring-unwanted-effect-runs

bschlintz:
Is there any update on this PR @xiao-lix? I have a need for this functionality in a project I'm working on with a customer.

extensions/applicationinsights-react-js/src/ReactPlugin.ts
```diff
@@ -20,6 +20,7 @@ export default class ReactPlugin extends BaseTelemetryPlugin {
 
     private _analyticsPlugin: IAppInsights;
     private _extensionConfig: IReactExtensionConfig;
+    private _logger: IDiagnosticLogger;
```
MSNev, line 23: You don't need to do this.

It's already exposed as
this.diagLog()

extensions/applicationinsights-react-js/src/ReactPlugin.ts
```diff
@@ -69,6 +71,15 @@ export default class ReactPlugin extends BaseTelemetryPlugin {
         }
     }
 
+    trackEvent(event: IEventTelemetry, customProperties?: ICustomProperties) {
+        if (this._analyticsPlugin) {
+            this._analyticsPlugin.trackEvent(event, customProperties);
+        } else {
+            this._logger.throwInternal(
```
MSNev, line 78: So this then becomes

this.diagLog().throwInternal(...)

MSNev: I'm not seeing any other blocking issues apart from the intro of the _logger.
For references the diaLog() takes care of the this getting called before initialization() as well.

Issue #1132, opened 2019-12-06
Enable support for reusing plugins in multiple instances of AppInsights


MSNev:
Support to enable reusing the same plugin in different instances of AppInsights, owned by the same application (but using different Instrumentation keys) is required so that they can reuse a single Plugin instance for all instances.

Currently, while each AppInsights instance allows passing in extension plugins, internally this actually calls "initialize()" on each plugin for every instance, even if it's already been initialized.

Support for this requires 2 changes
- Add an optional isInitialized() method on the plugin interface so that a plugin is only initialized once
- Add an extra parameter to the processTelemetry() method so that singleton (shared) style plugins can access the current AppInsights instance/config without the need to save this during the initialize() call.

MSNev:
Code is committed, will close the issue after we publish and official build (non-beta) with this change

MSNev:
Version 2.4.3 is now published to CDN and NPM and addresses this enhancement

PR #1133, opened 2019-12-06
Enable support for reusing plugins in multiple instances of AppInsights #1132

MSNev:
Latest version only address linting failures


AISKU/Tests/ApplicationInsightsDeprecatedTests.ts
```diff
@@ -150,7 +150,12 @@ export class ApplicationInsightsDeprecatedTests extends TestClass {
         if(this.successSpy.called) {
             let currentCount: number = 0;
             this.successSpy.args.forEach(call => {
-                currentCount += call[1];
+                call[0].forEach(message => {
+                    // Ignore the internal SendBrowserInfoOnUserInit message (Only occurs when running tests in a browser)
```
MSNev: As comment says, these tests where failing when running in a browser but passed fine from the command line -- this fixes that issue.

channels/applicationinsights-channel-js/Tests/Sender.tests.ts
```diff
@@ -60,7 +60,7 @@ export class SenderTests extends TestClass {
                     baseData: {}
                 };
                 try {
-                    this._sender.processTelemetry(telemetryItem);
+                    this._sender.processTelemetry(telemetryItem, null);
```
MSNev: The context is optional as existing plugin implementation may not pass it, so null is a valid value.
And it's really only required once we have a plugin that uses this value (i.e. a Singleton/shared plugin implementation)

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -150,12 +150,8 @@ export class Sender implements IChannelControlsAI {
      */
     private _timeoutHandle: any;
 
-    private _nextPlugin: ITelemetryPlugin;
```
MSNev: These will now be hidden, so any plugin which was relying on these internal fields WILL be broken.
I've added an accessor for the _logger -> log(), but the only *real* usage of the _nextPlugin *should* have been to check and call processTelemetry() which now has a helper processNext()

extensions/applicationinsights-react-js/test/ReactAI.test.ts
```diff
@@ -91,7 +91,7 @@ describe("ReactAI", () => {
     };
     core.initialize(config, [reactPlugin, channel]);
     // Mock logger
-    const loggerMock = reactPlugin["_logger"].throwInternal = jest.fn();
+    const loggerMock = core["logger"].throwInternal = jest.fn();
```
xiao-lix: Can you please explain why logger is update to core["logger"], shouldn't it be reactPlugin["log()"] or something? Thanks

MSNev: the syntax core["xxx"] is an index lookup for the property and could also be writter as core.xxx.
This can't be changed to "log()" as that is not a valid property name (core["log"] would be the correct syntax -- but that would return a reference to the function and not the property).

As internally the log() function uses the "core.logger" property and this unit test is adding a mock implementation of the throwInternal function, this needs to be as defined. So that when the log() function is called it will correctly call the mocked logger implementation.

This change was required as I'm starting to slowly hide the internal properties to provide better minification support.

shared/AppInsightsCore/src/JavaScriptSDK/TelemetryPluginProxy.ts
```diff
@@ -0,0 +1,114 @@
...
+                        // So rather than leave the pipeline dead in the water we call the next plugin
+                        _nextPlugin.processTelemetry(env, itemCtx);
+                    }
+                }
+            } else if (_nextPlugin) {
+                _hasRun = true;
+
+                // The underlying plugin is either not defined or does not have a processTelemetry implementation
+                // so we still want the next plugin to be executed.
+                _nextPlugin.processTelemetry(env, itemCtx);
```
MSNev: This case actually fixes an existing issue where if you added a new IPlugin which didn't implement the processTelemetry() between (based on the priority) 2 other ITelemetryPlugin instances the trailing chain would not get called and it would actually caused an exception.

shared/AppInsightsCore/src/JavaScriptSDK/TelemetryPluginProxy.ts
```diff
@@ -0,0 +1,114 @@
...
+                        // Backward compatibility setting the next plugin on the instance
+                        plugin.setNextPlugin(_nextPlugin);
+                    }
+
+                    plugin.processTelemetry(env, itemCtx);
+                } catch (error) {
+                    if (!_nextPlugin || !_nextPlugin.hasRun()) {
+                        // Either we have no next plugin or the current one did not attempt to call the next plugin
+                        // Which means the current one is the root of the failure so log/report this failure
+                        itemCtx.log().throwInternal(
```
MSNev: Try and log details about the actual plugin that is causing the issue

shared/AppInsightsCore/src/JavaScriptSDK/TelemetryPluginProxy.ts
```diff
@@ -0,0 +1,114 @@
...
+                            LoggingSeverity.CRITICAL,
+                            _InternalMessageId.PluginException,
+                            "Plugin [" + plugin.identifier + "] failed during processTelemetry - " + error);
+                    }
+
+                    if (_nextPlugin && !_nextPlugin.hasRun()) {
+                        // As part of the failure the current plugin did not attempt to call the next plugin in the cahin
+                        // So rather than leave the pipeline dead in the water we call the next plugin
+                        _nextPlugin.processTelemetry(env, itemCtx);
```
MSNev: Calling the next plugin (if present) so that the pipeline isn't dead and therefore the throwInternal() can still succeed.

shared/AppInsightsCore/src/JavaScriptSDK/BaseTelemetryPlugin.ts
```diff
@@ -0,0 +1,129 @@
...
+            // Does nothing as plugin sub-classes of this base class should be using 
+            // processNext() function for execution or the passed ITelemetryItemContext.getNext().getPlugin()
+            // if direct access to the next plugin is required. Note getNext() may return null.
+        };
+
+        _self.processNext = (env: ITelemetryItem, itemCtx: ITelemetryItemContext) => {
+            _self._getTelCtx(itemCtx).processNext(env);
+        };
+
+        _self._getTelCtx = (overrideCtx:ITelemetryItemContext = null) => {
```
MSNev: This needs work :-(
For proper backward compatibility I need to record and process the instance level nextPlugin chain when no telemetryItemContext is available.

shared/AppInsightsCore/src/JavaScriptSDK/BaseTelemetryPlugin.ts
```diff
@@ -0,0 +1,129 @@
...
+
+        _self.isInitialized = () => {
+            return _isinitialized;
+        }
+
+        _self.setInitialized = (isInitialized: boolean):void => {
+            _isinitialized = isInitialized;
+        }
+
+        _self.setNextPlugin = (next: ITelemetryPlugin) => {
```
MSNev: For proper backward compatibility, I need to record the instance level value as some plugins are relying on this and processing telemetry items outside of the currently implemented process loop.

shared/AppInsightsCore/src/JavaScriptSDK/TelemetryItemContext.ts
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        _self.setNext = (nextPlugin:ITelemetryPluginProxy) => {
+            _nextProxy = nextPlugin;
+        };
+
+        _self.processNext = (env: ITelemetryItem) => {
+            let nextPlugin = _nextProxy;
+
+            // Avoid double executing the same plugin
```
MSNev: Due to backward compatibility, we can't add this double execution check :-(

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/ITelemetryItemContext.ts
```diff
@@ -0,0 +1,72 @@
...
+import { IConfiguration } from './IConfiguration';
+import { ITelemetryItem } from './ITelemetryItem';
+import { IPlugin } from './ITelemetryPlugin';
+import { ITelemetryPluginProxy } from './ITelemetryPluginProxy';
+
+/**
+ * The current context for the current call to processTelemetry(), used to support sharing the same plugin instance
+ * between multiple AppInsights instances
+ */
+export interface ITelemetryItemContext {
```
MSNev: Renaming this to IProcessTelemetryContext to be more explicit about it's usage

extensions/applicationinsights-angular-js/src/AngularPlugin.ts
```diff
@@ -8,44 +8,39 @@ import {
...
 
-    initialize(config: IConfiguration & IConfig, core: IAppInsightsCore, extensions: IPlugin[]) {
-        this._extensionConfig =
-            config.extensionConfig && config.extensionConfig[this.identifier]
-                ? (config.extensionConfig[this.identifier] as IAngularExtensionConfig)
-                : { router: null };
-        this._logger = core.logger;
+    initialize(config: IConfiguration & IConfig, core: IAppInsightsCore, extensions: IPlugin[], pluginChain?:ITelemetryPluginChain) {
+        super.initialize(config, core, extensions, pluginChain);
+        let ctx = this._getTelCtx();
```
markwolff: Even though it's private, is there another name we can use other than "context"? This word already has meaning through the SDK. Maybe "chain" or "origin"? Not a blocker, since it can be changed in a different version

channels/applicationinsights-channel-js/Tests/Sender.tests.ts
markwolff: is including the `, null` necessary?

shared/AppInsightsCore/src/JavaScriptSDK/BaseTelemetryPlugin.ts
```diff
@@ -0,0 +1,164 @@
...
+     * returns that value (helps with minification for callers), otherwise it will return the configured
+     * context or a temporary one.
+     * @param currentCtx - [Optional] The current execution context
+     */
+    protected _getTelCtx: (currentCtx?:IProcessTelemetryContext) => IProcessTelemetryContext;
+
+    /**
+     * Internal helper to allow setting of the internal initialized setting for inherited instances and unit testing
+     */
+    protected setInitialized: (isInitialized: boolean) => void;
```
MSNev: Latest version changes order of these 2 methods to pass linting failure

shared/AppInsightsCore/src/JavaScriptSDK/BaseTelemetryPlugin.ts
```diff
@@ -0,0 +1,164 @@
...
+                // like a ITelemetryPlugin or ITelemetryPluginChain, just call processTelemetry
+                _nextPlugin.processTelemetry(env, null);
+            }
+        };
+
+        _self._getTelCtx = (currentCtx:IProcessTelemetryContext = null) => {
+            let itemCtx:IProcessTelemetryContext = currentCtx;
+            if (!itemCtx) {
+                let rootCtx = _rootCtx || new ProcessTelemetryContext(null, {}, _self.core);
+                // tslint:disable-next-line: prefer-conditional-expression
```
MSNev: Ignoring this rule in this case as I think it's clearer this way.

PR #1135, opened 2019-12-11
fix: dont block corr on empty location host

### LINKED ISSUE #1115: [BUG] enableCorsCorrelation doesn't work on ionic/cordova app
**Description/Screenshot**

I want to correlate my ionc app to my api server on Application Map of Application Insights. So I set "enableCorsCorrelation" to true in the config object and construct ApplicationInsights with the object. 

However, the configuration doesn't take effect when I build and run the android app on devices/emulator. I suspected the issue is caused by this line of code.

https://github.com/microsoft/ApplicationInsights-JS/blob/8d7fd1e43c41ee2e94c4cde3ddbcbed8ef9fc258/extensions/applicationinsights-dependencies-js/src/ajax.ts#L154-L155

```this.currentWindowHost ``` will be an empty string in runtime, since it is assigned to be window.location.host, [here](https://github.com/microsoft/ApplicationInsights-JS/blob/8d7fd1e43c41ee2e94c4cde3ddbcbed8ef9fc258/extensions/applicationinsights-dependencies-js/src/ajax.ts#L98), which is an empty string when running with cordova runtime (cordova-android@7.1.1).

**Steps to Reproduce**

 - OS/Browser: Android 4.4+
 - SDK Version [e.g. 22]: @microsoft/applicationinsights-web 2.3.1
 - How you initialized the SDK: 

```
new ApplicationInsights({ config: {
      instrumentationKey: 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx',
      correlationHeaderDomains: [domain], 
      enableCorsCorrelation: true
    }});
```

**Expected behavior**

"Request-Id" should be sent along with all requests to API server. 

**Additional context**

### PR DESCRIPTION
**Pull request by markwolff**
Fixes #1115 
Removes unneeded undefined check on `this.currentWindowHost` before checking if correlation headers should be included on `fetch` requests

### CODE REVIEW

File: extensions/applicationinsights-dependencies-js/src/ajax.ts [hunk 1]
```diff
@@ -153,7 +153,7 @@ export class AjaxMonitor implements ITelemetryPlugin, IDependenciesPlugin, IInst
             }
             return init;
         } else if (xhr) { // XHR
-            if (this.currentWindowHost && CorrelationIdHelper.canIncludeCorrelationHeader(this._config, xhr.ajaxData.getAbsoluteUrl(),
+            if (CorrelationIdHelper.canIncludeCorrelationHeader(this._config, xhr.ajaxData.getAbsoluteUrl(),
```
xiao-lix: If `this.currentWindowHost` is empty, then `canIncludeCorrelationHeader` will return false with this check [here ](https://github.com/microsoft/ApplicationInsights-JS/blob/e86462c8a430717d87f01c1864de890dfe29a979/shared/AppInsightsCommon/src/Util.ts#L646) then this if check is still false correct? Unless `requestHost` is empty too?

markwolff: Yes I don't think it will affect that check at all unless `requestHost` is somehow falsy as well. Since it comes from an actual request being made I don't think its possible. From what I can tell, the `this.currentWindow host &&` check isn't actually useful.

xiao-lix: yeah I agree to remove it, I mean if the case happens again (`this.currentWindowHost` is empty), it still can't pass this if check, the logic under if will not run as before. Is that the expected behavior?

markwolff: I've modified it to not pass the if check anymore unless enableCorsCorrelation is enable and it also passes the included/excluded domains check. I consider an empty `currentWindowHost` an undefined scenario, so my thought it to always reject it unless the user has opted into a particular CORS correlation scenario.

PR #1145, opened 2019-12-19
Bump version to 2.4.0-beta

MSNev:


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -585,6 +586,7 @@ export class ApplicationInsights extends BaseTelemetryPlugin implements IAppInsi
         }
 
         if (this.config.disableExceptionTracking === false &&
+            this.config.enableUnhandledPromiseRejectionTracking === true &&
```
MSNev: Added config so that this new feature is disabled by default

PR #1151, opened 2019-12-30
travis: speed up travis

markwolff:
Parallelize travis jobs


.travis/test_react_native.sh
```diff
@@ -0,0 +1,5 @@
+cd extensions/applicationinsights-react-native
```
xiao-lix: what about adding `test_angular.sh` like so: 
`
cd extensions/applicationinsights-angular
npm install
npm run build
npm run test
npm run lint
`
we can also adding it later after angular plugin is published.

markwolff: added `test_angular.sh`

.travis.yml
```diff
@@ -6,50 +6,7 @@ before_install:
...
-    #  script:
-    #  - cd extensions/applicationinsights-react-native
-    #  - npm install
-    #  - npm run build
-    #  - npm run test
-    #  - npm run lint
+    - script: ./.travis/test_rush.sh
+      name: "Tests"
+    - script: ./.travis/test_legacy.sh
+      name: "Legacy Tests"
```
hectorhdzg: Why react and angular are not running?

markwolff: React Native & Angular tests aren't including in CI as of yet. React runs as part of `test_rush`.

markwolff: Added react to CI. @xiao-lix can add Angular in separate PR once it is ready

PR #1152, opened 2020-01-02
docs: update pageview interfaces

markwolff:
- Document how to send custom pageview duration via APIs
- Add `properties` named argument to API-reference.
- Re-order telemetry initializer docs
- Add custom properties example
- Remove `customProperties`

### DIFF
--- a/API-reference.md
+++ b/API-reference.md
@@ -1,7 +1,7 @@
 ### trackPageView
 
 ```ts
-applicationInsights.trackPageView(pageView: IPageViewTelemetry, customProperties?: { [key: string]: any })
+applicationInsights.trackPageView(pageView: IPageViewTelemetry)
 ```
 
 The `IPageViewTelemetry` interface is below:
@@ -14,7 +14,9 @@ Parameter | Type | Description
 `pageType?` | string | **Optional**<br>Page Type string. Describes how you classify this page, e.g. errorPage, formPage, etc.
 `isLoggedIn?` | boolean | **Optional**<br>Whether or not the user is logged in
 `pageTags?` | dictionary | **Optional**<br>Property bag to contain an extension to domain properties - extension to Part B
+`properties?` | dictionary | **Optional**<br>Map of string to any: Additional data used to [filter pages](https://azure.microsoft.com/documentation/articles/app-insights-api-custom-events-metrics/#properties) in the portal. Defaults to empty.
 
+> *Note:* To send a custom duration (ms) of your pageview as an argument, it must be included in the `properties` named field. E.g `appInsights.trackPageView({ properties: { duration: 123.45 } });`.
 
 ### startTrackPage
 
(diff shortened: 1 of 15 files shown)


API-reference.md
```diff
@@ -1,7 +1,7 @@
 ### trackPageView
 
 ```ts
-applicationInsights.trackPageView(pageView: IPageViewTelemetry, customProperties?: { [key: string]: any })
```
xiao-lix: just curious, why are we removing `customProperties?` from readme? Isn't it still a param in code? 🤔

shared/AppInsightsCommon/src/Interfaces/IPageViewTelemetry.ts
```diff
@@ -31,6 +31,17 @@ export interface IPageViewTelemetry extends IPartC {
...
     isLoggedIn?: boolean;
+
+    /**
+     * Property bag to contain additional custom properties (Part C)
+     */
+    properties?: {
+        /**
+         * The number of milliseconds it took to load the page. Defaults to undefined. If set to default value, page load time is calculated internally.
+         */
+        duration?: number;
```
hectorhdzg: Is duration available to specify in PageViewPerformance events as well?, I can see is internal currently

README.md
```diff
@@ -118,6 +118,19 @@ appInsights.stopTrackEvent("event", {customProp1: "some value"});
 appInsights.flush();
 ```
 
+Custom properties can be included in your telemetry through the `properies` named argument. This can be done with *any* of the Track APIs.
```
hectorhdzg: typo "properties"

API-reference.md
markwolff: It is only used for 1P SDK and not meant to be exposed in this SDK's APIs. Although we still need to handle it for internal users, I think its best we just pick one way to do it instead of creating ambiguity and confusion with 2 (e.g. is there a difference between passing in one way or the other?) [Node.js SDK API](https://github.com/microsoft/ApplicationInsights-node.js/blob/be0f918c77aa7f0895f2784fa8c74ebc29131472/Declarations/Contracts/TelemetryTypes/Telemetry.ts#L12) uses `properties` named argument already, so I'm leaning that way. WDYT?

shared/AppInsightsCommon/src/Interfaces/IPageViewTelemetry.ts
markwolff: I'm not sure if the duration value from pageviewperformance ends up on the end-to-end trace gantt chart. AFAIK it does not and it uses the pageview duration instead.

README.md
markwolff: Nice. fixed.

API-reference.md
xiao-lix: Ok, makes sense. 👍

PR #1154, opened 2020-01-02
Queue events when track is called and not all extensions are initialized

hectorhdzg:
Adding internal queue to avoid sending events before all extensions are initialized

### DIFF
--- a/shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
+++ b/shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
@@ -27,9 +27,9 @@ export class BaseCore implements IAppInsightsCore {
 
     public _extensions: IPlugin[];
     public isInitialized: () => boolean;
+    protected _notificationManager: INotificationManager;
     private _eventQueue: ITelemetryItem[];
     private _allExtensionsInitialized: boolean;
-    protected _notificationManager: INotificationManager;
     private _channelController: ChannelController;
     private _setInit: (value: boolean) => void;
 
--- a/shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
+++ b/shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
@@ -27,7 +27,7 @@ export class BaseCore implements IAppInsightsCore {
 
     public _extensions: IPlugin[];
     public isInitialized: () => boolean;
-    private _eventQueue: Array<ITelemetryItem>;
+    private _eventQueue: ITelemetryItem[];
     private _allExtensionsInitialized: boolean;
     protected _notificationManager: INotificationManager;
     private _channelController: ChannelController;
--- a/shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
+++ b/shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
@@ -27,9 +27,11 @@ export class BaseCore implements IAppInsightsCore {
 
     public _extensions: IPlugin[];
     public isInitialized: () => boolean;
+    private _eventQueue: Array<ITelemetryItem>;
+    private _allExtensionsInitialized: boolean;
     protected _notificationManager: INotificationManager;
     private _channelController: ChannelController;
-    private _setInit: (value:boolean) => void;
+    private _setInit: (value: boolean) => void;
 
     constructor() {
         let _isInitialized = false;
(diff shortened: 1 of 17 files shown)


shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -181,4 +196,19 @@ export class BaseCore implements IAppInsightsCore {
 
         return new ProcessTelemetryContext(thePlugins, _this.config, _this);
     }
+
+    areExtensionsInitialized(): boolean {
+        let _this = this;
+        if (!_this._allExtensionsInitialized) {
+            let extensions = _this._extensions;
+            let isAnyExtensionNotInit = false;
+            _arrForEach(extensions, (ext: ITelemetryPlugin) => {
```
markwolff: Why is a foreach used? Does it compress better?

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -164,15 +167,27 @@ export class BaseCore implements IAppInsightsCore {
             telemetryItem.ver = "4.0";
         }
 
-        // Process the telemetry plugin chain
-        _this.getProcessTelContext().processNext(telemetryItem);
+        if (_this.areExtensionsInitialized()) {
+            // Release queue
+            if (_this._eventQueue.length > 0) {
+                _arrForEach(_this._eventQueue, (event: ITelemetryItem) => {
+                    _this.getProcessTelContext().processNext(telemetryItem);
```
markwolff: `event`?

hectorhdzg: Nice catch, test were still passing because it only verified the number of events, will update tests as well

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
hectorhdzg: I don't believe it compress better than a regular for, I don't have a particular reason to use it, is there any issue with it?

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -181,4 +196,19 @@ export class BaseCore implements IAppInsightsCore {
...
+        let _this = this;
+        if (!_this._allExtensionsInitialized) {
+            let extensions = _this._extensions;
+            let isAnyExtensionNotInit = false;
+            _arrForEach(extensions, (ext: ITelemetryPlugin) => {
+                if (!_isNullOrUndefined(ext.isInitialized) && !ext.isInitialized()) {
+                    isAnyExtensionNotInit = true;
+                }
+            });
+            _this._allExtensionsInitialized = isAnyExtensionNotInit ? false : true;
```
markwolff: `!isAnyExtensionNotInit` or `isAnyExtensionNotInit === false`?

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
markwolff: No issue with it, I think a `for...of` would make it not require the temp return var and be easier to read.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -181,4 +196,19 @@ export class BaseCore implements IAppInsightsCore {
 
         return new ProcessTelemetryContext(thePlugins, _this.config, _this);
     }
+
+    areExtensionsInitialized(): boolean {
+        let _this = this;
+        if (!_this._allExtensionsInitialized) {
+            let extensions = _this._extensions;
```
MSNev: During BaseCore initialization, this actually, won't include all of the extensions after initialization, it will only contain those "managed" by BaseCore and not the ones that have been delegated to the ChannelController

hectorhdzg: Removed this method entirely

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -28,16 +28,19 @@ export class BaseCore implements IAppInsightsCore {
     public _extensions: IPlugin[];
     public isInitialized: () => boolean;
     protected _notificationManager: INotificationManager;
+    private _eventQueue: ITelemetryItem[];
+    private _allExtensionsInitialized: boolean;
```
MSNev: Not needed

PR #1155, opened 2020-01-03
update angular plugin test

xiao-lix:


extensions/applicationinsights-angular-js/test-app/src/app/app.component.spec.ts
```diff
@@ -59,7 +59,7 @@ describe('Router: App', () => {
         }
       }
     }, core, []);
-    const angularConfig: IAngularExtensionConfig = angularPlugin['_extensionConfig'];
```
markwolff: what was the value of this before?

xiao-lix: the value was a private _extensionConfig: 
https://github.com/microsoft/ApplicationInsights-JS/commit/54e489d7f4ddca958423f25ae4211ad33bc7dc35#diff-d9d8329e74840d3b34ba24ebf17fe9afR23

The value was updated to use ctx.getExtCfg function.

PR #1159, opened 2020-01-14
Export Common Telemetry classes in Snippet

markwolff:
- Re-export Common Telemetry classes via `Microsoft.ApplicationInsights.Telemetry` as done in legacy SDK
- Fix issue in snippet causing `startTrackundefined` to be lazy defined


AISKU/snippet/snippet.js
```diff
@@ -46,7 +46,7 @@ var aisdk = window[aiName] || (function (aiConfig) {
     createLazyMethod("start" + trackPage);
     createLazyMethod("stop" + trackPage);
 
-    var trackEvent = track + method[0];
```
MSNev: Wow! very subtle error to spot!

markwolff: oops forgot to add this to the PR notes

Issue #1156, opened 2020-01-08
Allow to include response error data on failed requests


avilde:
Hi!

While working with [WebChat](https://github.com/microsoft/BotFramework-WebChat/#readme) and [AppInsights](https://github.com/microsoft/ApplicationInsights-JS#readme) we have found an issue in failed network request logs.
![image](https://user-images.githubusercontent.com/364880/71956116-f3c76880-31f2-11ea-8ae4-7e40c4d38292.png)

For us there is no good way to know what was the issue in case we got result code **403**.
Bellow is an example of how many cases there are for this one result code:
https://github.com/microsoft/BotFramework-DirectLineJS/issues/251#issuecomment-566170597

I did some debugging and found out that in result handler there is nothing done with the response itself:
https://github.com/microsoft/ApplicationInsights-JS/blob/c9f43a610ae8a0f2fbd976ef6107c58abf0e3ede/extensions/applicationinsights-dependencies-js/src/ajax.ts#L497

If I put breakpoint in this function I can see the response with error information:
![image](https://user-images.githubusercontent.com/364880/71958728-a7cbf200-31f9-11ea-9a81-775dfa27c6f3.png)

**What if there was a setting (flag) on app insights init config to include response in `ajaxData` object when return code is not e.g. `200 OK`?**

This way developer could write a telemetry plugin to process the `RemoteDependencyData` event and include any error data from the response.

Right now I'm forced to write my own override of AjaxMonitor and replace the original extension so I could find out what was the issue on client side.

PR #1161, opened 2020-01-17
include response error data

xiao-lix:
Include response error data in part C on dependency event when ajax respond code is not 200.

Closes: #1156

### DIFF
--- a/AISKU/Tests/applicationinsights.e2e.tests.ts
+++ b/AISKU/Tests/applicationinsights.e2e.tests.ts
@@ -61,6 +61,7 @@ export class ApplicationInsightsTests extends TestClass {
                 disableFetchTracking: false,
                 enableRequestHeaderTracking: true,
                 enableResponseHeaderTracking: true,
+                includeResponseErrorData: true,
                 maxBatchInterval: 2500,
                 disableExceptionTracking: false,
                 namePrefix: this.sessionPrefix,
--- a/AISKU/Tests/applicationinsights.e2e.tests.ts
+++ b/AISKU/Tests/applicationinsights.e2e.tests.ts
@@ -61,7 +61,6 @@ export class ApplicationInsightsTests extends TestClass {
                 disableFetchTracking: false,
                 enableRequestHeaderTracking: true,
                 enableResponseHeaderTracking: true,
-                includeResponseErrorData: true,
                 maxBatchInterval: 2500,
                 disableExceptionTracking: false,
                 namePrefix: this.sessionPrefix,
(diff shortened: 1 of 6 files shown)


extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -557,6 +559,11 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
                 }
             }
 
+            if (_self._config.includeResponseErrorData && xhr.status !== 200) {
```
ramthi: >_self._config.includeResponseErrorData && xhr.status !== 200 [](start = 16, length = 60)

Do we want to do this for all !=200 status codes or should we just check >=400 ?

MSNev: +1 we should only do this for status >= 400

As any response from 200 -> 299 is considered to be a success response, a common response is 204 == No Content.
And 300-> 399 are redirects

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -557,6 +559,11 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
+                dependency.properties = dependency.properties || {};
+                dependency.properties.responseError = xhr.statusText + " - " + xhr.responseText;
```
markwolff: nit: rename to `responseMessage` or `responseText`, to convey string type instead of Error.

AISKU/Tests/applicationinsights.e2e.tests.ts
```diff
@@ -61,6 +61,7 @@ export class ApplicationInsightsTests extends TestClass {
                 disableFetchTracking: false,
                 enableRequestHeaderTracking: true,
                 enableResponseHeaderTracking: true,
+                includeResponseErrorData: true,
```
markwolff: I think this should be renamed to `disableAjaxErrorStatusText` since it seems like something we would want to enable by default.

shared/AppInsightsCommon/src/Interfaces/IConfig.ts
```diff
@@ -294,6 +294,14 @@ export interface IConfig {
      */
     enableResponseHeaderTracking?: boolean;
 
+    /**
+     * @description An optional value that will track Resonse Error data through trackDependency function.
```
avilde: Small typo in description `Resonse`

PR #1167, opened 2020-01-27
Add better support for referencing global objects from a window and web workers

MSNev:


AISKU/Tests/TestFramework/TestClass.ts
```diff
@@ -80,24 +87,26 @@ class TestClass {
                                 nextTestStepTrigger.call(this);
                             }
                         } catch (e) {
-                            this._testCompleted();
+                            console.error("Failed: Unexpected Exception: " + e);
+                            this._testCompleted(true);
                             Assert.ok(false, e.toString());
-
+                            completed = true;
```
xiao-lix: Is this `completed` duplicated with `this._testCompleted`? 🤔

MSNev: No _testCompleted is a method, this is a flag to indicated that no additional steps should be processed.

AISKU/Tests/TestFramework/TestClass.ts
```diff
@@ -221,6 +240,32 @@ class TestClass {
...
+    }
+
+    public dumpPayloadMessages(spy:SinonSpy) {
+        let msg = "Sent Messages";
+        if (spy.called && spy.args && spy.args.length > 0) {
+            spy.args[0][0].forEach((value, idx) => {
+                msg += "\n" + idx + ":" + value;
+            });
+        }
+        Assert.ok(false, msg);
```
xiao-lix: Why is `msg` false, isn't it always has a value?

MSNev: So that the message is always logged into the output, as this is only called when the tests fail.

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -246,10 +246,12 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
         if (!_self.supportsFetch() || _self._fetchInitialized) {
             return;
         }
-        let _window = getWindow();
-        const originalFetch: (input?: Request | string, init?: RequestInit) => Promise<Response> = _window.fetch;
+
+        // Getting the global instance tp support web workers (which don't have window)
+        let global = getGlobal();
```
markwolff: This should be named `_global`?

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -128,10 +213,8 @@ export function hasJSON(): boolean {
  */
 export function getJSON(): JSON | null {
     if (hasJSON()) {
-        return JSON;
+        return JSON || getGlobalInst(strJSON);
```
markwolff: Is the `JSON ||` part here still necessary?

PR #1172, opened 2020-01-28
Update SameSite logic to handle UserAgents that don't support the attribute

MSNev:


shared/AppInsightsCommon/src/Util.ts
```diff
@@ -335,24 +341,85 @@ export class Util {
...
 
-        let location = getLocation();
-        if (location && location.protocol === "https:") {
-            secureAttrib = ";secure";
-            value = value + ";SameSite=None"; // SameSite can only be set on secure pages
-        }
+            let location = getLocation();
+            if (location && location.protocol === "https:") {
+                secureAttrib = ";secure";
+                if (!Util.disallowsSameSiteNone((getNavigator()||{} as Navigator).userAgent)) {
```
hectorhdzg: Is this something we can only call once instead of calling it every time we want to set a cookie?

shared/AppInsightsCommon/src/Util.ts
```diff
@@ -335,24 +341,85 @@ export class Util {
...
+     
+        // Cover Chrome 50-69, because some versions are broken by SameSite=None, and none in this range require it.
+        // Note: this covers some pre-Chromium Edge versions, but pre-Chromim Edge does not require SameSite=None, so this is fine.
+        // Note: this regex applies to Windows, Mac OS X, and Linux, deliberately.
+        if (userAgent.indexOf("Chrome/5") !== -1 || userAgent.indexOf("Chrome/6") !== -1) {
+            return true;
+        }
+     
+        // Unreal Engine runs Chromium 59, but does not advertise as Chrome until 4.23. Treat versions of Unreal
+        // that don't specify their Chrome version as lacking support for SameSite=None.
```
hectorhdzg: I cannot see this in https://www.chromium.org/updates/same-site/incompatible-clients are you using some other reference for this specific user agents?

shared/AppInsightsCommon/src/Util.ts
```diff
@@ -335,24 +341,85 @@ export class Util {
...
+        if (!_isString(userAgent)) {
+            return false;
+        }
+    
+        // Cover all iOS based browsers here. This includes:
+        // - Safari on iOS 12 for iPhone, iPod Touch, iPad
+        // - WkWebview on iOS 12 for iPhone, iPod Touch, iPad
+        // - Chrome on iOS 12 for iPhone, iPod Touch, iPad
+        // All of which are broken by SameSite=None, because they use the iOS networking stack
+        if (userAgent.indexOf("CPU iPhone OS 12") !== -1 || userAgent.indexOf("iPad; CPU OS 12") !== -1) {
```
hectorhdzg: Chromium page recommend more elaborated regex to identify these, are we sure this will be enough?

MSNev: Yes, these are simplified versions based on actual usages.

shared/AppInsightsCommon/src/Util.ts
MSNev: Good point -- let me change this.

MSNev: Updates

Issue #1126, opened 2019-12-03
Manage SameSite Cookie Settings


audacity76:
I was wondering how I can set the SameSite option for the Application Insights cookies. Chrome will change the behavior of SameSite cookies in Feb 2020 (https://www.chromestatus.com/feature/5088147346030592). Is there any documentation how the cookie options are set depending on the browser agent?

markwolff:
If you are on the latest version of the SDK, we are already [setting SameSite to `None`](https://github.com/microsoft/ApplicationInsights-JS/pull/1088/commits/12c6799f25d637425d81fbd3b80031b7fe54cd74). This only applies to new cookies, and existing cookies will not have the policy applied to them. From what I can tell the new policy only blocks the "creation" of non-policy'd cookies.

audacity76:
Thank you for your information. As far as I know the cookie setting depends on the browser agent as there are incompatible clients: https://www.chromium.org/updates/same-site/incompatible-clients

neolefty:
It looks like the fix for legacy cookies (issue #1112) is not in the latest NPM release 2.3.1:

* Fix: https://github.com/microsoft/ApplicationInsights-JS/pull/1112 was published 2019-11-08
* NPM package v2.3.1 was published on 2019-11-07: https://www.npmjs.com/package/@microsoft/applicationinsights-react-js/v/2.3.1

markwolff:
@neolefty The legacy cookie fix PR is (unreleased) patch for 1.0.20 iteration of the (legacy) SDK. For the 2.* version, the cookie change for this SDK is already released

hiraldesai:
I have an angular app using version 2.4.2 of the @microsoft/application-insights-web package and the SameSite issue doesn't happen in Chrome80 but I noticed that Edge Chromium Version 79.0.309.71 (Official build) (64-bit) is still writing no SameSite attribute that results in a warning (screenshot).

![image](https://user-images.githubusercontent.com/1005174/73583535-a16f1580-4447-11ea-9491-570fe544729c.png)

markwolff:
@hiraldesai Could you delete the cookie and see if it writes it with the correct property?

hiraldesai:
I had tried that before, tried it one more time. Same issue. 

![image](https://user-images.githubusercontent.com/1005174/73584008-3ffc7600-444a-11ea-9ba8-559e40e3cbd0.png)

markwolff:
Was it also an issue on 2.3.1?

[...]

PR #1174, opened 2020-01-29
Legacy Manage SameSite Cookie Settings #1126

MSNev:
Address UA validation when setting SameSite attribute on the cookie for the Legacy build

### DIFF
--- a/legacy/JavaScript/JavaScriptSDK.Tests/CheckinTests/Util.tests.ts
+++ b/legacy/JavaScript/JavaScriptSDK.Tests/CheckinTests/Util.tests.ts
@@ -506,6 +506,113 @@ class UtilTests extends TestClass {
                 Assert.equal("p.r.e.f.i.x.bing.com", UrlHelper.parseHost("http://wwW2.p.r.e.f.i.x.bing.com/"));
             }
         });
+
+        this.testCase({
+            name: "Check disableSameSiteCookie status",
+            test: () => {
+                let excludeValues = [
+                    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.140 Safari/537.36 Edge/17.17134",
+                    "Mozilla/5.0 (iPhone; CPU iPhone OS 12_1_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Mobile/15E148 Safari/604.1",
+                    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.140 Safari/537.36 Edge/18.17763",
+                    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36 Edge/16.16299",
+                    "Mozilla/5.0 (Windows NT 10.0; Win64; x64; WebView/3.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.140 Safari/537.36 Edge/17.17134",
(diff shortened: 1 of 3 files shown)


legacy/JavaScript/JavaScriptSDK.Tests/CheckinTests/Util.tests.ts
```diff
@@ -506,6 +506,113 @@ class UtilTests extends TestClass {
                 Assert.equal("p.r.e.f.i.x.bing.com", UrlHelper.parseHost("http://wwW2.p.r.e.f.i.x.bing.com/"));
             }
         });
+
+        this.testCase({
```
MSNev: Backported the SameSite code and tests added to the current version

legacy/JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -4,6 +4,13 @@
 /// <reference path="./Logging.ts" />
 /// <reference path="./UtilHelpers.ts" />
 
+
+function _endsWith(value:string, search:string) {
```
MSNev: Backported the SameSite code and tests added to the current version

legacy/JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -316,6 +325,65 @@ module Microsoft.ApplicationInsights {
...
+            // Unreal Engine runs Chromium 59, but does not advertise as Chrome until 4.23. Treat versions of Unreal
+            // that don't specify their Chrome version as lacking support for SameSite=None.
+            if (userAgent.indexOf("UnrealEngine") !== -1 && userAgent.indexOf("Chrome") === -1) {
+                return true;
+            }
+         
+            // UCBrowser < 12.13.2 ignores Set-Cookie headers with SameSite=None
+            // NB: this rule isn't complete - you need regex to make a complete rule.
+            // See: https://www.chromium.org/updates/same-site/incompatible-clients
+            if (userAgent.indexOf("UCBrowser/12") !== -1 || userAgent.indexOf("UCBrowser/11") !== -1) {
```
MSNev: While these are not exactly the same as that defined by https://www.chromium.org/updates/same-site/incompatible-clients this is using string validation rather than regex for better performance and the set of UA's is based on actual usage for AI and other MS sites.

legacy/JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -316,6 +325,65 @@ module Microsoft.ApplicationInsights {
...
+            // - Chromium on Mac OS X
+            // Because they do not use the Mac OS networking stack.
+            if (userAgent.indexOf("Macintosh; Intel Mac OS X 10_14") !== -1 && _endsWith(userAgent, "AppleWebKit/605.1.15 (KHTML, like Gecko)")) {
+                return true;
+            }
+         
+            // Cover Chrome 50-69, because some versions are broken by SameSite=None, and none in this range require it.
+            // Note: this covers some pre-Chromium Edge versions, but pre-Chromim Edge does not require SameSite=None, so this is fine.
+            // Note: this regex applies to Windows, Mac OS X, and Linux, deliberately.
+            if (userAgent.indexOf("Chrome/5") !== -1 || userAgent.indexOf("Chrome/6") !== -1) {
```
MSNev: While this will match Chrome versions 50, 67, 68 and 69 as incompatible (when they actually are compatible) is not expected to cause any issues as these older versions today are already functional without the SameSite attribute / restriction, which is only really required for the later versions).

legacy/JavaScript/JavaScriptSDK/Util.ts
```diff
@@ -329,7 +397,15 @@ module Microsoft.ApplicationInsights {
 
             if (Util.document.location && Util.document.location.protocol === "https:") {
                 secureAttrib = ";secure";
-                value = value + ";SameSite=None"; // SameSite can only be changed on secure pages
+                if (Util._uaDisallowsSameSiteNone === null) {
```
MSNev: Optimization, so we only perform this check once rather than for every cookie

Issue #1173, opened 2020-01-29
[BUG] Typescript build with 2.4.0 breaks


StefanSchoof:
**Description/Screenshot**
https://dev.azure.com/sschoof/espresso/_build/results?buildId=1514&view=logs&j=0a099ed4-e459-5b34-cc15-88dac43c3262&t=51a3d40f-3f24-5548-8c6e-04827d8ff082&l=37
**Steps to Reproduce**
Upgrade working Typescript build from 2.3.1 to 2.4.0
 - OS/Browser: n/a
 - SDK Version [e.g. 22]: n/a
 - How you initialized the SDK: n/a

**Expected behavior**
A version upgrade from 2.3 to 2.4 is not breaking

**Additional context**

MSNev:
This appears to be the side effect of a tighter restriction on the arguments from TS 3.7.3 being used as it compiles fine with TS 2.5.3..

I believe I understand the issue and will work on it in the morning.

MSNev:
@StefanSchoof can you confirm how are you consuming the module, while I've manually reviewed the method signatures and published a PR that will address the error, I was unable to reproduce the issue using multiple versions of TypeScript (including 3.7.3). From the error it appears that you are or maybe consuming and re-building/compiling the generated module which is causing the break as it validates against the *.d.ts? Can you confirm?

Either way, is it possible for you to assist by revalidating using this branch (MSNev/TypoFixesForTS3.7 instead of master) and using the generated code at all, this would be extremely helpful so we don't publish 2.4.1 (which we are planning to do tomorrow) and then find yet another blocking issue for you?

MSNev:
Fixed with v2.4.1

StefanSchoof:
Sorry, missed your questions. Too bad I missed a opportunity to help you.

Your fix wokred and with 2.4.1 my build is up again. Thanks for the quick fix.

PR #1175, opened 2020-01-29
[BUG] Typescript build with 2.4.0 breaks #1173

MSNev:
Address issue #1173 where the declaration of the methods does not match the actual implementation.

### DIFF
--- a/legacy/JavaScript/JavaScriptSDK/ajax/ajax.ts
+++ b/legacy/JavaScript/JavaScriptSDK/ajax/ajax.ts
@@ -85,7 +85,7 @@ module Microsoft.ApplicationInsights {
         private instrumentOpen() {
             var originalOpen = XMLHttpRequest.prototype.open;
             var ajaxMonitorInstance = this;
-            XMLHttpRequest.prototype.open = function (method, url, async) {
+            XMLHttpRequest.prototype.open = function (method, url, async?) {
                 try {
                     if (ajaxMonitorInstance.isMonitoredInstance(this, true) &&
                         (
--- a/shared/AppInsightsCommon/src/Util.ts
+++ b/shared/AppInsightsCommon/src/Util.ts
@@ -232,7 +232,7 @@ export class Util {
         const keys = [];
 
         if (Util.canUseSessionStorage()) {
-            for (const key in getGlobalInst("sessionStorage")) {
+            for (const key in getGlobalInst<any>("sessionStorage")) {
                 keys.push(key);
             }
         }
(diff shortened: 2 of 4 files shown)


AISKU/Tests/ApplicationInsightsDeprecatedTests.ts
```diff
@@ -1,7 +1,5 @@
 import { IAppInsightsDeprecated } from "../src/ApplicationInsightsDeprecated";
-import { ApplicationInsightsContainer } from "../src/ApplicationInsightsContainer";
-import { Snippet } from "../src/Initialization";
-import { Sender } from "@microsoft/applicationinsights-channel-js";
+import { ApplicationInsightsContainer, Snippet, Sender } from "@microsoft/applicationinsights-web";
```
MSNev: Changed the tests to consume the module rather than using the source code to try and repro the failure :-) It didn't, but this is will capture issues in the future where we don't export a class / code.

AISKU/Tests/package.json
```diff
@@ -0,0 +1,41 @@
+{
+    "name": "@microsoft/applicationinsights-web-tests",
```
MSNev: Adding new package.json so the tests reference the generated module.

legacy/JavaScript/JavaScriptSDK/ajax/ajax.ts
```diff
@@ -85,7 +85,7 @@ module Microsoft.ApplicationInsights {
         private instrumentOpen() {
             var originalOpen = XMLHttpRequest.prototype.open;
             var ajaxMonitorInstance = this;
-            XMLHttpRequest.prototype.open = function (method, url, async) {
+            XMLHttpRequest.prototype.open = function (method, url, async?) {
```
MSNev: Change backported from my TypeScript 3.7.x upgrade PR (Where I tested these changes) as the async is really optional.

shared/AppInsightsCommon/src/Util.ts
```diff
@@ -232,7 +232,7 @@ export class Util {
         const keys = [];
 
         if (Util.canUseSessionStorage()) {
-            for (const key in getGlobalInst("sessionStorage")) {
+            for (const key in getGlobalInst<any>("sessionStorage")) {
```
MSNev: Another TS 3.7.x required syntax change

shared/AppInsightsCore/src/JavaScriptSDK/ProcessTelemetryContext.ts
```diff
@@ -137,7 +137,7 @@ export class ProcessTelemetryContext implements IProcessTelemetryContext {
     /**
      * Create a new context using the core and config from the current instance
      */
-    public createNew: (plugins?:IPlugin[], startAt?:IPlugin) => IProcessTelemetryContext;
+    public createNew: (plugins?:IPlugin[]|ITelemetryPluginChain, startAt?:IPlugin) => IProcessTelemetryContext;
```
MSNev: This was the actual implementation failure

AISKU/Tests/ApplicationInsightsDeprecatedTests.ts
MSNev: Removing this change

Issue #1182, opened 2020-02-01
error TS2430: Interface 'Window' incorrectly extends interface 'WindowEventHandlers'


alvipeo:
I'm using it in angular 8 app and recently it started to throw this error:

>ERROR in node_modules/typescript/lib/lib.dom.d.ts:17123:11 - error TS2430: Interface 'Window' incorrectly extends interface 'WindowEventHandlers'.
  Types of property 'onunhandledrejection' are incompatible.
    Type '((this: Window, ev: PromiseRejectionEvent) => any) | null' is not assignable to type '((this: WindowEventHandlers, ev: PromiseRejectionEvent) => any) | null'.
      Type '(this: Window, ev: PromiseRejectionEvent) => any' is not assignable to type '(this: WindowEventHandlers, ev: PromiseRejectionEvent) => any'.
        The 'this' types of each signature are incompatible.
          Type 'WindowEventHandlers' is missing the following properties from type 'Window': Blob, TextDecoder, TextEncoder, URL, and 225 more.

>17123 interface __**Window**__ extends EventTarget, WindowTimers, WindowSessionStorage, WindowLocalStorage, WindowConsole, GlobalEventHandlers, IDBEnvironment, WindowBase64, GlobalFetch, WindowOrWorkerGlobalScope, WindowEventHandlers {

Now, I get this error only when I use `strict=true` in __tsconfig.json__. But it wokder with previous versions of this package without any issue.

I did check - if I remove `microsoft/applicationinsights-web` package and related code everything works. So this is clearly the proble with microsoft/applicationinsights-web.

Here's the angular service which causes the problem:

```

import { Injectable, OnInit } from "@angular/core";
import { ApplicationInsights } from "@microsoft/applicationinsights-web";
import { ActivatedRouteSnapshot, ResolveEnd, Router } from "@angular/router";
import { filter, map } from "rxjs/operators";
import { Subscription } from "rxjs";
import { environment } from "src/environments/environment";

@Injectable({
  providedIn: "root"
})
export class AppInsightsService {
  private routerSubscription: Subscription;

  private appInsights = new ApplicationInsights({
    config: {
      instrumentationKey: environment.production
        ? "b6f62105-85a5-4055-b483-466d57ebb263" // production
        : "b9eb5e3c-0581-4bb4-ad69-234d3ab7c8fd" // dev
    }
  });

  constructor(private router: Router) {
    this.appInsights.loadAppInsights();
[...]

PR #1183, opened 2020-02-03
error TS2430: Interface 'Window' incorrectly extends interface 'WindowEventHandlers' #1182

MSNev:


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -597,7 +593,7 @@ export class ApplicationInsights extends BaseTelemetryPlugin implements IAppInsi
             // We want to enable exception auto collection and it has not been done so yet
             const onunhandledrejection = "onunhandledrejection";
             const originalOnUnhandledRejection = _window[onunhandledrejection];
-            _window.onunhandledrejection = (error: PromiseRejectionEvent) => {
+            _window[onunhandledrejection] = (error: PromiseRejectionEvent) => {
```
MSNev: Changing to index assignment to avoid the TS error and this also means we don't need the additional declaration to add this for older TS versions.

PR #1190, opened 2020-02-13
stopTrackPage missing part c properties

### LINKED ISSUE #1189: Add the option to specify the refUri with stopTrackPage
**Problem**
Currently the stopTrackPage method does not support the overriding of the refUri property like the regular trackPageView does.
The [stopTrackPage method takes a name, url, customProperties and measurements arguments](https://github.com/microsoft/ApplicationInsights-JS/blob/master/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts#L347). But even if you include a refUri value in the customProperties, another refUri property is automatically set (not on the custom properties, rather on the pageView object itself) [in sendPageViewInternal](https://github.com/microsoft/ApplicationInsights-JS/blob/master/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts#L263) and it seems that the latter takes precedence, since that is the only value I can see in my Azure portal.

**Solution**
Provide an optional refUri parameter on stopTrackPage and set this property [in the page tracking callback](https://github.com/microsoft/ApplicationInsights-JS/blob/master/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts#L553). It's interface [already has refUri anyway](https://github.com/microsoft/ApplicationInsights-JS/blob/master/shared/AppInsightsCommon/src/Interfaces/IPageViewTelemetry.ts#L23).
```typescript
this._pageTracking.action = (name, url, refUri, duration, properties, measurements) => {
  // duration must be a custom property in order for the collector to extract it
  if (CoreUtils.isNullOrUndefined(properties)) {
      properties = {};
  }
  properties[durationProperty] = duration.toString();

  const pageViewItem: IPageViewTelemetry = {
      name,
      uri: url,
      refUri,
      properties,
      measurements
  };

  this.sendPageViewInternal(pageViewItem);
}
```

**Alternatives**
Another solution could be that the custom properties are also checked for the presence of a refUri property in [sendPageViewInternal](https://github.com/microsoft/ApplicationInsights-JS/blob/master/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts#L263) and the browser fallback is only used if refUri is set on neither the page view object itself or the custom proprties.

**Additional context**
We have an SPA, so the referrer value is only valid on the first page load.
The browser refUri fallback is also a full url that includes the domain name, but in our SPA the domain part is irrelevant since the same application runs on many different domains.

xiao-lix:
Add `properties` param for `stopTrackPage` function to capture part c customProperties.
Closes #1189

### DIFF
--- a/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
+++ b/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
@@ -557,7 +557,7 @@ export class ApplicationInsights extends BaseTelemetryPlugin implements IAppInsi
                 measurements
             };
 
-            this.sendPageViewInternal(pageViewItem);
+            this.sendPageViewInternal(pageViewItem, properties);
         }
 
         let _window = getWindow();
--- a/extensions/applicationinsights-analytics-js/Tests/ApplicationInsights.tests.ts
+++ b/extensions/applicationinsights-analytics-js/Tests/ApplicationInsights.tests.ts
@@ -531,7 +531,8 @@ export class ApplicationInsightsTests extends TestClass {
             duration: 200,
             properties: {
                 "property1": "5",
-                "property2": "10"
+                "property2": "10",
+                "refUri": "test.com"
             },
             measurements: {
                 "measurement": 300
@@ -556,7 +557,7 @@ export class ApplicationInsightsTests extends TestClass {
                 // act
                 appInsights.startTrackPage(testValues.name);
                 this.clock.tick(testValues.duration);
-                appInsights.stopTrackPage(testValues.name, testValues.url, testValues.properties);
+                appInsights.stopTrackPage(testValues.name, testValues.url, testValues.properties, testValues.measurements);
 
                 // verify
                 Assert.ok(spy.calledOnce, "stop track page view sent data");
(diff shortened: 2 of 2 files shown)


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -557,7 +557,7 @@ export class ApplicationInsights extends BaseTelemetryPlugin implements IAppInsi
                 measurements
             };
 
-            this.sendPageViewInternal(pageViewItem);
+            this.sendPageViewInternal(pageViewItem, properties);
```
markwolff: I think we need a test for this. Wouldn't we also want to send measurements as well?

xiao-lix: Good point, added some test, and double checked code. The reason `properties.refUri` was overwritten is because [this line](https://github.com/microsoft/ApplicationInsights-JS/blob/ebb1b6adc1b56fbcb4f086ef2baacad5cab1d7af/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts#L263). `measurements` filed is passed through `pageView` param then in saved in `baseData` [here](https://github.com/microsoft/ApplicationInsights-JS/blob/ebb1b6adc1b56fbcb4f086ef2baacad5cab1d7af/shared/AppInsightsCommon/src/TelemetryItemCreator.ts#L44).

PR #1192, opened 2020-02-18
Add to auto capture exception for ReactNative plugin

xiao-lix:


extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
```diff
@@ -75,6 +98,24 @@ export class ReactNativePlugin implements ITelemetryPlugin {
         this._device.deviceClass = newType;
     }
 
+    trackMetric(metric: IMetricTelemetry, customProperties: ICustomProperties) {
```
markwolff: private _trackMetric

extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
```diff
@@ -100,10 +141,23 @@ export class ReactNativePlugin implements ITelemetryPlugin {
         }
     }
 
+    /**
+     * Automatically collects unhandled JS exceptions and native exceptions
+     */
+    private _setExceptionHandlers() {
+        setJSExceptionHandler((error, isFatal) => {
+            this._trackException({exception: error});
```
markwolff: nit: whitespace
```suggestion
            this._trackException({ exception: error });
```

markwolff: What does `isFatal` mean? Will the app crash and telemetry be dropped? Would it require a flush after trackException?

extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
```diff
@@ -48,7 +57,21 @@ export class ReactNativePlugin implements ITelemetryPlugin {
             if (!this._config.disableDeviceCollection) {
                 this._collectDeviceInfo();
             }
+            
+            if (extensions) {
+                CoreUtils.arrForEach(extensions, ext => {
+                    const identifier = (ext as ITelemetryPlugin).identifier;
+                    if (identifier === 'ApplicationInsightsAnalytics') {
+                        this._analyticsPlugin = (ext as any) as IAppInsights;
```
markwolff: Are these casts required? Or can the `as any` be removed?

extensions/applicationinsights-react-native/README.md
```diff
@@ -36,6 +36,12 @@ By default, this plugin automatically collects
  - **Device Model Name** (iPhone XS, etc.)
  - **Device Type** (Handset, Tablet, etc.)
 
+ ## Uncaught Exceptions Collected
```
markwolff: Would the user need to install and link `react-native-exception-handler` like they would need to with the device-info package?

extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
```diff
@@ -7,11 +7,17 @@ import {
...
+    CoreUtils,
+    IDiagnosticLogger,
+    DiagnosticLogger,
+    LoggingSeverity,
+    _InternalMessageId
 } from '@microsoft/applicationinsights-core-js';
-import { ConfigurationManager, IDevice } from '@microsoft/applicationinsights-common';
+import { ConfigurationManager, IDevice, IMetricTelemetry, IAppInsights, IExceptionTelemetry } from '@microsoft/applicationinsights-common';
 import DeviceInfo from 'react-native-device-info';
+import { setJSExceptionHandler, setNativeExceptionHandler } from 'react-native-exception-handler';
```
markwolff: This is a breaking change for anyone that upgrades this plugin and does not have/want this package. Maybe we can lazy load this and the device-info package and only enable them if they are available?

extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
```diff
@@ -100,10 +141,23 @@ export class ReactNativePlugin implements ITelemetryPlugin {
...
 
+    /**
+     * Automatically collects unhandled JS exceptions and native exceptions
+     */
+    private _setExceptionHandlers() {
+        setJSExceptionHandler((error, isFatal) => {
+            this._trackException({exception: error});
+        }, true);
+        setNativeExceptionHandler(exceptionString => {
+            this._trackException({exception: new Error(exceptionString)});
```
markwolff: nit: whitespace

extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
xiao-lix: Just realized trackMetric is not used yet, removed it for now.

extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
xiao-lix: Got this error when removing `as any`: 
image

I think it's because ext is not necessarily analyticsPlugin? @MSNev can you please explain why we need two casts here? 🤔

extensions/applicationinsights-react-native/README.md
xiao-lix: yes you are right, added in next update. thanks

extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
xiao-lix: It's a self defined param of customized error handler ([example](`react-native-device-info`)), I think we don't need it here, removed it. thanks 👍

Issue #1194, opened 2020-02-20
XHR/Fetch enhancement - add additional telemetry from window.performance


MSNev:
Modern browsers support additional performance telemetry for XMLHttpRequest and fetch requests which are stored in the window.performance.getEntries() as resource entries.

Add an option to enable returning the additional details (when available) as part of the ajax requests for the XHR and fetch requests.

MSNev:
Version 2.5.2 is now fully deployed to the CDN, closing as this should address this issue.

PR #1195, opened 2020-02-20
XHR/Fetch enhancement - add additional telemetry from window.performance #1194

MSNev:


extensions/applicationinsights-dependencies-js/src/TraceParent.ts
```diff
@@ -12,24 +12,29 @@ export class Traceparent {
     private static DEFAULT_TRACE_FLAG = "01";
     private static DEFAULT_VERSION = "00";
     public spanId: string;
-    public traceFlag: string = Traceparent.DEFAULT_TRACE_FLAG;
```
MSNev: Moved this to inside the constructor and referenced as self.xxx as by default TypeScript adds this automatically to it's constructor definition as this.xxxx and the "this." can't be minified.

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -3,25 +3,114 @@
...
     CoreUtils, LoggingSeverity, _InternalMessageId,
     IAppInsightsCore, BaseTelemetryPlugin, ITelemetryPluginChain, IConfiguration, IPlugin, ITelemetryItem, IProcessTelemetryContext,
-    getLocation, getGlobal
+    getLocation, getGlobal, strUndefined, strPrototype, IInstrumentCallDetails, InstrumentFunc, InstrumentProto, getPerformance
 } from '@microsoft/applicationinsights-core-js';
-import { ajaxRecord } from './ajaxRecord';
+import { ajaxRecord, IAjaxRecordResponse } from './ajaxRecord';
 import { EventHelper } from './ajaxUtils';
 import { Traceparent } from './TraceParent';
+import dynamicProto from "@microsoft/dynamicproto-js";
```
MSNev: Major refactor to add XHR/Fetch window.performance additional metrics.
- Hiding private properties (by using dynamicProto)
- refactoring out duplicate code
Nett effect of this change on the 
Dependency plugin
- Unminified (browser) = additional code of just over 35kb
- Minified (browser) = 496 bytes LESS

AISKU
- Unminified (browser) = additional code of just under 35Kb
- Minified (browser) = 501 bytes LESS

extensions/applicationinsights-dependencies-js/src/ajaxRecord.ts
```diff
@@ -1,83 +1,351 @@
...
+export interface IAjaxRecordResponse {
+    statusText: string,
+    headerMap: Object,
+    correlationContext: string,
+    type?: string,
+    responseText?: string,
+    response?: Object
+}
+
+let strProperties = "properties";
```
MSNev: Using string property lookups for multiple references to support minification.

extensions/applicationinsights-dependencies-js/src/ajaxRecord.ts
```diff
@@ -1,83 +1,351 @@
...
+        let perfData = props[strPerf] = (props[strPerf] || {});
+        perfData[name] = value;
+        result = 1;
+    }
+
+    return result;
+}
+
+/** @ignore */
+function _populatePerfData(ajaxData:ajaxRecord, dependency:IDependencyTelemetry) {
```
MSNev: New perf data

extensions/applicationinsights-dependencies-js/src/ajaxRecord.ts
```diff
@@ -1,83 +1,351 @@
...
+        }
+    }
+
+    if (propsSet) {
+        dependency[strProperties] = props;
+    }
+}
 
 export class XHRMonitoringState {
-    public openDone: boolean = false;
```
MSNev: As eariler, moving populating the values so that Typescript doesn't place this in the constructor as "this.xxxx" and instead we can use self.xxxx

extensions/applicationinsights-dependencies-js/src/ajaxRecord.ts
```diff
@@ -1,83 +1,351 @@
...
+        // <summary>True, if onreadyStateChangeCallback function attached to xhr, otherwise false</summary>
+        self.stateChangeAttached = false;
+    }
 }
 
 export class ajaxRecord {
-    public completed = false;
-    public requestHeadersSize = null;
-    public requestHeaders = null;
-    public ttfb = null;
```
MSNev: Removed as not used

extensions/applicationinsights-dependencies-js/src/ajaxRecord.ts
```diff
@@ -1,83 +1,351 @@
...
     public spanID: string;
 
-    private _logger: IDiagnosticLogger;
-
-    constructor(traceID: string, spanID: string, logger: IDiagnosticLogger) {
-        this.traceID = traceID;
-        this.spanID = spanID;
-        this._logger = logger;
+    public getAbsoluteUrl(): string {
+        // @DynamicProtoStub -- DO NOT add any code as this will be removed during packaging
```
MSNev: Adding so that TypeScript will generate a d.ts file the same as previously, however, the Rush plugin will remove the generated stub during packaging.

e.g. TS will create the function 
ajaxRecord.prototype.getAbsoluteUrl = function() { ... }

But we will remove as this is just code bloat as the entire name can't be minified here, but by defining in the constructor and using dynamicProto() we remove this excessive "ajaxRecord.prototype." but we still end up with a class that defines the "getAbsoluteUrl" function on the ajaxRecord prototype

AISKU/package.json
```diff
@@ -49,6 +49,7 @@
         "typescript": "2.5.3"
     },
     "dependencies": {
+        "@microsoft/dynamicproto-js": "^0.5.1",
```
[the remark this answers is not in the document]
MSNev: As part of consuming the module (dist-esm folder) this would be a dependency and not a just a devDependency...

shared/AppInsightsCore/rollup.config.js
```diff
@@ -86,8 +95,14 @@ const nodeUmdRollupConfigFactory = (isProduction) => {
     nodeRollupConfig.plugins.push(
       uglify({
         ie8: true,
+        toplevel: true,
+        compress: {
+          passes:3,
+          unsafe: true,
```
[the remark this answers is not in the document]
MSNev: Supports better minification
- passes -- tries harder
- unsafe -- converts new Array(...) => [...]

topLevel however, might be bad -- will remove this one.

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -3,25 +3,114 @@
...
+            !_isNullOrUndefined(xhr.ajaxData.requestUrl)) {
+            result += "(url: '" + xhr.ajaxData.requestUrl + "')";
+        }
+    } catch (e) { }
+
+    return result;
+}
+
+/** @ignore */
+function _throwInternalCritical(ajaxMonitorInstance:AjaxMonitor, msgId: _InternalMessageId, message: string, properties?: Object, isUserAct?: boolean): void {
```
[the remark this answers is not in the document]
MSNev: For minification as they are called more than once so it removes several duplicate code definitions.

shared/AppInsightsCore/src/JavaScriptSDK/InstrumentHooks.ts
```diff
@@ -0,0 +1,235 @@
...
+}
+
+/**
+ * Intercept the named prototype functions for the target class / object
+ * @param target - The target object
+ * @param funcName - The function name
+ * @param callbacks - The callbacks to configure and call whenever the function is called
+ * @param checkPrototype - If the function doesn't exist on the target should it attempt to hook the prototype function
+ */
+export function InstrumentFunc(target:any, funcName:string, callbacks: IInstrumentHooksCallbacks, checkPrototype:boolean = true): IInstrumentHook {
```
markwolff: Do these functions make sense in core? Would it not make sense to include them in dynamicproto package?

shared/AppInsightsCore/src/JavaScriptSDK.Enums/LoggingEnums.ts
```diff
@@ -92,6 +92,7 @@ export const _InternalMessageId = {
     InvalidEvent: 70,
     FailedMonitorAjaxSetRequestHeader: 71,
     SendBrowserInfoOnUserInit: 72,
-    PluginException: 73
+    PluginException: 73,
+    InstrumentationError: 74,
```
markwolff: Where is this used?

shared/AppInsightsCore/src/JavaScriptSDK/InstrumentHooks.ts
MSNev: They belong here as DynamicProto doesn't do anything like what these functions are doing.

shared/AppInsightsCore/src/JavaScriptSDK.Enums/LoggingEnums.ts
MSNev: I appear to have removed the usage I did have in an earlier iteration -- will remove.

extensions/applicationinsights-dependencies-js/src/ajaxRecord.ts
```diff
@@ -1,83 +1,361 @@
...
+        }
+
+        propsSet |= _setPerfValue(props, strDuration, duration);
+        propsSet |= _setPerfValue(props, "perfTotal", duration);
+
+        var serverTiming = resourceEntry[strServerTiming];
+        if (serverTiming) {
+            let server = {};
+            _arrForEach(serverTiming, (value, idx) => {
+                let name = value[strName] || "" + idx;
```
MSNev: Should add some additional checks / conversion to make sure the name can be validly set as a key to avoid exceptions.

README.md
```diff
@@ -226,6 +226,9 @@ Most configuration fields are named such that they can be defaulted to falsey. A
 | enableRequestHeaderTracking | false | If true, AJAX & Fetch request headers is tracked, default is false.
 | enableResponseHeaderTracking | false | If true, AJAX & Fetch request's response headers is tracked, default is false.
 | enableAjaxErrorStatusText | false | Default false. If true, include response error data text in dependency event on failed AJAX requests.
+| enableAjaxPerfTracking | false | Default false. Flag to enable looking up and including additional browser window.performance timings in the reported ajax (XHR and fetch) reported metrics.
```
MSNev: Add this to the SKU's and typedoc.md files as well.

shared/AppInsightsCore/src/JavaScriptSDK/InstrumentHooks.ts
```diff
@@ -0,0 +1,235 @@
...
+                    h: []
+                };
+
+                // Override (hook) the original function
+                let newFunc = _createFunctionHook(aiHook);
+                newFunc[aiInstrumentHooks] = aiHook;        // Tag and store the function hooks
+                owner[funcName] = newFunc;
+            }
+
+            let theHook:IInstrumentHook = {
```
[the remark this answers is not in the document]
MSNev: Changed

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -3,25 +3,116 @@
...
+    ajaxMonitorInstance[strDiagLog]()[strThrowInternal](LoggingSeverity.CRITICAL, msgId, message, properties, isUserAct);
+}
+
+/** @ignore */
+function _throwInternalWarning(ajaxMonitorInstance:AjaxMonitor, msgId: _InternalMessageId, message: string, properties?: Object, isUserAct?: boolean): void {
+    ajaxMonitorInstance[strDiagLog]()[strThrowInternal](LoggingSeverity.WARNING, msgId, message, properties, isUserAct);
+}
```
xiao-lix: Can you please explain what does this syntax mean here `instance[]()[]()`? where is `ajaxMonitorInstance["dialog"]` come from? Thanks :)

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -50,749 +141,685 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
 
     public static getEmptyConfig(): ICorrelationConfig {
-        return {
-            maxAjaxCallsPerView: undefined,
-            disableAjaxTracking: undefined,
-            disableFetchTracking: undefined,
-            disableCorrelationHeaders: undefined,
-            distributedTracingMode: undefined,
-            correlationHeaderExcludedDomains: undefined,
-            appId: undefined,
```
xiao-lix: This `appId` has no default value in `defaultConfig`, should it be added in `emptyConfig` as `undefined`?

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -50,749 +141,685 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
-            enableCorsCorrelation: undefined,
-            correlationHeaderDomains: undefined,
```
xiao-lix: Same as `appId`

extensions/applicationinsights-dependencies-js/src/ajax.ts
MSNev: This 
instance["diagLog"]()["throwInternal"](....)
is effectively the same as this.
instance.diagLog().throwInternal(...)

The primary difference is that the first situation allows for better minification (only) when the quoted strings are used more than once as the variables holding the quoted strings can be minified to 1 (or 2 characters) while the direct names cannot.

I should also added that for minification you MUST use variables to hold the strings as using quote strings actually adds more characters than it removes, and uglify will actually convert direct quoted strings back to the full names anyway.

extensions/applicationinsights-dependencies-js/src/ajax.ts
MSNev: Nice catch!

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -50,749 +141,685 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
-    }
-
-    trackDependencyData(dependency: IDependencyTelemetry, properties?: { [key: string]: any }) {
-        this.trackDependencyDataInternal(dependency, properties);
-    }
-
-    public processTelemetry(item: ITelemetryItem, itemCtx?: IProcessTelemetryContext) {
-        this.processNext(item, itemCtx);
-    }
+        let _fetchInitialized:boolean = false;      // fetch monitoring initialized
```
xiao-lix: what's the difference to put these variables in constructor than leave outside? 🤔

MSNev: Outside the constructor they become object (this) level properties and therefore all references will need to address them as this.xxxxx

This leads to 2 problems (especially for private / internal properties).
1) The names can no longer be minified and must always be referenced as this.xxx (or self.xxxx)
2) They are now exposed and can be altered externally, part of the point of private properties is they should be internal to the implementation / class.

By moving them into the constructor the "scope" (visibility) of these variables is now limited to the within the constructor and any and all functions defined within the constructor.

Shorter answer
- Better minification
- Hiding private / internal properties, only used to track internal state.

README.md
MSNev: No need

extensions/applicationinsights-dependencies-js/src/ajax.ts
MSNev: It will get set to undefined by the method lookup.
It needs to be here as the code that populates the config that is used also iterates over the keys

Issue #1204, opened 2020-03-03
[BUG] When a fetch polyfill is installed the reporting endpoint is also causing events to be sent


MSNev:
**Steps to Reproduce**
Run in an environment that does not support fetch (eg, Phantom JS, React Native) and install a fetch polyfill (eg. whatwg-fetch).

Now whenever a fetch request is called from within the SDK that request is also instrumented (this is an extension and not part of the default configuration)

**Expected behavior**
The internal reporting requests should not cause additional requests to be sent.

MSNev:
We published version 2.5.1 to npm last night, which should address this issue. Once we complete the rollout to the CDN, I'll close this issue.

MSNev:
Version 2.5.2 is now fully deployed to the CDN, closing as this should address this issue.

PR #1205, opened 2020-03-03
[BUG] When a fetch polyfill is installed the reporting endpoint is also causing events to be sent #1204

MSNev:


common/Tests/External/whatwg-fetch.3.0.0.js
```diff
@@ -464,16 +464,15 @@
...
-      // -----------------------------------------------------------------------------
-      // TEST HOOK! -- NOT PART of the standard whatwg-fetch logic
-      // Adding so that we can test that only the fetch is getting logged!
       var xhr = new XHR();
 
       function abortXhr() {
         xhr.abort();
       }
 
-      xhr["Microsoft_ApplicationInsights_BypassAjaxInstrumentation"] = true;
```
MSNev: This hack is now no longer required as it's now built in :-)

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -29,16 +30,16 @@ let _objKeys = CoreUtils.objKeys;
...
-        result = false;
+    if (!_global || 
+            _isNullOrUndefined((_global as any).Request) ||
+            _isNullOrUndefined((_global as any).Request[strPrototype]) ||
+            _isNullOrUndefined(_global[strFetch])) {
+        return null;
     }
     
-    return result;
+    return _global[strFetch];
```
MSNev: Reworked slightly to help with minification

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -180,6 +181,7 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
         let _maxAjaxCallsPerView:number = 0;
         let _enableResponseHeaderTracking:boolean = false;
         let _hooks:IInstrumentHook[] = [];
+        let _disabledUrls:any = {};
```
MSNev: Defined as an object for performance, so that it can do an indexed lookup.
Did some testing using [] rather than an object and found IE in IE8 mode performance was seconds but as an object the lookup was ok.

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -456,30 +474,53 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
+                    isDisabled = xhr[DisabledPropertyName] === true;
+                } else if (!_isNullOrUndefined(request)) { // fetch
+                    // Look for DisabledPropertyName in either Request or RequestInit
+                    isDisabled = (typeof request === 'object' ? request[DisabledPropertyName] === true : false) ||
+                            (init ? init[DisabledPropertyName] === true : false);
+                }
+
+                if (isDisabled) {
+                    // Add the disabled url if not present
+                    if (!_disabledUrls[theUrl]) {
```
MSNev: Was a little concerned by the key being a full URL (eg. https://xxxxx.xxxx/xxxx etc), however, Chrome, IE (even in IE7 mode) and firefox all correctly handle the url as the key (with dots, slashes etc)

shared/AppInsightsCommon/src/Constants.ts
```diff
@@ -1,6 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+/**
+ * This is an internal property used to cause internal (reporting) requests to be ignored from reporting
+ * additional telemetry, to handle polyfil implementations ALL urls used with a disabled request will 
+ * also be ignored for future requests even when this property is not provided.
+ * Tagging as Ignore as this is an internal value and is not expected to be used outside of the SDK
+ * @ignore
```
hectorhdzg: Is this flag for typedoc to ignore?

extensions/applicationinsights-dependencies-js/src/ajax.ts
hectorhdzg: Did you manually verified or used some resource for it?

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -456,30 +474,53 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
                 }
             }
 
+            function _isDisabledRequest(xhr?: XMLHttpRequestInstrumented, request?: Request | string, init?: RequestInit) {
+                let isDisabled = false;
+                let theUrl:string = (!CoreUtils.isString(request) ? (request as Request).url || "" : request as string).toLowerCase();
+                let idx = theUrl.indexOf("?");
```
hectorhdzg: Do you want to remove hash parameter as well here? for custom endpoint scenarios to be extra safe

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -334,59 +334,76 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
-                            });
+                            }
+                        },
+                        // Create an error callback to report any hook errors
+                        hkErr: _createErrorCallbackFunc(_self, _InternalMessageId.FailedMonitorAjaxOpen, 
+                            "Failed to monitor Window.fetch, monitoring data for this fetch call may be incorrect.")
+                    }));
+    
+                    _fetchInitialized = true;
+                } else if ((fetch as any).polyfill) {
```
hectorhdzg: Is it possible to this to be called even if explicitly configured off using disableFetchTracking?

MSNev: Yes, I removed the explict check from the constructor and added it above.
And this is needed to ensure that we capture this state.

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -456,30 +474,53 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
                 }
             }
 
+            function _isDisabledRequest(xhr?: XMLHttpRequestInstrumented, request?: Request | string, init?: RequestInit) {
+                let isDisabled = false;
+                let theUrl:string = (!CoreUtils.isString(request) ? (request as Request).url || "" : request as string).toLowerCase();
```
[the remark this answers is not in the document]
MSNev: I actually pass the url as part of the call from the XHR, so no.

However, if they don't specify a url or pass undefined this would crash -- I'll add an extra check (if I don't have one already)

shared/AppInsightsCommon/src/Constants.ts
MSNev: Yes

extensions/applicationinsights-dependencies-js/src/ajax.ts
MSNev: Manually verified

extensions/applicationinsights-dependencies-js/src/ajax.ts
MSNev: Sounds like a plan as they could have no '?' but a hash.

PR #1206, opened 2020-03-04
[Release] Increase version to 2.5.0.

MSNev:


RELEASES.md
```diff
@@ -1,5 +1,35 @@
...
+
+### Changelog
+
+- #1204 When a fetch polyfill is installed there reporting endpoint is also causing events to be sent
+- #1202 ai_user and ai_session cookies not set #1203
+- #1201 add to auto track exceptions in react native plugin
+- #1199 Build is breaking when you do a "rush update --full --purge --recheck" due to tslib v1.11.0 update
+- #1194 XHR/Fetch enhancement - add additional telemetry from window.performance #1195
+- #1193 add sanitizer for operationName
+- #1192 Add to auto capture exception for ReactNative plugin
```
xiao-lix: this one was abandoned, could remove here. This was the method that use react-native-exception-handler package.

MSNev: ok, I'm just working locally to try and resolve this build issue.
Because react-native is not wired up with rush (and event when you do it fails) the build is failing because we don't have a reference to the updated 2.5.0 (as it's not published yet).

So I'm trying to see if I can do this in a single go rather than split this into separate PR (i.e. core 2.5.0 -> publish; and then react-native -> publish

Issue #1210, opened 2020-03-05
[BUG] Typescript error when using @microsoft/applicationinsights-web 2.5.0 and "noImplicitAny"/"strict" option


nicolashenry:
**Description/Screenshot**
There is a Typescript  error when using `@microsoft/applicationinsights-web` version `2.5.0` of  and "noImplicitAny"/"strict" option configured in tsconfig.json.
![image](https://user-images.githubusercontent.com/139160/75984979-b712a780-5eeb-11ea-884a-534dac633f71.png)

**Steps to Reproduce**
Use `@microsoft/applicationinsights-web` with version `2.5.0` and add "noImplicitAny"/"strict" option to "true" in tsconfig.json.

 - OS/Browser: Windows 7
 - SDK Version: 2.5.0
 - How you initialized the SDK: npm install @microsoft/applicationinsights-web

**Expected behavior**
No error

MSNev:
Investigating to see if we can enable this...

MSNev:
I'm adding noImplicitAny to the standard build config and fixing the breaks, however, adding strict is a very large issue, so not doing that now. Based on the error you are getting, I believe that this will unblock you once we republish with the fix

MSNev:
We published version 2.5.1 to npm last night, which should address this issue. Once we complete the rollout to the CDN, I'll close this issue (unless we here that it's still an issue)

nicolashenry:
It seems to be fixed :+1:

MSNev:
Version 2.5.2 is now fully deployed to the CDN, closing as this should address this issue.

PR #1211, opened 2020-03-05
[BUG] Typescript error when using @microsoft/applicationinsights-web 2.5.0 and "noImplicitAny"/"strict" option #1210

MSNev:


shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
```diff
@@ -11,32 +11,29 @@ const strAddEventHelper = "addEventListener";
...
+};
 
-function _isNullOrUndefined(value: any): boolean
-{
+const _isNullOrUndefined = (value: any): boolean => {
     return (_isUndefined(value) || value === null);
 }
 
-function _hasOwnProperty(obj:any, prop:string): boolean {
+const _hasOwnProperty = (obj:any, prop:string): boolean => {
```
MSNev: I initially attempted to also add "strict" set to true and this change was an attempt to work around TS complaining about evaluated values being possible undefined. While I undid the strict setting I havn't removed this change.

Currently performing comparisons against previous and this change, if this increases the total size of the prod packages or the generated code usage looks suspect I'll revert these

PR #1215, opened 2020-03-09
XHR object is not extensible on salesforce platform

dehru:
I'm unable to use ApplicationInsights-JS on the Salesforce Platform.  Salesforce has locked-down extending the XMLHttpRequest object with either of these methods:
```
Object.preventExtensions(XMLHttpRequest) 
// or 
Reflect.preventExtensions(XMLHttpRequest)
```
This change allows telemetry to be tracked on the Salesforce platform.  However, I'm not exactly sure the fallout of conditionally removing this property from xhr.  I'm submitting this for your feedback.  

Note: Also, I'm unable to satisfy the build because of ES3 support, though I believe I've sufficiently checked for existence before invoking the function.  The build continues to complain with: 

```
[Object.isExtensible(] is not supported in an ES3 environment, use a helper function or add explicit check for existence.
``` 

Thank you in advance for your time on my PR.

### DIFF
--- a/channels/applicationinsights-channel-js/src/Sender.ts
+++ b/channels/applicationinsights-channel-js/src/Sender.ts
@@ -539,7 +539,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
      */
     private _xhrSender(payload: string[], isAsync: boolean) {
         const xhr = new XMLHttpRequest();
-        if (typeof Object.isExtensible !== 'function' || (typeof Object.isExtensible === 'function' && Object.isExtensible(xhr))) {
+        if (CoreUtils.isExtensible(xhr)) {
             xhr[DisabledPropertyName] = true;
         }
         xhr.open("POST", this._senderConfig.endpointUrl(), isAsync);
--- a/channels/applicationinsights-channel-js/src/Sender.ts
+++ b/channels/applicationinsights-channel-js/src/Sender.ts
@@ -539,7 +539,9 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
      */
     private _xhrSender(payload: string[], isAsync: boolean) {
         const xhr = new XMLHttpRequest();
-        xhr[DisabledPropertyName] = true;
+        if (Object.isExtensible(xhr)) {
+            xhr[DisabledPropertyName] = true;
+        }
         xhr.open("POST", this._senderConfig.endpointUrl(), isAsync);
         xhr.setRequestHeader("Content-type", "application/json");
 
--- a/channels/applicationinsights-channel-js/src/Sender.ts
+++ b/channels/applicationinsights-channel-js/src/Sender.ts
@@ -539,7 +539,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
      */
     private _xhrSender(payload: string[], isAsync: boolean) {
         const xhr = new XMLHttpRequest();
-        if (Object.isExtensible(xhr)) {
+        if (typeof Object.isExtensible !== 'function' || (typeof Object.isExtensible === 'function' && Object.isExtensible(xhr))) {
             xhr[DisabledPropertyName] = true;
         }
         xhr.open("POST", this._senderConfig.endpointUrl(), isAsync);
--- a/shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
+++ b/shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
@@ -402,7 +402,7 @@ export class CoreUtils {
      * @param target The object on which to determine extensibility.
      * @returns True if the object is extensible.
      */
-    public static isExtensible<T>(obj:object) : boolean {
+    public static isExtensible(obj:object) : boolean {
         let extensibleProp = Object["isExtensible"];
         if (extensibleProp) {
             try {
--- a/shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
+++ b/shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
@@ -397,6 +397,26 @@ export class CoreUtils {
     }
 
     /**
+     * Try to determine if the object is extensible when run within an >= ES5 container.
+     * For < ES5 containers where the method does not exist, always return true.
+     * @param target The object on which to determine extensibility.
+     * @returns True if the object is extensible.
+     */
+    public static isExtensible<T>(obj:object) : boolean {
+        let extensibleProp = Object["isExtensible"];
+        if (extensibleProp) {
+            try {
+                return extensibleProp(obj);
+            } catch (e) {
+                // ES5 will throw a typeError if user passes a primitive
+                return false;
+            }
+        }
+        return true;
+    }
+
+
+    /**
      * Trys to add an event handler for the specified event to the window, body and document
      * @param eventName {string} - The name of the event
      * @param callback {any} - The callback function that needs to be executed for the given event


channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -539,7 +539,9 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
      */
     private _xhrSender(payload: string[], isAsync: boolean) {
         const xhr = new XMLHttpRequest();
-        xhr[DisabledPropertyName] = true;
+        if (typeof Object.isExtensible !== 'function' || (typeof Object.isExtensible === 'function' && Object.isExtensible(xhr))) {
+            xhr[DisabledPropertyName] = true;
```
MSNev: This property is required otherwise this would cause additional telemetry to be re-sent for the same request non-stop and unlike fetch, there is currently no additional alternative to stop this looping...

We took this approach (adding the DisabledPropertyName to the xhr) to avoid having to whitelist every possible URL used to report telemetry (not everyone uses the same/common URL, some people proxy the requests back to their own domain).

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -539,7 +539,9 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
      */
     private _xhrSender(payload: string[], isAsync: boolean) {
         const xhr = new XMLHttpRequest();
-        xhr[DisabledPropertyName] = true;
+        if (typeof Object.isExtensible !== 'function' || (typeof Object.isExtensible === 'function' && Object.isExtensible(xhr))) {
```
MSNev: While this is the correct and safe way to do this check, we have taken the approach of creating helper functions and then whitelisting specific usage.

See CoreUtils.objKeys(); CoreUtils.objDefineAccessors() etc for the helpers (Please create a helper for this method)

And then in the tools\rollup-es3 folder there is the source code for the "es3Poly()" and "es3Check()" rollup methods, both of these use tokens and exceptions listed in the Es3Tokens.ts file.

channels/applicationinsights-channel-js/src/Sender.ts
MSNev: We also assign the telemetry data against xhr as xhr.ajaxData (line 559 in the _openHandler()), removing this line would effectively make all XHR telemetry disabled...

Which means it's the same as setting the config value "disableAjaxTracking" to true to completely disable all XHR telemetry without any of these code changes. There are also some additional items within InstrumentHook.ts where we store the previous function and hook function calls against the individual methods of XHR.

Do you know if Salesforce platform also locks down window.fetch? As this would have a similar issue, So I'm not sure that this as is simple straight forward solution...

channels/applicationinsights-channel-js/src/Sender.ts
MSNev: While we prefer the helper approach, you can actually recode this portion "Object.isExtensible(xhr)" as "Object["isExtensible"](xhr) and this would fool the Es3Check() to allow the code. Which is why the wording for the error "or add explicit check for existence." you would still need the additional check that isExtensibile exists so that we maintain ES3 compatibility, we pick this up during code review.

channels/applicationinsights-channel-js/src/Sender.ts
dehru: @MSNev With the above change, i'm not seeing additional telemetry getting re-sent non-stop.  I see a pageview request.  And then we have a couple custom events that get batched and sent.  That all seems to be working....At least what I can see in the network terminal.  If it was a continual loop, i'd expect to see duplicate calls in the network tab of Chrome.

Screen Shot 2020-03-10 at 10 15 19 AM

If I run Object.isExtensible(fetch) it returns true;  Is there a way to configure this library to use fetch instead?

channels/applicationinsights-channel-js/src/Sender.ts
dehru: Thanks for your feedback.  I'll look at how to do this with a helper.  But it sounds like with your previous comments that this could cause some kind of looping behavior.  If I add this helper method and continue this effort, is this going to be possible to get this merged?

channels/applicationinsights-channel-js/src/Sender.ts
MSNev: It's probably not getting resent (looping) because of the other issue below where we populate the xhr.ajaxData -- this is caught internally. But this also means that automatic ajax instrumentation is not going to work either.

MSNev:
I need to create a patch version 2.5.2 to fix another issue, so I've created an issues which identifies the block you have and I'll try and unblock you (won't necessarily fix everything -- especially the automatic XHR tracking -- we can still work with this PR to find a solution for that if you wish

[BUG] App Insights fails when the XHR object is not extensible (or frozen) #1217

dehru:
Thanks for your help!  We are already using the App Insights nodejs sdk on our API and our VSCode extensions.  So adding telemetry to the browser-based portions of our app is the last dead-zone we have.

In the mean time, I'll work on the helper method you suggested above. And hopefully between that and what you can unblock for me, we can get something that works.

shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
```diff
@@ -396,6 +396,26 @@ export class CoreUtils {
         return false;
     }
 
+    /**
+     * Try to determine if the object is extensible when run within an >= ES5 container.
```
dehru: This is the helper I added to CoreUtils based on what I think you are suggesting.

shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
```diff
@@ -396,6 +396,26 @@ export class CoreUtils {
...
+     * For < ES5 containers where the method does not exist, always return true.
+     * @param target The object on which to determine extensibility.
+     * @returns True if the object is extensible.
+     */
+    public static isExtensible(obj:object) : boolean {
+        let extensibleProp = Object["isExtensible"];
+        if (extensibleProp) {
+            try {
+                return extensibleProp(obj);
```
MSNev: From a code minification perspective, this could be made *slightly* better by having 1 return statement

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -539,7 +539,9 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
      */
     private _xhrSender(payload: string[], isAsync: boolean) {
         const xhr = new XMLHttpRequest();
-        xhr[DisabledPropertyName] = true;
+        if (CoreUtils.isExtensible(xhr)) {
```
MSNev: The same problem can also occur if the object has bee frozen as well, in which case the the isExtensible() would still return true.

MSNev: The approach I took in my quick hack for this was to just wrap in a try / catch. My review is also fixing a lot of other stuff as well.

https://github.com/microsoft/ApplicationInsights-JS/pull/1218/files#diff-9e41298c570c651c39f1223741cb5cb7R553-R557

dehru: @MSNev Cool.  And you've added tests around this as well.  Do you think I should just close this PR and wait for your PR to merge and get released?

MSNev: Your call, but if my PR addresses your issues then perhaps close it and we can add the helpers leter.

MSNev: I'm also planning on getting my PR in today and releasing a version 2.5.2 with the fixes

dehru:
I tried your solution to just `try / catch` and it works fine.  We are still sending telemetry.  I'll close this.

Issue #1217, opened 2020-03-10
[BUG] App Insights fails when the XHR object is not extensible (or frozen)


MSNev:
(Taken from @dehru PR)

I'm unable to use ApplicationInsights-JS on the Salesforce Platform. Salesforce has locked-down extending the XMLHttpRequest object with either of these methods:

Object.preventExtensions(XMLHttpRequest) 
// or 
Reflect.preventExtensions(XMLHttpRequest)
This change allows telemetry to be tracked on the Salesforce platform. However, I'm not exactly sure the fallout of conditionally removing this property from xhr. I'm submitting this for your feedback.

Note: Also, I'm unable to satisfy the build because of ES3 support, though I believe I've sufficiently checked for existence before invoking the function. The build continues to complain with:

[Object.isExtensible(] is not supported in an ES3 environment, use a helper function or add explicit check for existence.
Thank you in advance for your time on my PR.

MSNev:
Fix is included in v2.5.2, will close this issues once this version is fully deployed to CDN

MSNev:
Version 2.5.2 is now fully deployed to the CDN, closing as this should address this issue.

Issue #1186, opened 2020-02-07
[BUG] App Insights initialization setting 'enableAjaxErrorStatusText' is not working


avilde:
**Description/Screenshot**
I was excited that I could replace my own workaround version to log failed ajax issues by retrieving also response body from XHR. #1156 

However I tested this new setting `enableAjaxErrorStatusText` and I could not get it to work. Instead I get an exception when remote dependency data is being logged.

Exception I found:
> "Failed to monitor XMLHttpRequest 'readystatechange' event handler, monitoring data for this ajax call may be incorrect." 
> Failed to read the 'responseText' property from 'XMLHttpRequest': The value is only accessible if the object's 'responseType' is '' or 'text' (was 'json').

**Steps to Reproduce**
* lunch application with app-insights turned on
* trigger any remote dependency where the response code for AJAX call is not successful.
* check application insights tracking information in Network Tab and in Azure Portal
* Instead of network dependency an exception is logged instead

- OS/Browser: 
Windows 10 Version 1909
Google Chrome Version 80.0.3987.87 (Official Build) (64-bit)

 - SDK Version (package.json):
    "@microsoft/applicationinsights-react-js": "^2.4.4",
    "@microsoft/applicationinsights-web": "^2.4.4"

 - How you initialized the SDK:
Here is my app insights telemetry class for logging client-side:
```Typescript
import { AppInsightsEventNames } from '../../constants';
import {
  ApplicationInsights,
  IConfiguration,
  IConfig,
} from '@microsoft/applicationinsights-web';
import { ReactPlugin } from '@microsoft/applicationinsights-react-js';
import { isDevEnvironment } from '../../utils/utils';

class TelemetryService {
  reactPlugin: ReactPlugin;
  appInsights!: ApplicationInsights;

  constructor() {
    this.reactPlugin = new ReactPlugin();
  }

  initialize(instrumentationKey: string) {
    const config = {
      instrumentationKey,
      maxBatchInterval: 0,
      enableRequestHeaderTracking: true,
      enableResponseHeaderTracking: true,
      enableAjaxErrorStatusText: true,
      enableDebugExceptions: isDevEnvironment(),
    } as IConfiguration & IConfig;

    this.appInsights = new ApplicationInsights({
      config: { ...config, extensions: [this.reactPlugin] },
    });

    this.appInsights.loadAppInsights();

    this.appInsights.trackEvent(
      { name: AppInsightsEventNames.AppInsightsInitialized },
[...]

PR #1218, opened 2020-03-10
[BUG] App Insights initialization setting 'enableAjaxErrorStatusText is not working #1186

MSNev:
[BUG] App Insights fails when the XHR object is not extensible (or frozen) #1217

### DIFF
--- a/channels/applicationinsights-channel-js/src/Sender.ts
+++ b/channels/applicationinsights-channel-js/src/Sender.ts
@@ -32,6 +32,16 @@ declare var XDomainRequest: {
 
 export type SenderFunction = (payload: string[], isAsync: boolean) => void;
 
+function _getResponseText(xhr: XMLHttpRequest | IXDomainRequest) {
+    try {
+        return xhr.responseText;
+    } catch (e) {
+        // Best effort, as XHR may throw while XDR wont so just ignore
+    }
+
+    return null;
+}
+
 export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
 
     public static constructEnvelope(orig: ITelemetryItem, iKey: string, logger: IDiagnosticLogger): IEnvelope {
@@ -335,7 +345,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
         if (xhr.readyState === 4) {
             let response: IBackendResponse = null;
             if (!this._appId) {
-                response = this._parseResponse(xhr.responseText || xhr.response);
+                response = this._parseResponse(_getResponseText(xhr) || xhr.response);
                 if (response && response.appId) {
                     this._appId = response.appId;
                 }
@@ -365,7 +375,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
             } else {
                 if (xhr.status === 206) {
                     if (!response) {
-                        response = this._parseResponse(xhr.responseText || xhr.response);
+                        response = this._parseResponse(_getResponseText(xhr) || xhr.response);
                     }
 
                     if (response && !this._senderConfig.isRetryDisabled()) {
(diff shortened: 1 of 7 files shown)


channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -539,12 +550,18 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
      */
     private _xhrSender(payload: string[], isAsync: boolean) {
         const xhr = new XMLHttpRequest();
-        xhr[DisabledPropertyName] = true;
-        xhr.open("POST", this._senderConfig.endpointUrl(), isAsync);
+        const endPointUrl = this._senderConfig.endpointUrl();
+        try {
+            xhr[DisabledPropertyName] = true;
+        } catch(e) {
+            // If the environment has locked down the XMLHttpRequest (preventExtensions and/or freeze), this would
```
markwolff: What sort of environments lock this down? How exactly do they lock it down? Would this be an issue for legacy SDK (we are fine to not update it, if so)?

extensions/applicationinsights-dependencies-js/Tests/Selenium/dependencies.tests.ts
```diff
@@ -1,6 +1,8 @@
-import { AjaxTests, AjaxPerfTests } from "./ajax.tests";
+import { AjaxTests, AjaxPerfTests, AjaxFrozenTests } from "./ajax.tests";
 
 export function runTests() {
     new AjaxTests().registerTests();
     new AjaxPerfTests().registerTests();
+    // The frozen tests are browser only as Phantom JS doesn't support enought of the spec to fail correctly
+    //new AjaxFrozenTests().registerTests();
```
markwolff: Are we meant to manually uncomment to test in browser? Is it possible to register these tests if we know we aren't in phantom?

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -65,6 +66,27 @@ function _supportsAjaxMonitoring(): boolean {
...
+        // Disable if the XmlHttpRequest can't be extended or hooked
+        try {
+            let xhr = new XMLHttpRequest();
+            xhr[strAjaxData] = {};
+
+            // Check that we can update the prototype
+            let theOpen = XMLHttpRequest[strPrototype].open;
+            XMLHttpRequest[strPrototype].open = theOpen;
+        } catch (e) {
+            // We can't decorate the xhr object so disable monitoring
```
markwolff: When/how would prototype object be locked down?

channels/applicationinsights-channel-js/src/Sender.ts
MSNev: Salesforce using Object.preventExtensions() and Object.freeze().

Its only now becoming an issue as environments start using these (ES5) features.
I don't think we want to backport to legacy, we should just take the approach that they should upgrade the SDK version if this situation occurs.

extensions/applicationinsights-dependencies-js/Tests/Selenium/dependencies.tests.ts
MSNev: Yes, as this is a patch I tested but didn't want to spend the time building the additional framework (at this point) to automate this.

extensions/applicationinsights-dependencies-js/src/ajax.ts
MSNev: The prototype is just an object, so it can also be locked down using the Object.preventExtensions(); Reflect.preventExtensions() and Object.freeze() directly on the XMLHttpRequest.prototype object.

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -359,12 +381,16 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
                 if (_config.disableFetchTracking === false) {
                     _hooks.push(InstrumentFunc(global, strFetch, {
                         // Add request hook
                         req: (callDetails: IInstrumentCallDetails, input, init) => {
                             let fetchData: ajaxRecord;
-                            if (_fetchInitialized && !_isDisabledRequest(null, input, init)) {
+                            if (_fetchInitialized && 
+                                    !_isDisabledRequest(null, input, init) && 
+                                    // If we have a polyfil and XHR instrumented then let XHR report otherwise we get duplicates
+                                    !(isPolyfill && _xhrInitialized)) {
```
MSNev: This is now required, as previously the XHR requests where actually throwing an internal exception (and not logging) because the responseType was "Blob".

With the responseType fix and not this check is was actually reporting the request for both the fetch and underlying (polyfilled) XHR request.

Issue #1232, opened 2020-03-31
[BUG] window.appInsights.properties is marked private


IGx89:
**Description/Screenshot**

README.md says that `window.appInsights.properties.context.telemetryTrace.traceID` is the way to get/set the trace ID, however the properties field is [marked private in the types](https://github.com/microsoft/ApplicationInsights-JS/blob/ebb1b6adc1b56fbcb4f086ef2baacad5cab1d7af/AISKU/src/Initialization.ts#L48) so I can't access it in my code. Either it's wrongly marked private, or I need a different way to get trace ID.

Thanks for your help! Use case is to show the trace ID in the error message that we show the user after a service call failure.

IGx89:
My ugly workaround:
```js
// Pull the W3C trace context ID from AI and add it to the error message.
// See https://github.com/microsoft/ApplicationInsights-JS/issues/1232 for why the any cast is needed.
// eslint-disable-next-line @typescript-eslint/no-explicit-any
const w3cTraceId = (window.appInsights as any).properties.context.telemetryTrace.traceID;
```

MSNev:
Hi,

After investigating this issue you don't need your ugly workaround, instead you will find that the appInsights object already has a public 'context' value which is a reference to the .properties.context instance.

So the issue is basically a documentation one (after calling loadAppInsights), so I'm going to update the README to read (as below) and ensure that the context value is assigned during the construction of the instance as opposed to only during the loadAppInsights() method as it is today.

`window.appInsights.context.telemetryTrace.traceID`

IGx89:
Thanks! `window.appInsights.context` is always undefined for me when using "Snippet Setup", so if your code change makes that work then that's exactly what I need :)

MSNev:
There was an issue with the way this value was propagated to the snippet, which I've fixed with the #1236 PR, this will be included in the next release

MSNev:
Fixed deployed as NPM and next CDN channels

MSNev:
Now also fully deployed to primary CDN

PR #1236, opened 2020-04-01
[BUG] window.appInsights.properties is marked private #1232

MSNev:


AISKU/src/Initialization.ts
```diff
@@ -278,6 +279,7 @@ export class Initialization implements IApplicationInsights {
         extensions.push(_this.properties);
         extensions.push(_this.dependencies);
         extensions.push(_this.appInsights);
+        _this.context = _this.properties.context;
```
MSNev, line 282: the context value of properties is only populated during the initialize() call

MSNev: And "this" is not the snippet, so the aisdk referenced from the snippet is not this object so context is undefined in that case. Need to extendthe ApplicationINsightsContainer to copy the post initialization properties back to the snippet.

Issue #1240, opened 2020-04-02
[BUG] Telemetry correlation headers are not included for all fetch requests


MSNev:
I am experiencing the same as @tcihak-fqa, and I have a similar setup.

After upgrading @microsoft/applicationinsights-web to v2.5.3 (from v2.4.4), correlation headers are missing in all `fetch` requests.

I tested to enable the new W3C distributed tracing mode
```javasript
{
    // ...
    distributedTracingMode: DistributedTracingModes.W3C
    // ...
}
```
...still no correlation headers of any sort.
No other distributedTracingMode options seem to work either. 

This must be a bug, it worked before upgrading.

@MSNev :
I'm running an ASP.NET core app with kestrel on https://localhost:5001.

No fetch polyfill is used (polyfill for fetch is dynamically loaded only if the browser does not have it natively - with the browser I use for testing the native fetch API is supported).

I use JS modules (traspiled from TypeScript) and import @microsoft/applicationinsights-web - it is included in webpack bundling from node_modules.

The HTML/CSS/JS code is hosted in the same web application; on the same domain/port as the API I'm calling with fetch (where the correlation headers are missing). No CORS should be required.

FYI: It made no difference to add "localhost:5001" to correlationHeaderDomains,

I had a few more configuration settings than in my code below, but I have tested stripping the configuration down to a minimum as follows and I still have the problem:

```javascript

import { ApplicationInsights, DistributedTracingModes } from "@microsoft/applicationinsights-web";

// ...
            const ai = new ApplicationInsights({
                config: {
                    //correlationHeaderDomains: ["localhost:5001"], // Made no difference
                    instrumentationKey: "the-instrumentation-key",
                    disableAjaxTracking: false,
                    disableCorrelationHeaders: false,
                    disableFetchTracking: false,
                    distributedTracingMode: DistributedTracingModes.AI, // .AI_AND_W3C or .W3C also does not add any correlation headers
                }
            });

            ai.loadAppInsights();


```

@josundt I think I found the root cause for your issue, basically the code is adding the headers to the intercepted fetch request *however* if you don't pass the optional 2nd parameter (init), it creates one internally and populated the headers on this object... BUT, it doesn't pass this onto the real fetch!!!

As a workaround, if you pass a value as part of this 2nd parameter the headers should be sent correctly.

MSNev:
Fixed deployed as NPM and next CDN channels

MSNev:
Now also fully deployed to primary CDN

Issue #1229, opened 2020-03-28
[BUG] Unable to include telemetry correlation headers


tcihak-fqa:
**Description**
We using this SDK in an SPA using fetch.
The requests to the my .NET Web APIs do not include the telemetry correlation headers ('Request-Id' and 'Request-Context').
Dependency telemetry is being collected just fine.

```
Request Headers
  Request URL: https://api.domain.com:8082/resource
  Request Method: GET
  Status Code: 200 
  Referrer Policy: no-referrer-when-downgrade
  :authority: api.domain.com:8082
  :method: GET
  :path: /resource
  :scheme: https
  accept: application/json
  accept-encoding: gzip, deflate, br
  accept-language: en-US,en;q=0.9,es;q=0.8,uk;q=0.7
  content-type: application/json
  origin: https://client.domain.com
  referer: https://client.domain.com
  sec-fetch-dest: empty
  sec-fetch-mode: cors
  sec-fetch-site: cross-site
  user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) 
  Chrome/80.0.3987.149 Safari/537.36

Response headers
  access-control-allow-credentials: true
  access-control-allow-origin: https://client.domain.com
  access-control-expose-headers: Request-Context
  content-encoding: gzip
  content-length: 361
  content-type: application/json; charset=utf-8
  date: Sat, 28 Mar 2020 17:18:46 GMT
  request-context: appId=cid-v1:<some id>
  server: Microsoft-IIS/10.0
  status: 200
  x-powered-by: ASP.NET
```

 - OS/Browser:
Windows / Chrome (Version 80.0.3987.149)

 - SDK Version [e.g. 22]:
Version 2.5.2

 - How you initialized the SDK:
NPM setup / webpack

```
import { ApplicationInsights } from '@microsoft/applicationinsights-web';

export class Logger {
  _ai;

  constructor() {
    this._init();
  }

  _init() {
    var options = {
     config: {
       instrumentationKey: 'my key',
       appId: 'my-app',
       autoTrackPageVisitTime: true,
       disableAjaxTracking: true,
       disableFetchTracking: false,
       enableCorsCorrelation: true,
       enableRequestHeaderTracking: true,
       enableDebug: true,
       loggingLevelConsole: 2,
    }
    this._ai = new ApplicationInsights(options);
    this._ai.loadAppInsights();
  }
[...]

PR #1241, opened 2020-04-03
[BUG] Telemetry correlation headers are not included for all fetch requests #1240

MSNev:


josundt:
Please see my additional comments added to the issue; maybe it could all be fixed in this PR:

https://github.com/microsoft/ApplicationInsights-JS/issues/1229#issuecomment-608419912

MSNev:
> Please see my additional comments added to the issue; maybe it could all be fixed in this PR:
> 
> [#1229 (comment)](https://github.com/microsoft/ApplicationInsights-JS/issues/1229#issuecomment-608419912)

After adding specific tests for this my original fix also resolved the Request() argument issues as well -- as the same basic problem was occuring, but I did find that when enabling enableRequestHeaderTracking the headers where not being included either -- now fixed

josundt:
When is the next planned release?
Will this PR be reviewed/accepted/merged before that?
I hope so since it is a blocking bug for many that should be fixed as soon as possible.

MSNev:
I was planning on starting the process for a dot release (2.5.4) as soon as these PR's (tslib and this one) are committed. So this should be fully deployed by the end of the week and optimistically we should have an NPM package later today (PST timezone)

Issue #1227, opened 2020-03-25
SPFX - undefinedundefined is not defined


enti333:
Hello. I try to use this code in my spfx application. 

import { ApplicationInsights } from '@microsoft/applicationinsights-web'

const appInsights = new ApplicationInsights({ config: {
  instrumentationKey: 'MY KEY'
  /* ...Other Configuration Options... */
} });
appInsights.loadAppInsights();
appInsights.trackPageView(); // Manually call trackPageView to establish the current user/session/pageview

Evrything is correctly deffined after my direct testing it was working. 

But if I insert this code to spfx in render function I get this error 

[SPLoaderError.loadComponentError]:
***Failed to load component "aea736a1-9c8e-4afc-891c-6487bb08d2a9" (OFormWebPart).
Original error: ***Failed to load entry point from component "aea736a1-9c8e-4afc-891c-6487bb08d2a9" (OFormWebPart).
Original error: Error loading https://component-id.invalid/aea736a1-9c8e-4afc-891c-6487bb08d2a9_0.0.1
	undefinedundefined is not defined

***INNERERROR:
***Failed to load entry point from component "aea736a1-9c8e-4afc-891c-6487bb08d2a9" (OFormWebPart).
Original error: Error loading https://component-id.invalid/aea736a1-9c8e-4afc-891c-6487bb08d2a9_0.0.1
	undefinedundefined is not defined
***CALLSTACK:
Error
    at SPLoaderError.SPError [as constructor] (https://localhost:4321/temp/workbench-packages/@microsoft_sp-loader/dist/sp-loader-assembly_default.js:8390:24)
    at new SPLoaderError (https://localhost:4321/temp/workbench-packages/@microsoft_sp-loader/dist/sp-loader-assembly_default.js:4036:28)
    at Function.ErrorBuilder.buildErrorWithVerboseLog (https://localhost:4321/temp/workbench-packages/@microsoft_sp-loader/dist/sp-loader-assembly_default.js:3789:21)
    at Function.ErrorBuilder.buildLoadComponentError (https://localhost:4321/temp/workbench-packages/@microsoft_sp-loader/dist/sp-loader-assembly_default.js:3697:21)
    at https://localhost:4321/temp/workbench-packages/@microsoft_sp-loader/dist/sp-loader-assembly_default.js:6275:47

When I remove this code my spfx is correctly working....
Hmm...What is wrong ... Thank you.

MSNev:
[...]

PR #1244, opened 2020-04-07
SPFX - undefinedundefined is not defined #1227

MSNev:


AISKU/rollup.config.js
```diff
@@ -19,28 +18,6 @@ const replaceValues = {
   "// Licensed under the MIT License.": ""
 };
 
-// Function to remove the @DynamicProtoStubs and rewrite the headers for the dist-esm files
```
MSNev, line 22: I believe the issue this problem was occuring for users directly consuming dist-esm modules files is because the previous change to change the copyright header (to include the version number) and the removal of the dynamicProto() stub functions in some cases caused some decoders of the map file throwing and reporting undefined into the output.

While I didn't repo the issue directly, I did find that by using this site to visual the map files showed a combination of an error for substr() and or additional "undefined" values being added to the end of the visualation.
http://sokra.github.io/source-map-visualization/#custom

tools/updateDistEsm/updateDistEsm.js
```diff
@@ -0,0 +1,138 @@
...
+  
+  const updateDistEsmFiles = (replaceValues, banner) => {
+    const dynRemove = dynamicRemove();
+    const files = globby.sync("./dist-esm/**/*.js");
+    files.map(inputFile => {
+      console.log("Loading - " + inputFile);
+      var src = fs.readFileSync(inputFile, "utf8");
+      var mapFile;
+      if (inputFile.endsWith(".js")) {
+        mapFile = inputFile.substring(0, inputFile.length) + ".map";
```
MSNev, line 51: Just add to the inputFile name

PR #1249, opened 2020-04-15
Identify whether the script is being consumed via the CDN or NPM package

MSNev:
…

### DIFF
--- a/AISKU/src/Initialization.ts
+++ b/AISKU/src/Initialization.ts
@@ -10,6 +10,8 @@ import * as Common from "@microsoft/applicationinsights-common"
 
 "use strict";
 
+let _internalSdkSrc: string;
+
 /**
  *
  * @export
@@ -18,6 +20,7 @@ import * as Common from "@microsoft/applicationinsights-common"
 export interface Snippet {
     config: IConfiguration & Common.IConfig;
     queue?: Array<() => void>;
+    sv?: string;
     version?: number;
 }
 
@@ -46,10 +49,12 @@ export class Initialization implements IApplicationInsights {
 
     private dependencies: DependenciesPlugin;
     private properties: PropertiesPlugin;
+    private _snippetVersion: string;
 
     constructor(snippet: Snippet) {
         let _this = this;
         // initialize the queue and config in case they are undefined
+        _this._snippetVersion = "" + (snippet.sv || snippet.version || "");
         snippet.queue = snippet.queue || [];
         snippet.version = snippet.version || 2.0; // Default to new version
         let config: IConfiguration & Common.IConfig = snippet.config || ({} as any);
@@ -269,6 +274,15 @@ export class Initialization implements IApplicationInsights {
 
         function _updateSnippetProperties(snippet: Snippet) {
             if (snippet) {
+                let snippetVer = "";
+                if (!CoreUtils.isNullOrUndefined(_this._snippetVersion)) {
+                    snippetVer += _this._snippetVersion;
+                }
+                if (legacyMode) {
+                    snippetVer += ".lg";
+                }
+                _this.properties.context.internal.snippetVer = snippetVer || "-";
+
                 // apply updated properties to the global instance (snippet)
                 for (const field in _this) {
                     if (CoreUtils.isString(field) && 
@@ -296,6 +310,9 @@ export class Initialization implements IApplicationInsights {
         // initialize core
         _this.core.initialize(_this.config, extensions);
         _this.context = _this.properties.context;
+        if (_internalSdkSrc) {
+            _this.context.internal.sdkSrc = _internalSdkSrc;
+        }
         _updateSnippetProperties(_this.snippet);
 
         // Empty queue of all api calls logged prior to sdk download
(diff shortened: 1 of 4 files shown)


extensions/applicationinsights-properties-js/src/TelemetryContext.ts
```diff
@@ -29,140 +29,159 @@ export class TelemetryContext implements ITelemetryContext {
...
+            if (CoreUtils.isString(internal.agentVersion)) {
+                event.tags[CtxTagKeys.internalAgentVersion] = internal.agentVersion; // not mapped in CS 4.0
+            }
+            if (CoreUtils.isString(internal.sdkVersion)) {
+                event.tags[CtxTagKeys.internalSdkVersion] = internal.sdkVersion;
             }
-            if (typeof this.internal.sdkVersion === "string") {
-                event.tags[CtxTagKeys.internalSdkVersion] = this.internal.sdkVersion;
+
+            if (event.baseType === _InternalLogMessage.dataType || event.baseType === PageView.dataType) {
```
MSNev, line 117: Limited to the Internal messages and initial page view

AISKU/src/Initialization.ts
```diff
@@ -411,3 +427,49 @@ export class Initialization implements IApplicationInsights {
...
+        }
+    } catch (e) {
+    }
+
+    if (sdkSrc) {
+        try {
+            let url = sdkSrc.toLowerCase();
+            if (url) {
+                let src = "";
+                if (url.indexOf("://az416426.vo.msecnd.net/") !== -1) {
```
MSNev, line 457: Only populate if this is from the standard CDN -- ignore non MS defined endpoints, and only include an anonymous name "cdn1" rather than the full domain/url

AISKU/src/Initialization.ts
```diff
@@ -411,3 +427,49 @@ export class Initialization implements IApplicationInsights {
...
+    try {
+        // Try and determine whether the sdk is being loaded from the CDN
+        // currentScript is only valid during initial processing
+        let scrpt = (document ||{} as any).currentScript;
+        if (scrpt) {
+            sdkSrc = scrpt.src;
+        // } else {
+        //     // We need to update to at least typescript 2.9 for this to work :-(
+        //     // Leaving as a stub for now so after we upgrade this breadcrumb is available
+        //     let meta = import.meta;
```
MSNev, line 445: Tried quite a few work-arounds to get this information with 2.5.3 (our current TS version), but nothing worked -- I could have done some sort of post processing (like the es3-poly), but that seems like overkill considering we *should* be upgrading to a later version of TS  sooner rather than later.

AISKU/src/Initialization.ts
```diff
@@ -411,3 +427,49 @@ export class Initialization implements IApplicationInsights {
...
+                        }
+                    }
+                }
+
+                if (src) {
+                    _internalSdkSrc = src + (isModule ? ".mod" : "");
+                }
+            }
+        } catch (e) {
+        }
```
markwolff, line 473: When caught, does that mean NPM? Is it possible to know NPM vs CDN vs unknown?

MSNev: This catch is unexpected and added purely to ensure that we don't crash, in which case this will mean that we don't know the outcome (same as what would happen with all IE and really old browsers).

So it's not possible to "always" determine npm, cdn or unknown. This is also why I added the snippet property as it should always be defined in which case we could assume that snippet users are using the primary CDN (vs their own).

markwolff: ok so for now it will report primary CDN or (NPM + custom CDN + unknown)?

MSNev: It will report primary CDN or nothing.
And when nothing we can assume that its NPM if also there is no snippet.

AISKU/src/Initialization.ts
```diff
@@ -46,10 +48,12 @@ export class Initialization implements IApplicationInsights {
 
     private dependencies: DependenciesPlugin;
     private properties: PropertiesPlugin;
+    private _snippetVersion: string;
 
     constructor(snippet: Snippet) {
         let _this = this;
         // initialize the queue and config in case they are undefined
+        _this._snippetVersion = "" + (snippet.version || "");
```
MSNev, line 56: This will need to be updated with the new snippet changes

AISKU/src/Initialization.ts
```diff
@@ -18,6 +20,7 @@ import * as Common from "@microsoft/applicationinsights-common"
 export interface Snippet {
     config: IConfiguration & Common.IConfig;
     queue?: Array<() => void>;
+    snipVer?: number;
```
MSNev, line 23: This is added in my snippet version updates #1258

AISKU/src/Initialization.ts
ramthi: If the script is consumed from CDN via the CName will we have to update this to match those domain names?

MSNev: Yes, once we have enabled additional common names this will need to be updated and I'm planning on enumerating the known names (which is why I called the existing one cdn1)

PR #1258, opened 2020-04-21
Update snippet to support reporting script load failures

MSNev:
- New Snippet Setup pollutes the global namespace (window) #974


AISKU/snippet/snippet.js
```diff
@@ -1,106 +1,304 @@
...
-            // Queue-up a call to the real method
-            appInsights.queue.push(function () {
-                // Invoke the real method with the captured original arguments
-                appInsights[name].apply(appInsights, originalArguments);
-            });
-        };
-    }
+﻿(function (win, doc, snipConfig) {
+    var locn = win.location;
+    var helpLink = "https://aka.ms/<TBD>";
```
MSNev, line 3: This still needs to be updated before I can check this in

AISKU/snippet/snippet.js
```diff
@@ -1,106 +1,304 @@
...
-    var appInsights = {
-        config: aiConfig
-    };
-    appInsights.initialize = true; // initialize sdk on download
-
-    // Assigning these to local variables allows them to be minified to save space:
-    var localDocument = document, localWindow = window, scriptText = "script";
-    setTimeout(function () {
-        var scriptElement = localDocument.createElement(scriptText);
-        scriptElement.src = aiConfig.url || "https://az416426.vo.msecnd.net/scripts/b/ai.2.min.js";
```
MSNev, line 14: I removed the aiConfig.url and replaced it with the snippetConfig.src.

I assume this is ok as users will be updating to this new snippet and the url is not used elsewhere.

shared/AppInsightsCore/src/JavaScriptSDK.Enums/LoggingEnums.ts
```diff
@@ -92,6 +92,7 @@ export const _InternalMessageId = {
     InvalidEvent: 70,
     FailedMonitorAjaxSetRequestHeader: 71,
     SendBrowserInfoOnUserInit: 72,
-    PluginException: 73
+    PluginException: 73,
+    SnippetScriptLoadFailure: 99
```
xiao-lix, line 96: why is this not 74? 😬

AISKU/src/Init.ts
```diff
@@ -35,7 +35,7 @@ try {
 
                 // overwrite snippet with full appInsights
                 // for 2.0 initialize only if required
```
xiao-lix, line 37: Also update the document here?

AISKU/snippet/snippet.js
```diff
@@ -1,106 +1,304 @@
...
+        // Only set if supplied or another name is defined to avoid polluting the global namespace
+        win[sdkInstanceName] = aiName;
     }
+    var aisdk = win[aiName] || (function (aiConfig) {
+        var loadFailed = false;
+        var handled = false;
+        var appInsights = {
+            initialize: true,   // initialize sdk on download
+            queue: [],
+            snipVer: 3.0,       // Track the actual snippet version for reporting.
```
xiao-lix, line 27: Does this snipVer need to update every time SDK updates, like all plugin versions?

shared/AppInsightsCore/src/JavaScriptSDK.Enums/LoggingEnums.ts
MSNev: I just wanted to move it to something that was obvious and easy to recall from an ICM :-)

AISKU/snippet/snippet.js
MSNev: The snipVer is only for supporting the version of the snippet that people are using, so ideally yes. But it's not a critical change. I'd update it only when we either fix some critical change or add some new feature(s). So that when reported we can just say -- updaste to the latest version of the snippet if they are expecting something that is supported in a later version.

AISKU/src/Init.ts
MSNev: Sure

AISKU/snippet/snippet.js
MSNev: After thinking about this for a while, I've re-added this so we don't break anyone.

AISKU/snippet/snippet.js
```diff
@@ -1,106 +1,304 @@
...
+            }
+        }
+
+        function _reportFailure(targetSrc) {
+            var conString = _parseConnectionString();
+            var iKey = conString[strInstrumentationKey] || aiConfig[strInstrumentationKey] || strEmpty;
+            var ingest = conString[strIngestionendpoint];
+            var endpointUrl = ingest ? ingest + "/v2/track" : config.endpointUrl; // only add /v2/track when from connectionstring
+
+            var message = "SDK LOAD Failure: Failed to load App Insights Sdk script (See stack for details)";
```
MSNev, line 78: Updated per Ram's email comments

AISKU/snippet/snippet.js
```diff
@@ -1,106 +1,304 @@
...
+
+        function _createException(iKey, message, targetSrc) {
+            var envelope = _createEnvelope(iKey, "Exception");
+            var data = envelope.data;
+            data.baseType = "ExceptionData";
+            data.baseData.exceptions = [{
+                typeName: "ScriptLoadFailed",
+                message: message.replace(/\./g, "-"),  // Replacing '.' characters as it causes the portal to hide the start of the message in the summary
+                hasFullStack: false,
+                stack: message + "\nSnippet failed to load [" + targetSrc + "] -- Telemetry is disabled\nHelp Link: " + helpLink + "\nHost: " + (locn && locn.pathname || "_unknown_"),
```
MSNev, line 149: Updated per Ram's email comments

AISKU/snippet/snippet.js
```diff
@@ -1,106 +1,304 @@
...
+        var loadFailed = false;
+        var handled = false;
+        var appInsights = {
+            initialize: true,   // initialize sdk on download
+            queue: [],
+            snipVer: 3.0,       // Track the actual snippet version for reporting.
+            version: 2.0,       // initialization version, if this is not 2.0 the previous scripts fail to initialize
+            config: aiConfig
+        };
+        function _parseConnectionString() {
```
ramthi, line 31: sorry Im confused how connectionstring would come into the picture when someone uses a snippet.  Isnt connectionstring something that is configured in the portal as an env variable and used when JS is injected?  Is there another use case i'm missing?  Direct question is:  Why are you adding connectionstring support as part of this change?

AISKU/snippet/snippet.js
ramthi: I still see App Insights.  We should call this Application Insights.  and capitialize Sdk?

AISKU/snippet/snippet.js
```diff
@@ -1,106 +1,304 @@
...
+    var aiName = snipConfig.name || "appInsights";  // provide non default instance name through snipConfig name value
+    if (snipConfig.name || win[sdkInstanceName]) {
+        // Only set if supplied or another name is defined to avoid polluting the global namespace
+        win[sdkInstanceName] = aiName;
     }
+    var aisdk = win[aiName] || (function (aiConfig) {
+        var loadFailed = false;
+        var handled = false;
+        var appInsights = {
+            initialize: true,   // initialize sdk on download
```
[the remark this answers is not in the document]
MSNev: Mostly no as this would break backward compatibility. Older scripts (and the current) require
queue - this is the documented way applications "wait" to process extra initialization until after the SDK is ready
initialize - without this the SDK wont call getAppInsights()
version - same as initialize, but also needs to be "2.0" (this is why I added snipVer so we can track the actual snippet version)
config - same again backward compatibility
snipVer - I could reduce to sv, would need to update other PR as well -- will consider

AISKU/snippet/snippet.js
MSNev: Lets discuss offline to clarify whether this is really required and if I've misunderstood the usage scenario

AISKU/snippet/snippet.js
markwolff: Would also add JavaScript SDK to the message, if it isn't visible anywhere else.

AISKU/snippet/snippet.js
```diff
@@ -1,106 +1,304 @@
...
+                hasFullStack: false,
+                stack: message + "\nSnippet failed to load [" + targetSrc + "] -- Telemetry is disabled\nHelp Link: " + helpLink + "\nHost: " + (locn && locn.pathname || "_unknown_"),
+                parsedStack: []
+            }];
+
+            return envelope;
+        }
+    
+        // Assigning these to local variables allows them to be minified to save space:
+        var targetSrc = aiConfig.url || snipConfig.src;
```
markwolff, line 157: for the stuff other than what you are adding for handling snippet failures, are there any other behavior changes?

AISKU/snippet/snippet.js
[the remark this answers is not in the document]
MSNev: Updated snipVer to sv

AISKU/snippet/snippet.js
MSNev: No - Apart from "adding" the snipConfig (src, name, ld, useXhr -- see Readmd.md for documentation) which mostly are just exposing the previously internal only values -- so we just define them once

AISKU/snippet/snippet.js
ramthi: We wanted to keep the text data to min so we can minimize on snippet bloat.  Fact that we are listing .js file in the error should be good enough IMO.

Issue #1274, opened 2020-05-28
[BUG] WithAITracking cannot wrap around functional components


Ishraque99:
**Description/Screenshot**
![image](https://user-images.githubusercontent.com/30386992/83147091-8fbe2600-a13a-11ea-9437-7bc0375cdedf.png)
Functional components (ala ES6) have undefined prototype and no constructor name. While we can provide a name parameter to WithAITracking I don't think it makes sense as these components are meant to be anonymous/nameless.

**Steps to Reproduce**
- create a functional component as such and wrap it `withAITracking()`
```TS
const Test = (props: Props) => {
    return (<div>
        <p>This is a test component</p>
    </div>);
}
...
export default withAITracking(reactPlugin, Test)
```

 - OS/Browser: Ubuntu 20.04/Firefox
 - applicationinsights-react-js version: ^2.5.4
 - How you initialized the SDK: following the basic usage guideline in the README for applicationinsights-react-js

**Expected behavior**
Component should load as normal

**Possible fix**
**Edit:** Added PR #1275 :)
I think it's a simple fix. Update WithAITracking as such.
```TS
  if (componentName === undefined || componentName === null || typeof componentName !== 'string') {
    componentName = Component.prototype && 
          Component.prototype.constructor && 
          Component.prototype.constructor.name ||
          'Anonymous component';
  }
```
instead of "Anonymous component" it could also be an empty string, just like the default for className.

MSNev:
Included in v 2.5.5 which is now fully deployed

Issue #1117, opened 2019-11-16
[BUG] [react-plugin] dist/applicationinsights-react-js.js is broken


kajoik:
**Description/Screenshot**
If I open dist/applicationinsights-react-js.js from the @microsoft/applicationinsights-react-js package (version 2.3.1), React references seem to be broken at least at L3315, 3308 and 3309. L3315 at least seem to work if I replace `undefined` with `React.default.Component`.

**Steps to Reproduce**
Run `npm pack @microsoft/applicationinsights-react-js@2.3.1`
Unpack and open the file mentioned above.

**Expected behavior**
References to React objects are being resolved correctly, for example as mentioned above.

**Additional context**
Pull request #920 seems to introduce these `undefined`s.

markwolff:
For a React app, I would imagine it would be using the `dist-esm` for the modules since they are packaged as ES Modules. Are you using webpack?

franzk:
I'm using a CRA, and it's not happening with the webpack dev server, but in jest. I know that the `withAITracking` wrapper is somehow highly optional in unit testing. Personally, I now added a handwritten equivalent for this function. Perhaps, the dist/ folder should either be fixed or removed, but feel free to close this issue if you think it's not relevant.

jmmccota:
I'm using default CRA and I'm getting an equivalent error.

```
  14 | 
  15 | export default function withAITracking(reactPlugin, Component, componentName, className) {
  16 |   if (componentName === undefined || componentName === null || typeof componentName !== 'string') {
> 17 |     componentName = Component.prototype.constructor.name;
  18 |   }
  19 | 
  20 |   if (className === undefined || className === null || typeof className !== 'string') {
```

irnc:
This bug was implicitly fixed by change introduced at https://github.com/microsoft/ApplicationInsights-JS/commit/826504ccbc49ca435a41a56d9b6c6ba9487ea5ca (`rush` not used anymore for react-js extension; causing `rollup` to run with node_modules local to react-js extension). This commit is released as part of `^3.0.0`, so install latest package version to fix browser build.

Issue was due to `namedExports` (see roollup config) configured for `node_modules/react/index.js` while during rush build (prior to 3.0.0) path is `../../common/temp/node_modules/react/index.js`, thus causing a mismatch.

PR #1275, opened 2020-05-28
fix for withAITracking wrapping functional components.

Ishraque99:
fix for issues #1274 and #1117

Can also use empty string instead of 'Anonymous component'. Let me know which is preferred.


extensions/applicationinsights-react-js/src/withAITracking.tsx
```diff
@@ -17,7 +17,10 @@ import ReactPlugin from './ReactPlugin';
 export default function withAITracking<P>(reactPlugin: ReactPlugin, Component: React.ComponentType<P>, componentName?: string, className?: string): React.ComponentClass<P> {
 
   if (componentName === undefined || componentName === null || typeof componentName !== 'string') {
-    componentName = Component.prototype.constructor.name;
+    componentName = Component.prototype && 
+          Component.prototype.constructor && 
+          Component.prototype.constructor.name || 
+          'Anonymous component';
```
MSNev, line 23: I'd go with with either just "Anonymous" or perhaps "Unknown"

Ishraque99: I like "Unknown". Updated.

Issue #1269, opened 2020-05-18
tslib v1.13.0 has breaking change


MSNev:
tslib v 1.13.0 is causing a compile error due to a breaking change

Compiling...
Using tsc v2.5.3
common/temp/node_modules/tslib/tslib.d.ts(37,78): error TS2304: Cannot find name 'PropertyKey'.
common/temp/node_modules/tslib/tslib.d.ts(37,103): error TS2304: Cannot find name 'PropertyKey'.

We have created an issue with tslib to try and get this fixed, if not we will need to roll back and be fixed to v 1.11.2. As a temporary workaround if you have this issue is to update your local package.json files to reference "1.11.2" instead of "^1.11.1"

Raised TSLib Issue
https://github.com/microsoft/tslib/issues/115

MSNev:
As this is the 2nd time in recent history and looking at how the tslib project is now structured, it appears that it's going to remain busted for us as we are currently targeting es3 as our base module to ensure that we don't accidently break our SDK user base that has users that are currently still relying on ES3 based browsers.

As such, I'm going to completely break our dependency on tslib, without the normal corresponding code bloat (as size is still a major concern). But I'm attempting to do this in a backward compatible way so that we don't break
- CDN only users
- Older TypeScript users ( < 3.x)
- More modern TypeScript users (i.e. 3.x)
- Anyone how is currently user *any* version of tslib 
  - as previously we broke users using TypeScript and a newer version of tslib then we needed to be pinned to (v1.10.1 from memory
  - This is also a major consideration for breaking the dependency, so we don't break anyone who wants/needs to use a version of tslib > 1.11.2 (the last known good version that we can compile with)

nicolashenry:
@MSNev Could you release a version still using tslib at least for esm version? The recent change create pollution in global object.

MSNev:
Created #1280 to investigate how we could possibly stop exposing the global functions.

nicolashenry:
> Is it just the __extends() or __assign() or are you seeing something else in the global scope

Both __extends and __assign but nothing else.

> Which modules / files are you ingesting (using) [i.e. just the AISku or the individual components, if the later which ones).

I am only using "@microsoft/applicationinsights-web" module (and indirectly its dependencies).

[...]

PR #1277, opened 2020-05-28
tslib v1.13.0 has breaking change #1269

MSNev:
- Removed Dependency on tslib


AISKU/Tests/Selenium/Tests.html
```diff
@@ -17,16 +17,6 @@
     <div id="qunit"></div>
     <div id="qunit-fixture"></div>
     <script>
-        function getParameterByName(name, url) {
```
MSNev, line 20: Moved to module loader script

common/Tests/Selenium/ModuleLoader.js
```diff
@@ -20,7 +61,10 @@ function ModuleLoader(config) {
             require([moduleDef.path], function (theModule) {
                 try {
                     if (moduleDef.run) {
-                        moduleDef.run(theModule);
+                        doModuleCb(moduleDef, theModule, function(theModule) {
```
MSNev, line 64: This was a bug fix that was stopping direct usage for the tests.

extensions/applicationinsights-angular-js/test-app/tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "experimentalDecorators": true,
     "module": "esnext",
     "moduleResolution": "node",
-    "importHelpers": true,
+    "importHelpers": false,
```
MSNev, line 12: This one is probably wrong -- will check and possible revert

gruntfile.js
```diff
@@ -34,12 +34,12 @@ module.exports = function (grunt) {
                 tsconfig: './shared/AppInsightsCore/tsconfig.json'
             },
             coretest: {
-                tsconfig: './shared/AppInsightsCore/src/JavaScriptSDK.Tests/tsconfig.json',
```
MSNev, line 37: I moved the core tests out from the src folder, so they are now in the same place as all of the other components

shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
```diff
@@ -1,10 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 "use strict";
-import { getWindow, getDocument, strUndefined, strObject, strFunction, strPrototype, getCrypto } from './EnvUtils';
+import { objCreateFn, strShimObject, strShimUndefined, strShimFunction, strShimPrototype } from "@microsoft/applicationinsights-shims";
```
MSNev, line 4: Reusing these properties to aid with minification for the full Sku's.

tools/rollup-es3/tsconfig.json
```diff
@@ -7,8 +7,8 @@
     "moduleResolution": "node",
     "target": "es3",
     "forceConsistentCasingInFileNames": true,
-    "importHelpers": true,
-    "noEmitHelpers": true,
+    "importHelpers": false,
+    "noEmitHelpers": false,
```
MSNev, line 11: This is the normal approach for breaking the tslib dependency, using this here as we don't have multiple __extend usages, so it's not as much of an issue for this development only tool

tools/shims/src/applicationinsights-shims.ts
```diff
@@ -0,0 +1,111 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
MSNev, line 1: This IS the main shim file that gets included

tools/shims/src/applicationinsights-shims.ts
```diff
@@ -0,0 +1,111 @@
...
+};
+
+export function __extendsFn(d: any, b: any) {
+    __extendStaticsFn(d, b);
+    function __() { this.constructor = d; }
+    // tslint:disable-next-line: ban-comma-operator
+    d[strShimPrototype] = b === null ? objCreateFn(b) : (__[strShimPrototype] = b[strShimPrototype], new (__ as any)());
+}
+
+let globalObj:any = getGlobal() || {};
```
MSNev, line 99: From here on is the primary "tricks" to wire up to the global __extends; __assign and local versions (lines 109, 110).
The primary change that this introduces (as opposed to not using tslib), is that moving forward *if* the consumer (of the npm packages) has defined their own __extends() and/or __assign() then this code WILL use them, previously the tslib versions where the only possible usages because tslib was a injected (via require) and used as tslib_1.__extend() rather than the global versions.

extensions/applicationinsights-angular-js/test-app/tsconfig.json
MSNev: Kept the removal of tslib, but set the noEmitHelpers to false, so that typescript drops out the __extends() rather than either always using tslib or using the shims modules, as that should only be required for internal AI usage and as long as an __extends() and/or __assign() exists everything should work.

PR #1278, opened 2020-06-02
Add Send Request notification

MSNev:
Adds a notification for when events are about to be sent to the server

### DIFF
--- a/channels/applicationinsights-channel-js/Tests/Sender.tests.ts
+++ b/channels/applicationinsights-channel-js/Tests/Sender.tests.ts
@@ -839,7 +839,7 @@ export class SenderTests extends TestClass {
         });
 
         this.testCase({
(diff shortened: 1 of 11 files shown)


channels/applicationinsights-channel-js/Tests/Sender.tests.ts
```diff
@@ -837,5 +837,109 @@ export class SenderTests extends TestClass {
                 Assert.equal("Microsoft.ApplicationInsights.iKey.Pageview", appInsightsEnvelope.name);
             }
         });
+
+        this.testCase({
+            name: "Channel Config: is sent when requests are being sent when requests exceed max batch size",
```
xiao-lix, line 842:
```suggestion
            name: "Channel Config: Notification is sent when requests are being sent when requests exceed max batch size",
```

Issue #1260, opened 2020-05-01
[BUG] Can't include Correlation Header on IE can fail


MSNev:
The following function in Utils.ts
const requestHost = UrlHelper.parseUrl(requestUrl).host.toLowerCase();

On IE can return the port number with the host rather than just the host.

Note: This *may* not be a significant issue or it only affects situations where the ajax request is using a different protocol (http/https/port) than the main hosting page -- investigating.

MSNev:
On investigating the criteria that is used when the SDK adds this value to the ajax request it appears that there has been a change in IE (as the SDK code around this has not change in over 8 months) where when an ajax request is created it checks that the ajax request is being sent to the same host and if not it does not include -- this IS what is happening here.

On doing some basic investigation using https protocol what is occurring is that (for example) when the host is "https://mydomain.com" and an ajax request is sent to "https://mydomain.com/api" when we query the URL for the ajax "host" IE returns "mydomain.com:443" and for the page it returns "mydomain.com" -- which don't match (as we support and verify different ports).

As a workaround for this issue you should be able to add the "correlationHeaderDomains" config to the Application Insights config (same place as the Instrumentation key)

So for a site hosted on "mydomain.com" using https should add something like the following.

aiConfig = {
    instrumentationKey: "INSTRUMENTATION_KEY",
    **correlationHeaderDomains: "mydomain.com:443"**
}

MSNev:
Included in v 2.5.5 which is now fully deployed

Issue #1251, opened 2020-04-15
[BUG] ajax.ts is using string trim() which is not supported on IE7/8


MSNev:
ajax.ts is using string trim() for headers.trim()

This will cause an exception and potential lockup on IE7/8 mode.

MSNev:
Included in v 2.5.5 which is now fully deployed

PR #1279, opened 2020-06-02
[Release] Increase to version 2.5.5

MSNev:
## 2.5.5

- #1278 Add optional 'eventsSendRequest' notification to NotificationManager
- #1269 TsLib v1.13.0 has breaking change (Remove dependency on TSLib)
  - Added new package **'applicationinsights-shims@1.0.0'**
- #991 Don't work with React HOOKS
  - #1120 Introducing React Hooks for AppInsights #1120
- #1274 Fix for withAITracking wrapping functional components.
- Using crypto to generate GUIDs when available (Make GUID more random)
- #1260 [BUG] Can't include Correlation Header on IE can fail
- #1258 Update snippet to support reporting script load failures
- #1251 [BUG] ajax.ts is using string trim() which is not supported on IE7/8
- #1249 Identify whether the script is being consumed via the CDN or NPM package
- Several minor documentation updates

### New Package applicationinsights-shims v1.0.0

- Created to remove dependency on TSLib as v 1.13.0 has introduced build breaks
  - provides internal implementation of __extends() and __assign() when no pre-existing version is present

### DIFF
--- a/extensions/applicationinsights-react-js/package.json
+++ b/extensions/applicationinsights-react-js/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@microsoft/applicationinsights-react-js",
-    "version": "2.5.4",
+    "version": "3.0.0",
     "description": "Microsoft Application Insights React plugin",
     "main": "dist/applicationinsights-react-js.js",
     "module": "dist-esm/applicationinsights-react-js.js",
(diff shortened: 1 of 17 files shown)


extensions/applicationinsights-react-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@microsoft/applicationinsights-react-js",
-    "version": "2.5.4",
+    "version": "2.5.5",
```
markwolff, line 3: I think this needs to be 3.0 bump because of typescript bump and react bump.

extensions/applicationinsights-react-js/package.json
```diff
@@ -47,8 +47,8 @@
     },
     "dependencies": {
         "@microsoft/applicationinsights-shims" : "1.0.0",
-        "@microsoft/applicationinsights-core-js": "2.5.4",
-        "@microsoft/applicationinsights-common": "2.5.4",
+        "@microsoft/applicationinsights-core-js": "2.5.5",
```
markwolff, line 50: Since this plugin is now decoupled from rush, need to use semver carets for now.

PR #1288, opened 2020-06-08
legacy: use w3c id for new operations

markwolff:
- Use longer w3c id for legacy operation id's to prevent conflicting op ids. If no conflicts occur, viewing app insights data on the Portal can use the new Indexer feature for significantly improved blade load times.

### DIFF
--- a/legacy/JavaScript/JavaScriptSDK/Context/Operation.ts
+++ b/legacy/JavaScript/JavaScriptSDK/Context/Operation.ts
@@ -17,10 +17,10 @@ module Microsoft.ApplicationInsights.Context {
         public syntheticSource: string;
 
         constructor() {
-            this.id = Util.newId();
+            this.id = Util.generateW3CId();
             if (window && window.location && window.location.pathname) {
                 this.name = window.location.pathname;
             }
         }
     }
-} 
+}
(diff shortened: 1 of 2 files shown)


legacy/JavaScript/JavaScriptSDK/Context/Operation.ts
```diff
@@ -17,10 +17,10 @@ module Microsoft.ApplicationInsights.Context {
         public syntheticSource: string;
 
         constructor() {
-            this.id = Util.newId();
+            this.id = Util.generateW3CId();
```
MSNev, line 20: So this will change the linking id for all things that use the operation. Such as ajax correlation requestId Header as "|<w3cId>.<Util.newId()>

Is this expected and won't break v1 users?

markwolff: Yes the expected behavior is for new headers going forward to be `|<Util.generateW3CId()>.whatever` vs `|<Util.newId()>.whatever`. This will not break anyone unless their app contains deep distributed trace and [the header eventually concats to be over 8k characters](https://stackoverflow.com/questions/686217/maximum-on-http-header-values#:~:text=No%2C%20HTTP%20does%20not%20define,headers%20size%20exceeds%20that%20limit.).

PR #1303, opened 2020-06-30
Task 7027291: Investigate CDN Configuration to support custom domain ...

MSNev:


AISKU/scripts/setActiveCdnVersion.ps1
```diff
@@ -0,0 +1,668 @@
...
+}
+
+Function SetProperties(
+    $stagedBlob,
+    $srcName,
+    $ver
+) {
+    $cloudBlob = $stagedBlob.ICloudBlob
+    $cloudBlob.FetchAttributes()
+    $cloudBlob.Properties.CacheControl = $cacheControl30Min
```
markwolff, line 515: Would it make sense for this to be configurable? I think it would be cool if this were set up in a multi phase deployment with 30min for initial release, and then 48?hours later is redeployed with X hours. WDYT?

AISKU/scripts/setActiveCdnVersion.ps1
```diff
@@ -0,0 +1,668 @@
...
+    foreach ($dataKey in $blob.ICloudBlob.Metadata.Keys) {
+        RemoveMetadata $cloudBlob $dataKey
+        $cloudBlob.Metadata.Add($dataKey, $blob.ICloudBlob.Metadata[$dataKey]) | Out-Null
+    }
+
+    $cloudBlob.SetProperties()
+    $cloudBlob.SetMetadata()
+}
+
+Function SetProperties(
```
markwolff, line 508: Is it possible to set an etag here too?

AISKU/scripts/setActiveCdnVersion.ps1
MSNev: We don't really want to make this value too big as it will take that long for the next release to get picked up, we could have a smaller rate (15min then 30min perhaps), but the downside is that every browser will potentially try to fetch the script again during that interval

I have another work item to minimize the size impacts so that the scripts are cached longer and/or the initial script is not so large.

AISKU/scripts/setActiveCdnVersion.ps1
MSNev: We could, but we don't need to. When the blob is uploaded the ETag is set automatically

markwolff: Is this a new change? I just checked and it was there, but I don't remember always seeing it there. nice!

PR #1305, opened 2020-06-30
feat: add timestamp to dependency if not specified

markwolff:
re-rev of #804


MSNev:
Reopening for release tracking for 1.0.22 legacy release

Issue #1089, opened 2019-10-16
Blocking certain URIs/Patterns from fetch tracking (patch included)


ewnavilae:
I have recently run into an issue where I had to block AppInsights from sending correlation headers to a specific URI via fetch. I understand I can stop these URIs from being tracked with a plugin that prevents this information from being sent to the telemetry server, however this doesn't stop fetch from being hooked and the headers from being sent. I couldn't find a documented feature that results in this behavior.

To work around it, I have applied the following patch using yarn patch-package:
https://gist.github.com/ewnavilae/07dd5aa58377db6c9ab6df116467b44f

If enough people value this change I may try to submit a pull request actually implementing this correctly. Is there an already implemented, supported way of doing this?

markwolff:
Does using `config.correlationHeaderExcludedDomains` solve this problem?

ewnavilae:
In my use case, I only need to filter a specific URL from a domain I am otherwise including. I may have tested it wrong but it didn't produce the behavior I required.

ewnavilae:
In my AppInsights configuration, I tried the following options:
`
          correlationHeaderDomains: [
           "my_api_server",
           ],
          correlationHeaderExcludedDomains: [
            "my_api_server/AMCalendarEventsAPI/IcsGenerator.ashx",
           ],
`
This still results in correlation headers being sent to that URI.

Am I doing something wrong here?

markwolff:
I believe you need to include the actual hostnames in the 2 config arrays since we do URL parsing on them during the comparison. e.g. `localhost:8080/my_api_server` and `production.endpoint.com/my_api_server`. Perhaps they should be updated to support wild carding of the hostname as well `*/my_api_server`

ewnavilae:
I'm sorry, I don't understand how that would get correlation headers to be sent to my API server on all URIs except the one I needed to blacklist.

I got this fixed by updating the problematic endpoint to allow the correlation headers.

Still, I believe a better mechanism to choose where to send correlation headers would be useful... Perhaps an option like shouldSendCorrelationHeaders whose signature would be `( url: string ) => boolean`? Not sure if it's worth implementing as it would seem this isn't highly requested.

markwolff:
See the implementation here. The URLs you configure with must include the hostname, and not just the route information.

[...]

PR #1306, opened 2020-06-30
release: 1.0.22

markwolff:
Contains #1288


MSNev:
This will also now include #1089 as it contained changes to the legacy package to include the exclusion checks

PR #1311, opened 2020-07-02
Allow the generated modules to extend the namespace defined by "name"  in rollup config -- rather than always replace.

MSNev:


markwolff:
I'm curious how we haven't affected the `Microsoft` namespace already. Are we also overwriting this namespace? This wouldn't be fixed by this config change right?

MSNev:
It because, this change only affects the final element of the name
Without the extend
(factory((global.Microsoft = global.Microsoft || {}, global.Microsoft.ApplicationInsights = {})));

With the extend
(factory((global.Microsoft = global.Microsoft || {}, global.Microsoft.ApplicationInsights = global.Microsoft.ApplicationInsights || {})));

So if we set the name to just "Microsoft" then we would have 😃

PR #1316, opened 2020-07-07
Update dependency version of DynamicProto

MSNev:


AISKU/package.json
```diff
@@ -50,7 +50,7 @@
         "typescript": "2.5.3"
     },
     "dependencies": {
-        "@microsoft/dynamicproto-js": "^0.5.2",
+        "@microsoft/dynamicproto-js": "^1.0.0",
```
MSNev, line 53: The actual generated code of v0.5.2 and v1.0.0 are identical, the only significant difference is that v1.0.0 removes tslib from it's peerDependency in it's package.json as it's not required by dynamicProto.

Issue #1321, opened 2020-07-20
[BUG] @microsoft/applicationinsights-web fails to initialize with latest version


jepetty:
**Description/Screenshot**
Starting this morning, our application insights instance began failing to initialize:
> TypeError: Cannot set property 'snippetVer' of undefined
	at _updateSnippetProperties (Initialization.js?e727:236)
	at Initialization.loadAppInsights (Initialization.js?e727:263)
	at new OnlineTelemetryClient (OnlineTelemetryClient.ts?ed0c:20)
	at Object.eval (Telemetry.ts?e430:72)
	at Generator.next (<anonymous>)
	at eval (Telemetry.ts?e430:5)
	at new Promise (<anonymous>)
	at __awaiter (Telemetry.ts?e430:5)
	at Object.initializeTelemetry (Telemetry.ts?e430:62)
	at eval (Extension.ts?7991:33)
	at Generator.next (<anonymous>)
	at eval (Extension.ts?7991:5)
	at new Promise (<anonymous>)
	at __awaiter (Extension.ts?7991:5)
	at activate (Extension.ts?7991:31)
	at Function._callActivateOptional (extHostExtensionService.ts:418)
	at Function._callActivate (extHostExtensionService.ts:407)
	at extHostExtensionService.ts:362

**Steps to Reproduce**
Initialize a new instance of Application Insights with a non-null snippet. Our snippet:
```ts
this.appInsights = new ApplicationInsights({
			config: {
				instrumentationKey,
				endpointUrl: 'https://vortex.data.microsoft.com/collect/v1',
				emitLineDelimitedJson: true,
				isCookieUseDisabled: true,
			},
		});
```

 - OS/Browser:
Windows - Chrome and Edge

 - SDK Version [e.g. 22]:
Using version 2.5.6 of the @microsoft/applicationinsights-web NPM package
**NOTE:** this works in version 2.5.3 (and I presume lower)

MSNev:
Hi @jepetty, 

This looks like where the issue is occurring
https://github.com/microsoft/ApplicationInsights-JS/blob/master/AISKU/src/Initialization.ts#L284

And looking at the code this looks like it an issue when running in an environment that doesn't support a window, as the internal property of the context object won't get populated.
https://github.com/microsoft/ApplicationInsights-JS/blob/master/extensions/applicationinsights-properties-js/src/TelemetryContext.ts#L37

[...]

PR #1322, opened 2020-07-20
[BUG] @microsoft/applicationinsights-web fails to initialize with latest version #1321

MSNev:


extensions/applicationinsights-properties-js/src/TelemetryContext.ts
```diff
@@ -30,11 +30,11 @@ export class TelemetryContext implements ITelemetryContext {
 
     constructor(logger: IDiagnosticLogger, defaultConfig: ITelemetryConfig) {
         let _self = this;
+        _self.application = new Application();
+        _self.internal = new Internal(defaultConfig);
```
MSNev, line 34: Always create the internal properties element

Issue #1324, opened 2020-07-21
[BUG] Type 'ReactNativePlugin' is not assignable to type 'ITelemetryPlugin'


nairmanu:
**Description/Screenshot**

Type 'ReactNativePlugin' is not assignable to type 'ITelemetryPlugin'.
  Types of property 'setNextPlugin' are incompatible.
    Type '(next: ITelemetryPlugin) => void' is not assignable to type '(next: ITelemetryPlugin | ITelemetryPluginChain) => void'.
      Types of parameters 'next' and 'next' are incompatible.
        Type 'ITelemetryPlugin | ITelemetryPluginChain' is not assignable to type 'ITelemetryPlugin'.
          Type 'ITelemetryPluginChain' is missing the following properties from type 'ITelemetryPlugin': priority, initialize, identifier

**Steps to Reproduce**
When I add the latest versions of applicationinsights-react-native and applicationinsights-web, I get the above error while trying to create a new instance of ApplicationInsights(). 
ReactNativePlugin is not accepted as a valid value for extensions array.

yarn add @microsoft/applicationinsights-react-native
yarn add @microsoft/applicationinsights-web

Versions installed are,
"@microsoft/applicationinsights-react-native": "^2.2.5",
"@microsoft/applicationinsights-web": "^2.5.6",

 - OS/Browser:
 - SDK Version [e.g. 22]:
 - How you initialized the SDK:

**Expected behavior**

**Additional context**

MSNev:
This issue is a method signature issue only, as a workaround you *should be able* and can safely cast the plugin to an ITelemetryPlugin when assigning the react-native to the extensions.

The ITelemetryPluginChain property is only used *if* you are creating a shared instance and reusing across multiple AI instances.

nairmanu:
thanks! the workaround seems to do the trick. Is there a plan to fix this in the future releases?

MSNev:
Yep, about to start working on it now -- should be in next release

nairmanu:
thank you!

MSNev:
PR is out -- please have a look as I've applied a couple of changes, specifically
- fixed this issue, by extending from the new base class
- refactored to use dynamicProto() -- for better minification
- as part of the refactoring ALL of the private properties are now closure based instance variables (part of the minification changes), but the big change is that they will no longer be available on the instance.

MSNev:
Fully Deployed to CDN and NPM as version 2.5.7

PR #1325, opened 2020-07-21
[BUG] Type 'ReactNativePlugin' is not assignable to type 'ITelemetryPlugin' #1324

MSNev:


extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
```diff
@@ -9,149 +9,170 @@ import {
...
+    IProcessTelemetryContext
 } from '@microsoft/applicationinsights-core-js';
 import { ConfigurationManager, IDevice, IExceptionTelemetry, IAppInsights, SeverityLevel, AnalyticsPluginIdentifier  } from '@microsoft/applicationinsights-common';
 import DeviceInfo from 'react-native-device-info';
 
 import { INativeDevice, IReactNativePluginConfig } from './Interfaces';
+import dynamicProto from '@microsoft/dynamicproto-js';
 
-export class ReactNativePlugin implements ITelemetryPlugin {
+export class ReactNativePlugin extends BaseTelemetryPlugin {
```
MSNev, line 23: Upgraded to use the new base class, so that it will properly support being used in a shared environment (reusing the same plugin from multiple app insight instances)

extensions/applicationinsights-react-native/src/ReactNativePlugin.ts
```diff
@@ -9,149 +9,170 @@ import {
...
-                    this._config[option]
-                );
+        super();
+
+        let _device: INativeDevice = {};
+        let _config: IReactNativePluginConfig = config || _getDefaultConfig();
+        let _analyticsPlugin: IAppInsights;
+        let _defaultHandler;
+    
+        dynamicProto(ReactNativePlugin, this, (_self, _base) => {
```
MSNev, line 37: Also, as I was touching the class -- implemented with dynamicProto() for minimifaction.

NOTE: The private properties are now within the constructor's closure -- so they will NOT be available on the instance for any form of overloading... If anyone is aware of these properties being used directly (even though they are private) then we will need to make them protected / public to properly support that usage pattern.

PR #1329, opened 2020-07-23
Instrumentation key validation

kryalama:
Specs taken from https://tools.ietf.org/html/rfc412
The definition of a VALID instrumentation key is as follows:

Not none
Not empty
Valid UUID according to this
Every character is a hex character [0-9a-f]
32 characters are separated into 5 sections via 4 dashes
First section has 8 characters
Second section has 4 characters
Third section has 4 characters, the first one in [1-5]
Fourth section has 4 characters, the first one in [89ab]
Fifth section has 12 characters

### DIFF
--- a/shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
+++ b/shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
@@ -57,6 +57,10 @@ export class BaseCore implements IAppInsightsCore {
             throw Error("Please provide instrumentation key");
         }
 
+        if(!_this.validateInstrumentationKey(config.instrumentationKey)) {
+            throw Error("invalid instrumentation key");
+        }
+
         _this.config = config;
         let channelController = _this._channelController;
 
@@ -192,6 +196,19 @@ export class BaseCore implements IAppInsightsCore {
         }
 
         return new ProcessTelemetryContext(thePlugins, _this.config, _this);
+    }
+
+    /**
+     * Validate UUID Format
+     * Specs taken from https://tools.ietf.org/html/rfc4122
+     */
+    validateInstrumentationKey(iKey:string) :boolean {
+        const regexp = new RegExp('^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$');
+        if (regexp.test(iKey)) {
+            return true;
+        } else {
+            return false;
+        }
     }
 
     protected releaseQueue() {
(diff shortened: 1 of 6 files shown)


shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -43,6 +44,7 @@ export class BaseCore implements IAppInsightsCore {
         _this.isInitialized = () => _isInitialized;
         _this._setInit = (value: boolean) => { _isInitialized = value; }
         _this._eventQueue = [];
+        _this._UUID_Regex='^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$';
```
MSNev, line 47: As the backend is most likely written in .Net, I think we should support all valid forms (except maybe the forms wrapped in { } brackets.

But I would expect that the hyphens should be optional, so I would think we should support these 2

Specifier | Description | Format
-- | -- | --
N | 32 digits | 00000000000000000000000000000000
D | 32 digits separated by hyphens | 00000000-0000-0000-0000-000000000000

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -194,6 +200,20 @@ export class BaseCore implements IAppInsightsCore {
         return new ProcessTelemetryContext(thePlugins, _this.config, _this);
     }
 
+    /**
+     * Validate UUID Format
+     * Specs taken from https://tools.ietf.org/html/rfc4122
+     */
+    validateInstrumentationKey(iKey:string) :boolean {
```
MSNev, line 207: It also might be worth adding a "disableIkeyValidation" which defaults to false -- just in case we miss something.

Alternate suggestion: default to calling throwInternal() to log an error rather than throwing an exception -- that way we won't break anyone.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -194,6 +200,20 @@ export class BaseCore implements IAppInsightsCore {
...
+        let _this = this;
+        const regexp = new RegExp(_this._UUID_Regex);
+        if (regexp.test(iKey)) {
```
markwolff, line 210: nit: `return regexp.test(iKey)`

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
markwolff: Agree with Nev here. Seems like a good usecase to tie this in with the `enableDebug` config, which replaces all `throwInternal()` calls with a `throw`

kryalama:
Creating a new pull request. Discarding this one.

PR #1339, opened 2020-08-05
[Release] [React] - Increase versions React 3.0.2

MSNev:
[Release] [ReactNative] - Increase versions ReactNative 2.2.7


MSNev:
This PR will continue to fail until I publish Core version 2.5.7 -- splitting this out to ensure that the react components are defined to use the latest rather than the previous.

Issue #1343, opened 2020-08-13
[BUG] IPerfManager interface - the create() function is defined to return an IPerfEvent and not an IPerfEvent?


MSNev:
there's a minor typescript problem with the IPerfManager interface - the create() function is defined to return an IPerfEvent and not an IPerfEvent?, so it can't return null/undefined (temporarily worked around by returning null as any as IPerfEvent)

MSNev:
v2.5.8 is now fully deployed

PR #1344, opened 2020-08-13
[BUG] IPerfManager interface - the create() function is defined to return an IPerfEvent and not an IPerfEvent? #1343

MSNev:


shared/AppInsightsCore/src/JavaScriptSDK/PerfManager.ts
```diff
@@ -148,19 +148,19 @@ export class PerfManager implements IPerfManager  {
 
         dynamicProto(PerfManager, this, (_self) => {
 
-            _self.create = (src: string, payloadDetails?: () => any, isAsync?: boolean): IPerfEvent => {
-                // TODO (newylie): at some point we will want to add additional configuration to "select" which events to instrument
+            _self.create = (src: string, payloadDetails?: () => any, isAsync?: boolean): IPerfEvent|null|undefined => {
+                // TODO (@MSNev): at some point we will want to add additional configuration to "select" which events to instrument
                 // for now this is just a simple do everything.
                 return new PerfEvent(src, payloadDetails, isAsync);
```
markwolff, line 154: When would this be null or undefined?

MSNev: As part of the documentation people can provide there own IPerfManager which can selectively opt-out of creating a IPerfEvent to indicate that it doesn't want to monitor the instrumented chunk of code.

OTEL are already creating their own version of IPerfManager and found this signature issue.

MSNev: But yes, for the default implementation it won't ever be, but for consistency with the *.d.ts, defining the return type to avoid issues in some environments and issues like #1324 which is just a signature issue

shared/AppInsightsCore/src/JavaScriptSDK/PerfManager.ts
```diff
@@ -148,19 +148,19 @@ export class PerfManager implements IPerfManager  {
 
         dynamicProto(PerfManager, this, (_self) => {
 
-            _self.create = (src: string, payloadDetails?: () => any, isAsync?: boolean): IPerfEvent => {
-                // TODO (newylie): at some point we will want to add additional configuration to "select" which events to instrument
+            _self.create = (src: string, payloadDetails?: () => any, isAsync?: boolean): IPerfEvent|null|undefined => {
```
markwolff, line 151:
```suggestion
            _self.create = (src: string, payloadDetails?: () => any, isAsync?: boolean): IPerfEvent |null |undefined => {
```

shared/AppInsightsCore/src/JavaScriptSDK/PerfManager.ts
```diff
@@ -177,11 +177,22 @@ export class PerfManager implements IPerfManager  {
...
 
-    public create(src: string, payload?: any, isAsync?: boolean): IPerfEvent {
+    /**
+     * Create a new event and start timing, the manager may return null/undefined to indicate that it does not 
+     * want to monitor this source event.
+     * @param src The source name of the event 
+     * @param payloadDetails - An optional callback function to fetch the payload details for the event.
+     * @param isAsync - Is the event occurring from a async event
+     */
+    public create(src: string, payload?: any, isAsync?: boolean): IPerfEvent|null|undefined {
```
markwolff, line 187:
```suggestion
    public create(src: string, payload?: any, isAsync?: boolean): IPerfEvent | null | undefined {
```

PR #1349, opened 2020-08-26
Update angular plugin track pageview logic and test

xiao-lix:
Previous pageView on angular plugin was tracked twice on page initial load : 
1. on plugin initialization 
2. on "NavigationEnd" router event happens

The change in this PR removes the second pageView when the first pageView already triggered.


extensions/applicationinsights-angular-js/src/AngularPlugin.ts
```diff
@@ -30,19 +32,24 @@ export default class AngularPlugin extends BaseTelemetryPlugin {
...
+            if (isPageInitialLoad) {
+                const pageViewTelemetry: IPageViewTelemetry = {
+                    uri: extConfig.router.url
+                };
+                this.trackPageView(pageViewTelemetry);
+            }
             extConfig.router.events.subscribe(event => {
-                if (event.constructor.name === "NavigationEnd") {
-                    // Timeout to ensure any changes to the DOM made by route changes get included in pageView telemetry
-                    setTimeout(() => {
```
markwolff, line 36: Should this `setTimeout` be kept around?

extensions/applicationinsights-angular-js/src/AngularPlugin.ts
```diff
@@ -30,19 +32,24 @@ export default class AngularPlugin extends BaseTelemetryPlugin {
...
-                        this.trackPageView(pageViewTelemetry);
-                    }, 500);
+                if (event.constructor.name === NAVIGATIONEND) {
+                    // for page initial load, do not call trackPageView twice
+                    if (isPageInitialLoad) {
+                        isPageInitialLoad = false;
+                        return;
+                    }
+                    const pageViewTelemetry: IPageViewTelemetry = { uri: extConfig.router.url };
+                    this.trackPageView(pageViewTelemetry);
```
markwolff, line 50: Update operation id before calling `trackPageView`

xiao-lix: oh I just pushed a commit before I saw your message. I added this, but find the properties plugin itself to update operation id. Can you please take a look if it makes sense. The `this._anaylticsPlugin['_properties']` returns `undefined`

extensions/applicationinsights-angular-js/src/AngularPlugin.ts
xiao-lix: Good question.. I actually don't understand why this setTimeout is required..

xiao-lix: I mean, reading the doc [here](https://angular.io/api/router/Event) , when router event type is "NavigationEnd" it means "navigation ends successfully", when navigation ends successfully then it's time to trigger track page view isn't it? 🤔

markwolff: I originally added the `setTimeout` in the analytics plugin because the route change events would trigger before the URL itself would change. This caused the pageview telemetry to have the previous URL attached to it, instead of the upcoming one. If the pageviews are correct in your test app, then I think the setTimeout can safely be removed.

extensions/applicationinsights-angular-js/src/AngularPlugin.ts
markwolff: There may be a bug in the analytics plugin, but this is where we search for the properties plugin there
https://github.com/microsoft/ApplicationInsights-JS/blob/ab359e7d6714577800672945107b5c7b0e5d1af0/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts#L624-L628

extensions/applicationinsights-angular-js/src/AngularPlugin.ts
xiao-lix: @markwolff Thanks for bring this up, it's a very good question.. I double checked with my test app and sample app, so the router events happen in this [sequence](https://angular.io/api/router/Event#description). And the `router.url` changes from old url to new url on `ActivationEnd`. And the event listener subscribes to `NavigationEnd` happens after `ActivationEnd` so it catches the new url. 🤓

Issue #1018, opened 2019-08-28
Error Mismatched anonymous define() module


mayong43111:
Dear,
  I have a angularjs + requirejs website, I use Pasting a script snippet at the <head> tag for monitor.  but some boy is error when he access this view. It's only the first access on new tab.

mayong43111:
> Dear,
> I have a angularjs + requirejs website, I use Pasting a script snippet at the tag for monitor. but some boy is error when he access this view. It's only the first access on new tab.

It's only IE 11.

mayong43111:
> Which script snippet are you using?

`<script type="text/javascript">
var sdkInstance="appInsightsSDK";window[sdkInstance]="appInsights";var aiName=window[sdkInstance],aisdk=window[aiName]||function(e){function n(e){t[e]=function(){var n=arguments;t.queue.push(function(){t[e].apply(t,n)})}}var t={config:e};t.initialize=!0;var i=document,a=window;setTimeout(function(){var n=i.createElement("script");n.src=e.url||"https://az416426.vo.msecnd.net/scripts/b/ai.2.min.js",i.getElementsByTagName("script")[0].parentNode.appendChild(n)});try{t.cookie=i.cookie}catch(e){}t.queue=[],t.version=2;for(var r=["Event","PageView","Exception","Trace","DependencyData","Metric","PageViewPerformance"];r.length;)n("track"+r.pop());n("startTrackPage"),n("stopTrackPage");var s="Track"+r[0];if(n("start"+s),n("stop"+s),n("setAuthenticatedUserContext"),n("clearAuthenticatedUserContext"),n("flush"),!(!0===e.disableExceptionTracking||e.extensionConfig&&e.extensionConfig.ApplicationInsightsAnalytics&&!0===e.extensionConfig.ApplicationInsightsAnalytics.disableExceptionTracking)){n("_"+(r="onerror"));var o=a[r];a[r]=function(e,n,i,a,s){var c=o&&o(e,n,i,a,s);return!0!==c&&t["_"+r]({message:e,url:n,lineNumber:i,columnNumber:a,error:s}),c},e.autoExceptionInstrumented=!0}return t}(
{
  instrumentationKey:"INSTRUMENTATION_KEY"
}
);window[aiName]=aisdk,aisdk.queue&&0===aisdk.queue.length&&aisdk.trackPageView({});
</script>`

Maus3rVonDutch:
[...]

PR #1352, opened 2020-08-27
add .cjs.js and .cjs.min.js

xiao-lix:


MSNev:
This should also address #1018

xiao-lix:
@MSNev should we merge this now ?

PR #1357, opened 2020-08-31
Error Mismatched anonymous define() module #1018

MSNev:
- Also publish an iife module format to skip the define loading

### DIFF
--- a/AISKU/rollup.config.js
+++ b/AISKU/rollup.config.js
@@ -18,13 +18,15 @@ const replaceValues = {
   "// Licensed under the MIT License.": ""
 };
 
-const browserRollupConfigFactory = (isProduction, libVersion = '2') => {
+const majorVersion = version.split('.')[0];
+
+const browserRollupConfigFactory = (isProduction, libVersion = '2', format = 'umd', postfix = '') => {
   const browserRollupConfig = {
     input: "dist-esm/Init.js",
     output: {
-      file: `browser/ai.${libVersion}.js`,
+      file: `browser/ai.${libVersion}${postfix}.js`,
       banner: banner,
-      format: "umd",
+      format: format,
       name: "Microsoft.ApplicationInsights",
       extend: true,
       freeze: false,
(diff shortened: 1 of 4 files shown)


AISKU/rollup.config.js
```diff
@@ -18,63 +18,15 @@ const replaceValues = {
   "// Licensed under the MIT License.": ""
 };
 
-const browserRollupConfigFactory = (isProduction, libVersion = '2') => {
```
MSNev, line 21: Reworking the scripts so we don't have too much duplication just for publishing

AISKU/rollup.config.js
```diff
@@ -168,12 +120,16 @@ updateDistEsmFiles(replaceValues, banner);
...
   browserRollupConfigFactory(false, version),
-  browserCJSRollupConfigFactory(true),
-  browserCJSRollupConfigFactory(false),
-  browserCJSRollupConfigFactory(true, version),
-  browserCJSRollupConfigFactory(false, version)
+  browserRollupConfigFactory(true, majorVersion, 'cjs', '.cjs'),
+  browserRollupConfigFactory(false, majorVersion, 'cjs', '.cjs'),
+  browserRollupConfigFactory(true, version, 'cjs', '.cjs'),
+  browserRollupConfigFactory(false, version, 'cjs', '.cjs'),
+  browserRollupConfigFactory(true, majorVersion, 'iife', '.gbl'),
```
MSNev, line 131: Publishing the iife format as "gbl" to 
- sort of align with what VueJs is call (they use .global)
- As this explicitly creates the Microsoft namespace a global highlighting that this will add global scope variables.

AISKU/scripts/setActiveCdnVersion.ps1
```diff
@@ -664,7 +664,8 @@ if ($container -eq "beta" -or $container -eq "next") {
 if ($files.ContainsKey($activeVersion) -ne $true) {
     Log-Failure "Version [$activeVersion] does not appear to be deployed to [$container]"
 } elseif ($files[$activeVersion].Count -ne 4 -and # Prior to 2.5.8
-        $files[$activeVersion].Count -ne 8) {     # Since 2.5.8
+        $files[$activeVersion].Count -ne 8 -and   # Since 2.5.8
+        $files[$activeVersion].Count -ne 12) {    # Since 2.5.8
```
MSNev, line 668: While we have already defined the "label" for 2.5.8, we are planning on sneaking this into the 2.5.8 release as it's only changing whats deployed and not any code.

MSNev: The only specific issue is that this new format won't appear in the NPM package until the next release version bump

AISKU/rollup.config.js
xiao-lix: Do we still create cjs files?

MSNev: Yes (lines 128-130), as it would be useful for people that want to "install" it as a module via the CDN.

PR #1373, opened 2020-09-16
Performance tests added

kryalama:
Performance tests added for
loadAppInsights, processTelemetry, flush and track calls.
Initial iteration is to determine the threshold for these calls.


AISKU/Tests/perfTests.ts
```diff
@@ -0,0 +1,363 @@
...
+      // Setup Notification Listener
+      this._sendNotifications = [];
+      this._notificationManager = new NotificationManager();
+      this._notificationManager.addNotificationListener({
+        perfEvent: (perfEvent: IPerfEvent): void => {
+          this._sendNotifications.push(perfEvent);
+        },
+      });
+
+      const init = new ApplicationInsights({
```
MSNev, line 59: As we are creating specific tests for initialization, we should move this into the specific tests, to allow for updating of the configurations

AISKU/Tests/perfTests.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  public registerTests() {
+    this.addPerfTests();
+  }
+
+  private addPerfTests(): void {
+    this.testCase({
+      name: "AISKU Perf Tests -> load AppInsights",
+      test: () => {
+        QUnit.assert.ok(this._sendNotifications.length === 1);
```
MSNev, line 103: This is what i mean by my comment above, this is testing the side effect of the testIntialize(), if some other test wants or needs to change the initialization it could break this.

solution: each test should setup and trigger the conditions it's testing.
We can still have common object creation and clearing out global values in testinitialize() though

AISKU/Tests/perfTests.ts
```diff
@@ -0,0 +1,363 @@
...
+          while (perfEvents.length > 0) {
+            const currEvent = perfEvents.shift();
+            if (currEvent.name.indexOf("processTelemetry") !== -1) {
+              totalTrackExtime += currEvent.exTime;
+            }
+            if (currEvent.childEvts && currEvent.childEvts.length > 0) {
+              perfEvents.push.apply(perfEvents, currEvent.childEvts);
+            }
+          }
+          console.log(
```
MSNev, line 152: General comment, we should have something that validates that the test actually managed to find the data it's looking for... Otherwise, it's possible that it could report a lower number than expected (because it found nothing)

AISKU/Tests/perfTests.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  private addPerfTests(): void {
+    this.testCase({
+      name: "AISKU Perf Tests -> load AppInsights",
+      test: () => {
+        QUnit.assert.ok(this._sendNotifications.length === 1);
+        QUnit.assert.ok(
+          this._sendNotifications[0].name.indexOf("AISKU.loadAppInsights") !==
+            -1,
+          "AppInsights Load time: " + this._sendNotifications[0].exTime
```
MSNev, line 107: I'm not sure that we want to track just the exclusive time (exTime) as we really want to validate (in this case) the total time it takes to initialize.

AISKU/Tests/perfTests.ts
```diff
@@ -0,0 +1,363 @@
...
+        },
+      ]
+        .concat(this.asserts(300))
+        .concat(() => {
+          let totalTrackExtime = 0;
+          let perfEvents = this._sendNotifications;
+          while (perfEvents.length > 0) {
+            const currEvent = perfEvents.shift();
+            if (currEvent.name.indexOf("processTelemetry") !== -1) {
+              totalTrackExtime += currEvent.exTime;
```
MSNev, line 146: extime doesn't seem  right here either -- especially as it (appears) to be only firing the parent events (looking at the config). This is also another reason for having the initialization locally as something like the firing of events can change the reported outcome.

Issue #1381, opened 2020-09-17
[BUG] NPM package for @microsoft/applicationinsights-angularplugin-js does not have a dist folder


OumarLM:
**Description/Screenshot**
The @microsoft/applicationinsights-angularplugin-js folder in `node-modules` does not have a `dist`
 folder
**Steps to Reproduce**
run npm install @microsoft/applicationinsights-angularplugin-js
and check node_modules. 

 - OS/Browser:
 - SDK Version [e.g. 22]:
 - How you initialized the SDK:

**Expected behavior**
not get : `Cannot find module '@microsoft/applicationinsights-angularplugin-js' or its corresponding type declarations.`
**Additional context**

MSNev:
Looks like the official build / publish pipeline is not building the angular plugin before packaging -- sorry

OumarLM:
Awesome!!!

MSNev:
Closing as 2.5.9 is now completely deployed to NPM and all CDN ednpoints

PR #1385, opened 2020-09-22
[BUG] NPM package for @microsoft/applicationinsights-angularplugin-js...

MSNev:
does not have a dist folder #1381


markwolff (approved):
Intentional lockfile check in?

Issue #1384, opened 2020-09-22
Add sideEffects field to applicationinsights-shims package.json


MSNev:
To make this package optimizeable via [webpack tree shaking](https://webpack.js.org/guides/tree-shaking/), please add a correct `sideEffects` field to the package.json file for this lib.

thanks to @johncrim for pointing this out as part of an issue raised in https://github.com/microsoft/DynamicProto-JS/issues/22

MSNev:
Closing as 2.5.9 is now completely deployed to NPM and all CDN ednpoints

PR #1386, opened 2020-09-23
Add sideEffects field to applicationinsights-shims package.json #1384

MSNev:


tools/shims/package.json
```diff
@@ -27,7 +27,10 @@
         "url": "https://github.com/microsoft/ApplicationInsights-JS/tree/master/tools/shims"
     },
     "license": "MIT",
-    "sideEffects": false,
+    "sideEffects": [
```
markwolff, line 30: Does this affect the sideEffects of consumers of this package?

MSNev: Yes, specifically [webpack tree shaking](https://webpack.js.org/guides/tree-shaking/),

markwolff: Do each of their sideEffects properties need to be updated as well?

MSNev: No, only on the the published module that actually have side effects that are not exported as variables. In this specific case the side effect is the setting / exposing of the global __extends() and __assign() functions. Just by loading the module.

markwolff (approved):

Issue #1391, opened 2020-09-28
Increase the randomness and size of the sessionId (newId())


MSNev:
For large customers using older or hosted browser instances they are seeing the same value being returned more than once within for the same PC over multiple days.

We need to update the length of the sessionId and use better randomness for the sessionId by default.

While also supporting a configuration to enable users to keep the existing length, so if their backend systems can't handle the increase in length they can configure it back to the current size.

MSNev:
After lots of investigation the issue one issue that was identified for randomness calculations is that Math.random() does not provide a good distribution (especially over multiple days) for IE 6, 7, 8 (especially on XP).

So as part of "fixing" this issue we will use a local javascript pseudo random number generator using the Mwc (Multiple With Carry) algorithm for IE only, all other browsers will continue to use the window.crypto.getRandomValues() or Math.random() as part of the call to CoreUtils.random32().

MSNev:
Closing as 2.5.9 is now completely deployed to NPM and all CDN ednpoints

PR #1392, opened 2020-09-30
Increase the randomness and size of the sessionId (newId()) #1391

MSNev:
The increased size of the newId() value (now length 22 was 5) will change the value reported for the following fields (once all cookies have expired)
- SessionId
- UserId


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewManager.ts
```diff
@@ -185,7 +187,7 @@ export class PageViewManager {
                                     pageViewPerformanceSent = true;
                                 }
                             }
-                        } else if (DateTimeUtils.GetDuration(start, +new Date) > maxDurationLimit) {
+                        } else if (start > 0 && DateTimeUtils.GetDuration(start, +new Date) > maxDurationLimit) {
```
MSNev, line 190: Yes, there is a case (mainly Old Hosted IE instances) where the performance.timing values may actually be negative

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
```diff
@@ -163,7 +163,7 @@ export class PageViewPerformanceManager {
         } else {
             // for other page views, don't report if it's outside of a reasonable range
             for (let i = 0; i < durations.length; i++) {
-                if (durations[i] >= this.MAX_DURATION_ALLOWED) {
+                if (durations[i] < 0 || durations[i] >= this.MAX_DURATION_ALLOWED) {
```
MSNev, line 166: Yes, there is a case (mainly Old Hosted IE instances) where the performance.timing values may actually be negative

extensions/applicationinsights-properties-js/src/Context/Session.ts
```diff
@@ -74,12 +75,14 @@ export class _SessionManager {
             this.initializeAutomaticSession();
         }
 
-        const now = DateTimeUtils.Now();
+        // Always using Date getTime() as there is a bug in older IE instances that causes the performance timings to have the hi-bit set eg 0x800000000 causing
+        // the number to be incorrect.
+        const now = new Date().getTime();
```
MSNev, line 80: As per above comment just trying to completely avoid the Hosted IE timing issue (DateTimeUtils.Now() tries to use it), and as this is just a calculation for the cookie expiry time we don't need the extra floating point precision anyway -- so just don't use it.

extensions/applicationinsights-properties-js/src/Context/Session.ts
```diff
@@ -166,9 +169,9 @@ export class _SessionManager {
     }
 
     private renew() {
-        const now = DateTimeUtils.Now();
+        const now = new Date().getTime();
```
MSNev, line 172: As above, avoid performance API bugs and we don't need the additional floating point precision.

extensions/applicationinsights-properties-js/src/Context/Session.ts
```diff
@@ -166,9 +169,9 @@ export class _SessionManager {
...
 
-        this.automaticSession.id = Util.newId();
+        this.automaticSession.id = Util.newId((this.config && this.config.idLength) ? this.config.idLength() : 22);
```
MSNev, line 174: Need to add this to the readme.

extensions/applicationinsights-properties-js/src/Context/Session.ts
```diff
@@ -196,15 +199,15 @@ export class _SessionManager {
...
 
         // if sessionExpirationMs is set to 0, it means the expiry is set to 0 for this session cookie
         // A cookie with 0 expiry in the session cookie will never expire for that browser session.  If the browser is closed the cookie expires.  
         // Another browser instance does not inherit this cookie.
         const UTCString = this.config.sessionExpirationMs() === 0 ? '0' : cookieExpiry.toUTCString();
-        Util.setCookie(this._logger, this._storageNamePrefix(), cookie.join('|') + ';expires=' + UTCString, cookieDomnain);
+        Util.setCookie(this._logger, this._storageNamePrefix(), cookie.join('|') + ';expires=' + UTCString, cookieDomain);
 
-        this.cookieUpdatedTimestamp = DateTimeUtils.Now();
+        this.cookieUpdatedTimestamp = new Date().getTime();
```
MSNev, line 210: And again as above comments on performance DateTime issues

extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -61,7 +61,7 @@ export class User implements IUser {
         this.config = config;
 
         if (!this.id) {
-            this.id = Util.newId();
+            this.id = Util.newId(config.idLength());
```
MSNev, line 64: config *should* always exist, but better add null checks

shared/AppInsightsCommon/src/Util.ts
```diff
@@ -479,53 +479,19 @@ export class Util {
...
-        return result;
-    }
+    public static newId = CoreUtils.newId;
 
     /**
      * generate a random 32bit number (-0x80000000..0x7FFFFFFF).
      */
     public static random32() {
-        return (0x100000000 * Math.random()) | 0;
+        return CoreUtils.random32(true);
```
MSNev, line 488: Moved the main implementation of these to CoreUtils, and added the ability to specify if you want signed or unsigned result, CoreUtils.random32(signed?: boolean). As this is a legacy version, this one will always returned a signed result as per documentation.

markwolff (approved):

PR #1395, opened 2020-10-05
Update publishing scripts to support automation

MSNev:
- Supports when the user is only a Contributor to the storage account
- Does not have access to create SasTokens or access subscription keys


markwolff (approved):
I'm no PS expert, but overall LGTM!

PR #1417, opened 2020-10-28
Adding nuspec for new Snippet nuget package

hectorhdzg:
Package will be published here https://www.nuget.org/profiles/AppInsightsSdk


hectorhdzg:
@MSNev this need to be published only when the snippet is updated, most likely will create another build just for this

MSNev (approved):

Issue #1091, opened 2019-10-17
Enable cookie support after the SDK has been initialized


wham:
We have a home page `/` with a cookie banner. We are not allowed to write cookies until the user consents. We initialize the SDK with `isCookieUseDisabled=true`. Once the user consents, we would like to immediately write the cookies so when they continue to the next page `/sign-in`, same session ID is used there.

We can call `appInsights.properties.context.sessionManager.update()` but there seems to be no way to set `CoreUtils._canUseCookies` to `true` once the SDK has been initialized.

markwolff:
Seems like a bug. Will get this in to next release (2.3.1)

Schmaga:
The fix did not make it into 2.3.1 as far as I can see. In the meantime, did anyone find a workaround for this, yet? Tried for two hours but failed miserably :)

Schmaga:
I managed to create a very hacky workaround. When using the following code, we seem to be able to activate cookies even after the SDK has been initialized.  It only works by recreating a completely new app insights instance. like this: 
```
activateCookies() {
    CoreUtils._canUseCookies = true;
    this.appInsights = new ApplicationInsights({
      config: {
        instrumentationKey: environment.appInsights.instrumentationKey,
        isCookieUseDisabled: false,
        enableAutoRouteTracking: true
      }
    });
    this.appInsights.loadAppInsights();
    this.appInsights.trackPageView();
}
```

MSNev:
Note: This workaround will break in the near future as we are planning on hiding all of the properties that start with "_" as they are considered to be private and should not be set directly.

To fix this properly we will need to add a setter method to be able to set/reset this value.

Schmaga:
Hopefully you will not hide it without fixing the bug itself. But thanks for the heads up :) Maybe a reloadConfig method or so might help, in case other settings are also affected by that "set-once, change-never"-behavior.

MSNev:
Noted.

Also while looking at some other indirectly related code, I just tripped over another issue with the workaround, because the _canUseCookies is actual an internal global (on the CoreUtils). *if* someone else creates a new instance of AppInsights and also sets the isCookieUseDisabled to true this will reset the _canUseCookies back to false (via calling CoreUtils.disableCookies()). 

This means that all of the cookie functions become no-op...

[...]

PR #1435, opened 2020-11-24
Fix #1434 add cookie path

### LINKED ISSUE #1434: Ability to specify cookie Path so that AI works behind App Gateway
Currently when creating cookies, the cookie path is hard-coded to `/`. This breaks cookies when an application using AI Client SDK is placed behind a reverse proxy such as App Gateway product in Azure. Applications behind the reverse proxy are available on sub-paths like /app1, /app2, and so on. Because AI cookies have path as /, they are shared between all apps. This results in the app that was refreshed last overwriting the session cookie for all other apps. That means AI session for the app that gets requested from server the most ends up with all the AI logs, and all other apps have their sessions deleted.

Possible solutions:
1. Do not set cookie path at all, browser will choose the current path to be the default. If AI SDK is initialized by /app1, cookie path will be set to /app1 by the browser, then app1 will track all of its AI cookies without other apps seeing and overwriting them.
2. Set cookie path to "", the behavior is identical to #1
3. Let cookie path be configurable for all cookies. You already have a setting for cookie domain, so you would just need to add one more setting for cookie path

Alternatives I considered:
1. Create our own fork of AI Client SDK and maintain it separately. We can choose to make #1, #2, or #3 change there and push it to our package server, then have everyone in the company use that. Obviously there are some maintenance effort issues with that
2. Implement #3 for this repo and submit it as a pull request. It could add a setting for cookie path and have it be '/' by default for backward compatibility. If set to null, cookie path will not be set (#1). That way the solution is both backward compatible and simple.
3. Write additional code and package it for our applications, that will initialize AI SDK with cookies disabled, and then create our own cookies. That has maintenance effort issues just like alternative #1 even though we avoid modifying AI Client SDK.

01binary:
Pull request for feature request #1434, adding cookie path to make AI client SDK work behind Azure Application Gateway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,7 @@ Most configuration fields are named such that they can be defaulted to falsey. A
 | enableSessionStorageBuffer | true | Default true. If true, the buffer with all unsent telemetry is stored in session storage. The buffer is restored on page load |
 | isCookieUseDisabled | false | Default false. If true, the SDK will not store or read any data from cookies.|
 | cookieDomain | null | Custom cookie domain. This is helpful if you want to share Application Insights cookies across subdomains. |
+| cookiePath | `/` | Custom cookie path. This is helpful if you want to control the path the cookie is available on. |
 | isRetryDisabled | false | Default false. If false, retry on 206 (partial success), 408 (timeout), 429 (too many requests), 500 (internal server error), 503 (service unavailable), and 0 (offline, only if detected) |
 | isStorageUseDisabled | false | If true, the SDK will not store or read any data from local and session storage. Default is false. |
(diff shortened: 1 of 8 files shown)


01binary:
I am looking at build errors, haven't found what was wrong yet - it's using a built system I am not familiar with.

MSNev:
I'm going to include the ability to set the path as part of #1091 which I'm doing in a working branch (MSNev/CookieHandling) -- note this includes several changes that will be split out into several PR's (i.e. it has an amount of minification work that is making the size of a PR too large) -- it does not yet support the suggested changes yet.

MSNev:
Assigning this PR to myself (so it doesn't get lost/forgotten)

01binary:
I understand you will use this PR to look at proposed changes but it will never be merged - closing - thank you!

PR #1439, opened 2020-12-03
Added missing property in IEventTelemetry documentation

Andy9FromSpace:
Added missing property in IEventTelemetry documentation


API-reference.md
```diff
@@ -142,6 +142,7 @@ Parameter | Type | Description
 ---|---|---
 `name` | string | **Required**<br>An event name string.
 `properties?` | dictionary | **Optional**<br>Map of string to any: Additional data used to [filter pages](https://azure.microsoft.com/documentation/articles/app-insights-api-custom-events-metrics/#properties) in the portal. Defaults to empty.
+`measurements?` | dictionary | **Optional**<br>Map of string to any: Additional data (metrics) used to [filter pages](https://azure.microsoft.com/documentation/articles/app-insights-api-custom-events-metrics/#properties) in the portal. Defaults to empty.
```
MSNev, line 145: This is actually excluded from the documentation as it's deprecated

MSNev: If you would like to still see it in the documentation please add strikeout around it to indicate that it's deprecated (using double ~~ around the fields.

~~strikeout example~~

Andy9FromSpace: I think it will help other developers as well to know that this field is deprecated. I updated the API reference accordingly.

MSNev (approved):

PR #1444, opened 2020-12-15
Fix for Click Plugin Issues

kryalama:
The following issues were fixed:
1. pageName populated based on doc.title. Same logic used in analytics extension.
2. Fixed click event name
3. Event data with no data/empty data will be discarded instead of populating them as "undefined"

+

Fixed Typo in 'disableInstrumentaionKeyValidation' config property


extensions/applicationinsights-clickanalytics-js/src/DataCollector.ts
```diff
@@ -115,15 +115,8 @@ export function getPageName(config: IClickAnalyticsConfiguration, overrideValues
     } else if (config.coreData && config.coreData.pageName) {
         return config.coreData.pageName;
     } else {
-        let loc = getLocation() || ({} as Location);
-        var pagename = loc.pathname || "";
```
MSNev, line 119: This feels like we should have a config for this, to allow users to determine whether they use the document title vs the URL. 

This is especially true nowadays that most browsers don't show or minimize the document title -- so it may not always be present.

hectorhdzg (approved):
LGTM

extensions/applicationinsights-clickanalytics-js/src/DataCollector.ts
kryalama: I was thinking users can override the pageName using the config.coreData.pageName config.

kryalama: We provide a callback [config](https://github.com/microsoft/ApplicationInsights-JS/blob/9271cf7bf59495e0d696688054d29b14316fc098/extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts#L116) to override the default pageName capturing behavior.

MSNev (approved):

MSNev:
Looks like it was because of this comment "pageName populated based on doc.title. Same logic used in analytics extension.", I don't remember why this was needed...

Can you check with hector and unless this would cause the event to fail to be ingested by the back end adding as an off by default options (I think) would be good.

PR #1457, opened 2021-01-13
Click Analytics Plugin Update

kryalama:
Following updates have been made:
1. Change the way custom event name is populated. Previosuly pageName is used as custom event name.
2. Added pageName as new custom attribute
3. Added parentId as new custom attribute
4. Added some more tests

### DIFF
--- a/extensions/applicationinsights-clickanalytics-js/package.json
+++ b/extensions/applicationinsights-clickanalytics-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft/applicationinsights-clickanalytics-js",
-  "version": "2.5.11",
+  "version": "2.5.10",
   "description": "Microsoft Application Insights Click Analytics extension",
   "main": "dist/applicationinsights-clickanalytics-js.js",
   "module": "dist-esm/applicationinsights-clickanalytics-js.js",
--- a/extensions/applicationinsights-clickanalytics-js/package.json
+++ b/extensions/applicationinsights-clickanalytics-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft/applicationinsights-clickanalytics-js",
-  "version": "2.5.10",
+  "version": "2.5.11",
   "description": "Microsoft Application Insights Click Analytics extension",
   "main": "dist/applicationinsights-clickanalytics-js.js",
   "module": "dist-esm/applicationinsights-clickanalytics-js.js",
--- a/extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts
+++ b/extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts
@@ -39,10 +39,10 @@ export interface IClickAnalyticsConfiguration {
      */
     defaultRightClickBhvr?: string | number;
     /**
-     * Flag to drop events that donot have custom event names, no parentId and no data in content (basically no useful click data).
+     * Flag to drop events that do not have custom event names, no parentId and no data in content (basically no useful click data).
      * Default will be false
      */
-    disableUndefinedEventsTracking?: boolean;
+    dropInvalidEvents?: boolean;
 }
 
 /**
--- a/extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts
+++ b/extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts
@@ -38,6 +38,11 @@ export interface IClickAnalyticsConfiguration {
      * value will be ovverriden if the element has the data-*-bhvr attribute present.
      */
     defaultRightClickBhvr?: string | number;
+    /**
+     * Flag to drop events that donot have custom event names, no parentId and no data in content (basically no useful click data).
+     * Default will be false
+     */
+    disableUndefinedEventsTracking?: boolean;
 }
 
 /**
--- a/extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts
+++ b/extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts
@@ -45,9 +45,9 @@ export interface IClickAnalyticsConfiguration {
  */
 export interface ICustomDataTags {
     /**
-     * When a particular element is not tagged with content name prefix or content name prefix is not provided by user, this flag is used to collect standard HTML attribute for contentName.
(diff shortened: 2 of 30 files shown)


extensions/applicationinsights-clickanalytics-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft/applicationinsights-clickanalytics-js",
-  "version": "2.5.10",
+  "version": "2.5.11",
```
MSNev, line 3: We don't update this separately, we update all of the versions are part of the Release PR

extensions/applicationinsights-clickanalytics-js/src/handlers/DomContentHandler.ts
```diff
@@ -75,21 +71,24 @@ export class DomContentHandler implements IContentHandler {
                 }
             } else {
                 // traverse up the DOM to find the closest parent with data-* tag defined
-                contentElement = walkUpDomChainWithElementValidation(element, this._isTracked, dataTagPrefix);
-                elementContent = extend(elementContent, this._populateElementContentwithDataTag( contentElement, element, dataTagPrefix, parentDataTagPrefix));
+                //contentElement = walkUpDomChainWithElementValidation(element, this._isTracked, dataTagPrefix);
```
MSNev, line 74: Don't we still want to do this -- just not for the id

extensions/applicationinsights-clickanalytics-js/package.json
kryalama: Reverted to 2.5.10

extensions/applicationinsights-clickanalytics-js/src/handlers/DomContentHandler.ts
kryalama: We are doing this only when user has specified the parentDataTag. This tag does two things. IF present it auto populates parent content and also used as an identifier to stop traversing up the DOM to capture content name and value of elements when encountered with this tag.

extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts
```diff
@@ -38,16 +38,21 @@ export interface IClickAnalyticsConfiguration {
      * value will be ovverriden if the element has the data-*-bhvr attribute present.
      */
     defaultRightClickBhvr?: string | number;
+    /**
+     * Flag to drop events that donot have custom event names, no parentId and no data in content (basically no useful click data).
+     * Default will be false
+     */
+    disableUndefinedEventsTracking?: boolean;
```
kryalama, line 45: I am open to a better name, if you think its not very intuitive

MSNev: It does seem a bit long, but it is descriptive -- not sure about describing them as undefined though -- as we are talking about events here are some suggestions
- trackUndefined -- default to true
- ignoreUndefined -- default to false
- ignoreInvalid -- default to false
- dropInvalid -- default to false
- skipInvalid -- default to false

extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts
```diff
@@ -38,16 +38,21 @@ export interface IClickAnalyticsConfiguration {
      * value will be ovverriden if the element has the data-*-bhvr attribute present.
      */
     defaultRightClickBhvr?: string | number;
+    /**
+     * Flag to drop events that donot have custom event names, no parentId and no data in content (basically no useful click data).
```
MSNev, line 42: nit: donot => don't or do not

extensions/applicationinsights-clickanalytics-js/src/events/PageAction.ts
```diff
@@ -21,23 +21,26 @@ export class PageAction extends WebEvent {
...
+        this._populateEventDataIfPresent(event.data, 'actionType', pageActionEvent.actionType);
+        this._populateEventDataIfPresent(event.data, 'behavior', pageActionEvent.behavior);
+        this._populateEventDataIfPresent(event.data, 'clickCoordinates', pageActionEvent.clickCoordinates);
+        this._populateEventDataIfPresent(event.data, 'content', pageActionEvent.content);
+        this._populateEventDataIfPresent(event.data, 'targetUri', pageActionEvent.targetUri);
         this._populateEventDataIfPresent(event.data, 'timeToAction', pageActionEvent.timeToAction);
         this._populateEventDataIfPresent(event.data, 'refUri', pageActionEvent.refUri);
+        this._populateEventDataIfPresent(event.data, 'pageName', pageActionEvent.pageName);
+        this._populateEventDataIfPresent(event.data, 'parentId', pageActionEvent.parentId);
         for (let property in properties) {
```
MSNev, line 45: As properties is optional it might not exist.
These 2 lines (for..in and hasOwnProperty) can be replased by using objForEachKey() (from CoreUtils) which also checks if the target (properties) exists.

extensions/applicationinsights-clickanalytics-js/src/events/PageAction.ts
```diff
@@ -115,4 +133,15 @@ export class PageAction extends WebEvent {
         }
     }
 
+    private _sanitizePageActionEventContent(pageActionContent: any) {
+        if(pageActionContent) {
+            delete pageActionContent.id;
+            delete pageActionContent.parentid;
+            delete pageActionContent.parentname;
+            if(isValueAssigned(this._config.dataTags.parentDataTag)) {
```
MSNev, line 143: Can there be a case where _config or _config.dataTags is null or uindefined? If so we should guard against this throwing a JS error

MSNev (approved):
Approved with suggestions

extensions/applicationinsights-clickanalytics-js/src/Interfaces/Datamodel.ts
kryalama: thanks for your suggestion. dropInvalidEvents looks better I guess

extensions/applicationinsights-clickanalytics-js/src/events/PageAction.ts
kryalama: Actually by default we create an empty object this._config.dataTags

MSNev (approved):

PR #1468, opened 2021-02-05
Enable fetch automatic dependency tracking by default

hectorhdzg:
Fetch API is widely used so enabling this automatic tracking by default


README.md
```diff
@@ -275,7 +275,7 @@ Most configuration fields are named such that they can be defaulted to falsey. A
 | samplingPercentage | 100 | Percentage of events that will be sent. Default is 100, meaning all events are sent. Set this if you wish to preserve your datacap for large-scale applications. |
 | autoTrackPageVisitTime | false | If true, on a pageview,the previous instrumented page's view time is tracked and sent as telemetry and a new timer is started for the current pageview. Default is false. |
 | disableAjaxTracking | false | If true, Ajax calls are not autocollected. Default is false. |
-| disableFetchTracking | true | If true, Fetch requests are not autocollected. Default is true |
+| disableFetchTracking | false | If true, Fetch requests are not autocollected. Default is false |
```
MSNev, line 278: It might be worth calling out that any environment that uses a polyfil this may cause duplicate dependency logging.

For react native I detect the default polyfil (whatwg-fetch) so this is avoided, but other polyfils may have an issue

xiao-lix, line 278: Update the doc to "default is false" in `IConfig.ts`:
```
/**
     * @description If true, Fetch requests are not autocollected. Default is **true**.
     * @type {boolean}
     * @memberof IConfig
     * @defaultValue true
     */
    disableFetchTracking?: boolean;
```

MSNev (changes requested):
Based on off-line discussions about possible breaking change (because the default will change), we need to delay and/or find a different way to change this 😞

MSNev (approved):

PR #1469, opened 2021-02-08
CoreUtils - Tree-Shaking ennhancements

MSNev:
- extract and export each CoreUtils function as first class
- stop using CoreUtils class within core, using the new direct functions

PR #1474, opened 2021-02-12
Stamp Specific endpoint changes

kryalama:
[Change spec](https://microsoft.sharepoint.com/teams/OneDSJavascript/_layouts/15/Doc.aspx?sourcedoc={b0af8db9-909e-4bf4-9c04-2fa435460c7e}&action=edit&wd=target%28Notes.one%7C32cb49ff-af84-4654-aca3-76973d0907eb%2FStamp%20Specific%20Endpoint%20support%7C422648d7-1046-4ac3-bec0-4cde3b1395a1%2F%29&wdorigin=703)

//TODO
1) Once we have the new api ' /v2.1/track' available , make necessary updates to SDK.
2) Add  unit tests if possible
3) Once Breeze have the ability to accept query parameter 'redirect=false', update the beacon API code and the snippet code to call the new url with redirect=false parameter.


MSNev:
Note: it looks like the test pass failed for the same issue I talked about in standup today.

My PR #1469 has the temporary fix for this -- it's a bit big though, but if you have a chance to review would be good.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -357,6 +361,15 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
             _self._xhrReadyStateChange = (xhr: XMLHttpRequest, payload: string[], countOfItemsInPayload: number) => {
                 if (xhr.readyState === 4) {
                     let response: IBackendResponse = null;
+
+                    // check if the xhr's responseURL is same as endpoint url
+                    // TODO after 10 redirects force send telemetry with'redirect=false' as query parameter. 
+                    if(xhr.responseURL !== _self._senderConfig.endpointUrl()) {
```
MSNev, line 367: Probably should check that
- the responseURL is not empty and look valid (some XHR polyfill's might not set this etc)

Additional considerations.
- Do we want to retry the original URL if the redirected one also continues to fail
- Do we only want to update on a successful send (200-299 response status code)

MSNev:
> Note: it looks like the test pass failed for the same issue I talked about in standup today.

Fixes for the test failures (dts generation) are now checked in, to rebasing should resolve the test issues

channels/applicationinsights-channel-js/src/Sender.ts
kryalama: I think updating the url only after we get a successful xhr send is a better approach.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -386,6 +400,11 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
                                 _InternalMessageId.TransmissionFailed, `. Offline - Response Code: ${xhr.status}. Offline status: ${Offline.isOffline()}. Will retry to send ${payload.length} items.`);
                         }
                     } else {
+
+                        // check if the xhr's responseURL is same as endpoint url
+                        // TODO after 10 redirects force send telemetry with 'redirect=false' as query parameter.
```
MSNev, line 405: What's the timeframe for this todo?

MSNev (approved):

Issue #1473, opened 2021-02-10
[BUG] New dts gneration doesn't work when the environment doesn't have powershell


MSNev:
The dtsgen.ps1 script called from the default build task fails in environments when powershell is not available.
- github build environment checks run on linux.

The script should be converted from powershell to JS so it can natively run.

MSNev:
Upgrading to a p1 as this is causing the github build to fail silently as it's built on a ubuntu instance.

MSNev:
Adding "fixed - waiting release" for tracking, but as this is a build issue it is technically already fixed and active.

MSNev:
v2.6.0 is now fully deployed to all CDN endpoints

PR #1475, opened 2021-02-17
[BUG] New dts gneration doesn't work when the environment doesn't have powershell #1473

MSNev:
- Convert PowerShell script to a nodejs one.


scripts/dtsgen.js
```diff
@@ -0,0 +1,162 @@
+//
```
MSNev, line 1: converted the dtsgen.ps1 file (in same folder)

hectorhdzg (approved):

Issue #1478, opened 2021-02-19
[Bug] Ajax tracking for XHR and fetch is not always setting the start time correctly


MSNev:
We were having issues with pageview where its "eventTime" in appinsights was actually the end of the event, while in dotnet core appinsights the eventTime seems to be the start of the event.. I tested this by setting the startTime of a dependency telemetry manually with the correct value with a matching pageView request.

![image](https://user-images.githubusercontent.com/8827790/108437993-7a79c080-72a2-11eb-98d6-97ef48c5ba71.png)

Here B is the correct time, since it aligns with the backend services timestamps, while A is the incorrect pageview time which is perfectly off by its duration.

Managed to fix this by adding a TelemetryInitializer and overriding the .startTime property to follow the AJAX codepath shown as resolved for this issue https://github.com/microsoft/ApplicationInsights-JS/issues/628.

`appInsights.addTelemetryInitializer(TelemetryInitializer);`

```
const setStartTimeToCorrectValue = (envelope: ITelemetryItem, duration: number) => {
  const newStartTime = new Date((new Date(envelope.time) as any) - duration);
  envelope.baseData.startTime = newStartTime;
  envelope.time = newStartTime.toISOString();
};

export default (envelope: ITelemetryItem) => {
  if (envelope.baseType === 'RemoteDependencyData' && envelope.baseData.duration) {
    const duration = envelope.baseData.duration;
    setStartTimeToCorrectValue(envelope, duration);
  } else if (envelope.baseType === 'PageviewData' && envelope.data.duration) {
    const duration = envelope.data.duration;
    setStartTimeToCorrectValue(envelope, duration);
  }
};
```

Note that I also overrode dependency values, since the auto tracking of FETCH calls were also wrong for us. They would start in appinsights a duration time after what our downstream services reported.

_Originally posted by @mRowlinson27 in https://github.com/microsoft/ApplicationInsights-JS/issues/781#issuecomment-781720528_

MSNev:
This is a very good catch / observation. Drilling into the code the problem is that ajax.ts trackDependencyDataInternal() sets the startTime to now when no startTime is not present -- which is mostly correct for manually calling trackDependencyData

But, the XHR and fetch also doesn't set the the startTime (unless) you have enabled "enableAjaxPerfTracking" AND it managed to find the performance resource entry.

It should be setting the startTime to the "requestSentTime" of the ajax record to report the correct value.

MSNev:
[...]

PR #1485, opened 2021-02-23
[Bug] Ajax tracking for XHR and fetch is not always setting the starttime correctly #1478

MSNev:
- Add some minification fixes

### DIFF
--- a/extensions/applicationinsights-dependencies-js/src/ajaxRecord.ts
+++ b/extensions/applicationinsights-dependencies-js/src/ajaxRecord.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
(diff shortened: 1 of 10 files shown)


extensions/applicationinsights-dependencies-js/src/ajaxRecord.ts
```diff
@@ -293,6 +293,12 @@ export class ajaxRecord {
                     method: self.method,
                     [strProperties]: { HttpMethod: self.method }
                 } as IDependencyTelemetry;
+
+                if (self.requestSentTime) {
```
hectorhdzg, line 297: I suppose this is the actual fix, how was startTime being set before?

MSNev: Yes it is. The previous issue was that it wasn't getting set until the trackDependency call occurred, I've kept that code because it can still be called by external users and not set.

https://github.com/microsoft/ApplicationInsights-JS/blob/master/extensions/applicationinsights-dependencies-js/src/ajax.ts#L352-L354

hectorhdzg (approved):

PR #1487, opened 2021-02-25
add angular plugin deprecated message

xiao-lix:


extensions/applicationinsights-angularplugin-js/README.md
```diff
@@ -1,3 +1,4 @@
+> Note: Angular Plugin is moved to its own repo [here](https://github.com/microsoft/applicationinsights-angularplugin-js). This Plugin is deprecated. Please refer to the new repo for usage instructions, and open issue there.
```
MSNev, line 1: It probably better to phrase as moved as the plugin itself is not deprecated, just where the code lives :smile:

xiao-lix: That makes a lot sense... XD 
What do you think of the update? Thanks.

MSNev (approved):

PR #1488, opened 2021-02-26
Additional TreeShaking and minification for Properties, Common and Analytics

MSNev:


AISKU/src/Initialization.ts
```diff
@@ -463,10 +463,14 @@ export class Initialization implements IApplicationInsights {
 
                 // Back up the current session to local storage
                 // This lets us close expired sessions after the cookies themselves expire
-                const ext = appInsightsInstance.appInsights.core['_extensions'][PropertiesPluginIdentifier];
-                if (ext && ext.context && ext.context._sessionManager) {
-                    ext.context._sessionManager.backup();
-                }
+                arrForEach(appInsightsInstance.appInsights.core['_extensions'], (ext: any) => {
```
xiao-lix, line 466: Does this change add extra loop compared with previous code? Looks like previously you only need to check if "ext" exists, with current change, you'll loop the extensions array and check every ext's identifier. 🤔

MSNev: The previous code actually didn't work, because it "assumed" that the array contained a key with the string value of PropertiesPluginIdentifier... Which is doesn't.

So the loop is required to find the properties plugin within the array -- this part is actually a bug fix that was found by adding additional type checks caused by the refactoring.

common/Tests/Framework/src/AITestClass.ts
```diff
@@ -15,6 +15,15 @@ export interface FakeXMLHttpRequest extends XMLHttpRequest {
     respond: (status: number, headers: any, body: string) => void;
 }
 
+function _getObjName(target:any, unknownValue?:string) {
+    if (target.hasOwnProperty("prototype")) {
+        // Look like a prototype
+        return target.name || unknownValue || ""
+    }
+
+    return (((target || {})["constructor"]) || {}).name || unknownValue || "";
```
xiao-lix, line 24: Is this always the object type name? like "Object", "String", "Number", etc?

MSNev: Depending on the runtime environment, es3 doesn't have obj.constructor (which is why the ((target || {})["constructor"]) || {}).

I also pinched this from what I did in dynamicProto 😄

xiao-lix (approved):

kryalama (approved):

extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -2,9 +2,23 @@
 // Licensed under the MIT License.
 
 import { ITelemetryConfig } from '../Interfaces/ITelemetryConfig';
-import { Util, IUser, IUserContext } from '@microsoft/applicationinsights-common';
+import { Util, IUserContext } from '@microsoft/applicationinsights-common';
```
ramthi, line 5: Was this Util import missed?

MSNev: No, the class is still using the Util.removeStorage(), I have not **yet** refactored out all of the Util functions like I did for CoreUtils as the Cookie Manager change is also refactoring the Util set / get cookie functions, so I didn't want to keep adding onto this PR and complicate the merge than it already was 😄

PR #1489, opened 2021-02-27
bug fix - track dependency name overwrites data

### LINKED ISSUE #1239: [BUG] Dependency tracking data not as expected
**Description/Screenshot**
I am attempting to set up AppInsights to work with Axios interceptors.  So that when we receive a bad response/error from a dependent web service it is logged through App Insights directly.

Example of telemetry object used in `trackDependencyData()` - 
```
{
  "name": "Browser.DependencyFailure",
  "id": "1",
  "responseCode": 400,
  "duration": 238,
  "success": false,
  "type": "get",
  "target": "https://<actualUrl>!poly?$select=*,&$filter=$id+in+'dc7c26d6-d97f-4ef0-be15-7d549b5a5e44']",
  "correlationContext": "appId=cid-v1:a5aa0fa1-9dab-4c2d-92fb-c798aad3f142",
  "data": "{\"errorId\":null,\"errorMessage\":\"Unexpected character at index 0 in the specified literal. Expected: [ But was: '\",\"errorDescription\":null}"
}
```
Example of what actually gets sent in network tab - 
```
{
  "baseType": "RemoteDependencyData",
  "baseData": {
    "id": "1",
    "ver": 2,
    "name": "Browser.DependencyFailure",
    "resultCode": "400",
    "duration": "00:00:00.102",
    "success": false,
    "data": "Browser.DependencyFailure",
    "target": "<base-url> | appId=cid-v1:a5aa0fa1-9dab-4c2d-92fb-c798aad3f142",
    "type": "get",
    "properties": {},
    "measurements": {}
  }
}
```
**Steps to Reproduce**
Instantiate app insights.
Create `IDependencyTelemetry `object
send telemetry object to `appInsights.trackDependencyData(...)`
In my case, dependency is a non MS URL.  It's just a dependent web service we use.

 - OS/Browser:  Windows 10 / Chrome 80
 - SDK Version [e.g. 22]: 2.5.3
 - How you initialized the SDK:
From inside an app insights wrapper class like so - 
```
public constructor(
    appInsightsKey: string, 
    private readonly _routeService: RouteService, 
    private readonly _authnService: AuthenticationService, 
  ) {
    this._appInsights = new ApplicationInsights({
      config: {
        instrumentationKey: appInsightsKey,
        disableAjaxTracking: true,
        disableDataLossAnalysis: true,
        disableFetchTracking: true,
        disableExceptionTracking: false,
        disableTelemetry: false,
        isBeaconApiDisabled: !navigator.sendBeacon,
      }
    });
    this._appInsights.loadAppInsights();
}
```
**Expected behavior**
`trackDependencyData()` sends exactly what I give it through the `IDependencyTelemetry` interface.

**Additional context**

I'm not sure if I'm using this incorrectly or if I should just use the `TrackEvent` api or if I have to add in the Dependency Plugin (which I can't find much documentation about) or what.  Any guidance would be helpful.

In addition.  Your ReadMe API file doesn't match the types in the package.  Which confuses me a bit more. -
```
import { IPartC } from './IPartC';
/**
 * DependencyTelemetry telemetry interface
 */
export interface IDependencyTelemetry extends IPartC {
    id: string;
    name?: string;
    duration?: number;
    success?: boolean;
    responseCode: number;
    correlationContext?: string;
    type?: string;
    data?: string;
    target?: string;
}
```
Vs. ReadMe - 
image

xiao-lix:
Fix #1239

### DIFF
--- a/API-reference.md
+++ b/API-reference.md
@@ -122,12 +122,15 @@ The `IDependencyTelemetry` interface is described below
 Parameter | Type | Description
 ---|---|---
 `id` | string | **Required**<br>Unique id, this is used by the backend to correlate server requests.
-`responseCode` | number | **Required**<br>Response code returned by the dependency request (e.g., `200` for a success)
-`absoluteUrl?` | string | **Optional**<br>Absolute url used to make the dependency request
-`success?` | boolean | **Optional**<br>Whether or not the request was successful or not (e.g., `responseCode` in the range 200-299)
-`commandName?` | string| **Optional**<br>Command used to make the dependency request
+`responseCode` | number | **Required**<br>Response code returned by the dependency request (e.g., `200` for a success).
+`name?` | string | **Optional**<br>Name of the command initiated with this dependency call.
 `duration?` | number | **Optional**<br>Elapsed time of request & reply
-`method?` | string | **Optional**<br>Represents request verb (GET, POST, etc.)
+`success?` | boolean | **Optional**<br>Whether or not the request was successful or not (e.g., `responseCode` in the range 200-299)
+`startTime?` | Date | **Optional**<br>Dependency start timestamp.
+`correlationContext?` | string | **Optional**<br>correlation context from the server.
+`type?` | string | **Optional**<br>Dependency type name.
+`data?` | string | **Optional**<br>Command initiated by this dependency call.
+`target?` | string | **Optional**<br>Target site of a dependency call.
 `properties?` | dictionary | **Optional**<br>Map of string to any: Additional data used to [filter pages](https://azure.microsoft.com/documentation/articles/app-insights-api-custom-events-metrics/#properties) in the portal. Defaults to empty.
 
 ### trackEvent
--- a/API-reference.md
+++ b/API-reference.md
@@ -129,7 +129,7 @@ Parameter | Type | Description
 `startTime?` | Date | **Optional**<br>Dependency start timestamp.
 `correlationContext?` | string | **Optional**<br>correlation context from the server.
 `type?` | string | **Optional**<br>Dependency type name.
-`data?` | string | **Optional**<br>Command initiated by this dependency call.
+`data?` | string | **Internal**<br>Command initiated by this dependency call. This is not a user settable field. `data` is automatically set based on the dependency name (if available). Please include additional custom properties as part of the `properties` value below.
 `target?` | string | **Optional**<br>Target site of a dependency call.
 `properties?` | dictionary | **Optional**<br>Map of string to any: Additional data used to [filter pages](https://azure.microsoft.com/documentation/articles/app-insights-api-custom-events-metrics/#properties) in the portal. Defaults to empty.
 
--- a/channels/applicationinsights-channel-js/Tests/Sender.tests.ts
+++ b/channels/applicationinsights-channel-js/Tests/Sender.tests.ts
@@ -445,7 +445,7 @@ export class SenderTests extends TestClass {
 
                 // Assert baseData
                 QUnit.assert.ok(baseData.name);
-                QUnit.assert.equal("Some name given", baseData.data);
+                QUnit.assert.equal("some data", baseData.data);
                 QUnit.assert.equal("some id", baseData.id);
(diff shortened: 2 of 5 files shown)


channels/applicationinsights-channel-js/Tests/Sender.tests.ts
```diff
@@ -445,7 +445,7 @@ export class SenderTests extends TestClass {
 
                 // Assert baseData
                 QUnit.assert.ok(baseData.name);
-                QUnit.assert.equal("Some name given", baseData.data);
+                QUnit.assert.equal("some data", baseData.data);
```
xiao-lix, line 448: Hi @MSNev , I noticed you committed last time on this file. Do you know why we assign "name" value to "data" value when "data" value has content?

shared/AppInsightsCommon/src/Util.ts
```diff
@@ -743,8 +743,8 @@ export class CorrelationIdHelper {
 }
 
 export class AjaxHelper {
-    public static ParseDependencyPath(logger: IDiagnosticLogger, absoluteUrl: string, method: string, commandName: string) {
-        let target, name = commandName, data = commandName;
+    public static ParseDependencyPath(logger: IDiagnosticLogger, absoluteUrl: string, method: string, commandName: string, customData: string) {
+        let target, name = commandName, data = customData || commandName;
```
xiao-lix, line 747: Update to use "data" value when "data" property exists, otherwise fallback to use commandName <-- the previous behavior.

MSNev: Looking back at the history of this field (takes quite a bit of digging), the change to have the "data" field being the sanitized version of the "name" is a requirement of downstream systems.

And by default the remote dependency doesn't populate the "data" field (I think because it's derived from name)

shared/AppInsightsCommon/src/Telemetry/RemoteDependencyData.ts
```diff
@@ -52,8 +52,8 @@ export class RemoteDependencyData extends GeneratedRemoteDependencyData implemen
 
         this.type = DataSanitizer.sanitizeString(logger, requestAPI);
 
-        const dependencyFields = AjaxHelper.ParseDependencyPath(logger, absoluteUrl, method, commandName);
-        this.data = DataSanitizer.sanitizeUrl(logger, commandName) || dependencyFields.data; // get a value from hosturl if commandName not available
+        const dependencyFields = AjaxHelper.ParseDependencyPath(logger, absoluteUrl, method, commandName, data);
+        this.data = DataSanitizer.sanitizeUrl(logger, data) || dependencyFields.data; // get a value from hosturl if commandName not available
```
MSNev, line 56: This would be a breaking change and looking at the history would cause issues.

MSNev (changes requested):
This would cause internal downstream issues as is.

API-reference.md
```diff
@@ -122,12 +122,15 @@ The `IDependencyTelemetry` interface is described below
...
-`commandName?` | string| **Optional**<br>Command used to make the dependency request
+`responseCode` | number | **Required**<br>Response code returned by the dependency request (e.g., `200` for a success).
+`name?` | string | **Optional**<br>Name of the command initiated with this dependency call.
 `duration?` | number | **Optional**<br>Elapsed time of request & reply
-`method?` | string | **Optional**<br>Represents request verb (GET, POST, etc.)
+`success?` | boolean | **Optional**<br>Whether or not the request was successful or not (e.g., `responseCode` in the range 200-299)
+`startTime?` | Date | **Optional**<br>Dependency start timestamp.
+`correlationContext?` | string | **Optional**<br>correlation context from the server.
+`type?` | string | **Optional**<br>Dependency type name.
+`data?` | string | **Optional**<br>Command initiated by this dependency call.
```
MSNev, line 132: While this is defined on the interface, it doesn't seem that it's supported as a user settable field. It's defined on the interface so the internal code has correct type checking.

So comment should either be removed or changed to **Internal**

MSNev (approved):

PR #1490, opened 2021-03-02
update react to v17.0.1

kryalama:
Tested the plugin with latest changes.
https://reactjs.org/blog/2020/08/10/react-v17-rc.html#other-breaking-changes


MSNev (approved):
To fix the build issues you will need to
- remove the es3Check() from the applicationinsights-react-js/rollup.config.js (as this is not supported on ES3 (IE 7/8))
- Add a note to the README.md stating that this extension will no longer work with ES3 by default because react 17.x does not support ES3 without the react-app-polyfill. (I don't this is a breaking issue as react browser support by default is IE9+ (ES5)

kryalama:
> To fix the build issues you will need to
> 
> * remove the es3Check() from the applicationinsights-react-js/rollup.config.js (as this is not supported on ES3 (IE 7/8))
> * Add a note to the README.md stating that this extension will no longer work with ES3 by default because react 17.x does not support ES3 without the react-app-polyfill. (I don't this is a breaking issue as react browser support by default is IE9+ (ES5)

Looks like we already have second point stated in the readme:
image

MSNev (approved):

Issue #1393, opened 2020-10-02
[BUG] enableAutoRouteTracking should also update the Operation Name with the hashroute


markwolff:
**Description/Screenshot**
`enableAutoRouteTracking: true` should update `ai.operation.name` to include the hashroute in the pathname

https://github.com/microsoft/ApplicationInsights-JS/blob/ae71ebf806da82e149b0af60f00d70b2c1d3e3f0/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts#L650-L653

xiao-lix:
Fixed with #1510

MSNev:
v2.6.2 is now fully deployed

PR #1510, opened 2021-03-24
update operation name

xiao-lix:
Solve #1393

### DIFF
--- a/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
+++ b/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
@@ -650,9 +650,12 @@ export class ApplicationInsights extends BaseTelemetryPlugin implements IAppInsi
                         _window.addEventListener(_self.config.namePrefix + "locationchange", () => {
                             if (_properties && _properties.context && _properties.context.telemetryTrace) {
                                 _properties.context.telemetryTrace.traceID = generateW3CId();
-                                (_location && _location.pathname && _location.hash) ? 
-                                    _properties.context.telemetryTrace.name = _location.pathname.concat(_location.hash) :
-                                    _properties.context.telemetryTrace.name = _location && _location.pathname || "_unknown_";
+                                let traceLocationName = "_unknown_";
+                                if (_location && _location.pathname) {
+                                    traceLocationName = _location.pathname + (_location.hash || "");
+                                }
+
+                                _properties.context.telemetryTrace.name = traceLocationName;
                             }
                             if (_currUri) {
                                 _prevUri = _currUri;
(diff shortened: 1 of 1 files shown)


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -647,7 +647,7 @@ export class ApplicationInsights extends BaseTelemetryPlugin implements IAppInsi
                         _window.addEventListener(_self.config.namePrefix + "locationchange", () => {
                             if (_properties && _properties.context && _properties.context.telemetryTrace) {
                                 _properties.context.telemetryTrace.traceID = generateW3CId();
-                                _properties.context.telemetryTrace.name = _location && _location.pathname || "_unknown_";
+                                _properties.context.telemetryTrace.name = (_location && _location.pathname).concat(_location.hash) || "_unknown_";
```
MSNev, line 650: location does not exist in IE8 and below or Firefox < v14 etc.
To avoid a crash we need to ensure that location exists (which is happening for pathname) 
-- additional, if pathname is not defined this will throw because concat is not a function of undefined.

MSNev: Although, because this is in the location change event that "unlikely"

MSNev: Do we have an Issue for this?
If so can we please include the number in the commit. And this *may* also not be something everyone wants as a change. So we should add this as a config, so it can be disabled for users that don't want this capability

xiao-lix: @MSNev thanks for the review. Yes there's an open issue here: #1393 
Should I added it in the comment above, do you usually add issue number in each commit? 

I updated the check, can you please have a second look - as for the config check - the issue says this should be added when `enableAutoRouteTracking` is true, I assume there's no need to add another config? But I do get your point. Should we add new config for it ? Thanks!

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -650,7 +650,9 @@ export class ApplicationInsights extends BaseTelemetryPlugin implements IAppInsi
                         _window.addEventListener(_self.config.namePrefix + "locationchange", () => {
                             if (_properties && _properties.context && _properties.context.telemetryTrace) {
                                 _properties.context.telemetryTrace.traceID = generateW3CId();
-                                _properties.context.telemetryTrace.name = _location && _location.pathname || "_unknown_";
+                                (_location && _location.pathname && _location.hash) ?
```
MSNev, line 653: Hmm, way to complicated and hard to read -- also not very minifiable. What about doing this instead
```
let traceLocationName = "_unknown_";
if (_location && _location.pathname) {
    traceLocationName = _location.pathname + (_location.hash || "");
}

_properties.context.telemetryTrace.name = traceLocationName;
```

MSNev (approved):

Issue #1518, opened 2021-03-30
[BUG] v2.6.0 is not re-hydrating the automatic session id correctly for each request


MSNev:
The automatic session re-hydration is incorrectly resetting the value for each request, due to a bug in the validation logic

MSNev:
v2.6.1 is now fully deployed to NPM and all CDN endpoints

PR #1519, opened 2021-03-30
[Release][Core] Increase version to v2.6.1

MSNev:
- [BUG] v2.6.0 is not re-hydrating the automatic session id correctly for each request #1518


extensions/applicationinsights-properties-js/src/Context/Session.ts
```diff
@@ -169,9 +169,8 @@ export class _SessionManager {
                             _logger.throwInternal(LoggingSeverity.WARNING,
                                 _InternalMessageId.SessionRenewalDateIsZero,
                                 "AI session renewal date is 0" + sessionReset);
-                        } else if (session.id) {
-                            // If we don't have an id then considered it to be expired
-                            // Only assign the values if everything looks good
+                        } else if (tokens[0]) {
```
MSNev, line 172: This is where the bug was introduced - as the caller of this function only calls _initializeAutomaticSession() if (!session.id) -- so this was always false and subsequently got removed from the code.

PR #1531, opened 2021-04-13
bug fix - update network time calculation in pv performance manager

### LINKED ISSUE #1440: [BUG] PageViewPerformanceManager.populatePageViewPerformanceEvent() is always returning zero for the network time
In the populatePageViewPerformanceEvent method of PageViewPerformanceManager there are a couple of issues that need addressing.
1) For Browsers that support the Navigation Timing perf event the calculation of the "networkConnect" time via 
var network = DateTimeUtils.GetDuration(navigationTiming.startTime, navigationTiming.connectEnd);
is always zero as the GetDuration() only returns a value if BOTH start and end are not zero, null or undefined, however, the startTime (according to https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming) is now always zero
2) The scoping of the variables total, network, request, response, dom are incorrect, while the code seems to be working (with the current version) we should fix the scoping of these variables so they are not "defined" in the if / else clause to avoid a future break by a version of TypeScript, Uglify or browser JS implementation.

xiao-lix:
Close #1440

### DIFF
--- a/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
+++ b/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
@@ -37,22 +37,36 @@ export class PageViewPerformanceManager {
          *  |             ||             |              |         |-loadEventEnd
          *  |---network---||---request---|---response---|---dom---|
          *  |--------------------------total----------------------|
+         * 
+         *  total = The difference between the load event of the current document is completed and the first recorded timestamp of the performance entry : https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings#duration
+         *  network = Redirect time + App Cache + DNS lookup time + TCP connection time 
+         *  request = Request time : https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings#request_time
+         *  response = Response time
+         *  dom = Document load time : https://html.spec.whatwg.org/multipage/dom.html#document-load-timing-info
+         *      = Document processing time : https://developers.google.com/web/fundamentals/performance/navigation-and-resource-timing/#document_processing
+         *      + Loading time : https://developers.google.com/web/fundamentals/performance/navigation-and-resource-timing/#loading
          */
         const navigationTiming = this.getPerformanceNavigationTiming();
         const timing = this.getPerformanceTiming();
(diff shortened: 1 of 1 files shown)


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewPerformanceManager.ts
```diff
@@ -57,7 +57,10 @@ export class PageViewPerformanceManager {
         if (navigationTiming || timing) {
             if (navigationTiming) {
                 total = navigationTiming.duration;
-                network = total - dateTimeUtilsDuration(navigationTiming.connectEnd, navigationTiming.loadEventEnd);
+                let afterNetwork = dateTimeUtilsDuration(navigationTiming.connectEnd, navigationTiming.loadEventEnd);
+                if (!isNotNullOrUndefined(afterNetwork)) {
```
MSNev, line 61: This still won't populate the values... As per the bug details, startTime is always zero so dateTimeUtilsDuration() will always return null...

xiao-lix: Yeah so I avoid using startTime. Instead of checking the difference of `startTime` between `connectEnd` (which is always null), the network time is `total` minus `afterNetwork`. No `startTime` involved here. Did I miss something... 
image

xiao-lix: Oh I guess this is comparing my first commit with my second commit... I added a null check for afterNetwork, making sure `total - afterNetwork` doesn't make network = total (total - null = total). If afterNetwork is null, network is still 0.

MSNev: Oh, I missed that part I was just looking how it was getting set and not the value.
Using afterNetwork - total, I think is also not quite the correct approach as it might potentially include additional time on some browsers, I was thinking this change should be something a bit simplier.
```
if (navigationTiming.startTime === 0) {
   network = navigationTiming.connectEnd;
} else {
  // Support older browsers where the startTime is not zero
  network = dateTimeUtilsDuration(navigationTiming.startTime, navigationTiming.connectEnd);
}
```

MSNev (approved):

Issue #1528, opened 2021-04-09
[BUG] `correlationHeaderExcludePatterns` is not honored in Ajax calls


jpfifer:
**Description/Screenshot**

The value of `correlationHeaderExcludePatterns` in `Snippet` is not honored by the SDK.

It appears that the [`getDefaultConfig()` in `AjaxMonitor`](https://github.com/microsoft/ApplicationInsights-JS/blob/22d3787cb9b328bbf3cab0ee38a686d4779f8031/extensions/applicationinsights-dependencies-js/src/ajax.ts#L154) doesn't include an empty value for `correlationHeaderExcludePatterns`. This causes the [configuration loading](https://github.com/microsoft/ApplicationInsights-JS/blob/22d3787cb9b328bbf3cab0ee38a686d4779f8031/extensions/applicationinsights-dependencies-js/src/ajax.ts#L217) to never see the configuration.

**Steps to Reproduce**
1. Set `correlationHeaderExcludePatterns` to `[/.*/]`
  This should cause no correlation headers to be sent at all
2. Run the application and verify that the correlation headers are being sent.

 - OS/Browser: macOS Chrome and Firefox
 - SDK Version [e.g. 22]: 2.5.10 and 2.6.1
 - How you initialized the SDK:
```
config: {
        instrumentationKey,
        extensions: [reactPlugin],
        extensionConfig: {
          [reactPlugin.identifier]: { history }
        },
        disableFetchTracking: false,
        enableCorsCorrelation: true,
        disableAjaxTracking: false,
        enableRequestHeaderTracking: true,
        enableResponseHeaderTracking: true,
        correlationHeaderExcludedDomains: [`...`],
        correlationHeaderExcludePatterns: [new RegExp('...')],
        distributedTracingMode: DistributedTracingModes.AI
      }
```

**Expected behavior**
Correlation headers should not be sent when the URI matches the exclude pattern

MSNev:
I think this is related to #1524

jpfifer:
I just upgraded to the current version and it `correlationHeaderExcludePatterns` appears to be working correctly.

MSNev:
v2.6.2 is now fully deployed

PR #1532, opened 2021-04-14
[BUG] address issues #1517 add unload listener and #1524 config from snippet

### LINKED ISSUE #1517: [BUG] addHousekeepingBeforeUnload should also be listening to the 'unload' event
Add and export the following helper to listen for all "unload" / "navigate" events

```
export function addPageUnloadEventListener(listener: any): boolean {
    // Hook the unload event for the document, window and body to ensure that the client events are flushed to the server
    // As just hooking the window does not always fire (on chrome) for page navigations.
    let pageUnloadAdded = addEventHandler("beforeunload", listener);
    pageUnloadAdded = addEventHandler("unload", listener) || pageUnloadAdded;
    pageUnloadAdded = addEventHandler("pagehide", listener) || pageUnloadAdded;
    return pageUnloadAdded;
}
```

And update `addHousekeepingBeforeUnload()` in AISKU/Initialization.ts to use it

### LINKED ISSUE #1524: [BUG] Config items are not functional for current snippet disableFlushOnBeforeUnload, disableFlushOnBeforeUnload and maybe others
See history in #1516

When using 
- a current snippet it initializes the extension/ai-analytics-js/ApplicationInsights instance as appInsightsInstance.appInsights which is NOT initializing the disable* config items... (This is part of the bug).
- But when using the legacy snippet this initializes the ApplicationInsightsDeprecated instance as appInsightsInstance.appInsights which initializes the disable* config items and the extension/ai-analytics-js/ApplicationInsights instance is linked as appInsightsInstance.appInsightsNew as this is where most of the work is done.

This looks like it's been broken like this for a while (so not going to elevate this to a P1), but linking to #1517 as there is a bunch of code that needs fixing here to restore compatibility.

xiao-lix:
1. Add event listener on `unload` event to address #1517 
2. Add to initialize config items to address #1524 
3. Add to initialize correlationHeaderExcludePatterns config item in ajax.ts to address #1528


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -76,6 +76,22 @@ export class ApplicationInsights extends BaseTelemetryPlugin implements IAppInsi
         config.enableAutoRouteTracking = stringToBoolOrDefault(config.enableAutoRouteTracking);
         config.namePrefix = config.namePrefix || "";
 
+        config.maxBatchSizeInBytes = config.maxBatchSizeInBytes > 0 ? config.maxBatchSizeInBytes : 102400; // 100kb
```
MSNev, line 79: Why are we doing all of these?

Setting the defaults for lines 80, 82 I get as they are not provided otherwise, but the other checks to "set" the existing value strictly to a boolean or like on line 88 to the same value doesn't seem right...

It is also really bad from a minification perspective as all of the long names on the config elements can't be compressed.

MSNev: ok, just went and looked at the bug and the implementation...
As the usage within the code iterates over the list of "keys" on the object returned by getDefaultConfig(), this sort of works -- it's just not very nice.

MSNev: Ok, so this line ```_self.config = ctx.getExtCfg<IConfig>(identifier)``` (Line 498 on my repo) deals with the "identifier" version of things as this will just return those values.... So it's just the root level ones that we are interested in that are getting missed (which is what the default stuff is effectively doing)...

I'm wondering whether we can do something really simple like... (_assignFn comes form the shims)
_self.config = __assignFn(ApplicationInsights.getDefaultConfig(), config, ctx.getExtCfg<IConfig>(identifier));

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts
```diff
@@ -76,6 +76,10 @@ export class ApplicationInsights extends BaseTelemetryPlugin implements IAppInsi
         config.enableAutoRouteTracking = stringToBoolOrDefault(config.enableAutoRouteTracking);
         config.namePrefix = config.namePrefix || "";
 
+        config.enableDebug = stringToBoolOrDefault(config.enableDebug);
+        config.disableFlushOnBeforeUnload = stringToBoolOrDefault(config.disableFlushOnBeforeUnload);
+        config.disableFlushOnUnload = stringToBoolOrDefault(config.disableFlushOnUnload, config.disableFlushOnBeforeUnload);
```
xiao-lix, line 81: @MSNev Sorry Nev for the confusion... I see what's going on here. 

For other configurations I added before, they're not analytics extension related and should not bring in here. 

`maxBatchSizeInBytes, maxBatchInterval, isBeaconApiDisabled, enableSessionStorageBuffer, isRetryDisabled, disableTelemetry` are `sender config` and initialized [here](https://github.com/microsoft/ApplicationInsights-JS/blob/3d4a6fa42a7a25ceedbf3131598819249797ad47/channels/applicationinsights-channel-js/src/Sender.ts#L81). And when user passes in user defined config, it receives [here](https://github.com/microsoft/ApplicationInsights-JS/blob/3d4a6fa42a7a25ceedbf3131598819249797ad47/channels/applicationinsights-channel-js/src/Sender.ts#L220).

 `maxAjaxCallsPerView , disableCorrelationHeaders, enableCorsCorrelation, correlationHeaderExcludedDomains, disableAjaxTracking` are `remote dependency extension config` , initialized [here ](https://github.com/microsoft/ApplicationInsights-JS/blob/3d4a6fa42a7a25ceedbf3131598819249797ad47/extensions/applicationinsights-dependencies-js/src/ajax.ts#L153)and received user defined [here](https://github.com/microsoft/ApplicationInsights-JS/blob/3d4a6fa42a7a25ceedbf3131598819249797ad47/extensions/applicationinsights-dependencies-js/src/ajax.ts#L218). 

And `disableCookiesUsage` is handled by the cookie manager you created. 

But for the above three configs, this is the place to initialize them. And on line 505, getDefaultConfig is missing the param `config` so the user defined configs dropped!

I think we need this config.[key] pattern because we don't need all keys from config, like "extensionConfig" is in config object too but we only need to initialize some of the configs... does that make sense?

MSNev (approved):

PR #1544, opened 2021-04-23
Add more files in publish and setActiveVersion scripts

kryalama:


extensions/applicationinsights-debugplugin-js/scripts/setActiveCdnVersion.ps1
```diff
@@ -717,7 +717,7 @@ if ($container -eq "beta" -or $container -eq "next") {
 
 if ($files.ContainsKey($activeVersion) -ne $true) {
     Log-Failure "Version [$activeVersion] does not appear to be deployed to [$container]"
-} elseif ($files[$activeVersion].Count -ne 4) {
+} elseif ($files[$activeVersion].Count -ne 12) {
```
MSNev, line 720: This will need to be 4 or 12 as currently we only have 4 published and it will be the next cycle before we publish 12.

Otherwise, this will fail for the same reason as the click analytics and nothing will get promoted.

MSNev: Or we just wait until the next cycle before adding the new files and this change for the debug plugin.

MSNev (approved):

Issue #1546, opened 2021-04-23
[Typings] Update the TypeScript typings to identify the readonly properties/fields and dynamic values of the snippet


MSNev:

MSNev:
v2.6.3 is now fully deployed

PR #1548, opened 2021-04-23
[Typings] Update the TypeScript typings to identify the readonly properties/fields and dynamic values of the snippet #1546

xiao-lix:
Address #1546


shared/AppInsightsCommon/src/Interfaces/ITelemetryContext.ts
```diff
@@ -40,7 +40,7 @@ export interface ITelemetryContext {
     /**
      * The object describing a user tracked by this object.
      */
-    user: IUserContext;
+    readonly user: IUserContext;
```
MSNev, line 43: There will be a LOT more than just these.

We need to go over all of the API surface and allocate the correct settings.

eg. All of the TelemetryContext objects should be readonly,

MSNev (approved):

PR #1555, opened 2021-04-27
[Release][React] Increase React Plugin to version 3.1.2

MSNev:
- Remove unnecessary shims import, fix packaging and tests


tools/rollup-plugin-uglify3-js/dist/esm/rollup-plugin-uglify3-js.js
```diff
@@ -0,0 +1,51 @@
+/*!
```
MSNev, line 1: It's not ideal to check this in, but for now it's the quickest way to get the react and react-native components packaging correctly -- without the need to recompile everything as this module (rollup-uglify3) is not currently published.

kryalama (approved):

Issue #1545, opened 2021-04-23
Cannot modify the request headers and cookies when using a custom endpoint


llatinov:
I am using the AppInsights tracking in a React application. This is my setup
1. I cannot expose the connectionString as a configuration in the React application, because this means everybody reading my React code will have my connections string.
2. I have created a separate proxying microservice and I send the tracking data to it
```
const appInsights = new ApplicationInsights({
	config: {
		instrumentationKey: 'TO_BE_HANDLED_BY_BACKEND',
		disableInstrumentationKeyValidation: true,
		endpointUrl: configApi.msApplicationInsigts.trackUrl
	}
})
```
3. I still want to secure my microservice and prevent misuse. I am limited to company internal rules, so in the requests, I am always sending:
3.1. Secure authentication cookie
3.2. Special header config value that complements the cookie
4. The above setup works fine when I use fetch() method.

With the current setup defined, my Application Insights tracking is not working, because I am not able to modify the outgoing request. I have found addTelemetryInitializer which gives me access to the request payload but I need to be able to configure the outgoing request, similar to how I am able to modify the fetch:
```
const options: RequestInit = {
	credentials: 'include',
	mode: 'cors',
	headers: {
		'X-Custom-Header': 'value'
	}
}

return fetch(url, options)
```

MSNev:
Hi, there are a couple of things here
- There is now security risk in exposing connection string and instrumentationKey as this information does not provide any potential attacker access to your collected data, it only provides the ability to "upload" telemetry. So the only risk here is that someone could send bogus/excessive telemetry to your endpoint.
- Currently the sender only supports XDR (for IE9 or lower) and XHR (XMLHttpRequest), we have an issue open to add an option to use fetch #1268, but that is not currently scheduled
- Looking at the code I can't identify and "easy" and supportable way within the SDK today to allow adding additional headers to the sender 😞 , I'll add this as part of the work required for #1268 as that will require a bunch of work anyway.

So having said above, let me provide an approach that you might be able to implement as we do understand the concerns you are talking about.
- One immediate option that comes to mind would be to effectively "patch" the XHR request (as the SDK does), by overriding the XHR implementation or just "hooking" the XHR send function using the same mechanism that the SDK does.
Simplistically, we provide an ```InstrumentProto()``` helper that you could use to provide a callback that would allow you to add the additional headers.
```
InstrumentProto(XMLHttpRequest, "send", {
[...]

Issue #1268, opened 2020-05-13
Investigate and add a sender that uses fetch when XMLHttpRequest is not available


MSNev:
The sender codes does not include a fetch sender, so if you have not configured the beacon sender and XmlHttpRequest is not available (web workers) then the requests won't get sent.

The issue with the beacon sender is that it's a fire and forget situation, so if the server fails to accept the request the telemetry events are lost.

We should also add the ability for the user to provide extra headers

MSNev:
v2.6.3 is now fully deployed

PR #1556, opened 2021-04-27
[Enhancement] Add a sender that uses fetch when XMLHttpRequest is not available

xiao-lix:
1. Add a fetch sender - Address #1268 
2. Add the ability for the user to provide extra headers - Address #1545 

TODO: add instruction in README (Configuration section & Basic usage?)
e.g.
```
const appInsights = new ApplicationInsights({ config: {
      instrumentationKey: <YOUR_IKEY>,
      isBeaconApiDisabled: true,
      customHeaders: [
        {
          header: 'test1',
          value: 'test1'
        }
      ]
    } });

const sender = this.appInsights.getSender();
sender.addHeader('test2', 'value2');
this.appInsights.loadAppInsights();
```


channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -229,11 +237,17 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
...
+                        this.addHeader(customHeader.header, customHeader.value);
+                    });
+                }
+
                 if (!_self._senderConfig.isBeaconApiDisabled() && isBeaconApiSupported()) {
                     _self._sender = _beaconSender;
                 } else {
-                    if (typeof XMLHttpRequest !== undefined) {
-                        const xhr:any = getGlobalInst("XMLHttpRequest");
+                    if (typeof XMLHttpRequest !== "undefined") {
```
xiao-lix, line 249: Found a bug here - need to check type `"undefined"` instead of `undefined`. Otherwise this if check is always true.

channels/applicationinsights-channel-js/Tests/Sender.tests.ts
```diff
@@ -247,6 +247,179 @@ export class SenderTests extends TestClass {
...
+                    sender.flush();
+                } catch(e) {
+                    QUnit.assert.ok(false);
+                }
+
+                QUnit.assert.equal(false, sendBeaconCalled, "Beacon API is disabled, Beacon API is not called");
+                QUnit.assert.equal(0, this._getXhrRequests().length, "xhr sender is not called");
+                QUnit.assert.ok(fetchstub.called, "fetch sender is called");
+                // store it back
+                (window as any).XMLHttpRequest = fakeXMLHttpRequest;
```
xiao-lix, line 294: Need to store the xhr request back to fake request - otherwise it impacts the following tests.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -201,6 +205,10 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
             };
 
             _self.teardown = _notImplemented;
+
+            _self.addHeader = (name: string, value: string) => {
```
xiao-lix, line 209: @MSNev Does it make sense to add endpointUrl check on this method? With current implementation, even the endpoinUrl is Breeze, user can call `sender.addHeader` to add custom headers and fails track.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -229,11 +237,17 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
                         _InternalMessageId.InvalidInstrumentationKey, "Invalid Instrumentation key "+config.instrumentationKey);
                 }
 
+                if (!isInternalApplicationInsightsEndpoint(_self._senderConfig.endpointUrl()) && _self._senderConfig.customHeaders() && _self._senderConfig.customHeaders().length > 0) {
```
xiao-lix, line 240: `customHeaders` are only added to request when endpointUrl is not Breeze.

AISKU/src/Initialization.ts
```diff
@@ -506,6 +508,11 @@ export class Initialization implements IApplicationInsights {
         }
     }
 
+    public getSender(): Sender {
+        let _self = this;
```
MSNev, line 512: It doesn't make sense (from a minification perspective) to have a local ```_self``` and assignment here... just use 
```return this._sender```

File: AISKU/src/Initialization.ts [hunk 6]
```diff
@@ -110,6 +111,7 @@ export class Initialization implements IApplicationInsights {
 
     private dependencies: DependenciesPlugin;
     private properties: PropertiesPlugin;
+    private sender: Sender;
```
MSNev, line 114: Can we rename this to ```_sender```, I just like to call out private properties with the prefix '_' so that when reading later code it's obvious.

renaming the dependencies and properties, would be nice --- but you may have to update tests if you do that.
Eventually, this will probably be converted to use dynamicProto() and the privates become internal properties that will not be exposed.

channels/applicationinsights-channel-js/src/Sender.ts
MSNev: No, I don't think so... If the request fails because they add (effectively) invalid headers for breeze then this would be considered a programming error (I think). Most people won't need to or use this..

channels/applicationinsights-channel-js/src/Sender.ts
MSNev: getGlobalInst("XMLHttpRequest") should return either null or undefined (don't remember) if it's not defined -- so you shouldn't need the typeof check...

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -545,8 +505,71 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
...
+                        }
+                    } else {
+                        _consecutiveErrors = 0;
+                        _self._onSuccess(payload, countOfItemsInPayload);
+                    }
+                }
+            }
+
+            function _checkAndUpdateEndPointUrl(responseURL: string) {
+                const responseUrl = responseURL;
```
MSNev, line 572: Why the local assignment? As your passing the string value now -- just rename the param.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -622,6 +649,49 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
...
+                    requestHeaders.append(headerName, _headers[headerName]);
+                });
+                const init = {
+                    method: "POST",
+                    headers: requestHeaders,
+                    body: plainTextBatch
+                }
+                const request = new Request(endPointUrl, init);
+
+                fetch(request).then(response => {
```
kryalama, line 677: As per stamp specific redirects, @MSNev, @xiao-lix  I am thinking if its ideal to automatically follow the redirect with the following fetch option? redirect: 'follow' .

API-reference.md
```diff
@@ -196,6 +196,22 @@ Adds a telemetry initializer to the collection. Telemetry initializers will be c
 before the telemetry item is pushed for sending.
 If one of the telemetry initializers returns false or throws an error, then the telemetry item will not be sent.
 
+### getSender
```
xiao-lix, line 199: @MSNev do you think this is the proper place to add the new API instruction? Thanks.

README.md
```diff
@@ -310,6 +310,7 @@ Most configuration fields are named such that they can be defaulted to falsey. A
...
 | idLength | numeric | 22 | [Optional] Identifies the default length used to generate new random session and user id's. Defaults to 22, previous default value was 5 (v2.5.8 or less), if you need to keep the previous maximum length you should set this value to 5.
+| customHeaders | `[{header: string, value: string}]` | undefined | [Optional] The ability for the user to provide extra headers when using a custom endpoint.
```
xiao-lix, line 313: Add new configuration here.

API-reference.md
MSNev: Yep, when we check this in lets also update the Issue with a link to this file so it can be more easily found.

MSNev (approved):

MSNev:
I've updated the issues to tag them with the next milestor "2.x.x (Next Release)" as this makes it a LOT easier to keep track of the items for the release notes.

Also once this is checked in, please keep the issues Open (re-open if git-hub closes them) and add the tag "fixed - waiting for release" as again this is for tracking and it makes it obvious that the issue still exists in all released versions.

MSNev (approved):

PR #1570, opened 2021-05-27
Ignore auth headers in AjaxDependencyPlugin

Karlie-777:


shared/AppInsightsCommon/src/Interfaces/ICorrelationConfig.ts
```diff
@@ -39,5 +39,10 @@ export interface ICorrelationConfig {
      */
     ajaxPerfLookupDelay?: number;
 
-    correlationHeaderDomains?: string[]
+    correlationHeaderDomains?: string[];
+
+    /**
+     * Response and request headers to be excluded from ajax tracking data.
+     */
+    ignoreHeaders?: string[];
```
MSNev, line 47: Looks good, can you also add this to the different documentation places (*.md) any of them that already have a reference to the "enableRequestHeaderTracking" and the response ones.

And include the currently known defaults to the documentation in the *,md and on this item.

MSNev: The documentation should also mention that if they want to include/override the default values they will need to add this to the config as an empty array (as the getConfig() helper function will ignore null and undefined values)

MSNev: Here is a list of some headers that we should exclude
https://www.loginradius.com/blog/async/everything-you-want-to-know-about-authorization-headers/

MSNev: Credentials is a request header that looks like it's used by ASP.Net and WWW-Authenticate is a response header

Karlie-777: it seems that `Authorization: <type> Credential=......`  Credentials might be a parameter of Authorization, so I didn't include Credentials. All header class in ASP.Net https://docs.microsoft.com/en-us/dotnet/api/microsoft.net.http.headers.headernames?view=aspnetcore-5.0#fields

MSNev (approved):

Issue #1472, opened 2021-02-09
Is there a way to disableAjaxTracking for specific ajax requests


thoo1:
**Is your feature request related to a problem? Please describe.**
I want to be able to manually log to dependencies table for some requests for my app, for parts of the app that I can control. I would also like to keep the automatic ajax tracking for other requests of my app.

When I manually log, I would like to turn off the automatic logging that app insights sdk does, so that I do not get two logs for the same api call.

**Describe the solution you'd like**
A flag or something I can pass in the XHR object that disables the automatic logging. Or any other solution

**Describe alternatives you've considered**
Accomplishing this by monkey patching xhr to read some dummy header that flags to the telemetry initializer to not log the automatically collected log.

**Additional context**

MSNev:
This depends on the URL you are using.

There is an "internal" property which we use to tag XHR and fetch requests to our backend domains so that we don't end up in a loop.

And for performance, once a domain has been identified as Disabled ALL requests to that domain (even without the flag) will NEVER be instrumented

The code for the flag is here https://github.com/microsoft/ApplicationInsights-JS/blob/master/extensions/applicationinsights-dependencies-js/src/ajax.ts#L532

But as I said above this is designed as an *internal* switch, so there are no guarantee's that mechanism is not changed in the future (at which point you would be broken).

thoo1:
I see, thanks for that. Do you think this could be a feature that is supported? The `xhr[DisabledPropertyName] = true` would work for my use case.

Our situation is that we want to keep the app insight auto logging on for all requests that are outside our code's control (to make sure we don't drop any logs and get a full picture of api calls being made). At the same time, we want to have custom dependency logging for code that we do control, but we dont need a duplicate log from app insight's auto logging

MSNev:
Tagging this as an enhancement, to be able to configure which URL's are tracked -- I actually thought we had this already but I'm probably mis-remembering or something.

MSNev:
Reopening for release tracking

xiao-lix:
> Reopening for release tracking

oops... sorry I thought this was released. 😅

MSNev:
Actually, it looks like it might be as the commit was in June -- which means it probably got missed in the release notes for the July release @xiao-lix can you please check and if it's not in the release notes lets update them and then re-close this issue

xiao-lix:
@MSNev it should be included in this release - https://github.com/microsoft/ApplicationInsights-JS/releases/tag/2.6.4
[...]

PR #1582, opened 2021-06-15
Add config to exclude a specific request auto tracking

xiao-lix:
#1472


extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -561,11 +564,11 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
                 // check that this instance is not not used by ajax call performed inside client side monitoring to send data to collector
                 if (!isNullOrUndefined(xhr)) {
                     // Look on the XMLHttpRequest of the URL string value
-                    isDisabled = xhr[DisabledPropertyName] === true || theUrl[DisabledPropertyName] === true;
+                    isDisabled = xhr[DisabledPropertyName] === true || theUrl[DisabledPropertyName] === true || xhr[_excludeRequestFromAutoTrackingPropertyName] === true || theUrl[_excludeRequestFromAutoTrackingPropertyName] === true;
```
xiao-lix, line 567: @MSNev hi Nev, can you please explain what does `theUrl[DisabledPropertyName]` mean? Both `theUrl` and `DisabledPropertyName` are string, one string will never has a string property isn't it? Should we remove both the checks here? And my IDE complains this line also: 
image

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -548,6 +551,20 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
             function _isDisabledRequest(xhr?: XMLHttpRequestInstrumented, request?: Request | string, init?: RequestInit) {
                 let isDisabled = false;
                 let theUrl:string = ((!isString(request) ? ((request ||{}) as Request).url || "" : request as string) ||"").toLowerCase();
+
+                // check excludeRequestFromAutoTrackingRegex before stripping off any query string
+                arrForEach(_excludeRequestFromAutoTrackingRegex, (regex: string | RegExp) => {
+                    if (!isDisabled) {
+                        const regexp = new RegExp(regex);
```
MSNev, line 558: Looking at the documentation new RexExp(regExp), passing a Regular expression was only added in ES6 (so this won't work in older browsers, we should add an isString() check.

shared/AppInsightsCommon/src/Interfaces/IConfig.ts
```diff
@@ -118,6 +118,14 @@ export interface IConfig {
      */
     disableFetchTracking?: boolean;
 
+    /**
+     * @description Provide a way to exclude specific route from automatic tracking for XMLHttpRequest or Fetch request.
+     * @type {string[] | RegExp[]}
+     * @memberof IConfig
+     * @defaultValue undefined.
+     */
+     excludeRequestFromAutoTrackingRegex?: string[] | RegExp[];
```
MSNev, line 127: Lets be consistent with the other item and call this "Patterns" rather than "Regex"

extensions/applicationinsights-dependencies-js/src/ajax.ts
xiao-lix: I see, then should the parameter be "string[]" only?

MSNev: No, both is a nice option.

xiao-lix: Had a second thought.. I think it's good to provide both types as parameter thus provide users more flexibility right? I updated the checking 😬

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -548,6 +551,32 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
+                    if (!isDisabled) {
+                        if (isRegExp) {
+                            const revertRegexObj = JSON.parse(regex as string);
+                            const regexp = new RegExp(revertRegexObj.source, revertRegexObj.flags);
+                            isDisabled = regexp.test(theUrl);
+                        } else {
+                            const regexp = new RegExp(regex);
+                            isDisabled = regexp.test(theUrl);
+                        }
+                    }
```
xiao-lix, line 572: Got the convert RegExp to string then back method from this [answer](https://stackoverflow.com/questions/39154255/converting-regexp-to-string-then-back-to-regexp).
Doc points the RegExp does have [toString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/toString) method but it converts a regex to a string directly and if convert it back it becomes `"\\xxx\\"`. 
@MSNev do you think it's better to use `JSON.stringify` or to trim the extra `"\"` with `toString` method or there's a better way? Thanks~

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -548,6 +551,32 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
             function _isDisabledRequest(xhr?: XMLHttpRequestInstrumented, request?: Request | string, init?: RequestInit) {
                 let isDisabled = false;
                 let theUrl:string = ((!isString(request) ? ((request ||{}) as Request).url || "" : request as string) ||"").toLowerCase();
+
+                // check excludeRequestFromAutoTrackingPatterns before stripping off any query string
+                arrForEach(_excludeRequestFromAutoTrackingPatterns, (regex: string | RegExp) => {
+                    let isRegExp = false;
+                    if (!isString(regex)) {
```
MSNev, line 558: If it's a regex we should just reuse it rather than pull it apart and recreate it
```
let theRegEx = regex;
if (isString(regex)) {
    theRegEx = new RegEx(regex);
}

isDisabled = theRegEx.test(theUrl);

Comment (xiao-lix) [reply extensions/applicationinsights-dependencies-js/src/ajax.ts]: Ohhhhh...!!! I just made things more complicated... 😅

File: shared/AppInsightsCommon/src/Interfaces/ICorrelationConfig.ts [hunk 6]
```diff
@@ -45,4 +45,11 @@ export interface ICorrelationConfig {
...
      */
     ignoreHeaders?: string[];
+
+    /**
+     * Provide a way to exclude specific route from automatic tracking for XMLHttpRequest or Fetch request.
+     * For an ajax / fetch request that the request url matches with the regex patterns, auto tracking is turned off.
+     * Default is undefined.
+     */
+    excludeRequestFromAutoTrackingPatterns?: string[] | RegExp[];
 }
```
MSNev, line 55: Add same comment to IConfig and the ReadMe table 😉

MSNev (approved):

Issue #1580, opened 2021-06-15
[BUG] URL without host fails in CanIncludeCorrelationHeader


panda7789:
**Description**

When fetching url without host like this:
`"../../api/v01/as/descriptions/Attributes?className=Polozky_smlouvy"`
in method canIncludeCorrelationHeader it fails on parsing host, because url doesnt have host (missing www.host.com on begin).

![image](https://user-images.githubusercontent.com/6593380/121994909-6b190880-cda6-11eb-970f-da06eaa631f8.png)
![image](https://user-images.githubusercontent.com/6593380/121994946-79ffbb00-cda6-11eb-8077-9e0a00592673.png)

**Steps to Reproduce**

 - OS/Browser: Edge
 - SDK Version [e.g. 22]: `@microsoft/applicationinsights-web:` `2.6.2`
 - How you initialized the SDK: 
 ```javascript
 new ApplicationInsights({
        config: {
            instrumentationKey: iKey,
            disableCorrelationHeaders: false,
            disableFetchTracking: false,
            disableAjaxTracking: false,
            enableRequestHeaderTracking: true,
            maxAjaxCallsPerView: 200,
            correlationHeaderDomains: ["app2.loc", "app1.loc", "portal.loc/api/"]
        }
```

**Expected behavior**

It shouldn't fail, local api calls should be ok.

MSNev:
Agree it should not fail / throw

MSNev:
Now fully deployed to all CDN endpoints

PR #1583, opened 2021-06-17
URL without host should not fails canIncludeCorrelationHeader right away

xiao-lix:
#1580


shared/AppInsightsCommon/src/UrlHelperFuncs.ts
```diff
@@ -65,7 +65,7 @@ export function urlGetCompleteUrl(method: string, absoluteUrl: string) {
 
 // Fallback method to grab host from url if document.createElement method is not available
 export function urlParseHost(url: string, inclPort?: boolean) {
-    let fullHost = urlParseFullHost(url, inclPort);
+    let fullHost = urlParseFullHost(url, inclPort) || "";
```
xiao-lix, line 68: Add `| ""` means when url has no host, the `urlParseFullHost` method returns `null`. In that case, `null` passed back to [here](https://github.com/microsoft/ApplicationInsights-JS/blob/2612a3cc285526fb66464f860d049ba3d145180b/shared/AppInsightsCommon/src/Util.ts#L351) but `null.toLowerCase()` is wrong. So always return an empty string when host is not detected.

shared/AppInsightsCommon/src/Util.ts
```diff
@@ -238,9 +238,9 @@ export interface IUtil {
 
 export const Util: IUtil = {
     NotSpecified: strNotSpecified,
-    createDomEvent: createDomEvent,
+    createDomEvent,
```
xiao-lix, line 241: Fix some lint errors

shared/AppInsightsCommon/src/Util.ts
```diff
@@ -355,7 +355,7 @@ export const CorrelationIdHelper: ICorrelationIdHelper = {
             requestHost = (urlParseFullHost(requestUrl, true) || "").toLowerCase();
         }
 
-        if ((!config || !config.enableCorsCorrelation) && requestHost !== currentHost) {
+        if ((!config || !config.enableCorsCorrelation) && (requestHost && requestHost !== currentHost)) {
```
xiao-lix, line 358: ONLY check if the requestHost equals currentHost when requestHost exists. For the case that requestHost not exist (empty string when host not exist) skip the check. Otherwise it will returns false right away for the cases that url without host.

kryalama (approved):

MSNev (approved):

Issue #1586, opened 2021-06-17
[BUG] namePrefix is not getting assigned to ai_user Cookie


ThaminduDilshan:
**Description/Screenshot**
As for the documentation we can set a name postfix for localStorage and cookie name (by passing the config parameter `namePrefix`). But this postfix only get assigned to the `ai_session` cookie and not for the `ai_user` cookie.

I'm using the javascript sdk with react plugin for a react frontend application.

 - OS/Browser: Google chrome 91.0.4472.101
 - SDK Version [e.g. 22]: 2.6.2
 - How you initialized the SDK: NPM Setup

**Expected behavior**
After removing cookies and letting sdk to recreate them, following are the cookie names,
- `ai_session<postfix>`
- `ai_user`

MSNev:
Now fully deployed to all CDN endpoints

PR #1587, opened 2021-06-17
ai_user cookie should use userCookiePostfix for user cookie storage

xiao-lix:
#1586


extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -61,10 +61,14 @@ export class User implements IUserContext {
     constructor(config: ITelemetryConfig, core: IAppInsightsCore) {
         let _logger = safeGetLogger(core);
         let _cookieManager: ICookieMgr = safeGetCookieMgr(core);
+        let _storageNamePrefix: () => string;
 
         dynamicProto(User, this, (_self) => {
+            _self.config = config;
```
kryalama, line 67: Have following questions:
1. In the docs , its mentioned that namePrefix is expected to be a string, but here it looks like callback function.
2. Also does the namePrefix have to be appended to all cookies? or is it only ai_user and ai_session?
3. nit: its confusing to use namePrefix in configuration, but its actually used as postfix

MSNev: 1. We wrap it internall with a function, so the value passed in from the config is a string
2. I have the same question as we have existing users already deployed with this and this cookie is LOOOOONG lasting, we should not automatically add it -- there should be a config otherwise we are going to cause an issue with anyone that is using this. Also, by not including the prefix on this cookie it can (and is) being shared across multiple sites.
3. Agree, but its not easy to change this we could introduce an new config value which defaults to re-using the namePrefix -- but this assumes we can stop using namePrefix which we cant as it's also used for storage.
- Options, add new configs ```cookiePostfix``` (defaults to the same as namePrefix if not provided, so the ai_session keeps working) and either ```prefixUserCookie:boolean``` which defaults to false or split them ```userCookiePostfix``` which defaults to an empty string, for this later one would recommend using ```sessionCookiePostfix``` instead of a generic ```cookiePostfix```

extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -61,10 +61,14 @@ export class User implements IUserContext {
...
+            _storageNamePrefix = () => _self.config.namePrefix && _self.config.namePrefix() ? User.userCookieName + _self.config.namePrefix() : User.userCookieName;
```
MSNev, line 68: As mentioned above, we should not just start applying this for everyone.

MSNev (dismissed):
We need to add a config for this to avoid breaking anyone.

extensions/applicationinsights-properties-js/src/Context/User.ts
xiao-lix: Thanks for the suggestions Nev. I updated the logic as below: 

- Add `sessionCookiePostfix` and `userCookiePostfix`.

For session cookie:
- if both `sessionCookiePostfix` and `namePrefix` is configured, then ignore `namePrefix` and use `sessionCookiePostfix`.
- if `sessionCookiePostfix` not configured but `namePrefix` is configured, then use `namePrefix`. 
- if none is configured, no suffix is added.

For user cookie: 
- `namePrefix` has no effect on user cookie.
- if `userCookiePostfix` is configured, append `userCookiePostfix` to `ai_user` cookie name.
- if none is configured, no suffix is added.

shared/AppInsightsCommon/src/Interfaces/IConfig.ts
```diff
@@ -311,6 +311,22 @@ export interface IConfig {
      */
     namePrefix?: string;
 
+    /**
+     * @description An optional value that will be used as name postfix for session cookie name.
+     * @type {string}
+     * @memberof IConfig
+     * @defaultValue null
+     */
+    sessionCookiePostfix?: string;
```
MSNev, line 320: Add details about the fallback logic and also update the READMD.md with the details, so that users know about how this works.

I've also started adding some indication on when any new configs where added (I think mostly the internal code)

extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -61,10 +61,15 @@ export class User implements IUserContext {
     constructor(config: ITelemetryConfig, core: IAppInsightsCore) {
         let _logger = safeGetLogger(core);
         let _cookieManager: ICookieMgr = safeGetCookieMgr(core);
+        let _storageNamePrefix: () => string;
 
         dynamicProto(User, this, (_self) => {
+            _self.config = config;
+            const userCookiePostfix = (_self.config.userCookiePostfix && _self.config.userCookiePostfix()) ? _self.config.userCookiePostfix() : "";
+            _storageNamePrefix = () => User.userCookieName + userCookiePostfix;
```
kryalama, line 69: nit: rename this to userCookieNameWithPostFix?

kryalama (approved):

extensions/applicationinsights-properties-js/src/Context/User.ts
MSNev: No -- don't increase the length of the config name as this directly affects minification.
userCookiePostfix is specific enough.

kryalama: I was referring to _storageNamePrefix :smile: nvm I am ok with previous name

PR #1588, opened 2021-06-17
addSizeFileCheck

Karlie-777:


AISKU/Tests/AISKUSize.Tests.ts
```diff
@@ -0,0 +1,46 @@
+import { TestClass } from "./TestFramework/TestClass";
+import { Assert } from "./TestFramework/Assert";
+import * as pako from "pako";
+
+export class AISKUSizeCheck extends TestClass {
+    private readonly MAX_DEFLATE_SIZE = 40;
+    private readonly filePath = "../../dist/applicationinsights-web.min.js";
```
MSNev, line 7: This is actually the wrong file it should be
../../browser/ai.2.min.js

extensions/applicationinsights-analytics-js/Tests/AnalyticsExtensionSize.tests.ts
```diff
@@ -0,0 +1,45 @@
+import { Assert, AITestClass } from "@microsoft/ai-test-framework";
+import * as pako from "pako";
+
+export class AnalyticsExtensionSizeCheck extends AITestClass {
+    private readonly MAX_DEFLATE_SIZE = 20;
+    private readonly filePath = "../../dist/applicationinsights-analytics-js.min.js";
```
MSNev, line 6: This one is a little different in that we should check this file AND the ../../browser/xxx version

AISKU/Tests/AISKUSize.Tests.ts
MSNev: Actually, lets check both of the files

extensions/applicationinsights-properties-js/Tests/Selenium/propertiesSize.tests.ts
```diff
@@ -0,0 +1,45 @@
+import { Assert, AITestClass } from "@microsoft/ai-test-framework";
+import * as pako from "pako";
+
+export class PropertiesExtensionSizeCheck extends AITestClass {
+    private readonly MAX_DEFLATE_SIZE = 15;
+    private readonly filePath = "../../dist/applicationinsights-properties-js.min.js";
```
MSNev, line 6: Probably the same for all of these check both the /dist/ which is the raw compiled version that someone *might* include vs the /browser/ version which is really the packaged version where the size is the critical issue

MSNev (approved):

PR #1596, opened 2021-06-29
Apply the eslint fixes (from Component Governance Checks)

MSNev:
- Update ts-plugin dependency version


AISKU/src/Initialization.ts
```diff
@@ -1,23 +1,23 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { 
-    IConfiguration, AppInsightsCore, IAppInsightsCore, LoggingSeverity, _InternalMessageId, ITelemetryItem, ICustomProperties, 
-    IChannelControls, hasWindow, hasDocument, isReactNative, doPerf, IDiagnosticLogger, INotificationManager, objForEachKey, proxyAssign,
+import {
+    IConfiguration, AppInsightsCore, IAppInsightsCore, LoggingSeverity, _InternalMessageId, ITelemetryItem, ICustomProperties,
+    IChannelControls, hasWindow, hasDocument, isReactNative, doPerf, IDiagnosticLogger, INotificationManager, objForEachKey, proxyAssign,
```
MSNev, line 6: These are just mixed whitespace issues (had spaces and tabs)

extensions/applicationinsights-debugplugin-js/src/components/LogEntry.ts
```diff
@@ -81,7 +81,7 @@ export class LogEntry {
                     ms = "0" + ms;
                 }
 
-                ms = ms.replace(/(.)(\d{3})+$/g, '$1.$2s')
+                ms = ms.replace(/(.)(\d{3}){1}$/g, '$1.$2s')
```
MSNev, line 84: The regex changes are security related and a little more complicated and the changes are 2 fold
- Unnecessary escaping (additional \ when not needed), this *could* allow someone to "fiddle" with the resulting regex (in memory) to change it's behavior
- When using * and +, depending on the input string unconstrained matching can cause the regex to use excessive memory, cpu or just fail.

hectorhdzg (approved):

PR #1617, opened 2021-07-22
Add Stale Issue / PR GitHub Action

MSNev:
- Starting with high values because we have not had this before
- Initial values should be 300 days => stale and 30 days => close

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,26 @@
+name: 'Stale issue handler'
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 7 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4.0.0
+        id: stale
+        with:
+          stale-issue-message: 'This Issue will be closed in 30 days. Please remove the "Stale" label or comment to avoid closure with no action.'
+          stale-pr-message: 'This PR will be closed in 30 days. Please remove the "Stale" label or comment to avoid closure with no action.'
+          days-before-stale: 300
+          days-before-close: 30
+          stale-issue-label: 'stale'
+          stale-pr-label: 'stale'
+          close-issue-label: 'closed'
+          close-pr-label: 'closed'
+          exempt-issue-labels: 'bug,enhancement,documentation,waiting,keep'
+          exempt-all-milestones: true
+          exempt-all-assignees: true
+      - name: Print outputs
+        run: echo ${{ join(steps.stale.outputs.*, ',') }}


.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4.0.0
+        id: stale
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 300 days with no activity. Remove stale label or comment or this will be closed in 30 days'
+          stale-pr-message: 'This PR is stale because it has been open for 300 days with no activity. Remove stale label or comment or this will be closed in 30 days'
```
ramthi, line 15: NIT: Suggestion on the last sentence: 'This PR will be closed in 30 days.  Please remove the "Stale" label or comment to avoid closure with no action.'

.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+    - cron: '0 7 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4.0.0
+        id: stale
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 300 days with no activity. Remove stale label or comment or this will be closed in 30 days'
```
ramthi, line 14: NIT: Suggestion 'This Issue will be closed in 30 days. Please remove the "Stale" label or comment to avoid closure with no action.'

ramthi (changes requested):
Added a couple of suggestion for text change.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+      - uses: actions/stale@v4.0.0
+        id: stale
+        with:
+          stale-issue-message: 'This Issue will be closed in 30 days. Please remove the "Stale" label or comment to avoid closure with no action.'
+          stale-pr-message: 'This PR will be closed in 30 days. Please remove the "Stale" label or comment to avoid closure with no action.'
+          days-before-stale: 300
+          days-before-close: 30
+          stale-issue-label: 'stale'
+          stale-pr-label: 'stale'
+          close-issue-label: 'closed'
```
lzchen, line 20: I don't think closed issues/prs need a label.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,26 @@
...
+          close-pr-label: 'closed'
```
lzchen, line 21: I would add a message instead on automatic close. See [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-python/blob/main/.github/workflows/close-stale-issues.yml#L14)

.github/workflows/stale.yml
MSNev: I added this so that we can track (at least initially) the set of issues / PR's that get closed automatically vs all of the others that we have and will close manually (which won't have these labels). According to the documentation the action will also remove this if it gets re-opened.

So eventually (once we are in a much better sate) I agree this will not be needed but as we currently have 117 open (most of which I expect will eventually get closed by this) and 622 Closed I want an easier way to identify these auto-closed items.

.github/workflows/stale.yml
MSNev: We already have a message for when it goes stale, which when closed would be the last message. So I don't think that it's necessary...

I think that by using the label it's a bit more concise and also more searchable without getting false positive matches.

lzchen (approved):

kryalama (approved):

ramthi (approved):

PR #1619, opened 2021-07-30
Update stale action to hit more items

MSNev:


.github/workflows/stale.yml
```diff
@@ -13,6 +13,7 @@ jobs:
         with:
           stale-issue-message: 'This Issue will be closed in 30 days. Please remove the "Stale" label or comment to avoid closure with no action.'
           stale-pr-message: 'This PR will be closed in 30 days. Please remove the "Stale" label or comment to avoid closure with no action.'
+          operations-per-run: 200
```
MSNev, line 16: This defaults to 30 and is not hitting all of the old items

kryalama (approved):

Issue #1561, opened 2021-05-05
How to enrich dependencies logs with context at the beginning of api call?


thoo1:
**Is your feature request related to a problem? Please describe.**
API calls take some time. We have a use case where we'd like to capture some context at the beginning of the api call, and enrich the dependencies log with this context. We're finding that our current implementation is not using the public interface of app insights sdk and is pretty fragile.

**Describe the solution you'd like**
A way to hook into the start of the API call which app insights SDK has patch over the fetch and xhr functions. Then add some metadata to the telemetryItem object which would log as part of the "customDimensions"

**Describe alternatives you've considered**
Currently we do this by 
- patching over fetch and xhr functions in a similar way as the app insights sdk
- in these patched functions we get the added "Request-Id" header that app insights adds, and use that as a key to add the context we want to enrich the log with, into a map.
- so the map looks like {string: Context} where the string is the "Request-Id"
- then in the telemetry initializer, we do a lookup in the map using `telemetryItem.baseData.id` to find the context and enrich the log

This is pretty fragile and isn't part of the interface of application insights sdk.

**Additional context**
Please feel free to contact me on teams (my alias is thoo) for more info

MSNev:
Agree, I just went digging into the code to see if there would be an easier way than what you have already done and I can't find any other safer way that wouldn't get broken.

So yes, adding a callback to the config would probably be the easiest way to do this, and there are a couple of places in the code (_openHandler() and _createFetchRecord()) where it would make sense to have the callback called so the returned map could be added to the internal ajaxRecord, and them populated into the event -- which would still facilitate using a telemetryInitializer() to munge the data if required.

MSNev:
v2.7.0 is not fully deployed to NPM and all CDN endpoints

PR #1624, opened 2021-08-03
Provide a way to enrich dependencies logs with context at the beginning of api call

xiao-lix:
- Add a new configuration to allow users to enrich custom properties on requests.
- The configuration is a call back function that add extra properties to the dependency data, and it works on both xhr and fetch requests.

Related Issue: How to enrich dependencies logs with context at the beginning of api call? #1561 

**Update:** 
The basic usage would be adding configurations when AI gets initialized: 
```
addRequestContext: (requestContext) => {
    if (requestContext.xhr) {
        return {
            test: "ajax context",
            xhrStatus: requestContext.status,
            xhrTimeout: requestContext.xhr.timeout,
            xhrResonseURL: requestContext.xhr.responseURL
        }
    }
    if (requestContext.request && requestContext.response) {
        return {
            test: "fetch context",
            fetchStatus: requestContext.status,
            fetchRequestURL: requestContext.request,
            fetchReponseType: requestContext.response.type,
            fetchResponseStatusText: requestContext.response.statusText 
        }
    }
}
```
And in the payload, you'll find the specified dimensions are captured in properties field: 
**eg. Fetch**
image

**eg. Ajax**
image


extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -176,7 +176,8 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
             ignoreHeaders:[
                 "Authorization", 
                 "X-API-Key",
-                "WWW-Authenticate"]
+                "WWW-Authenticate"],
+            addContextOnRequests: undefined
```
MSNev, line 180: As this is specific for ajax requests we should probably target the name to that
perhaps
- addAjaxContext
- addAjaxProperties
- etc

We don't want to get too big on the name as this will be public and unminifiable, but we also want it to be specific / recognizable... Trying to avoid the what does that mean questions when some sees it.

shared/AppInsightsCommon/src/Interfaces/ICorrelationConfig.ts
```diff
@@ -55,8 +55,14 @@ export interface ICorrelationConfig {
     excludeRequestFromAutoTrackingPatterns?: string[] | RegExp[];
 
     /**
-     * Provide a way to enrich dependencies logs with context at the beginning of api call.
+     * Provide a way to enrich dependencies logs with context at the beginning of ajax call.
      * Default is undefined.
      */
-     addContextOnRequests?: () => ICustomProperties;
+     addAjaxContext?: (xhr?: XMLHttpRequest) => ICustomProperties;
```
MSNev, line 61: Can we have this as 1 function that takes a new "request context" object as if we want to pass something more (like detected performance values) this would require a breaking change to the signature of this function.

The new "request context" object (interface) can contain all of the fields as optional (xhr, fetch request, fetch response, etc)

MSNev (approved):

streamingpro:
Thanks for these changes. Is there a way to access request body as part of xhr? We have single endpoint for all the dependency calls (API), but the request body has the specific method name. Currently, AI log shows single operation name and we are not able to figure out the actual method until we look at end-to-end transaction. If there is a way, we want to get the method name from request body and append it to the operation name before sending the telemetry. Thank you

MSNev:
Sorry for the delay as I've been away.

I just went through the code and I can't see a current way to directly access the send body of the request.
We could probably keep a reference of the request around so that once we get the response it could be passed onto the addRequestContext function... But that will have memory impacts as the formatted body (JSON) will hang around longer.

Another possible way (as you own the backend) would be to return a response header with the extra information that you want to include and then in the addRequestContext() you ```should``` then be able to access the response headers and figure out what you want to log??

Issue #1526, opened 2021-04-08
[TypeScript Compile Error] Property 'sessionManager' does not exist on type 'ITelemetryContext'


thoo1:
**Description/Screenshot**
- when trying to access the automatic session id, by referencing `appInsights.context.sessionManager.automaticSession.id`, typescript throws an error "Property 'sessionManager' does not exist on type 'ITelemetryContext'"

**Steps to Reproduce**
-  try to reference `appInsights.context.sessionManager`
 - OS/Browser: N/A
 - SDK Version [e.g. 22]: "@microsoft/applicationinsights-web": "^2.6.1",
 - How you initialized the SDK: 
```
    this.appInsights = new ApplicationInsights({
      config: {
        instrumentationKey: Environment.config.appInsightsInstrumentationKey,
        endpointUrl: Environment.config.appInsightsEndpointUrl,
        extensions: [this.reactPlugin],
        extensionConfig: {
          [this.reactPlugin.identifier]: { history }
        },
        disableFetchTracking: false,
        disableAjaxTracking: false,
        enableCorsCorrelation: true,
        disableCorrelationHeaders: false,
        enableAutoRouteTracking: true,
        maxAjaxCallsPerView: -1,
        enableResponseHeaderTracking: true,
        enableAjaxPerfTracking: true
      }
    });
```

**Expected behavior**
 - no typescript error
 
**Additional context**
- this way of getting the session_Id was brought up in an issue here: #943

MSNev:
This is correct `ITelemetryContext` doesn't define or require a sessionManager as not all instances have it, this is because not all usage patterns actually include the PropertiesPlugin (which is where the sessionManager implementation is coming from), even though in your case the `context` does actually have the property.

For example, if you load this same Sku from a legacy snippet then the `appInsights.context` does NOT contain the `sessionManager` as the context instance is not that of the PropertiesPlugin. This is the primary reason it's not on the `ITelemetryContext` interface.

And as part of the tree-shaking improvements in the latest release the interfaces where better defined to identify these cases... Hence your current compile error.

[...]

Issue #1256, opened 2020-04-16
[DOCS] How to get ai session id for debug purposes


jean-moldovan:
Hello!

I'd love to spit out ai session id to the console for debug / investigation purposes. Currently it is not clear where to get it from.

I'm on a SPA app and this is how I initialize app insights:
```
  const appInsights = new ApplicationInsights({
    config: {
      instrumentationKey: 'xxxx'
    }
  })

  appInsights.loadAppInsights()
```

Then I'd track the page on route change:
```
  router.beforeEach((to, from, next) => {
    appInsights.startTrackPage(to.fullPath)
    next()
  })

  router.afterEach((to, from) => {
    appInsights.stopTrackPage(from.fullPath)
  })
```

1) How can I access ai session id from my js code?
2) If I track the page manually on route change do I still need to initialize the session by calling `trackPageView ` on app insights load?
```
appInsights.loadAppInsights();
appInsights.trackPageView();
```

Thanks!

MSNev:
Not sure I'm going to completly answer your question here but..

For  if you want to link your initial server request to the page request you would need to do something like the following after you have initialized the sdk and before the first track call (i.e. the page view -- otherwise that won't be linked correctly)

if (aisdk.queue && aisdk.queue.length === 0) {
    aisdk.properties.context.telemetryTrace.traceID = "<server operation id>"
    aisdk.properties.context.telemetryTrace.parentID = "<server request id>"
    aisdk.properties.context.telemetryTrace.name = "<server operation name>"
}

If you don't want to link you server and client requests together then you can also reference the above values (aisdk.properties.context.telemetryTrace.xxxx) to get the values. The trace id I don't believe is set until after the first request (event) is sent.

jean-moldovan:
Found an answer in https://github.com/microsoft/ApplicationInsights-JS/issues/943.
Closing.

Issue #943, opened 2019-06-21
How to read or get appInsights's  SessionID


satyarapelly:
Hi, 

We are trying to session Id i.e. ai.session.id value after we load the appInsights using appInsights.loadAppInsights().

Please help us in getting the ai.session.id value using appInsights objects in Javascript.

Thanks
satya R

markwolff:
Hi you can use 
`appInsights.context.sessionManager.automaticSession`

PR #1627, opened 2021-08-04
Add sesId to allow access to sessionManager session info

xiao-lix:
- To fix this issue:  #1526 
- And this issue:  #943 could be used as: `appInsights.context.sesId()` instead.
- This issue is also related:  #1256


extensions/applicationinsights-properties-js/src/TelemetryContext.ts
```diff
@@ -62,8 +64,10 @@ export class TelemetryContext implements IPropTelemetryContext {
                 // If customer set session info, apply their context; otherwise apply context automatically generated
                 if (session && isString(session.id)) {
                     setValue(getSetValue(evt.ext, Extensions.AppExt), "sesId", session.id);
+                    this.sesId = () => session.id ;
```
MSNev, line 67: While technically this will work as applySessionContext() is currently always called after the automaticSession is updated this would be better as a first class function and have the apply use it so that everyone should be refrencing in the same way and this will (hopefully) avoid issues in the future about usages.

There will still be issues around when the automatic session is updated because the cookie expires etc, but we could address this by the below

```typescript
_self.sesId = () => {
    let session = _self.session;
    let sesId = null;
    
    // If customer set session info, apply their context; otherwise apply context automatically generated
    if (session && isString(session.id)) {
        sesId = session.id;
    } else {
        // Gets the automatic session if it exists or an empty object
        let autoSession = (_self.sessionManager || {}).automaticSession;
    
        sesId = autoSession && isString(autoSession.id) ? autoSession.id : null;
    }

    return sesId;
}

_self.applySessionContext = (evt: ITelemetryItem, itemCtx?: IProcessTelemetryContext) => {
    setValue(getSetValue(evt.ext, Extensions.AppExt), "sesId", _self.sesId(), isString);
}

File: extensions/applicationinsights-properties-js/Tests/Unit/src/properties.tests.ts [hunk 2]
```diff
@@ -36,6 +36,7 @@ export class PropertiesTests extends AITestClass {
         this.addUserTests();
         this.addDeviceTests();
         this.addTelemetryTraceTests();
+        this.addSessionTests();
```
MSNev, line 39: Note: If this PR is related to an issue, please use the issue name and reference the issue number as it makes the release notes a LOT easier to create and for people to follow.

If not, please raise an Issue and reference it as well.

xiao-lix: Got it, I updated the description, is that correct?

MSNev: And the title

extensions/applicationinsights-properties-js/src/TelemetryContext.ts
xiao-lix: This is nice, I didn't think of this. So when the automatic session is updated, will the applySessionContext method gets called again? Is that when sesId got updated?

MSNev: Currently yes, it happens here (https://github.com/microsoft/ApplicationInsights-JS/blob/master/extensions/applicationinsights-properties-js/src/PropertiesPlugin.ts#L90-L94) where the call to _processTlemetryInternal() calls applySessionContext().

So currently it only gets updated with each event, so someone could call sesId() before an event right on the cusp of the rollover. So calling after the event will always have correct value

xiao-lix: I see, that makes sense. Thanks~ Updated to take the sesId out as a function.

shared/AppInsightsCommon/src/Interfaces/ITelemetryContext.ts
```diff
@@ -61,4 +61,9 @@ export interface ITelemetryContext {
      * application id obtained from breeze responses. Is used if appId is not specified by root config
      */
     appId: () => string;
+
+    /**
+     * session id obtained from session manager.
+     */
+    sesId: () => string;
```
MSNev, line 68: Just noticed this and I'm in 2 minds on whether to leave like this or make it optional.
Why? 
Technically, this is an exposed interface as the ```context``` of an app insights instance and while we don't __expect__ that anyone would provide their own version (to overwrite ours) having this as required make this a breaking change.

Which is where the 2 minds comes from
1) This is an internal interface and people should not be updating this property -- therefore breaking (required method) is fine
2) If there is someone who wants or needs to do this they will be (technically) broken (until they add a function to do this)..

Did we marked the ```context``` accessor to this as readonly?

I'm also thinking about pushing the TypeScript 4 update as the next release which would bump us to at least 2.7 (maybe 3.0) so this change would (I think) be ok for a jump to 2.7 or 3.0 -- lets talk about it in standup tomorrow.

xiao-lix: What is "context accessor" , I understand this `sesId` is the accessor to session id, and it's not readonly because it can be reassigned right? (eg. each time applySessionContext is called). And sure, let's talk about the update tomorrow.

MSNev: ah, today you can get to the automatic session id via
```appInsights.context.sessionManager.automaticSession```

So what I mean by the ```context accessor``` is the ```context``` property on the ```appInsights``` instance 😄 , it will probably be defined on several interfaces (standard, deprecated and maybe the snippet). There may also be a AppInsightsCore version as well... I suspect it's not currently ```readonly``` (but it probably should be)

MSNev (approved):

PR #1628, opened 2021-08-09
Add setUserCookie method on cookie manager

xiao-lix:


MSNev (changes requested):
This issue is not about adding another method to the cookie manager (which we should not do), as it already has all of the necessary set methods for cookies required.

The suggestion I had in the issue was to add this method to the User object (same as the setAuthenticatedUser()) as the application just wants to "re-enable" cookies and as such trigger that any configured cookies can now be re-created.

So this would apply to both ```ai_user``` and ```ai_session```, so it might mean that we put this on the telemetry context rather than the user -- etc.

extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -176,4 +196,8 @@ export class User implements IUserContext {
     public clearAuthenticatedUserContext() {
         // @DynamicProtoStub -- DO NOT add any code as this will be removed during packaging
     }
+
+    public update() {
```
MSNev, line 200: While this does get removed after converted to JS, the function signature needs to match as this is used by TypeScript to generate the *.d.ts files.

xiao-lix: ah sorry that was a typo. Added it, thanks.

extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -197,7 +197,7 @@ export class User implements IUserContext {
         // @DynamicProtoStub -- DO NOT add any code as this will be removed during packaging
     }
 
-    public update() {
+    public update(userId?: string) {
```
MSNev, line 200: please add the missing ';' after the bracket on line 178 (above) to avoid the Code scanning warning -- since we are touching this file.

MSNev (approved):

MSNev (approved):

Issue #1471, opened 2021-02-09
Convert undefined to blank in customDimensions?


coreyroth:
**Is your feature request related to a problem? Please describe.**
I have a number of customDimensions in trackPageView, trackEvent, etc.  Given this SDK is JavaScript based, dealing with undefined is a common thing.  When we send a value of undefined, it shows up in our data model as undefined as well.  I'm then reading the data through the API or Power BI and undefined.  That doesn't look good on reports and we have to do comparisons to both '' and undefined.

I can work around this easily enough, but it's a bit of a mess when I am sending 50+ properties in an event.  Does it make sense for the SDK to strip undefined and send '' instead of undefined?

**Describe the solution you'd like**
I would like the SDK or the API to convert undefined to a blank ('') value so it doesn't show up in the data model.  This could be an option on the SDK perhaps in case people don't want this behavior.  

**Describe alternatives you've considered**
I can work around this of course by wrapping the values I am sending with a function to detect undefined.  I could also use a telemetryInitializer to work around it as well.

MSNev:
v2.7.0 is not fully deployed to NPM and all CDN endpoints

coreyroth:
Thanks for including this!

MSNev:
It may take up to around 30 minutes for all CDN endpoints to pick up the new bits (depending where you are located)

coreyroth:
I finally have had a chance to test this with 2.7.2 and I'm still seeing values listed as "undefined" in customDimensions.  Is there a trick to making this work?

MSNev:
Reopened so we can investigate further @Karlie-777 do you have some cycles to investigate this see the PR #1630 to see what was changed

Karlie-777:
Hi @coreyroth 
The current npm version only works with undefined value but not value equal to string "undefined".
so for example, 
```appInsights.trackEvent({ name: "Custom event" , properties:{"prop1": undefined, "prop2": "undefined"}});```

`prop1` will be replaced but not `prop2`.

coreyroth:
Hi @Karlie-777, perfect.  That's my understanding as well.  Using v2.7.2.

I initialize my instance with something like:

`      this.appInsights = new ApplicationInsights({
        config: {
          instrumentationKey: instrumentationKey,
          autoTrackPageVisitTime: true,
          disableAjaxTracking: true,
          convertUndefined: ''
        }
      });`

[...]

PR #1630, opened 2021-08-10
Convert undefined custom properties to empty string

xiao-lix:
To address this issue: Convert undefined to blank in customDimensions #1471 

Solution: Add a sender configuration `convertUndefined` that customer can set custom defined value and convert undefined field to the value they need.

Sample Usage: 
- AI set up:
```
const ai = new ApplicationInsights({
    config: {
        instrumentationKey: '<YOUR_IKEY>',
        maxBatchInterval: 100,
        disableFetchTracking: false,
        convertUndefined: 'test'
    }
});
ai.loadAppInsights();
```

- send a event track: 
`ai.trackEvent({name: "test-event", properties: {test2: undefined}});`
- without the configuration `convertUndefined` payload : 
image
- with configuration `convertUndefined` payload:
image


channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
```diff
@@ -183,6 +191,7 @@ export class DependencyEnvelopeCreator extends EnvelopeCreator {
         const customMeasurements = telemetryItem[strBaseData].measurements || {};
         const customProperties = telemetryItem[strBaseData][strProperties] || {};
         EnvelopeCreator.extractPropsAndMeasurements(telemetryItem.data, customProperties, customMeasurements);
+        EnvelopeCreator.convertPropsUndefinedToBlank(customProperties);
```
MSNev, line 194: This needs to be behind a configuration as this will affect the performance and not everyone will want this behavior. It might even be worth having the config define the "value" used for undefined.

channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
```diff
@@ -162,7 +170,7 @@ export abstract class EnvelopeCreator {
 
     protected _logger: IDiagnosticLogger;
 
-    abstract Create(logger: IDiagnosticLogger, telemetryItem: ITelemetryItem): IEnvelope;
+    abstract Create(logger: IDiagnosticLogger, telemetryItem: ITelemetryItem, customDefinedValue?: any): IEnvelope;
```
MSNev, line 173: nit: ```customUndefinedValue```

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -106,7 +107,8 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
             instrumentationKey: undefined,
             namePrefix: undefined,
             samplingPercentage: undefined,
-            customHeaders: undefined
+            customHeaders: undefined,
+            convertCustomDimensionUndefinedField: undefined,
```
MSNev, line 111: nit: Can we make this a little smaller as this is public and therefore all references will never get minified.

channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
xiao-lix: I mean the value is a custom defined value to replace the "undefined", should this be "customUndefinedValue"?

channels/applicationinsights-channel-js/src/Sender.ts
xiao-lix: Yeah.. agree.. what about "convertUndefined"?

channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
MSNev: I think it should be ```customUndefinedValue``` as this is the value that will be used when the value is ```undefined```/

MSNev (approved):

Issue #1585, opened 2021-06-17
ai_user cookie not present after reenabling the cookie


julen-folky:
My app (Vue) starts with the cookies disabled (`disableCookiesUsage` set to `true`).
When the user consents cookies, the app enables the cookies (`getCookieMgr().setEnabled(true)`).
After a breve navigation the `ai_session` is set but the `ai_user` is not.
I recently moved to the lastest version of the package (from 2.5.X to 2.6.2).
Is this an issue, somehting I am missing or just the `ai_user` is no longer used?

MSNev:
Simply, re-enabling cookies does not cause the code that creates cookies to execute -- likewise disabling cookies doesn't delete them.

The ai_user cookie is still used, but it's only "created" as part of the creating (loading) of the SDK...

The assumption is that the next time your app/page loads cookies will be enabled and therefore the ai_user would get created...

As your using Vue, can I assume your case is a SPA and you won't be "reloading" and therefore "creating" an instance?

julen-folky:
Exactly, after the first load the page will not be load again.
If the ai_user is not set, I guess some info is not going to be collected (before your reply I assume that if ai_session would be in charge of that in 2.6.2).
Should I manually set the cookie? But, how can I know its value?
(BTW, thank you for your help)

MSNev:
users and sessions are slightly different concepts.
So with no user cookie the user would be considered to be effectively "anonymous" (same as if the user blocked cookies on the browser) and therefore they would be lumped together (I'm not sure how this is visualized in the portal)
And the session is "refreshed" after around 30mins (depending on whether the user stays active)

In theory you could set the value yourself but that is not supported and will be potentially broken in a future release.

We already have a ```setAuthenticatedUser``` which you could possibly use as a fudge to workaround this (when stored in the cookie it is a session cookie)
```_self.setAuthenticatedUserContext = (authenticatedUserId: string, accountId?: string, storeInCookie = false) => {```

I'm going to mark this Issue as an enhancement on the basis that we should add a new function suggestion ```setUserCookie``` to handle this scenario. Not currently sure on the timing of when this might get added.

MSNev:
v2.7.0 is not fully deployed to NPM and all CDN endpoints

PR #1632, opened 2021-08-11
ai_user cookie not present after reenabling the cookie #1585

MSNev:
- Add optimization to avoid always updating the cookie value

### DIFF
--- a/extensions/applicationinsights-properties-js/src/Context/User.ts
+++ b/extensions/applicationinsights-properties-js/src/Context/User.ts
@@ -56,7 +56,15 @@ export class User implements IUserContext {
      */
     public accountAcquisitionDate: string;
 
+    /**
+     * A flag indicating whether this represents a new user
+     */
     public isNewUser = false;
+
+    /**
+     * A flag indicating whether the user cookie has been set
+     */
+    public isUserCookieSet = false;
 
     constructor(config: ITelemetryConfig, core: IAppInsightsCore) {
         let _logger = safeGetLogger(core);
@@ -75,6 +83,8 @@ export class User implements IUserContext {
                 const params = cookie.split(User.cookieSeparator);
                 if (params.length > 0) {
                     _self.id = params[0];
+                    // we already have a cookie
+                    _self.isUserCookieSet = !!_self.id;
                 }
             }
 
@@ -98,7 +108,7 @@ export class User implements IUserContext {
                 // set it to 365 days from now
                 // 365 * 24 * 60 * 60 = 31536000 
                 const oneYear = 31536000;
-                _cookieManager.set(_storageNamePrefix(), cookie, oneYear);
+                _self.isUserCookieSet = _cookieManager.set(_storageNamePrefix(), cookie, oneYear);
             }
 
             if (!_self.id) {
@@ -172,9 +182,12 @@ export class User implements IUserContext {
             };
 
             _self.update = (userId?: string) => {
-                let user_id = userId ? userId : _generateNewId();
-                let user_cookie = _generateNewCookie(user_id);
-                _setUserCookie(user_cookie.join(User.cookieSeparator));
+                // Optimizations to avoid setting and processing the cookie when not needed
+                if (_self.id !== userId || !_self.isUserCookieSet) {
+                    let user_id = userId ? userId : _generateNewId();
+                    let user_cookie = _generateNewCookie(user_id);
+                    _setUserCookie(user_cookie.join(User.cookieSeparator));
+                }
             };
         });
     }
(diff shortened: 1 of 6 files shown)


extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -75,6 +83,8 @@ export class User implements IUserContext {
                 const params = cookie.split(User.cookieSeparator);
                 if (params.length > 0) {
                     _self.id = params[0];
+                    // we already have a cookie
+                    _self.isUserCookieSet = !!_self.id;
```
MSNev, line 87: If we successfully read a value from the cookie then we assume that the cookie was set.

extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -98,7 +108,7 @@ export class User implements IUserContext {
                 // set it to 365 days from now
                 // 365 * 24 * 60 * 60 = 31536000 
                 const oneYear = 31536000;
-                _cookieManager.set(_storageNamePrefix(), cookie, oneYear);
+                _self.isUserCookieSet = _cookieManager.set(_storageNamePrefix(), cookie, oneYear);
```
MSNev, line 111: Use new response that tells us if the cookie was really set.

extensions/applicationinsights-properties-js/src/Context/User.ts
```diff
@@ -172,9 +182,12 @@ export class User implements IUserContext {
             };
 
             _self.update = (userId?: string) => {
-                let user_id = userId ? userId : _generateNewId();
-                let user_cookie = _generateNewCookie(user_id);
-                _setUserCookie(user_cookie.join(User.cookieSeparator));
+                // Optimizations to avoid setting and processing the cookie when not needed
+                if (_self.id !== userId || !_self.isUserCookieSet) {
```
MSNev, line 186: Add an extra performance based check so we don't try and set the cookie if
- The value was already set
- The new userId is the same as the previous one.

extensions/applicationinsights-properties-js/src/PropertiesPlugin.ts
```diff
@@ -92,14 +92,15 @@ export default class PropertiesPlugin extends BaseTelemetryPlugin implements IPr
...
+                    let userCtx = theContext.user;
+                    if (userCtx && !userCtx.isUserCookieSet) {
+                        userCtx.update(theContext.user.id);
                     }
     
                     _processTelemetryInternal(event, itemCtx);
     
-                    if (theContext.user && theContext.user.isNewUser) {
-                        theContext.user.isNewUser = false;
+                    if (userCtx && userCtx.isNewUser) {
```
MSNev, line 102: Simple minification improvement which also has a miner perf improvement.

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/ICookieMgr.ts
```diff
@@ -20,8 +20,9 @@ export interface ICookieMgr {
      * @param maxAgeSec - [optional] The maximum number of SECONDS that this cookie should survive
      * @param domain - [optional] The domain to set for the cookie
      * @param path - [optional] Path to set for the cookie, if not supplied will default to "/"
+     * @returns - True if the cookie was set otherwise false (Because cookie usage is not enabled or available)
      */
-    set(name: string, value: string, maxAgeSec?: number, domain?: string, path?: string): void;
+    set(name: string, value: string, maxAgeSec?: number, domain?: string, path?: string): boolean;
```
MSNev, line 25: While this is a change to the signature of the function, it should be backward compatible as previously it was void.

MSNev: This is also used by internally as well, so this should be compatible with that as well.

shared/AppInsightsCore/src/JavaScriptSDK/CookieMgr.ts
```diff
@@ -190,6 +191,8 @@ export function createCookieMgr(rootConfig?: IConfiguration, logger?: IDiagnosti
                 let setCookieFn = cookieMgrConfig.setCookie || _setCookieValue;
                 setCookieFn(name, _formatCookieValue(theValue, values));
             }
+
+            return result;
```
xiao-lix, line 195: Do you forget to set result = true if cookie manager is enabled? Looks like the result will always be false.

MSNev: nice catch!

MSNev: Added

xiao-lix (approved):

PR #1645, opened 2021-08-26
[Feature] Add Statsbeat Network data collection

xiao-lix:
TODOs and FYI: 
1. Need to discuss what count as exceptions on web. eg. Do we want to count all `throwInternals` as exception for statsbeat?
2. Need to add more config check, eg. sample rate. Now only checks if it's disabled by user.
3. For telemetries sent by Web SDK, all envelops have a name with such format: `Microsoft.ApplicationInsights.<ikey>.Metric`. This seems too long for statsbeat, but it's required by the SDK if I understand correctly. Reference [here](https://github.com/microsoft/ApplicationInsights-JS/blob/6596e4e4541b78920bf4c1617270c652ddeda6c7/channels/applicationinsights-channel-js/src/EnvelopeCreator.ts#L60).
4. For Web SDK, Network statsbeat are sent as a single telemetry (eg. countRequest, countRetry, countException, etc), the counts are put in custom dimensions - `data.baseData.properties`. Thus the queries to get statsbeat data from Web SDK will be different.

A sample payload: 
image


channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,162 @@
...
+    private _config: IConfiguration & IConfig;
+    private _isEnabled: boolean;
+
+    // Custom dimensions
+    private _cikey: string;
+
+    public initialize(config: IConfiguration & IConfig, core: IAppInsightsCore, extensions: IPlugin[], pluginChain?:ITelemetryPluginChain, endpoint?: string) {
+        this._networkCounter = new NetworkStatsbeat(endpoint);
+        this._sender = new Sender(undefined);
+        this._sender.initialize(config, core, extensions, pluginChain, endpoint);
```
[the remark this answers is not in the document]
xiao-lix: Is there a preference? I tried both options locally, and I think the first option (to pass null as ikey) is easier.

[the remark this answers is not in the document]
MSNev: Passing the extra param is easier, but I think the config is probably cleaner as it means we don't need the extra check/code in Sender

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -427,20 +439,23 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
             /**
              * error handler
              */
-            _self._onError = (payload: string[], message: string, event?: ErrorEvent) => {
+            _self._onError = (payload: string[], message: string, event?: ErrorEvent, endpointUrl?: string) => {
```
lzchen, line 442: How come you need to pass in the `endpointUrl` for every callback? Isn't there someway to access the `Config` globally, so you can just access `endpointUrl` within statsbeat itself?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
+import { NetworkStatsbeat } from "./NetworkStatsbeat";
+import { Sender } from "./Sender";
+import { StatsbeatCounter } from "./Constants";
+import { IConfig, Metric } from "@microsoft/applicationinsights-common";
+import { IConfiguration, IAppInsightsCore, IPlugin, ITelemetryPluginChain, ITelemetryItem, objKeys, dateNow,  } from '@microsoft/applicationinsights-core-js';
+import dynamicProto from '@microsoft/dynamicproto-js';
+
+export class Statsbeat {
+    public static INSTRUMENTATION_KEY = "c4a29126-a7cb-47e5-b348-11414998b11e";
```
lzchen, line 9: Any reason for these to be public?

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -264,6 +269,11 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
                         }
                     }
                 }
+
+                // initialize statsbeat instance last after sender config is fully populated
+                if (_self._statsbeat && config.disableStatsbeat !== true) {
```
lzchen, line 274: Why even create the `Statsbeat` object if config is set to disabled? Why can't initializing happen on creation instead of explicitly calling `initialize`? Is this a design pattern used across the sdk?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+        dynamicProto(Statsbeat, this, (_self, _base) => {
+            _self.initialize = (config: IConfiguration & IConfig, core: IAppInsightsCore, extensions: IPlugin[], pluginChain?:ITelemetryPluginChain, endpoint?: string) => {
+                _networkCounter = new NetworkStatsbeat(endpoint);
+                _sender = new Sender(undefined);
+                let senderConfig = {...config};
+                senderConfig.instrumentationKey = Statsbeat.INSTRUMENTATION_KEY;
+                _sender.initialize(senderConfig, core, extensions, pluginChain, endpoint);
+                _statsbeatMetrics = {};
+                _config = config;
+                _isEnabled = _config.disableStatsbeat !== true;
```
lzchen, line 32: Is there a point to all the initializing code above this line if `disableStatsbeat` is `true`?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+                    }
+                }
+            }
+
+            _self.isInitialized = (): boolean => {
+                return !!_isEnabled;
+            }
+
+            _self.countRequest = (endpoint: string, duration: number, success: boolean) => {
+                if (!_isEnabled) {
```
lzchen, line 48: I don't think `_isEnabled` is needed to be checked for each if we just don't initialize at the beginning.

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+                    if (!_handle) {
+                        _handle = setInterval(() => {
+                            this.trackShortIntervalStatsbeats();
+                        }, Statsbeat.STATS_COLLECTION_SHORT_INTERVAL);
+                    }
+                }
+            }
+
+            _self.isInitialized = (): boolean => {
+                return !!_isEnabled;
```
lzchen, line 44: Shouldn't this be `isEnabled`? What's the relationship between being `initialized` and being `enabled`?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+            _self.trackShortIntervalStatsbeats = (): void => {
+                _trackRequestDuration();
+                _trackRequestsCount();
+                _sendStatsbeats();
+            }
+
+            function _getCustomProperties() {
+                _cikey = _config.instrumentationKey;
```
lzchen, line 93: Missing some properties?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+                    };
+                    _sender.processTelemetry(statsbeat);
+                }
+                _statsbeatMetrics = {};
+                _sender.triggerSend();
+            }
+
+            function _trackRequestDuration() {
+                var currentCounter = _networkCounter;
+                currentCounter.time = dateNow();
```
lzchen, line 130: Will this throw an exception if `_networkCounter` is not available (not instantiated)?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+                _sendStatsbeats();
+            }
+
+            function _getCustomProperties() {
+                _cikey = _config.instrumentationKey;
+            }
+
+            function _getNetworkStatsbeatCounter(host: string): NetworkStatsbeat {
+                // Check if counter is available
+                if (_networkCounter && _networkCounter.host === host) {
```
lzchen, line 98: If the `host` doesn't match the calling endpoint, we create a new instance? When would this happen?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+            function _trackRequestDuration() {
+                var currentCounter = _networkCounter;
+                currentCounter.time = dateNow();
+                var intervalRequests = (currentCounter.totalRequestCount - currentCounter.lastRequestCount) || 0;
+                var elapsedMs = currentCounter.time - currentCounter.lastTime;
+                var averageRequestExecutionTime = ((currentCounter.intervalRequestExecutionTime - currentCounter.lastIntervalRequestExecutionTime) / intervalRequests) || 0;
+                currentCounter.lastIntervalRequestExecutionTime = currentCounter.intervalRequestExecutionTime; // reset
+                if (elapsedMs > 0 && intervalRequests > 0) {
```
lzchen, line 135: Shouldn't `averageRequestExecutionTime` be the one that should be checked for negative?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+                    _statsbeatMetrics.properties = _statsbeatMetrics.properties || {};
```
lzchen, line 136: How come it's just the properties that are not populated if `if (elapsedMs > 0 && intervalRequests > 0) {` is not true? Shouldn't the metric just not be sent?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+export class Statsbeat {
+    public static INSTRUMENTATION_KEY = "c4a29126-a7cb-47e5-b348-11414998b11e";
+    public static STATS_COLLECTION_SHORT_INTERVAL: number = 900000; // 15 minutes
+    public static NETWORK = "Network";
+
+    constructor() {
+        let _networkCounter: NetworkStatsbeat;
+        let _sender: Sender;
+        let _handle: any;
+        let _statsbeatMetrics: { properties?: {} };
```
lzchen, line 17: Nit: Can you use `metrics` instead of `properties`?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+                    "cikey": _cikey,
+                }
+                if (objKeys(_statsbeatMetrics)) {
+                    let statsbeat: ITelemetryItem = {
+                        iKey: Statsbeat.INSTRUMENTATION_KEY,
+                        name: Statsbeat.NETWORK,
+                        baseData: {
+                            name: Statsbeat.NETWORK,
+                            average: 0,
+                            properties: {"host": _networkCounter.host, ..._statsbeatMetrics.properties, ...networkProperties},
```
lzchen, line 118: Is this logic correct? I believe each entry in `_statsbeatMetrics` is supposed to be sent as a separate metric, not as a custom dimension?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+                        },
+                        baseType: Metric.dataType
+                    };
+                    _sender.processTelemetry(statsbeat);
```
lzchen, line 122: Is `processTelemetry` for telemetry processors? Are `statsbeat` metrics ever processed?

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -724,11 +754,19 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
...
              */
-            function _resendPayload(payload: string[], linearFactor: number = 1) {
+            function _resendPayload(payload: string[], linearFactor: number = 1, endpointUrl: string, status?: number) {
                 if (!payload || payload.length === 0) {
                     return;
                 }
         
+                var endpointHost = urlParseUrl(endpointUrl).hostname;
+                if (_self._statsbeat && _self._statsbeat.isInitialized()) {
+                    _self._statsbeat.countRetry(endpointHost);
```
lzchen, line 764: I think `statsbeat` metrics are mutually exclusive. Meaning if status code is `429`, it will count as throttle instead of retry.

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+                }
+                _statsbeatMetrics = {};
```
lzchen, line 124: Does this have to be reset everytime? The values are just overwritten right?

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -608,9 +631,16 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
                     _self._buffer.markAsSent(payload);
                     // no response from beaconSender, clear buffer
                     _self._onSuccess(payload, payload.length);
+                    if (_self._statsbeat && _self._statsbeat.isInitialized()) {
+                        let endTime = dateNow();
+                        _self._statsbeat.countRequest(url, endTime - startTime, true);
```
lzchen, line 636: We only increment counts for `beaconSender`?

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -608,9 +631,16 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
...
+                    }
                 } else {
-                    _xhrSender(payload, true);
+                    _xhrSender(payload, true, startTime);
                     _self.diagLog().throwInternal(LoggingSeverity.WARNING, _InternalMessageId.TransmissionFailed, ". " + "Failed to send telemetry with Beacon API, retried with xhrSender.");
+                    if (_self._statsbeat && _self._statsbeat.isInitialized()) {
```
lzchen, line 641: Failure to send with beacon api counts as exception? What about the result of xhrSender? Do we count that as well?

channels/applicationinsights-channel-js/src/Statsbeat.ts
MSNev: Nice catch, we should just make this a const at the top of the file.

channels/applicationinsights-channel-js/src/Statsbeat.ts
MSNev: This is defining the "type" the actual value of _statsbeatMetric will be undefined

channels/applicationinsights-channel-js/src/Statsbeat.ts
MSNev: They won't match if during a previous batch send Breeze redirected the response.

channels/applicationinsights-channel-js/src/Statsbeat.ts
xiao-lix: Thanks, modified.

channels/applicationinsights-channel-js/src/Sender.ts
xiao-lix: The design is to pass a statsbeat instance through sender and use that for each AI. And statsbeat itself has a sender as well, we figured out this way to differentiate these two senders by checking if a statsbeat instance is passed in.

channels/applicationinsights-channel-js/src/Statsbeat.ts
xiao-lix: I realized when it comes into statsbeat's initialize function, the config.disableStatsbeat is either `false` or `undefined`, it will never be `true`. Because in sender, we only call statsbeat.initialize when the config is false or undefined. But the flag is useful to check if statsbeat is enabled. So to answer Leighton's question - the sender needs to be initialized for statsbeat because this is inside statsbeat's initialize function, but as it confuses you, it may not necessary to check config, just assign `_isEnabled = true` here would be good I think. @MSNev what do you think?

channels/applicationinsights-channel-js/src/Statsbeat.ts
xiao-lix: They're the same. You mean to update the function name to isEnabled?

channels/applicationinsights-channel-js/src/Statsbeat.ts
xiao-lix: Yeah for web sdk, we decide to group the metrics into one telemetry - this would cause the query change. I listed a point in description for this change.

channels/applicationinsights-channel-js/src/Statsbeat.ts
xiao-lix: It's not necessarily have all the properties for each telemetry. so clear it up each time.

channels/applicationinsights-channel-js/src/Sender.ts
xiao-lix: Failed to send with any sender counts as exception. For other senders, they call `_onError` inside which is calling statsbeat.countException. Are you saying if sendbeacon fails then xhrsender also fails. In that case we'll double counting exceptions? I'm not sure if this count as duplicate - because they're both exceptions isn't it? @MSNev what do you think?

channels/applicationinsights-channel-js/src/Statsbeat.ts
```diff
@@ -0,0 +1,203 @@
...
+                _sender = new Sender(undefined);
+                let senderConfig = {...config};
+                senderConfig.instrumentationKey = Statsbeat.INSTRUMENTATION_KEY;
+                _sender.initialize(senderConfig, core, extensions, pluginChain, endpoint);
+                _statsbeatMetrics = {};
+                _config = config;
+                _isEnabled = _config.disableStatsbeat !== true;
+                if (_isEnabled) {
+                    _getCustomProperties();
+                    if (!_handle) {
```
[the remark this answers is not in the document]
xiao-lix: hmm, how would initialize a single statsbeat multiple times happen? The statsbeat.initialize is called inside sender.initialize, and sender is only initialize once. And `_statsbeat` is a private property of sender, you can't call its initialize method outside of sender, thus it's not possible to be initialized multiple times. @MSNev can you please confirm if I understand correct? Thanks~

channels/applicationinsights-channel-js/src/Statsbeat.ts
xiao-lix: Yeah... thanks for pointing it out, this is a todo item I forgot to mention. I'll follow up with Nev and Hector:
- what should "language" property be for web? "JavaScript"?
- what should "sdkVersion" property be for web? 
- what about os and runtimeVersion?

channels/applicationinsights-channel-js/src/Statsbeat.ts
xiao-lix: @hectorhdzg can you please help here thanks! Should it be different from node sdk?

channels/applicationinsights-channel-js/src/Sender.ts
xiao-lix: Because for xdr senders, endpointUrl could be different. Eg. for xdr the endpoint url is reprocessed [here](https://github.com/microsoft/ApplicationInsights-JS/blob/44ec248075402dd0827fa8eaeb15e398e9d0be8e/channels/applicationinsights-channel-js/src/Sender.ts#L839). @MSNev please correct me if I understand wrong... or is it ok to ALWYAS get endpointUrl from `config.endpointUrl === undefined ? "https://dc.services.visualstudio.com/v2/track" : config.endpointUrl`  thanks~

MSNev: As the XDR is just ripping off the leading protocol "https", I think we can just always use the config, as this is only for IE9 or lower.

channels/applicationinsights-channel-js/src/Statsbeat.ts
MSNev: My 2c
- Language => JavaScript
- sdkVersion => The current sdk version (we set this in the ext.sdk.version)
- os/runTimeVersion => Either empty or Browser (We don't really want to start completely cracking the UserAgent to figure this out), and collector will do that for use anyway.

channels/applicationinsights-channel-js/src/Statsbeat.ts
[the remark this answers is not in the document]
MSNev: As any embedded object inside of sender, there will actually be no way to ```initialize``` more than once as the framework blocks multiple initialization of components (they have there own isInitialized()) flag, so I don't think we need that for this as well (based on the embedded design)

channels/applicationinsights-channel-js/src/Sender.ts
MSNev: There will be no double counting, the Sender only uses 1 transport, beacon, xhr, xdr or fetch.

channels/applicationinsights-channel-js/src/Statsbeat.ts
MSNev: Yeh, just assign to true.

channels/applicationinsights-channel-js/src/Statsbeat.ts
xiao-lix: Had a second thought, decide to keep the original name to be compliant with other plugins. and isEnabled is to be compliant with node sdk.

MSNev:
Now superceeded by #2489

Issue #1647, opened 2021-08-30
[BUG] customProperties parameter missing from trackException function


devin5885:
**Description/Screenshot**
In some places the trackException function includes the customProperites parameter, in other places it does not. Without this parameter it does not seem to be possible to pass custom properties when logging an exception.

**Steps to Reproduce**
n/a

 - OS/Browser: Windows 10/Chrome
 - SDK Version [e.g. 22]: latest
 - How you initialized the SDK: yes
n/a

**Expected behavior**
trackException should include this parameter in all cases.

**Additional context**

MSNev:
Looks like it is missing from the ```AISKU/src/Initialization.ts``` version, not seeing anything obvious as to why this was not included in this version.

It's specifically not included in the ```AISKU/src/ApplicationInsights/Deprecated.ts``` due to backward compatibility using the older snippet and upgrading from v1.

Tagging this issue as a Bug so we add the additional customProperties?: ICustomProperties to ```Initialization``` only

MSNev:
v2.7.0 is not fully deployed to NPM and all CDN endpoints

PR #1651, opened 2021-08-31
[BUG] customProperties parameter missing from trackException function #1647

MSNev:


AISKU/Tests/applicationinsights.e2e.tests.ts
```diff
@@ -280,6 +306,38 @@ export class ApplicationInsightsTests extends TestClass {
...
+                        columnNumber: 53,
+                        error: undefined,
+                        evt: null
+                    } as IAutoExceptionTelemetry;
+    
+                    try {
+                        exception = e;
+                        this._ai.trackException({ exception: autoTelemetry }, { custom: "custom value" });
+                    } catch (e) {
+                        console.log(e);
```
xiao-lix, line 332: nit: Are the console logs on purpose? But I guess it's fine to have in tests.

xiao-lix (approved):

AISKU/Tests/applicationinsights.e2e.tests.ts
ramthi: Assert.ok(false) fails the test correct? and the stack is displayed as the error message.  What is the purpose of console.log?

MSNev: Correct, console.log is for easier debugging the failure in a browser.

Issue #1667, opened 2021-09-12
Allow properly disposing AI


olee:
I am using application insights through npm in a web application and also running the project in nodejs for jest tests.
However, because [`AppInsightsCore.pollInternalLogs`](https://github.com/microsoft/ApplicationInsights-JS/blob/master/shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts#L72-L94) creates a `setInterval` which handle is never stored and there is no way to dispose of it, jest never manages to shut down properly.

Would you mind properly storing the returned handle from the `setInterval` and provide some kind of `dispose` function to stop this polling?

Alternatively, could it be possible to add an option by eg. setting `diagnosticLogInterval` to `0` to completely disable this polling?
From my tests I noticed that this internal queue never seems to be used anyway and always stays empty.

EDIT: For now I managed to disable this feature by executing this code before AI initialization:
```ts
        const ai = new ApplicationInsights({ config: { /* ... */ } });
        // Disable pollInternalLogs to prevent jest from properly shutting down
        ai.pollInternalLogs = () => undefined;
        ai.loadAppInsights();
```

kryalama:
Setting the interval to 0 might not solve the problem. setInterval actually has a minimum amount which is implicit, and depends on the browser. I believe it is around 10ms. We can just return without creating a setInterval when diagnosticLogInterval is set to 0.

olee:
@kryalama that was of course what I was implying.
If it worked that way, I would have solved it by setting it to 0 of course 😂

PR #1674, opened 2021-09-17
expose internal log poller

kryalama:
Address https://github.com/microsoft/ApplicationInsights-JS/issues/1667


olee:
I don't think it's necessary to expose the internal timer but instead just a method to stop the internal polling.
That would also allow the framework to properly know whether the timer is running or not which would not work when exposing the timer handle and just letting users cancel it.

AISKU/src/Initialization.ts
```diff
@@ -463,7 +464,7 @@ export class Initialization implements IApplicationInsights {
     }
 
     public pollInternalLogs(): void {
-        this.core.pollInternalLogs();
+        this._internalLogPoller = this.core.pollInternalLogs();
```
MSNev, line 467: Lets not expose this as it will restrict what and how we might refactor this in the future.

Lets change to have a ```this.core.stopPollInternalLogs()```, lets also fix that if ```pollInternalLogs``` is called more than once it stops the previous generated interval.

MSNev (changes requested):
Lets not expose the internal interval timer value

AISKU/src/Initialization.ts
```diff
@@ -527,7 +537,7 @@ export class Initialization implements IApplicationInsights {
     private getSKUDefaults() {
         let _self = this;
         _self.config.diagnosticLogInterval =
-            _self.config.diagnosticLogInterval && _self.config.diagnosticLogInterval > 0 ? _self.config.diagnosticLogInterval : 10000;
+            _self.config.diagnosticLogInterval && _self.config.diagnosticLogInterval >= 0 ? _self.config.diagnosticLogInterval : 10000;
```
MSNev, line 540: When diagnosticLogInterval === 0 the following part will be false, so the extra = will never get evaluated

> _self.config.diagnosticLogInterval

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IDiagnosticLogger.ts
```diff
@@ -33,7 +34,7 @@ export interface IDiagnosticLogger {
     /**
      * The internal logging queue
      */
-    queue: _InternalLogMessage[];
+    queue:ILogMessageQueue;
```
MSNev, line 37: As a public property this is a breaking change, while we don't expect people to be access it we can't make this change

MSNev: To implement the queue length checks / limits we will need to do this as a helper function to update the check and reduce the queue.

The only other "safe" way to do this would be to construct a new object that implements the COMPLETE array accessors and functions which from a space and performance perspective is not viable and we would also need to use "get" which is not compatible with our current ES3 limitation 😢

AISKU/src/Initialization.ts
```diff
@@ -463,7 +464,16 @@ export class Initialization implements IApplicationInsights {
...
-        this.core.pollInternalLogs();
+        // If pollInternalLogs is called more than once, existing log poller should be stopped before starting another log poller.
+        if(this._internalLogPoller) {
+            this.stopPollingInternalLogs();
+        }
+        this._internalLogPoller = this.core.pollInternalLogs();
+    }
+
+    public stopPollingInternalLogs(): void {
+         this.core.stopPollingInternalLogs(this._internalLogPoller);
```
MSNev, line 475: Probably add an if check on whether we have a timer?

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IAppInsightsCore.ts
```diff
@@ -74,6 +74,8 @@ export interface IAppInsightsCore extends IPerfManagerProvider {
 
     pollInternalLogs?(eventName?: string): number;
 
+    stopPollingInternalLogs?(internalLogPoller?: number): void;
```
MSNev, line 77: This is still exposing the internal timer...

Lets move the _internalLogPoller to this class so then the stop doesn't take any arguments

MSNev: And as the AppInsightsCore is using dynamicProto we won't need to expose _internalLogPoller on the instance either, it can remain within the closure -- for even better minification

MSNev (approved):

PR #1694, opened 2021-10-19
Governance Updates -- update used dependencies

MSNev:


MSNev:
This doesn't address all of the component governance issues as some of the dependencies (one being the current npm) has not yet released a version that addresses all of them.

So this addresses those that it can.

Issue #1680, opened 2021-10-01
[BUG] trackMetric does not track stdDev nor sum


stefanedwards:
When tracking a (custom) metric, the standard deviation is not tracked (as `valueStdDev`) and `valueSum` is identical to `value`.
Additionally, when `min`-value equals 1, 

**Steps to Reproduce**

 - OS/Browser: Google Chrome 93.0.4577.82 (64-bit), Windows 10
 - SDK Version [e.g. 22]: javascript:2.7.0
 - How you initialized the SDK:

```
<!DOCTYPE html>
<html>
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
  <script src="https://js.monitor.azure.com/scripts/b/ai.2.7.0.min.js"></script>
  <script>
  
  let msg = { config: {
    "instrumentationKey": "<<>>",
    "appId": "Debug Metrics"
}}
  
let init = new Microsoft.ApplicationInsights.ApplicationInsights(msg);
let appInsights = init.loadAppInsights();  
  </script>
</head>
<body>
	<script>
	appInsights.trackMetric({name: 'my_metric', average: 1, sampleCount: 2, min: 1, max: 2, stdDev: 1.0});
	appInsights.flush();
	</script>
</body>
</html>
```

When viewing the tracked metric in Application Insights on portal.azure.com, the values do not match what was provided.

![Metric in Application Insights log on portal.azure.com](https://user-images.githubusercontent.com/3592346/135607240-96a94916-ecec-4fb7-ae95-aaa48694d032.png)

**Expected behavior**

1. It should be possible to track the `sum`.
2. When retrieving the metric in the Application Insights logs on portal.azure.com, the `value`, `valueCount`, `valueSum`, `valueMin`, `valueMax`, and `valueStdDev` should match (within some reasonable precision) to the values submitted from the JS script. 

**Additional context**
The documentation on `appInsights.trackMetric` is entirely missing (for the JS-part), and this is my best guess for actually getting something through. A lot of other attempts (e.g. `trackMetric('name',value)` didn't even raise an error -- but nothing appeared to be sent).

I have further found, that the second argument to `trackMetric` can be used to send additional properties (i.e. `customDimensions`), such as `appInsights.trackMetric({name: 'bacon', average: 5, sampleCount: 6, min: 1, max: 15}, {PageName: "foo", "PageUrl": "/", "questionable_field": "zap" });`

Karlie-777:
[...]

PR #1701, opened 2021-10-28
[BUG] trackMetric does not track stdDev nor sum #1680

MSNev:
- add (missing) stdDev sending for metrics


dballenger85:
This change has broken the ability to track custom properties/dimensions as part of TrackMetrics API.   You missed a change to the MetricEnvelopeCreator function in EnvelopeCreator.ts

PR #1715, opened 2021-11-13
Address Security issues identified by internal component governance services

MSNev:


AISKU/Tests/Unit/src/SnippetInitialization.Tests.ts
```diff
@@ -116,7 +116,7 @@ export class SnippetInitializationTests extends AITestClass {
 
             this.testCaseAsync({
                 name: "[" + snippetName + "] : Public Members exist",
-                stepDelay: 1,
+                stepDelay: 100,
```
MSNev, line 119: Adding test delay as these are randomly failing (assuming that sometimes it was taking more than 1ms)

kryalama (approved):

PR #1717, opened 2021-11-16
Downgrade to npm 6 to avoid CVE-2021-43616

MSNev:


MSNev:
This introduces other critical failures 😢

PR #1726, opened 2021-12-03
In getting IE version, prefer Trident version over MSIE version.

cfeltner:
Embedded IE browser control includes higher Trident version than MSIE version, so updated getIEVersion to prefer Trident version over MSIE version.

### DIFF
--- a/shared/AppInsightsCommon/Tests/Unit/src/Util.tests.ts
+++ b/shared/AppInsightsCommon/Tests/Unit/src/Util.tests.ts
@@ -536,6 +536,7 @@ export class UtilTests extends AITestClass {
                 Assert.equal(8, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 5.1; Trident/4.0)"));
                 Assert.equal(8, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.1; Trident/4.0)"));
                 Assert.equal(8, Util.getIEVersion("Mozilla/5.0 (compatible; MSIE 8.0; Windows NT 6.0; Trident/4.0; WOW64; Trident/4.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729; .NET CLR 1.0.3705; .NET CLR 1.1.4322)"));
+                Assert.equal(8, Util.getIEVersion("Mozilla/4.0 (Compatible; MSIE 8.0; Windows NT 5.2; Trident/6.0)"));
                 Assert.equal(9, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 9.0; Windows NT 6.0)"));
                 Assert.equal(9, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 9.0; Windows NT 6.1)"));
                 Assert.equal(10, Util.getIEVersion("Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.1; WOW64; Trident/6.0)"));
@@ -548,7 +549,7 @@ export class UtilTests extends AITestClass {
 
                 const origDocMode = document['documentMode'];
                 document['documentMode'] = 11;
-
+ 
                 Assert.equal(11, Util.getIEVersion("Mozilla/4.0 (Compatible; MSIE 8.0; Windows NT 5.2; Trident/6.0)"));
                 Assert.equal(11, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.2; Win64; x64; Trident/7.0; .NET4.0C; .NET4.0E)"));
 
--- a/shared/AppInsightsCommon/Tests/Unit/src/Util.tests.ts
+++ b/shared/AppInsightsCommon/Tests/Unit/src/Util.tests.ts
@@ -531,14 +531,14 @@ export class UtilTests extends AITestClass {
                 }
 
                 Assert.equal(7, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 7.0b; Windows NT 6.0)"));
-                Assert.equal(7, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.0; WOW64; Trident/4.0;)"));
                 Assert.equal(7, Util.getIEVersion("Mozilla/5.0 (compatible; MSIE 7.0; Windows NT 6.0; en-US)"));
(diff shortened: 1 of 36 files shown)


shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -311,13 +311,14 @@ export function getIEVersion(userAgentStr: string = null): number {
     }
 
     var ua = (userAgentStr || "").toLowerCase();
-    if (strContains(ua, strMsie)) {
-        return parseInt(ua.split(strMsie)[1]);
-    } else if (strContains(ua, strTrident)) {
+    // Checking for Trident first since embedded IE browser control includes higher version than MSIE.
```
MSNev, line 314: This is by design as by default an embedded IE (WebBrowser) actually runs in IE7 mode, so detecting based on the Trident version would result in potentially the wrong values being used.

Returning a X-UA-Compatible met tag in your HTML (should) cause IE to update the "version" of MSIE that it's "running" as.

MSNev:
A better approach for this (might be) to add code to review the `document.documentMode` and `document.compatMode`, but generally this has not been required and the extra code would be unnecessary for most users.

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
cfeltner: We are sending a X-UA-Compatible meta tag with the following contents:
`http-equiv="X-UA-Compatible" content="IE=edge"`
  
But this is the User Agent string that we are getting back:
`Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.2; Win64; x64; Trident/7.0; .NET4.0C; .NET4.0E)`

It is my understanding that this should cause it to use the highest document mode supported which in this case should be IE 11 which matches up with the Trident version (7+4=11) but not the MSIE version. This is why I proposed changing it to prefer the Trident version which takes into account the X-UA-Compatible meta tag.

If the X-UA-Compatible meta tag is not used, then it would run in the default IE7 mode.

cfeltner:
> A better approach for this (might be) to add code to review the document.documentMode and document.compatMode, but generally this has not been required and the extra code would be unnecessary for most users.

Are you proposing changing getieversion to return document.documentMode instead of Trident and MSIE versions? Seems like this would be the same as returning the Trident version.

MSNev:
In your current case because you have the X-UA-Compat meta tag this is probably returning the same values, but for other users that are not including the header this would fall back to the current "mode".

And the returned value of this function is used to switch some features on/off (both in this codebase and another internal variant), so just changing to always use Trident (could) end up causing exceptions because some code would start to assume things and objects that don't exist.

So I guess the real question is why do you need to know this? As ___most___ of the code only uses the response as a boolean flag to determine if this is IE or not.

However, I think you have identified what I think is an underlying bug as the Ajax tracking looks like it will always be disabled in this case.

So, to come back to your initial question, yes adding a documentMode check would work around this using something like this so it still falls back to the current behavior. Which I think would then fix the above, handle your case and also ensure that other users without (or a different) meta-tag are also handled correctly.

I'm thinking something like this to minimize the code impact and handle cases where the SDK is not used in a full browser environment where the "document' may not exist or is not available.

```JS
if (strContains(ua, strMsie)) {
        let doc = getDocument();
        return Math.Max(parseInt(ua.split(strMsie)[1]), (doc ? (doc.documentMode || 0) : 0));
    } else if (strContains(ua, strTrident)) {
```

shared/AppInsightsCommon/Tests/Unit/src/Util.tests.ts
```diff
@@ -531,12 +531,11 @@ export class UtilTests extends AITestClass {
                 }
 
                 Assert.equal(7, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 7.0b; Windows NT 6.0)"));
-                Assert.equal(7, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.0; WOW64; Trident/4.0;)"));
                 Assert.equal(7, Util.getIEVersion("Mozilla/5.0 (compatible; MSIE 7.0; Windows NT 6.0; en-US)"));
+                Assert.equal(7, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.0; WOW64; Trident/4.0;)"));
                 Assert.equal(8, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 5.1; Trident/4.0)"));
                 Assert.equal(8, Util.getIEVersion("Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.1; Trident/4.0)"));
                 Assert.equal(8, Util.getIEVersion("Mozilla/5.0 (compatible; MSIE 8.0; Windows NT 6.0; Trident/4.0; WOW64; Trident/4.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729; .NET CLR 1.0.3705; .NET CLR 1.1.4322)"));
-                Assert.equal(8, Util.getIEVersion("Mozilla/4.0 (Compatible; MSIE 8.0; Windows NT 5.2; Trident/6.0)"));
```
MSNev, line 539: I know your also checking this one below with the document mode enabled, but can we please keep this one here as well

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -311,8 +312,10 @@ export function getIEVersion(userAgentStr: string = null): number {
     }
 
     var ua = (userAgentStr || "").toLowerCase();
+    // Also check for documentMode in case X-UA-Compatible meta tag was included in HTML.
     if (strContains(ua, strMsie)) {
-        return parseInt(ua.split(strMsie)[1]);
+        let doc = getDocument() || {} as Document;
```
MSNev, line 317: 👍

MSNev (approved):

MSNev (approved):

MSNev:
Created Bug #1735 for this so that the release notes can include the details on why this change is being added.

MSNev:
Looks good to me @cfeltner do you have any additional issues you want to address before we commit?

Note: while I've tagged this for the next release which is not scheduled until the end of Jan (early feb availability) once this is committed it will end up in the next nightly build/publish (which we have not documented yet as still working out a few kinks in the pipeline, but it is available via NPM and CDN versions). The nightly build/publish is not meant to be used for production but rather for early testing/validation.

cfeltner:
@MSNev, great I have no additional changes to this PR. Thanks for the info.

PR #1758, opened 2022-02-14
Additional Performance enhancements to use provided functions rather than internal polyfill's

MSNev:


shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
```diff
@@ -44,17 +109,17 @@ export function isNotNullOrUndefined(value: any): boolean {
 }
 
 export function hasOwnProperty(obj: any, prop: string): boolean {
-    return obj && ObjHasOwnProperty.call(obj, prop);
+    return !!(obj && ObjHasOwnProperty.call(obj, prop));
```
MSNev, line 112: The !! is to ensure that if undefined or null is passed the return is still a boolean value (false in both of those cases)

shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
```diff
@@ -116,9 +181,17 @@ export function detachEvent(obj: any, eventNameWithoutOn: string, handlerRef: an
  */
 export function normalizeJsName(name: string): string {
     let value = name;
-    let match = /([^\w\d_$])/g;
-    if (match.test(name)) {
-        value = name.replace(match, "_");
+
+    if (value && isString(value)) {
```
MSNev, line 185: Updated this as in a future PR I'm using this a lot more and the original version didn't quite matching the JavaScript specification.

Like a variable must not "start" with a digit, so while "0" is value "0a" is not.

shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
```diff
@@ -211,21 +310,21 @@ export function strContains(value: string, search: string) {
...
+    return !!(obj && objToString(obj) === "[object Date]");
 }
 
 /**
  * Check if an object is of type Array
  */
-export function isArray(obj: any): boolean {
-    return objToString(obj) === "[object Array]";
+export function isArray<T = any>(obj: any): obj is Array<T> {
+    return !!(_isArray ? _isArray(obj) : (obj && objToString(obj) === "[object Array]"));
```
MSNev, line 320: using !! as (obj && ...) could return null or undefined if passed.

shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
```diff
@@ -268,35 +367,71 @@ export function isSymbol(value: any): boolean {
     return typeof value === "symbol";
 }
 
+/**
+ * Checks if the type of the value is a normal plain object (not a null or data)
+ * @param value
+ */
+export function isPlainObject(value: any): boolean {
```
MSNev, line 374: Leveraged most of this logic from jQuery to specifically identify ONLY objects and not other objects (arrays etc) that return a typeof value === "object"

shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
```diff
@@ -268,35 +367,71 @@ export function isSymbol(value: any): boolean {
...
+}
+
+
 /**
  * Convert a date to I.S.O. format in IE8
  */
 export function toISOString(date: Date) {
     if (date) {
-        if (date[strToISOString]) {
+        if (_dataToISOString) {
```
MSNev: This can be crunched

Karlie-777 (approved):

PR #1765, opened 2022-02-18
Another fix for objKeys() should be Object.keys()

MSNev:
- Previous version fails if the object passed in contains a "keys" property or function.


shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
```diff
@@ -36,8 +35,9 @@ const strToString = "toString";
 const strConstructor = "constructor";
  
 const _objDefineProperty = ObjDefineProperty;
-const _objFreeze = ObjClass["freeze"];
-const _objSeal = ObjClass["seal"];
+const _objFreeze = ObjClass.freeze;
+const _objSeal = ObjClass.seal;
+const _objKeys = ObjClass.keys;
```
MSNev, line 40: These are all static functions and using ["name"] was 4 more characters than really required. Used that notation previously because of the version of typescript we used. (v2 from memory)

PR #1766, opened 2022-02-19
Beta Part 1: Part of Mega Dynamic Load/Unload support

MSNev:
- Refactor TelemetryPluginChain ready to start supporting load/unload
- Move TelemetryInitializer to BaseCore
- add getPlugin (will be used for remove/unload this plugin)


AISKU/Tests/Unit/src/AISKUSize.Tests.ts
```diff
@@ -2,7 +2,7 @@ import { AITestClass, Assert } from "@microsoft/ai-test-framework";
 import * as pako from "pako";
 
 export class AISKUSizeCheck extends AITestClass {
-    private readonly MAX_DEFLATE_SIZE = 40;
+    private readonly MAX_DEFLATE_SIZE = 41;
```
MSNev, line 5: This is going to creep up a bit until I've merged in all changes, then I'll spend some time reducing the total size again.

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IAppInsightsCore.ts
```diff
@@ -72,6 +75,13 @@ export interface IAppInsightsCore extends IPerfManagerProvider {
      */
     removeNotificationListener?(listener: INotificationListener): void;
 
+    /**
+     * Add a telemetry processor to decorate or drop telemetry events.
+     * @param telemetryInitializer - The Telemetry Initializer function
+     * @returns - A ITelemetryInitializerHandler to enable the initializer to be removed
+     */
+    addTelemetryInitializer(telemetryInitializer: TelemetryInitializerFunction): ITelemetryInitializerHandler | void;
```
MSNev, line 83: This is moving the telemetry initializer to the Core so all Sku's will have the telemetry initializer functionality not just the Web Analytics.

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IAppInsightsCore.ts
```diff
@@ -80,4 +90,10 @@ export interface IAppInsightsCore extends IPerfManagerProvider {
      * Return a new instance of the IProcessTelemetryContext for processing events
      */
     getProcessTelContext() : IProcessTelemetryContext;
+
+    /**
+     * Find and return the (first) plugin with the specified identifier if present
+     * @param pluginIdentifier
+     */
+    getPlugin<T extends IPlugin = IPlugin>(pluginIdentifier: string): ILoadedPlugin<T>;
```
MSNev, line 98: Creating a unified "lookup" for a plugin, this happens throughout the code by lookup at the passed config. But as we are going to be updating the loaded extensions that's going to get problematic in later PR's

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IProcessTelemetryContext.ts
```diff
@@ -27,12 +33,12 @@ export interface IProcessTelemetryContext {
...
      * Gets the current core config instance
      */
-    getCfg: ()=> IConfiguration;
+    getCfg: () => IConfiguration;
 
     /**
      * Gets the named extension config
      */
-    getExtCfg: <T>(identifier: string, defaultValue?:T|any) => T;
+    getExtCfg: <T>(identifier: string, defaultValue?: T | any, mergeDefault?: GetExtCfgMergeType) => T;
```
MSNev, line 41: Moving some code that was repeated across the code bases into the central location.

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IProcessTelemetryContext.ts
```diff
@@ -41,7 +47,7 @@ export interface IProcessTelemetryContext {
      * @param field The config field name
      * @param defaultValue The default value to return if no defined config exists
      */
-    getConfig: (identifier: string, field: string, defaultValue?: number | string | boolean) => number | string | boolean;
+    getConfig: (identifier: string, field: string, defaultValue?: number | string | boolean | string[] | RegExp[] | Function) => number | string | boolean | string[] | RegExp[] | Function;
```
MSNev, line 50: Adding some more of the actual types so that users don't need to say "any" or "unknown"

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IProcessTelemetryContext.ts
```diff
@@ -73,4 +86,9 @@ export interface IProcessTelemetryContext {
      *                  order then the next plugin will be NOT set.
      */
     createNew: (plugins?:IPlugin[]|ITelemetryPluginChain, startAt?:IPlugin) => IProcessTelemetryContext;
+
+    /**
+     * Set the function to call when the current chain has executed all processNext or unloadNext items.
+     */
+    onComplete: (onComplete: () => void) => void;
```
MSNev, line 93: Adding a "callback" that will be called once all of the plugins have been processed.

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -5,25 +5,20 @@ import { BaseCore } from "./BaseCore";
...
 
 "use strict";
 
 export class AppInsightsCore extends BaseCore implements IAppInsightsCore {
     constructor() {
         super();
-        /**
-         * Internal log poller
-         */
-         let _internalLogPoller: number = 0;
```
MSNev, line 26: Moving to BaseCore.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -231,13 +247,13 @@ export class BaseCore implements IAppInsightsCore {
             };
 
             _self.getPerfMgr = (): IPerfManager => {
-                if (!_perfManager) {
+                if (!_perfManager && !_cfgPerfManager) {
```
MSNev, line 250: Will now support the concept of "config" based perf Manager and manually set perf Manager, as once we start allowing unload / re-initialize and config updates we need to track and keep any manually set one.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -249,14 +265,182 @@ export class BaseCore implements IAppInsightsCore {
...
+            _self.stopPollingInternalLogs = (): void => {
+                if(_internalLogPoller) {
+                    clearInterval(_internalLogPoller);
+                    _internalLogPoller = 0;
+                    _flushInternalLogs();
+                }
+            }
+
+            // Add addTelemetryInitializer
+            proxyFunctions(_self, () => _telemetryInitializerPlugin, [ "addTelemetryInitializer" ]);
```
MSNev, line 310: This effectively "adds" the `addTelemetryInitializer` function to this instance and as we are within the `dynamicProto` section it will also become a prototype level function.

This helper `proxyFunctions` avoids a lot of the boilerplate code needed to "forward" the function call and arguments using something like

```ts
_self.addTelemetryInitializer = function(telemetryInitializer: TelemetryInitializerFunction) {
    _telemetryInitializerPlugin.addTelemetryInitializer(telemetryInitializer);
}
```

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -249,14 +265,182 @@ export class BaseCore implements IAppInsightsCore {
...
+
+            _self.getPlugin = _getPlugin;
+
+            function _initDefaults() {
```
MSNev, line 314: Just setting up a pattern so that during initial construction and after getting "unloaded" this function will be called this is to avoid having two places in the code and the potential side effects (bugs) thie might cause.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -249,14 +265,182 @@ export class BaseCore implements IAppInsightsCore {
...
+                _coreExtensions = null;
+                _configExtensions = null;
+                _channelControl = null;
+                _channelConfig = null;
+                _channelQueue = null;
+                _internalLogsEventName = null;
+            }
+
+            // Initialize or Re-initialize the plugins
+            function _initPluginChain(config: IConfiguration) {
```
MSNev, line 338: One of the big changes in this PR is that the PluginChain can now be created once and reused for every event, previously a new context and chain were created for every event. - That was a long standing performance bug.

The intent of the "context" is so that there is one for each event so as each event is processed through the system it can handle both the old (prior to the context `setNextPlugin()` only plugins) and the newer `ctx.processNext()` so that the SDK can keep track of and update any required context.
Such as the upcoming the order of execution completely changing during async (add/remove plugins) operations in the same running plugin and SDK instance.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -249,14 +265,182 @@ export class BaseCore implements IAppInsightsCore {
...
+            
+                _coreExtensions = theExtensions.core;
+                _pluginChain = null;
+            
+                // Sort the complete set of extensions by priority
+                let allExtensions = theExtensions.all;
+
+                // Initialize the Channel Queues and the channel plugins first
+                _channelQueue = objFreeze(createChannelQueues(_channelConfig, allExtensions, config, _self));
+                _channelControl = createChannelControllerPlugin(_channelQueue, _self);
```
MSNev, line 350: Removing the `ChannelController` class, but keeping the functionality the same by using an interface `IChannelController` definition, which is a lot smaller for minification.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -249,14 +265,182 @@ export class BaseCore implements IAppInsightsCore {
...
+                // Required to allow plugins to call core.getPlugin() during their own initialization
+                _self._extensions = sortPlugins(allExtensions);
+
+                // Initialize the controls
+                _channelControl.initialize(config, _self, allExtensions);
+                
+                initializePlugins(_self.getProcessTelContext(), allExtensions);
+
+                // Now reset the extensions to just those being managed by Basecore
+                _self._extensions = objFreeze(sortPlugins(_coreExtensions || [])).slice();
```
MSNev, line 365: Now freezing the internal extensions to make it immutable, so that during load/unload coding I make sure that when there is an order change internal values are updated correctly.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -249,14 +265,182 @@ export class BaseCore implements IAppInsightsCore {
...
+                arrForEach(_self._extensions, (ext: any) => {
+                    if (ext.identifier === pluginIdentifier) {
+                        thePlugin = ext;
+                        return -1;
+                    }
+                });
+
+                if (thePlugin) {
+                    theExt = {
+                        plugin: thePlugin as T
```
MSNev, line 381: This is where I'll be adding extra functions like `remove` so only this returned plugin is affected.

I'm taking this approach vs adding extra "removeFromCore" on the plugin as it then becomes "messy" for the when plugin needs to implement it's own `unload` or `teardown` functionality as this way it doesn't have to tell the core that it's no longer active because it will be the core telling the plugin to unload.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -249,14 +265,182 @@ export class BaseCore implements IAppInsightsCore {
...
+                    }
+                }
+
+                return theExt;
+            }
+
+            function _getPluginChain() {
+                if (!_pluginChain) {
```
MSNev, line 389: Just a helper, to lazily create the current chain to avoid initializing a new one every time something changes even if now events are processed.

- Think adding or removing multiple plugins on the fly.

shared/AppInsightsCore/src/JavaScriptSDK/BaseCore.ts
```diff
@@ -249,14 +265,182 @@ export class BaseCore implements IAppInsightsCore {
...
+                            baseData: { message: logMessage.message }
+                        };
+                        _self.track(item);
+                    });
+
+                    queue.length = 0;
+                }
+            }
+
+            function _initDebugListener(config: IConfiguration) {
```
MSNev, line 419: Just FYI - the DebugListener is used by the chrome extension and debug plugins to listen to internal debug events. So we can publish critical issues etc.

shared/AppInsightsCore/src/JavaScriptSDK/BaseTelemetryPlugin.ts
```diff
@@ -3,16 +3,18 @@
...
 import { IAppInsightsCore } from "../JavaScriptSDK.Interfaces/IAppInsightsCore"
 import { IConfiguration } from "../JavaScriptSDK.Interfaces/IConfiguration";
 import { IDiagnosticLogger } from "../JavaScriptSDK.Interfaces/IDiagnosticLogger";
 import { IPlugin, ITelemetryPlugin } from "../JavaScriptSDK.Interfaces/ITelemetryPlugin";
 import { ITelemetryItem } from "../JavaScriptSDK.Interfaces/ITelemetryItem";
 import { IProcessTelemetryContext } from "../JavaScriptSDK.Interfaces/IProcessTelemetryContext";
 import { ITelemetryPluginChain } from "../JavaScriptSDK.Interfaces/ITelemetryPluginChain";
-import { ProcessTelemetryContext } from "./ProcessTelemetryContext";
-import { isFunction, isNullOrUndefined, setValue } from "./HelperFuncs";
+import { createProcessTelemetryContext } from "./ProcessTelemetryContext";
```
MSNev, line 14: While the `ProcessTelemetryContext` class still exists (for backward compatibility only), changing to no longer directly use it for minification. The class itself uses the `createProcessTelemetryContext` to generate it's actual usage

shared/AppInsightsCore/src/JavaScriptSDK/BaseTelemetryPlugin.ts
```diff
@@ -77,15 +79,38 @@ export abstract class BaseTelemetryPlugin implements ITelemetryPlugin {
...
+        _initDefaults();
+
+        dynamicProto(BaseTelemetryPlugin, _self, (_self) => {
+
+            _self.initialize = (config: IConfiguration, core: IAppInsightsCore, extensions: IPlugin[], pluginChain?:ITelemetryPluginChain): void => {
+                _setDefaults(config, core, pluginChain);
+                _isinitialized = true;
+            }
+        
+            _self._addHook = (hooks: IInstrumentHook | IInstrumentHook[]) => {
```
MSNev, line 99: This will be used to "remember" hooks that have been added for the plugin, so when it's unloaded the hooks will be automatically removed, so each plugin doesn't have to remember them itself.

shared/AppInsightsCore/src/JavaScriptSDK/ChannelController.ts
```diff
@@ -1,135 +1,177 @@
...
-const ChannelValidationMessage = "Channel has invalid priority";
+export const ChannelControllerPriority = 500;
+const ChannelValidationMessage = "Channel has invalid priority - ";
+
+export interface _IInternalChannels {
+    queue: IChannelControls[];
+    chain: ITelemetryPluginChain;
+}
 
-export class ChannelController extends BaseTelemetryPlugin {
```
MSNev, line 22: Completely removing the class as it's internal only, so no need to keep it as it should never be exposed.

shared/AppInsightsCore/src/JavaScriptSDK/ChannelController.ts
```diff
@@ -1,135 +1,177 @@
...
-                let extensionQueue:IChannelControls[] = [];
-                arrForEach(extensions as IChannelControls[], plugin => {
-                    if (plugin.priority > ChannelControllerPriority) {
-                        extensionQueue.push(plugin);
+
+            _processChannelQueue(null, (chainCtx: IProcessTelemetryContext) => {
+                chainCtx.iterate<IChannelControls>((plugin) => {
+                    if (plugin.flush) {
+                        waiting ++;
+                        plugin.flush(isAsync, () => {
```
MSNev, line 140: Investigation note: What to do if a channel doesn't take callback function or call it!!!

shared/AppInsightsCore/src/JavaScriptSDK/CoreUtils.ts
```diff
@@ -182,33 +182,6 @@ export function perfNow(): number {
     return dateNow();
 }
 
-/**
- * Generate random base64 id string.
- * The default length is 22 which is 132-bits so almost the same as a GUID but as base64 (the previous default was 5)
- * @param maxLength - Optional value to specify the length of the id to be generated, defaults to 22
- */
-export function newId(maxLength = 22): string {
```
MSNev, line 190: Moved to RandomHelper, to avoid some circular dependencies that are going to crop up later.

shared/AppInsightsCore/src/JavaScriptSDK/RandomHelper.ts
```diff
@@ -52,12 +52,14 @@ export function randomValue(maxValue: number) {
  * @param signed - True to return a signed 32-bit number (-0x80000000..0x7FFFFFFF) otherwise an unsigned one (0x000000..0xFFFFFFFF)
  */
 export function random32(signed?: boolean) {
-    let value;
+    let value = 0;
```
MSNev, line 55: Update to handle edge cases where the built in random functions can return zero when excessively called. So in this case this will fall back to the none hardware methods.

shared/AppInsightsCore/src/JavaScriptSDK/TelemetryInitializerPlugin.ts
```diff
@@ -0,0 +1,119 @@
...
+                        }
+                    }
+                }
+
+                if (!doNotSendItem) {
+                    _self.processNext(item, itemCtx);
+                }
+            };
+
+            _self.unload = (itemCtx: IProcessTelemetryContext, isAsync: boolean): void => {
```
MSNev, line 84: Ok, this is the signature of the new `unload` function that each plugin will be able to implement.
I thought I'd removed all of them from this PR (in an attempt to keep this PR smaller).

shared/AppInsightsCore/src/JavaScriptSDK/ChannelController.ts
MSNev: Will need to re-work this as the default "Sender" for AI actually doesn't handle any of the arguments for `flush()` so will probably need to add a timeout or just return once all of the channels have had their `flush()` calls triggered...

MSNev: Separating the ChannelController to be a more specific IChannelControls instance so that it can be passed a timeout, also changing the existing `flush` so the functions can return an optional boolean to indicate that they "will" call the callback. This way if the existing flush returns nothing or false the controller will start a timer (defaulting to 5 seconds) to "wait" for all channels to finish flushing before calling any callback. This is SUPER important for a later PR that will correctly handle unloading the entire SDK.

hectorhdzg (approved):
LGTM, it would be good discuss all changes into detail once all parts are ready

PR #1768, opened 2022-02-26
Beta Part 2: Part of Mega Dynamic Load/Unload support

MSNev:
- Add Event Namespace support
- Minification of constant values
- Add part of the unload functionality (required for unified `teardown()` functionality)


shared/AppInsightsCore/src/JavaScriptSDK/InternalConstants.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// Note: DON'T Export these const from the package as we are still targeting ES3 this will export a mutable variables that someone could change!!!
+
+export const strEmpty = "";
```
ramthi, line 6: NIT: how about naming these like so?  EMPTY, .... REMOVE_NOTIFICATION_LISTENER, etc..

PR #1771, opened 2022-02-28
remove article link in the ReadMe

Karlie-777:


README.md
```diff
@@ -156,7 +156,7 @@ appInsights.trackPageView();
 
 #### (Alternative Setup Method) Include AI JS SDK script and initialize statically
 
-Use this approach if you would like to host AI JS SDK script on your endpoint or bundle it with other scripts. One popular example is Cordova applications (see [this blog post](http://www.teamfoundation.co.za/2016/02/application-insights-and-typescript/). After JS script has loaded, include the following snippet to initialize Application Insights:
+Use this approach if you would like to host AI JS SDK script on your endpoint or bundle it with other scripts. After JS script has loaded, include the following snippet to initialize Application Insights:
```
MSNev, line 159: I think the "After JS script has loaded," doesn't seem right as line 162 is actually loading the SDK 😄 so even the existing comment on line 161 is wrong (IMHO anyway).

Maybe also update the fragment to highlight that line 162 is "loading" the SDK from the CDN and then a new comment (between the script tags) highlighting that bit that will "initialize the SDK" (after it's loaded)?

MSNev (approved):

PR #1786, opened 2022-03-18
Beta: Minification improvements

MSNev:
- Remove Generated classes
- And Convert to interfaces rather than classes


MSNev:
Moving this to `master` from `beta`

PR #1791, opened 2022-03-30
Merge remote-tracking branch 'upstream/beta' into MSNev/MergeBetaToMaster

MSNev:


AISKU/Tests/Unit/src/ApplicationInsightsDeprecatedTests.ts
```diff
@@ -19,7 +19,11 @@ export class ApplicationInsightsDeprecatedTests extends AITestClass {
     
     constructor() {
         super("ApplicationInsightsDeprecatedTests");
+
+        // this.assertNoEvents = false;
```
MSNev, line 23: This is a left over from before fixing all cleanup paths -- probably should cleanup before release (proposing not this PR)

AISKU/Tests/Unit/src/AISKUSize.Tests.ts
```diff
@@ -1,14 +1,33 @@
 import { AITestClass, Assert } from "@microsoft/ai-test-framework";
+import { strUndefined } from "@microsoft/applicationinsights-core-js";
 import * as pako from "pako";
+import { Snippet } from "../../../src/Initialization";
 
 export class AISKUSizeCheck extends AITestClass {
-    private readonly MAX_DEFLATE_SIZE = 40;
+    private readonly MAX_DEFLATE_SIZE = 44;
```
ramthi, line 7: This I guess would go down after minification work!  Should we create a task and reference here?

MSNev: Yes, that is the plan -- still working on these so I don't have final numbers yet. But I plan to reduce once all of the minifications are checked in.

It does appear that Core will be larger as that is where most of the changes are located, but as I'm not using namespaces tree-shaking will remove anything unused.

ramthi (approved):
This is mostly code that has already been PRd.  I scanned for anything obvious.  We should merge, test and then go from there.  Let's chat about if there is any stabilization phase we can create for release from the main branch.

PR #1793, opened 2022-03-31
Master Minification Improvements

MSNev:
- Remove Generated classes
- And Convert to interfaces rather than classes


hectorhdzg (approved):
LGTM

PR #1803, opened 2022-04-13
Add Automatic Name replacing / Crunching

MSNev:


AISKU/src/ApplicationInsightsDeprecated.ts
```diff
@@ -1,11 +1,11 @@
 import {
```
MSNev, line 1: Because the Minify script automatically "adds" the name replacements before compiling and then removes them after compiling it also "updates" the imports of every file by

- Removed all unused imports (unless it contains comments -- it leaves these alone)
- Sorts the imported values, with "default" imports `import xxx from yyy;` at the top
- Sorts the imports with all "@" imports at the top

AISKU/src/InternalConstants.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
MSNev, line 1: Any strings added to the "InternalConstants.ts" are automatically used to scan for and replace any possible references

channels/applicationinsights-channel-js/Tests/Unit/src/Sender.tests.ts
```diff
@@ -1280,7 +1280,6 @@ export class SenderTests extends AITestClass {
                 let appInsightsCore = new AppInsightsCore();
                 appInsightsCore.logger = new DiagnosticLogger();
                 let messageId: _InternalMessageId = _InternalMessageId.InvalidInstrumentationKey;
-                let logInternalSpy = this.sandbox.spy(appInsightsCore.logger, 'logInternalMessage');
```
MSNev, line 1283: Removed as the internal workings of the logger no longer calls the "public" method (part of the optimization)

channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
```diff
@@ -263,7 +259,6 @@ export function PageViewEnvelopeCreator(logger: IDiagnosticLogger, telemetryItem
     EnvelopeCreatorInit(logger, telemetryItem);
 
     // Since duration is not part of the domain properties in Common Schema, extract it from part C
-    let strDuration = "duration";
```
MSNev, line 266: Moved to InternalConstants.ts, so all usages in the project will use this value.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -45,6 +41,9 @@ function _getResponseText(xhr: XMLHttpRequest | IXDomainRequest) {
 }
 
 function _getDefaultAppInsightsChannelConfig(): ISenderConfig {
+    let defaultValue: string;
```
MSNev, line 44: Because the "names" get replaced as part of minification this identified that the `undefined` was implicitly being cast to `any`, so to get around this we need to define the actual types.

channels/applicationinsights-channel-js/src/TelemetryProcessors/Sample.ts
```diff
@@ -16,8 +16,8 @@ export class Sample implements ISample {
         let _logger = logger || safeGetLogger(null);
         
         if (sampleRate > 100 || sampleRate < 0) {
-            _logger.throwInternal(LoggingSeverity.WARNING,
-                _InternalMessageId.SampleRateOutOfRange,
+            _logger.throwInternal(eLoggingSeverity.WARNING,
+                _eInternalMessageId.SampleRateOutOfRange,
```
MSNev, line 20: Changing all references from the full name to the const enum which is replaced with the numeric value during compile.

common/config/rush/command-line.json
```diff
@@ -11,6 +11,16 @@
             "ignoreMissingScript": false,
             "allowWarningsInSuccessfulBuild": true
         },
+        {
+            "commandKind": "bulk",
+            "name": "mintest",
```
MSNev, line 16: Adding a new test pass which runs the tests agains the minified version of the core code.

common/config/rush/command-line.json
```diff
@@ -41,6 +51,26 @@
             "summary": "Run all build for all packages",
             "name": "rebuild",
             "allowWarningsInSuccessfulBuild": true
-        }  
+        },
+        {
+            "commandKind": "bulk",
+            "name": "ai-min",
```
MSNev, line 57: adding new targets `rush ai-min` and `rush ai-unmin` runs the minification and unminification for all files.

This is necessary for
- When the compile fails you may need to manually run the `unmin` to revert the file back to the original source
- Validating that running multiple times in either direction doesn't corrupt the file.

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewManager.ts
```diff
@@ -166,13 +166,13 @@ export class PageViewManager {
                             if (!pageViewPerformance.isValid && !pageViewSent) {
                                 // If navigation timing gives invalid numbers, then go back to "override page view duration" mode.
                                 // That's the best value we can get that makes sense.
-                                customProperties["duration"] = customDuration;
+                                customProperties.duration = customDuration;
```
MSNev, line 169: This line is now converted automatically to `customProperties[strDuration] = customDuration;` because the `strDuration` is present in the `InternalConstants.ts`

gruntfile.js
```diff
@@ -1,4 +1,39 @@
 module.exports = function (grunt) {
+
+    const aiCoreDefaultNameReplacements = [
```
MSNev, line 3: These files are read and processed dynamically, so all of the values listed in the source code cause the automatic replacements of any possible values.

The difference between `core` and the rest is that the `core` has the main exported definitions and therefore needs to import from the local file, while the rest import from the module.

gruntfile.js
```diff
@@ -1,4 +1,39 @@
...
+        {
+            src: "../shared/AppInsightsCore/src/JavaScriptSDK.Enums/SdkCoreNames.ts",
+            import: "@microsoft/applicationinsights-core-js"
+        },
+        {
+            src: "../shared/AppInsightsCore/src/JavaScriptSDK.Enums/ConfigEnums.ts",
+            import: "@microsoft/applicationinsights-core-js"
+        },
+        {
+            src: "../../shared/AppInsightsCore/src/JavaScriptSDK.Enums/SdkCoreNames.ts",
```
MSNev, line 24: The different "depths" of the references are due to the project layout so

`./AISKU` only needs to use `../shared/...` to get to the Core source code while
`./extensions/xxxx` needs to use `../../shared/...` to get there.

gruntfile.js
```diff
@@ -1,4 +1,39 @@
...
+            import: "@microsoft/applicationinsights-core-js"
+        },
+        {
+            src: "../../shared/AppInsightsCore/src/JavaScriptSDK.Enums/ConfigEnums.ts",
+            import: "@microsoft/applicationinsights-core-js"
+        }
+    ];
+
+    const aiInternalConstants = [
+        "./src/InternalConstants.ts"
```
MSNev, line 34: The location within each project (module) where to find the file(s) that identify constant strings to be dynamically read.

gruntfile.js
```diff
@@ -36,13 +76,74 @@ module.exports = function (grunt) {
...
 
         for (var key in modules) {
             if (modules.hasOwnProperty(key)) {
                 var modulePath = modules[key].path;
                 var moduleCfg = modules[key].cfg;
+                var packageJsonFile = modulePath + '/package.json';
+
+                if (grunt.file.exists(packageJsonFile)) {
+                    // Read the actual module version from the package.json
+                    var pkg = grunt.file.readJSON(modulePath + '/package.json');
```
MSNev, line 110: Just adding the minify tasks, referencing the package.json as the utility packages we don't want or need to convert them.

gruntfile.js
```diff
@@ -63,15 +164,89 @@ module.exports = function (grunt) {
...
-                //     };
-                // }
+                var addQunit = false;
+                var testRoot = "" 
+                if (modules[key].testHttp !== false) {
+                    testRoot = "http://localhost:9001/" 
+                }
+
+                var testUrl = testRoot + modulePath + "/test/UnitTests.html";
+                if (grunt.file.exists(modulePath + '/test/tsconfig.json')) {
```
MSNev, line 174: Refactoring the tests grunt definition (already done this for internal variant).

gruntfile.js
```diff
@@ -170,153 +381,93 @@ module.exports = function (grunt) {
...
             var actions = [
                 "eslint-ts:" + name + "-lint-fix"
             ];
     
+            var aiMinifyConfig = theBuildConfig["ai-minify"] || {};
             var gruntTsConfig = theBuildConfig["ts"];
             if (replaceName) {
+
                 actions.push("string-replace:" + replaceName);
+                if (aiMinifyConfig[name]) {
```
MSNev, line 404: When adding the minification task, it will actually "compile" the source file twice with and without minification to check that the code and references are valid and not broken before or after

MSNev:
Closing for now as this work is being moved to a different release and possible branch.

Issue #1807, opened 2022-04-19
[BUG]  Angular project doesn't build after install latest version  v.2.8.0


cmonzon44:
Angular project doesn't build after install latest version  v.2.8.0

![image](https://user-images.githubusercontent.com/11430219/164023959-1de266ec-95b6-41ea-8f72-dd346a256a02.png)

ERROR in node_modules/@microsoft/applicationinsights-core-js/types/JavaScriptSDK.Enums/EnumHelperFuncs.d.ts:33:30 - error TS1005: ',' expected.

     [key in keyof E]: [eValue: E[keyof E], mapValue: V[keyof V]];

I should to set version 2.7.4 to build correctly again.

MSNev:
Yes, we will prepare a release which uses 2.8.0

javier-caparo:
Same here for us, using "typescript: ~3.8.3" in our package.json , which calls this dependency: "Microsoft Application Insights JavaScript SDK - Core" ( @microsoft/applicationinsights-core-js , version : 2.8.0)

-----
ERROR in node_modules/@microsoft/applicationinsights-core-js/types/JavaScriptSDK.Enums/EnumHelperFuncs.d.ts:33:30 - error TS1005: ',' expected.

33     [key in keyof E]: [eValue: E[keyof E], mapValue: V[keyof V]];

                           ~

node_modules/@microsoft/applicationinsights-core-js/types/JavaScriptSDK.Enums/EnumHelperFuncs.d.ts:33:52 - error TS1005: ',' expected.

33     [key in keyof E]: [eValue: E[keyof E], mapValue: V[keyof V]];

-----

Workaround:

a) we created the package-lock.json and then modifying the dependency version of @microsoft/applicationinsights-core-js to 2.5.10.

and then execute 
b) npm ci 

c) npm build

it works!!

MSNev:
v2.8.1 of the AngularPlugin which uses AI v2.7.4 is now published
https://www.npmjs.com/package/@microsoft/applicationinsights-angularplugin-js/v/2.8.1

We plan to also release a v2.9.0 (which uses AI v2.8.0 shortly as well)

MSNev:
@javier-caparo Can you upgrade to TypeScript 4.x? Or do you need to stay on TS 3.x?

MSNev:
[...]

PR #1810, opened 2022-04-19
v2.8.0 has incompatible TypeScript 3.x type declaration

MSNev:
- [BUG] Angular project doesn't build after install latest version v.2.8.0 #1807

### DIFF
--- a/shared/AppInsightsCore/src/JavaScriptSDK.Enums/EnumHelperFuncs.ts
+++ b/shared/AppInsightsCore/src/JavaScriptSDK.Enums/EnumHelperFuncs.ts
@@ -45,7 +45,7 @@ export function createEnumMap<E, I = keyof E>(values: { [key in keyof E]: E[keyo
  * @param values - The values to populate on the new object
  * @returns
  */
-export function createValueMap<E, V = E>(values: { [key in keyof E]: [ eValue: E[keyof E], mapValue: V[keyof V]] }): V {
+export function createValueMap<E, V = E>(values: { [key in keyof E]: [ E[keyof E], V[keyof V] ] }): V {
     let mapClass: any = {};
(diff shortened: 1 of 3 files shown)


shared/AppInsightsCore/src/JavaScriptSDK.Enums/EnumHelperFuncs.ts
```diff
@@ -45,7 +45,7 @@ export function createEnumMap<E, I = keyof E>(values: { [key in keyof E]: E[keyo
  * @param values - The values to populate on the new object
  * @returns
  */
-export function createValueMap<E, V = E>(values: { [key in keyof E]: [ eValue: E[keyof E], mapValue: V[keyof V]] }): V {
+export function createValueMap<E, V = E>(values: { [key in keyof E]: [ E[keyof E], V[keyof V] ] }): V {
```
MSNev, line 48: I Rolled back to TS 3.8.3, reworked the definition to something that works for both TS 3.x and TS 4.x

tools/chrome-debug-extension/src/telemetryViewerPopup.tsx
```diff
@@ -113,7 +113,7 @@ export const TelemetryViewerPopup = (): React.ReactElement => {
     React.useEffect(() => {
         versionCheck();
 
-        let configurationTypeToSet = undefined;
+        let configurationTypeToSet: ConfigurationType = undefined;
```
MSNev, line 116: TS 3.x actually complained about this and set the type of this variable to string | undefined -- which was wrong.
While this doesn't affect anyone consumers I fixed anyway so that i could validate that the entire build works.

Karlie-777 (approved):

Issue #1812, opened 2022-04-21
[BUG] Browser exceptions are no longer automatically tracked after 2.8.0


WoodyWoodsta:
**Description/Screenshot**
Using `2.8.0`, browser exceptions are no longer automatically caught and tracked. No error handler is present on the `window`.

**Steps to Reproduce**
1. Install `2.8.0` and throw an exception in the browser. Check Application Insights failures panel for errors. No exceptions are presented. No exceptions are added to the `/track` network request payload.
2. Install `2.7.4` and repeat. Exceptions are tracked, and the handler is present on the `window`.

 - OS/Browser: MacOS/Chrome
 - SDK Version: `2.8.0`
 - How you initialized the SDK:
```ts
export const appInsights = new ApplicationInsights({
  config: {
    instrumentationKey,
    extensions: [reactPlugin],
    enableDebug,
    loggingLevelConsole: enableDebug ? 2 : 0,
    loggingLevelTelemetry: enableDebug ? 2 : 1,
    autoTrackPageVisitTime: true,
    enableAutoRouteTracking: true,
    enableDebugExceptions: enableDebug,
    enableUnhandledPromiseRejectionTracking: true,
    enableCorsCorrelation: false,
    enableRequestHeaderTracking: true,
    enableResponseHeaderTracking: true,
  },
});
```

`enableDebug=true`

**Expected behavior**

As above, the exceptions thrown in the browser are tracked - the handler is present on the `window`.

MSNev:
OK, the issue seems to be the way we are now hooking the window.onerror.
With a local repro I am getting the error events.

I'll restore to the "standard" method of hooking the onError rather than the monkey patching that 2.8.0 is currently using (this was done to support unloading)

MSNev:
Actually it looks a little more nuanced as the hooking mechanism does assign the entry -- but, we also tag the endpoint that this is hooked -- so I suspect that part is causing the hook to fail *sigh*.

WoodyWoodsta:
@MSNev Thank you for raising #1815 to fix this issue :)

MSNev:
v2.8.1 is now fully deployed

MSNev:
Rolled back CDN to v2.7.4 due to #1822

MSNev:
On investigating #1822, while the fix for this issue doesn't throw anymore it also appears that it's still not correctly hooking the onerror function either

WoodyWoodsta:
[...]

PR #1815, opened 2022-04-21
[BUG] Browser exceptions are no longer automatically tracked after 2.8.0 #1812

MSNev:


shared/AppInsightsCore/src/JavaScriptSDK/InstrumentHooks.ts
```diff
@@ -109,15 +110,17 @@ function _createFunctionHook(aiHook:IInstrumentHooks) {
...
-        try {
-            funcArgs.rslt = theFunc.apply(funcThis, orgArgs);
-        } catch (err) {
-            // Report the request callback
-            funcArgs.err = err;
-            _doCallbacks(hooks, funcArgs, cbArgs, hookCtx, CallbackType.FunctionError);
-
-            // rethrow the original exception so anyone listening for it can catch the exception
-            throw err;
+        if (theFunc) {
```
MSNev, line 113: This was 1 issue in that there was no previous "event" to call, so `theFunc.apply(...)` would throw and call the error handling hook and not the processing hook

shared/AppInsightsCore/src/JavaScriptSDK/InstrumentHooks.ts
```diff
@@ -183,44 +227,7 @@ export function InstrumentFunc(target:any, funcName:string, callbacks: IInstrume
         if (owner) {
             let fn = owner[funcName]
             if (typeof fn === strShimFunction) {
```
MSNev, line 229: This was another issue, in that not all environments would return the "window.onerror" (and other events) as functions, they get returned as Object. Which lead to the hook never getting installed

shared/AppInsightsCore/src/JavaScriptSDK/InstrumentHooks.ts
```diff
@@ -250,3 +257,23 @@ export function InstrumentFuncs(target:any, funcNames:string[], callbacks: IInst
...
+
+/**
+ * Add an instrumentation hook to the provided named "event" for the target class / object, this doesn't check whether the
+ * named "event" is in fact a function and just assigns the instrumentation hook to the target[evtName]
+ * @param target - The target object
+ * @param evtName - The name of the event
+ * @param callbacks - The callbacks to configure and call whenever the function is called
+ * @param checkPrototype - If the function doesn't exist on the target should it attempt to hook the prototype function
+ */
+export function InstrumentEvent(target: any, evtName: string, callbacks: IInstrumentHooksCallbacks, checkPrototype:boolean = true): IInstrumentHook {
```
MSNev, line 269: Added this "new" function which doesn't "check" if the target is a function and instead just applies the hook to the named "property" of the "target" (which is called from the SDK for events).

hectorhdzg (approved):

PR #1817, opened 2022-04-22
deprecate ikey

Karlie-777:


shared/AppInsightsCommon/src/TelemetryItemCreator.ts
```diff
@@ -35,6 +35,11 @@ export function createTelemetryItem<T>(item: T,
     if (item[strIkey]) {
         iKey = item[strIkey];
         delete item[strIkey];
+        //TODO: add connectionString Config
+        var _console = typeof console !== strUndefined ? console : null;
```
MSNev, line 39: Same comment as above -- this will always occur as it will always be populated -- even when using a connection string.

AISKU/src/Initialization.ts
```diff
@@ -179,6 +179,11 @@ export class Initialization implements IApplicationInsights {
                 const ingest = cs.ingestionendpoint;
                 config.endpointUrl = ingest ? `${ingest}/v2/track` : config.endpointUrl; // only add /v2/track when from connectionstring
                 config.instrumentationKey = cs.instrumentationkey || config.instrumentationKey;
+            } else {
+                var _console = typeof console !== strUndefined ? console : null;
```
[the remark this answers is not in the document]
Karlie-777: so the check condition is `if (config.connectionString){....} else {warn("....")}`, I thought the warning message should be populated when `config.connectionString` is not provided.😊

shared/AppInsightsCommon/src/TelemetryItemCreator.ts
Karlie-777: so it is in the `createTelemetryItem` function, I thought `ikey` here should be overwritten ikey instead of the one from configuration.
```
  let iKey = "";
    if (item[strIkey]) {
        iKey = item[strIkey];
        delete item[strIkey];
        //TODO: add connectionString Config
        var _console = typeof console !== strUndefined ? console : null;
        if (_console && _console.warn) {
            _console.warn("Instrumentation key based ingestion will be no longer supported.");
        }
    }

```

AISKU/src/Initialization.ts
```diff
@@ -179,6 +179,11 @@ export class Initialization implements IApplicationInsights {
...
+                if (_console && _console.warn) {
+                    _console.warn("Instrumentation key based ingestion will be no longer supported, please use connection string instead.");
```
ramthi, line 185: Make sure we either include a aka link (go.microsoft.com) pointing to the deprecation notice for ikeys.  I think this is one potential document we can refer to: https://docs.microsoft.com/en-us/azure/azure-monitor/app/migrate-from-instrumentation-keys-to-connection-strings

In any case " no longer supported" is not accurate.  We stop supporting it sometime in future.

AISKU/src/Initialization.ts
```diff
@@ -189,6 +189,14 @@ export class Initialization implements IApplicationInsights {
             _core = new AppInsightsCore();
             _self.core = _core;
 
+        
+            if (!config.connectionString) {
+                _throwInternal(_core.logger,
+                    eLoggingSeverity.CRITICAL,
+                    _eInternalMessageId.InstrumentationKeyDeprecation,
+                    "Instrumentation key based ingestion support will end, please use connection string instead. More information at: https://docs.microsoft.com/en-us/azure/azure-monitor/app/migrate-from-instrumentation-keys-to-connection-strings");
```
MSNev, line 197: Can we create an aka.ms alias for this as it will be a LOT shorter as strings can't be minified, so this is a lot of extra text.

Mabe something like
"Instrumentation key support will end soon, see: aka.ms/xxxxxx"

MSNev (approved):

AISKU/src/Initialization.ts
```diff
@@ -189,6 +189,14 @@ export class Initialization implements IApplicationInsights {
             _core = new AppInsightsCore();
             _self.core = _core;
 
+        
+            if (!config.connectionString) {
```
MSNev, line 193: We talked about also adding a config to disable this (for large customers who don't want to be annoyed by this message)

Karlie-777: I will create another PR to add the config! Thanks Nev!

PR #1829, opened 2022-05-04
Extracting HOC tracked component class base for re-use

bittola:
While withAITracking function is working in most of the cases, it's not useful in a scenario when a content of the wrapped component is not physically located under the component's DOM tree. 
An example of that is Modal component from Microsoft Fluent UI framework but there are many other similar cases too.
Therefore, a developer may sometimes need to create a custom version of the HOC function, which for example, as in the case of the Modal component would wrap a content instead of the component itself, however still have the same tracking functionality as the app insights library HOC function.

The change here extract the HOC class so it can be re-used and customized by consumers of the library

### DIFF
--- a/extensions/applicationinsights-react-js/src/withAITracking.tsx
+++ b/extensions/applicationinsights-react-js/src/withAITracking.tsx
@@ -2,9 +2,91 @@
 // Licensed under the MIT License.
 
 import { IMetricTelemetry } from '@microsoft/applicationinsights-common';
-import { CoreUtils } from '@microsoft/applicationinsights-core-js';
+import { dateNow } from '@microsoft/applicationinsights-core-js';
 import * as React from 'react';
 import ReactPlugin from './ReactPlugin';
+
+/**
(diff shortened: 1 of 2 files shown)


extensions/applicationinsights-react-js/src/withAITracking.tsx
```diff
@@ -2,10 +2,92 @@
...
+  protected _idleStartTimestamp: number = 0;
+  protected _lastActiveTimestamp: number = 0;
+  protected _totalIdleTime: number = 0;
+  protected _idleCount: number = 0;
+  protected _idleTimeout: number = 5000;
+  protected _intervalId?: any;
+  protected _componentName: string;
+  protected _reactPlugin: ReactPlugin;
+
+  public constructor(props: P, reactPlugin: ReactPlugin, componentName: string,) {
```
MSNev, line 25: nit: trailing comma?

MSNev (approved):

Issue #1852, opened 2022-06-09
[BUG] Snippet initialization with IE8 fails with minified code (works with un-minified code)


MSNev:
There is an issue with the snippet version check ApplicationInsightsContainer when using === and !== matching the version with "2.0".

This check works with unminified version of SDK works without any issues in IE8/9 mode, but causes the SDK to fail to initialize properly when the minified version is used.

This is because of a difference in the way the minified code performs the 2 numeric comparisons for the snippet version for (version !== 2.0 and version === 2.0) which causes this to fail for IE8/9 and all other browsers.

This is because the snippet initializes the version as 2 (not 2.0 -- subtle issue) and IE8/9 treat this differently (integer vs float) for the way these are checked in the minified code, which results in this check failing only for the minified code -- it works for the un-minified code.

The correct fix for this (because we can't update all of the deployed snippet versions) is to change these 2 comparisons to use `!=` and `==` instead of `!==` and `===`.

As a workaround for anyone hitting this issue and assuming you CAN update the snippet used is to set the snippet version to 2.0 instead of 2.
So changing this minified code `,sv:"5",version:2,config:d` to `,sv:"5",version:2.0,config:d`

MSNev:
This is quite a bit more complicated than originally determined.
For the current version on the CDN, the only really work-around it to use the 2.5 version from the CDN as it's not the version number from the snippet as originally diagnosed.

The underlying root cause is a combination of items, simplistically, it is not a bug in any 1 component but the way all 3 are interacting.

This issue only affects the pre-minified bundle which is included in the NPM packages and deployed to the CDN, the main un-minified code works with IE8 and depending on the minifier you use and the level of minification performed will determine whether the issue reproduces.

What is the issue?
- As part of minification for Application Insights we use @microsoft/dynamicProto to remove the un-minifier `Class.prototype.xxxx` in the generated code.
- This works fine for un-minified code
- For minification we use Uglify-JS v3 and as part of the "minification" process when a "class" doesn't have any static or class level functions (which is what happens when we use "dynamicProto") the "class" effectively becomes an object that is "linked" to the `class` function during creation.
- All of the above works perfectly on all browsers (minified or un-minified) "except" for IE8, where the minification step of creating an object `class` does not "keep" the link with the "constructor" (class) function, so when dynamicProto successfully creates the prototype functions the object is not linked with the prototype, so the `class` functions don't exist..

[...]

PR #1854, opened 2022-06-14
[BUG] Snippet initialization with IE8 fails with minified code (works with un-minified code) #1852

MSNev:


extensions/applicationinsights-react-js/package.json
```diff
@@ -53,7 +53,7 @@
         "ts-jest": "^27.0.7",
         "typescript": "^4.3.4",
         "tslib": "^2.0.0",
-        "uglify-js": "^3.11.0",
+        "uglify-js": "3.16.0",
```
MSNev, line 56: Pinning to the version of uglify-js that has been tested with this work-around, so that a future change doesn't introduce another issue.

This pinning will be removed for the beta, as we are dropping ES3 (IE8) support.

Karlie-777 (approved):

ramthi (approved):

Issue #1636, opened 2021-08-16
[BUG] measurements not being sent when using stopTrackEvent(name, properties, measurements);


CarbonCollins:
**Description/Screenshot**
We are not receiving any measurements in AppInsights when using the `@microsoft/applicationinsights-web` NPM package, specifically the  `startTrackEvent(name);` and `stopTrackEvent(name, properties, measurements);` functions. We get the event created and the custom properties are attached without an issue, however, the custom metrics is left empty in AppInsights.

**Steps to Reproduce**

Initialise AppInsights using:
```
const appInsights = new ApplicationInsights({
  config: {
    connectionString: '[theConnectionString]',
    enableAutoRouteTracking: true,
    disableFetchTracking: false,
    namePrefix: 'someprefix'
  }
});
```
start and then stop tracking an event with custom properties and custom measurements:
```
appInsights.startTrackEvent('someEvent');

appInsights.stopTrackEvent('someEvent', { someProperty: 'hi' }, { someMeasurement: 12 });
```
wait for the event to appear in app insights and notice the custom properties is populated with `{ someProperty: 'hi' }` but custom measurements is blank

 - OS/Browser: `Firefox 90`/`Chrome 91`
 - SDK Version [e.g. 22]:  `2.6.5`
 - How you initialized the SDK: using `@microsoft/applicationinsights-web` (see example above)

**Expected behavior**
Numerical measurements are sent to App Insights when using `startTrackEvent(name);` and `stopTrackEvent(name, properties, measurements);` functions.

**Additional context**
I spent a little bit of time drilling into the packages in my node_modules folder and found two things which may help this issue: 

1) In the `@microsoft/applicationinsights-web` package in the `applicationinsights-web/src/Initialization.ts` file seems to have a `Todo: Fix to pass measurements once type is updated` on the `stopTrackEvent` function

2) In the `@microsoft/applicationinsights-analytics-js` package in the `applicationinsights-analytics-js/src/JavaScriptSDK/ApplicationInsights.ts` file seems to be where the `stopTrackEvent` call ends up, however, in the `Timing` class we do not seem to be passing the measurements record anywhere so I would hazard a guess that the package is not even sending the measurements to App Insights at all.

CarbonCollins:
We are still having this issue

MSNev:
@Karlie-777 can you please investigate.
@CarbonCollins do you have a simple repo (beyond the above) that @Karlie-777 could use?

PR #1859, opened 2022-06-29
fix stopTrackEvent

Karlie-777:


Karlie-777:
https://github.com/microsoft/ApplicationInsights-JS/issues/1636

Issue #1862, opened 2022-07-06
[BUG] Remote Dependency requests don't "always" have the correct ai.operation.id tag (page view race condition)


MSNev:
When an ajax (XHR/fetch) request occurs it is provided with the "requestId" and "traceparent" headers for distributed tracing downstream based on the current "traceId" of the page (which is the correct behavior) and when the response for the request is received the SDK will cause an event to be sent via the `trackInternalDependency` function.

However, this call will only occur after BOTH of the following
- The response is received (even a failed / timed out response)
- if `enableAjaxperfTracking` is enabled after the runtime populates the performance event for the request OR the SDK times out waiting for the value (defaults to 3 attempts every 25ms (75ms total)).

Additionally, during the construction of the `event` (that is passed to the track call) and it getting stored in the internal buffer (waiting to be sent) the event will be populated with a bunch of standard values including the `tag` `ai.operation.id` which is the current value at the point of the `trackXXXX()` call.

Normally, this is not an issue 
-- unless a page navigation occurs OR you set / change the `traceID` for the page (between) the time the ajax was "sent" to the time that it's `trackXXX()` call is made. In this case the `ai.operation.id` tag value sent for the request will be that of the "current page" and not the page that triggered the event.

To address this issue fix required (further analysis / investigation required), looks like it would be to 
- Set the `ext.trace` values on the event before calling `traceInternalDependency`
- Update the [TelemetryContext](https://github.com/microsoft/ApplicationInsights-JS/blob/master/extensions/applicationinsights-properties-js/src/TelemetryContext.ts#L146-L150) code to use `getSetValue()` instead of `setValue()` (which always set the value) or add a `isNullOrUndefined` value check argument to to the `setVeluer` helper. Otherwise, this code will ALWAYS overwrite any value on the event.

PR #1869, opened 2022-07-30
[BUG] Remote Dependency requests don't "always" have the correct ai.operation.id tag (page view race condition) #1862

MSNev:


extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -23,7 +23,6 @@ const AJAX_MONITOR_PREFIX = "ai.ajxmn.";
 const strDiagLog = "diagLog";
 const strAjaxData = "ajaxData";
 const strFetch = "fetch";
-const strTrackDependencyDataInternal = "trackDependencyDataInternal"; // Using string to help with minification
```
MSNev, line 26: Removing and letting the automatic dynamic name logic work

tools/grunt-tasks/minifyNames.js
```diff
@@ -1010,7 +1010,7 @@ function findPossibleFileConstants(parsedFile, theNames, ignoreConst, ignoreName
 }
 
 var dynamicIndex = 0;
-function createDynamicName(value) {
+function createDynamicName(value, isDupe) {
```
MSNev, line 1013: This was generating the same "name" for the same string with a different case
eg. `traceId` and `traceID`

Karlie-777 (approved):

PR #1882, opened 2022-08-16
[Task]14569737:  Add support to throttle the deprecation message

Karlie-777:
[draft]
Specs Overview: [shared/AppInsightsCommon/src/Interfaces/IThrottleMgr.ts](https://github.com/microsoft/ApplicationInsights-JS/pull/1882/files/0b19df2686333318d78d2210db61c3a3e195cd0c#diff-47e400f1c578b07a6eb851a986322fd8bd248c674d9efbfcfd67feb625f88bd7)


shared/AppInsightsCommon/src/Enums.ts
```diff
@@ -47,3 +47,8 @@ export const DistributedTracingModes = createEnumStyle<typeof eDistributedTracin
     W3C: eDistributedTracingModes.W3C
 });
 export type DistributedTracingModes = number | eDistributedTracingModes;
+
+export const enum IThrottleMsgKey {
```
MSNev, line 51: Do these need to be strings?

Ideally, we want const enum's to be numeric as this produces smaller code.

shared/AppInsightsCommon/src/Interfaces/IThrottleMgr.ts
```diff
@@ -0,0 +1,112 @@
...
+    * Default: send data on 28th every 3 month each year
+    */
+    interval?: IThrottleInterval;
+}
+
+
+/**
+* Identifies date object
+*/
+export interface IthrottleDate {
```
MSNev, line 79: If this is representing a general Date, then we can probably just use the JS Date object

shared/AppInsightsCommon/src/ThrottleMgr.ts
```diff
@@ -0,0 +1,186 @@
...
+    IAppInsightsCore, IDiagnosticLogger, _eInternalMessageId, _throwInternal, eLoggingSeverity, isNullOrUndefined, safeGetLogger, strTrim
+} from "@microsoft/applicationinsights-core-js";
+import { IThrottleMsgKey } from "./Enums";
+import { IthrottleDate, IthrottleLocalStorageObj, IthrottleMgrConfig, IthrottleResult } from "./Interfaces/IThrottleMgr";
+import { utlCanUseLocalStorage, utlGetLocalStorage, utlSetLocalStorage } from "./StorageHelperFuncs";
+
+const THROTTLE_STORAGE_PREFIX = "appInsightsThrottle";
+
+
+export class ThrottleMgr {
```
MSNev, line 11: Tests showing expected usage would be good and helpful in validating and following the flow.

shared/AppInsightsCommon/src/ThrottleMgr.ts
```diff
@@ -0,0 +1,251 @@
...
+        function _getLocalStorageName(msgKey: IThrottleMsgKey, prefix?: string) {
+            let fix = isNotNullOrUndefined(prefix)? prefix : "";
+            if (msgKey) {
+                return THROTTLE_STORAGE_PREFIX + fix + "-" + msgKey;
+            }
+            return null;
+        }
+
+        // transfer local storage string value to object that identifies start date and current count
+        function _getLocalStorageObj(value: string, logger: IDiagnosticLogger, storageName: string) {
```
MSNev, line 183: I missed this previously, noticed as part of going through the tests.
lets store this as a single JSON blob, so that the "storageName" is a key into the blob. And then we can have the values in the blob as straight forward values, this then reduces the validate required a little.

shared/AppInsightsCommon/src/Interfaces/IThrottleMgr.ts
```diff
@@ -0,0 +1,109 @@
...
+     * Identifies max times items can be sent within a month
+     * Default: 1
+     */
+    maxTimesPerMonth?: number;
+}
+
+/**
+* Identifies basic config
+*/
+export interface IthrottleMgrConfig {
```
MSNev, line 50: nit: Should be IThro...

shared/AppInsightsCommon/src/Interfaces/IThrottleMgr.ts
```diff
@@ -0,0 +1,109 @@
...
+    * Identifies frequency of items sent
+    * Default: send data on 28th every 3 month each year
+    */
+    interval?: IThrottleInterval;
+}
+
+/**
+* Identifies object for local storage
+*/
+export interface IthrottleLocalStorageObj {
```
MSNev, line 77: same as above "T" not "t"

shared/AppInsightsCommon/src/Interfaces/IThrottleMgr.ts
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * identifies previous triggered throttle date
+     */
+    preTriggerDate?: Date;
+}
+
+/**
+* Identifies throttle result
+*/
+export interface IthrottleResult {
```
MSNev, line 98: dito

shared/AppInsightsCommon/src/applicationinsights-common.ts
```diff
@@ -58,7 +59,8 @@ export { IPropertiesPlugin } from "./Interfaces/IPropertiesPlugin";
 export { IUser, IUserContext } from "./Interfaces/Context/IUser";
 export { ITelemetryTrace, ITraceState } from "./Interfaces/Context/ITelemetryTrace";
 export { IRequestContext } from "./Interfaces/IRequestContext";
-export { eDistributedTracingModes, DistributedTracingModes } from "./Enums";
+export { IthrottleLocalStorageObj, IthrottleMgrConfig, IthrottleResult} from "./Interfaces/IThrottleMgr";
```
MSNev, line 62: Do we need to also export IThrottleLimit and IThrottleInterval?

We should if they will be referenced by using components (which I *think* they will)

shared/AppInsightsCommon/Tests/Unit/src/ThrottleMgr.tests.ts
```diff
@@ -0,0 +1,642 @@
...
+const compareDates = (date1: Date, date: string | Date) => {
+    try {
+        if (date1 && date) {
+            let date2 = typeof date == "string"? new Date(date) : date;
+            return date1.getUTCFullYear() === date2.getUTCFullYear() &&
+            date1.getUTCMonth() === date2.getUTCMonth() &&
+            date1.getUTCDate() === date2.getUTCDate();
+        }
+    } catch (e) {
+        console.log("compare dates error" + e);
```
MSNev, line 18: suggestion: As a test helper perhaps pass a additional "expected" match value and this this can be changed into an assert with the expected value "true/false" so if it's not as expected it would cause the test to fail.

shared/AppInsightsCommon/src/ThrottleMgr.ts
```diff
@@ -0,0 +1,285 @@
...
+        }
+        
+        function _sendMessage(msgID: _eInternalMessageId, logger: IDiagnosticLogger, message: string, severity?: eLoggingSeverity) {
+            _throwInternal(logger,
+                severity || eLoggingSeverity.CRITICAL,
+                msgID,
+                message);
+        }
+
+        // NOTE: config.limit.samplingPercentage unit is 1000.
```
MSNev, line 279: I'm thinking maybe we should rename this to something like `samplingRate` and in the description we identify this as a `percentage` to the 4 decimal places (I don't think 2 decimal places is enough)

including examples like 1 is the same as specifying 0.0001%

shared/AppInsightsCommon/Tests/Unit/src/ThrottleMgr.tests.ts
```diff
@@ -1,23 +1,25 @@
...
         if (date1 && date) {
             let date2 = typeof date == "string"? new Date(date) : date;
-            return date1.getUTCFullYear() === date2.getUTCFullYear() &&
+            isSame = date1.getUTCFullYear() === date2.getUTCFullYear() &&
             date1.getUTCMonth() === date2.getUTCMonth() &&
             date1.getUTCDate() === date2.getUTCDate();
         }
     } catch (e) {
-        console.log("compare dates error" + e);
+        Assert.ok(false,"compare dates error" + e);
```
MSNev, line 19: Not exactly what I was thinking...

```ts
const compareDates = (date1: Date, date: string | Date, expectedSame: boolean)
...
} catch(e) {
    // This is ok as we are saying that any exception should cause the test to fail
    Assert.ok(false,"compare dates error" + e);
}
```

And then
```ts
Assert.equal(isSame, expectedSame, "Checking that the dates where as expected");
```

MSNev (approved):

PR #1888, opened 2022-08-18
V3 Beta - Remove Legacy code and ES3 support from V2 SDK

MSNev:
- Remove ES3 polyfills (Breaking change from V2)
- Remove Deprecated (v1) API compatibility layer (Breaking change from V2)
- Remove unused namespace globals (Breaking change from V2)
- Document breaking changes


AISKU/rollup.config.js
```diff
@@ -3,7 +3,7 @@ import { uglify } from "@microsoft/applicationinsights-rollup-plugin-uglify3-js"
 import replace from "@rollup/plugin-replace";
 import cleanup from "rollup-plugin-cleanup";
 import dynamicRemove from "@microsoft/dynamicproto-js/tools/rollup/node/removedynamic";
-import { es3Poly, es3Check, importCheck } from "@microsoft/applicationinsights-rollup-es3";
```
MSNev, line 6: Actually, I should replace these with the new es5 versions -- I removed these before I decided to branch the es3 -> es5.
While there are no current automatic polyfills there are some checks that should still occurr.

AISKU/src/Initialization.ts
```diff
@@ -57,7 +54,7 @@ export interface Snippet {
 }
 
 export interface IApplicationInsights extends IAppInsights, IDependenciesPlugin, IPropertiesPlugin {
-    appInsights: ApplicationInsights;
+    appInsights: AnalyticsPlugin;
```
Karlie-777, line 57: I am just curious, `ApplicationInsights` includes deprecated instance, so we switch to `analyticsplugin` directly here?

MSNev: For this one I renamed the exported object from "ApplicationInsights" to "AnalyticsPlugin" a few releases ago and as part of this (for backward compatibility) it also exported the "AnalyticsPlugin" as "ApplicationInsights" (so 2 names referencing the same class).

PR #1904, opened 2022-09-19
Fixed release notes, previously used a deprecated file.

josephsivits:
The file linked in the 'Changelist' had been deprecated, moved to https://github.com/microsoft/ApplicationInsights-JS/releases , an active and dynamically supported list of releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -662,7 +662,7 @@ It is expected that most users will be using the `Public` URL, however, it is al
 
 ## Release Notes
 
-[Changelist](./CHANGELIST.md)
+[Releases](https://github.com/microsoft/ApplicationInsights-JS/releases)
 
 ## Browser Support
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -662,7 +662,7 @@ It is expected that most users will be using the `Public` URL, however, it is al
 
 ## Release Notes
 
-[Changelist](./CHANGELIST.md)
```
MSNev, line 665: We should keep this, but it needs to point to [./RELEASES.md](the https://github.com/microsoft/ApplicationInsights-JS/blob/master/RELEASES.md) instead of ./CHANGELIST.md

README.md
```diff
@@ -662,7 +662,7 @@ It is expected that most users will be using the `Public` URL, however, it is al
...
+[Releases](https://github.com/microsoft/ApplicationInsights-JS/releases)
```
MSNev, line 665: But I like having this reference. suggest changing the "Changelist" above to "Release Notes" perhaps

MSNev (approved):

PR #1906, opened 2022-09-26
[Task]14569737: add throttle cdn config

Karlie-777:


tools/throttle/scripts/listCdnVersions.ps1
```diff
@@ -0,0 +1,114 @@
...
+    if ([string]::IsNullOrWhiteSpace($container) -ne $true -and "beta","next","public", "dev", "nightly" -NotContains $container) {
+        Write-LogFailure "[$($container)] is not a valid value, must be beta, next or public"
+    }
+}
+
+Function Get-AllVersionFiles(
+    [system.collections.generic.dictionary[string, system.collections.generic.list[hashtable]]] $files,
+    [string] $storagePath
+) {
+    Get-VersionFiles $files "$storagePath/ext" "ai.throttle." $null
```
MSNev, line 58: We will actually put this in the $storagePath -- so remove the /ext (which is for extensions)

tools/config/config.json
```diff
@@ -0,0 +1,49 @@
+{
+    "name":"Application Insights Config",
+    "description":" Application Insights JavaScript SDK - Web Config",
+    "copyright": "(c) Microsoft and contributors. All rights reserved.",
+    "author": "Microsoft Application Insights Team",
+    "throttleConfig": {
+        "default": {
```
MSNev, line 7: Do we want the default to be effectively disabled

MSNev (approved):

PR #1930, opened 2022-10-18
Remove the source-map-loader from the package.json as we don't use it.

MSNev:


MSNev:
This is a Component Governance issue where it's failing because of the listed version

Karlie-777 (approved):

PR #1945, opened 2022-11-15
[Beta] Plugin does not wire itself into the Channel Configuration when...

MSNev:
…both channels and channel extensions are supplied


channels/tee-channel-js/src/TeeChannel.ts
```diff
@@ -0,0 +1,211 @@
...
+    ITelemetryUpdateState, createProcessTelemetryContext, initializePlugins, onConfigChange, proxyFunctions
+} from "@microsoft/applicationinsights-core-js";
+import { arrForEach, isArray, objDeepFreeze, objFreeze, throwError } from "@nevware21/ts-utils";
+import { ChannelControllerPriority, IChannelController, _IInternalChannels, createChannelControllerPlugin } from "./ChannelController";
+import { ITeeChannelConfig } from "./Interfaces/ITeeChannelConfig";
+
+const ChannelValidationMessage = "Channel has invalid priority - ";
+
+const defaultTeeChannelConfig: IConfigDefaults<ITeeChannelConfig> = objDeepFreeze({
+});
```
MSNev, line 18: I need to add the "defaults" in here (even if they are undefined) so that they become dynamic.

extensions/applicationinsights-debugplugin-js/src/DebugPlugin.ts
```diff
@@ -194,8 +194,8 @@ export default class DebugPlugin extends BaseTelemetryPlugin {
                         });
                     }
 
-                    if (isFunction(core.getTransmissionControls)) {
-                        let channelControls: IChannelControls[][] = core.getTransmissionControls();
+                    if (isFunction(core["getTransmissionControls"])) {
```
MSNev, line 197: Keeping a reference for "getTransmissionControls" for backward compatibility

hectorhdzg (approved):
LGTM

shared/AppInsightsCore/src/JavaScriptSDK/ChannelController.ts
```diff
@@ -1,274 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// // Licensed under the MIT License.
-
-import { ITimerHandler, arrForEach, isArray, objFreeze, scheduleTimeout, throwError } from "@nevware21/ts-utils";
```
MSNev, line 4: Ok, I think I've managed to re-instate the git mv to keep the history of this file -- will publish an update shortly

MSNev: :-( Ok, local shows this as a git move/rename but here as a delete and add...

PR #1957, opened 2022-12-08
[Beta][Task]15982350: Dynamic config for Dependencies plugin

Karlie-777:


extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -271,7 +271,7 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
     public static identifier: string = "AjaxDependencyPlugin";
 
     public identifier: string = AjaxMonitor.identifier;
-
+    public readonly config: IConfig;
```
MSNev, line 274: Lets not add this (unless we "really" need to) as this is now accessible via the `core.config` as they are all the same

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -555,83 +560,88 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
-                                    !_isDisabledRequest(null, input, init) &&
-                                    // If we have a polyfil and XHR instrumented then let XHR report otherwise we get duplicates
-                                    !(isPolyfill && _xhrInitialized)) {
-                                let ctx = callDetails.ctx();
-                                fetchData = _createFetchRecord(input, init);
-                                let newInit = _self.includeCorrelationHeaders(fetchData, input, init);
-                                if (newInit !== init) {
-                                    callDetails.set(1, newInit);
+                _self._addHook(onConfigChange(_extensionConfig, () => {
+                    _disableFetchTracking = !!_extensionConfig.disableFetchTracking;
```
MSNev, line 564: Oh, nice catch... If this is disabled on start and changed later

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -555,83 +552,88 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
+                        // disabled request URLS (i.e. internal urls) to ensure we don't end up in a constant loop
+                        // of reporting ourselves, for example React Native uses a polyfill for fetch
+                        // Note: Polyfill implementations that don't support the "poyyfill" tag are not supported
+                        // the workaround is to add a polyfill property to your fetch implementation before initializing
+                        // App Insights
+                        _addHook(InstrumentFunc(global, strFetch, {
+                            ns: _evtNamespace,
+                            req: (callDetails: IInstrumentCallDetails, input, init) => {
+                                // Just call so that we record any disabled URL
+                                _isDisabledRequest(null, input, init);
```
MSNev, line 631: I don't think we need to do this???
As won't this cause the XHR request to also then get ignored even though we want it to be?

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -555,83 +552,88 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
+    
+                        _fetchInitialized = true;
+                    } else if (isPolyfill) {
+                        // If fetch is a polyfill we need to capture the request to ensure that we correctly track
+                        // disabled request URLS (i.e. internal urls) to ensure we don't end up in a constant loop
+                        // of reporting ourselves, for example React Native uses a polyfill for fetch
+                        // Note: Polyfill implementations that don't support the "poyyfill" tag are not supported
+                        // the workaround is to add a polyfill property to your fetch implementation before initializing
+                        // App Insights
+                        _addHook(InstrumentFunc(global, strFetch, {
```
MSNev, line 627: This could also get called more than once when the referenced config items are called more than once.

extensions/applicationinsights-dependencies-js/src/ajax.ts
MSNev: I think this case is handled on line 568 with `!(isPolyfill && _xhrInitialized)) {`

MSNev: Having said that we had this previously so this probably just needs an additional flag so we only call the `_addHook` once.

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -644,82 +646,88 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
-                                let ajaxData = xhr[strAjaxData];
-                                if (!_isDisabledRequest(xhr, url) && _isMonitoredXhrInstance(xhr, true)) {
-                                    if (!ajaxData || !ajaxData.xhrMonitoringState.openDone) {
-                                        // Only create a single ajaxData (even when multiple AI instances are running)
-                                        _openHandler(xhr, method, url, async);
+                _self._addHook(onConfigChange(_extensionConfig, () => {
+                    _disableAjaxTracking = !!_extensionConfig.disableAjaxTracking;
+                    _enableRequestHeaderTracking = _extensionConfig.enableRequestHeaderTracking;
+
+                    if (_supportsAjaxMonitoring(_self) && !_disableAjaxTracking && !_xhrInitialized) {
```
MSNev, line 653: We probably want to put this part `if (_supportsAjaxMonitoring(_self)` outside of the onConfigChange as this won't change as it's based on the support of the runtime.

MSNev: i.e. don't even "hook" if XHR is not supported

MSNev (approved):

PR #1966, opened 2023-01-03
[Beta][Task]15982296: Dynamic config for Sender

Karlie-777:


channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -239,18 +244,77 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
                         }
                     });
 
+                    // Only update the endpoint if the original config !== the current config
+                    // This is so any redirect endpointUrl is not overwritten
+                    if (_orgEndpointUrl !== senderConfig.endpointUrl) {
+                        if (_orgEndpointUrl) {
+                            // TODO: add doc to remind users to flush before changing endpoint
+                            _self._buffer.clear();
```
MSNev, line 252: I don't think we want to clear the buffer, as this could cause the loss of events, the TODO is valid but we just tell them that if they don't existing queued events will also be sent to the new endpoint.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -239,18 +244,77 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
...
+                        // case1 (Pre and Now enableSessionStorageBuffer settings are same)
+                        // if namePrefix or eventsLimitInMem or emitLineDelimitedJson is changed, transfer current buffer to new buffer
+                        // else no action needed
+
+                        //case2 (Pre and Now enableSessionStorageBuffer settings are changed)
+                        // transfer current buffer to new buffer
+                       
+                        if (shouldUpdate) {
+                            _self.pause();
+                            let unsentPayload = _self._buffer.getItems().slice(0);
```
MSNev, line 287: We also need to track the "sent" but not acknowledged items and "move" them as well as if the request that sent them fails they will also need to be re-sent. It might be easier to add a `copy` function to the Buffer implementations that way we can test them also.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -590,6 +634,8 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
                         _self._appId = response.appId;
                     }
                 }
+                //let isOrgEndpoint = responseUrl !== _orgEndpointUrl;
+                let isOrgEndpoint = true;
```
MSNev, line 638: Lets not worry about this case and just document as a "potential" issue that when changing the endpoint if a retry is required that any previous events may (if not dropped) be sent to the new endpoint.

Basically, this is where we say that it you want the events to only be sent to the original endpoint that you should either
- Create a 2nd instance and only use that instance
- Flush and wait for the send to be successful (not sure if we currently has a callback on the flush that returns only after all events are flushed -- we do for 1DS).

channels/applicationinsights-channel-js/src/SendBuffer.ts
```diff
@@ -124,6 +129,13 @@ abstract class BaseSendBuffer {
         
                 return null;
             };
+
+            _self.deepCopy = (buffer: BaseSendBuffer) => {
```
MSNev, line 133: Suggestion: Lets either
- rename this to `deepCopyTo` 
- reverse the logic so that it copies the passed buffer to this instance (which is the default way I started reading this), as to me deepCopy means either create a deep copy of yourself and return it or create a deepCopy of the passed buffer.
- change this so that it's just `copy` or `clone` or something like that where it returns a new instance etc. (just thinking out aloud)

MSNev: `copyFrom` might be another good name for this

channels/applicationinsights-channel-js/Tests/Unit/src/Sender.tests.ts
```diff
@@ -48,13 +66,506 @@ export class SenderTests extends AITestClass {
...
+                    arrBuffer.enqueue(val);
+                });
+                arrBuffer.deepCopy(sessionBuffer);
+                QUnit.assert.deepEqual( sessionBuffer.getItems(), payload, "payload should be same");
+                QUnit.assert.deepEqual(this._getBuffer(BUFFER_KEY, logger), payload, "session storage buffer is set");
+                arrBuffer.enqueue("payload");
+                QUnit.assert.deepEqual(arrBuffer.getItems().length, 7, "arrBuffer length");
+                QUnit.assert.deepEqual( sessionBuffer.getItems().length, 6, "copy is deep copy");
+
+                utlSetSessionStorage(logger, BUFFER_KEY,JSON.stringify([]));
```
MSNev, line 476: This should also check that the SessionStorage was updated and contains the expected values

channels/applicationinsights-channel-js/Tests/Unit/src/Sender.tests.ts
```diff
@@ -48,13 +66,506 @@ export class SenderTests extends AITestClass {
...
+                arrForEach(payload, (val) =>{
+                    sessionBuffer.enqueue(val);
+                });
+                QUnit.assert.deepEqual(this._getBuffer(BUFFER_KEY, logger), payload, "session storage buffer is set");
+                sessionBuffer.deepCopy(sessionBufferCopy);
+                QUnit.assert.deepEqual(sessionBufferCopy.getItems(), payload, "payload should be same");
+                QUnit.assert.deepEqual(sessionBuffer.getItems(), [], "original session storage buffer should be clear");
+                QUnit.assert.deepEqual(this._getBuffer(BUFFER_KEY, logger), payload, "session storage should not be changed");
+
+                utlSetSessionStorage(logger, BUFFER_KEY,JSON.stringify([]));
```
MSNev, line 564: Likewise we should check that all of the stuff stored in SessionStorage for the old one has been duplicated in the new one

channels/applicationinsights-channel-js/Tests/Unit/src/Sender.tests.ts
```diff
@@ -48,13 +66,506 @@ export class SenderTests extends AITestClass {
...
+                let sessionBufferCopy = new SessionStorageSendBuffer(logger, config);
+             
+                sessionBuffer.deepCopy(sessionBufferCopy);
+                QUnit.assert.deepEqual(sessionBufferCopy.getItems(), [], "payload should be empty");
+                QUnit.assert.deepEqual(this._getBuffer(BUFFER_KEY, logger), [], "session storage buffer should be empty");
+                let payload = ["payload1", "payload2", "payload3", "payload4", "payload5", "payload6"];
+                arrForEach(payload, (val) =>{
+                    sessionBuffer.enqueue(val);
+                });
+                QUnit.assert.deepEqual(this._getBuffer(BUFFER_KEY, logger), payload, "session storage buffer is set");
```
MSNev, line 558: It should also include items "marked as sent" as that is really where we have a risk of loosing events in this scenario

MSNev (approved):

PR #1967, opened 2023-01-06
[Beta][Task]15982342: Dynamic config for extension click analytics

Karlie-777:


extensions/applicationinsights-clickanalytics-js/src/ClickAnalyticsPlugin.ts
```diff
@@ -5,19 +5,74 @@
...
+    eferrerUri: hasDocument ? document.referrer : "",
+    requestUri: "",
+    pageName: "",
+    pageType: ""
+} as ICoreData;
+
+const defaultValues: IConfigDefaults<IClickAnalyticsConfiguration> = objDeepFreeze({
+    autoCapture: true,
+    callback: {
+        pageActionPageTags: () => null,
```
MSNev, line 48: I think all of these should just be defined as null and not functions that return null.

extensions/applicationinsights-clickanalytics-js/src/ClickAnalyticsPlugin.ts
```diff
@@ -5,19 +5,74 @@
...
+    behaviorValidator: (key:string) => key || "",
+    defaultRightClickBhvr: "",
+    dropInvalidEvents : false,
+    urlCollectHash: false,
+    urlCollectQuery: false
+});
+
+function _setProp(val: Object, def: Object): Object {
+    removeInvalidElements(val);
+    if (JSON.stringify(def) !== "{}") {
```
MSNev, line 64: This would better being either `if (def) {` or `if (isObject(def)) {` as `JSON.stringify()` can be expensive.

extensions/applicationinsights-clickanalytics-js/src/ClickAnalyticsPlugin.ts
```diff
@@ -99,13 +150,41 @@ export class ClickAnalyticsPlugin extends BaseTelemetryPlugin {
...
+                    let ctx = createProcessTelemetryContext(null, config, core);
+                    let _config = ctx.getExtCfg(_identifier, defaultValues);
+
+                    let logger = _self.diagLog();
+                    _contentHandler = new DomContentHandler(_config, logger);
+                    let metaTags = _contentHandler.getMetadata();
+                    _pageAction = new PageAction(this, _config, _contentHandler, _config.callback.pageActionPageTags, metaTags, logger);
+    
+                    // Default to DOM autoCapture handler
+                    _autoCaptureHandler = _autoCaptureHandler ? _autoCaptureHandler : new AutoCaptureHandler(_self, _config, _pageAction, logger);
```
MSNev, line 168: This might be problematic, as if any config changes the _pageAction will be assigned a new object, but this won't get reassigned.

extensions/applicationinsights-clickanalytics-js/src/ClickAnalyticsPlugin.ts
```diff
@@ -99,13 +150,41 @@ export class ClickAnalyticsPlugin extends BaseTelemetryPlugin {
...
+                    if (_config.autoCapture) {
+                        _autoCaptureHandler.click();
```
MSNev, line 170: This is also problematic as if the "autoCapture" is true and unchanged any other config changes would cause another click event to be triggered.

extensions/applicationinsights-clickanalytics-js/src/common/Utils.ts
```diff
@@ -265,50 +265,50 @@ export function validateContentNamePrefix ( config: IClickAnalyticsConfiguration
...
+//     for (const attribute in defaultConfig) {
+//         if (typeof defaultConfig[attribute] === "object") {
+//             attributesThatAreObjectsInConfig.push(attribute);
+//         }
+//     }
+
+//     if (overrideConfig) {
+//         // delete attributes that should be object and
+//         // delete properties that are null, undefined, ''
+//         removeNonObjectsAndInvalidElements(overrideConfig, attributesThatAreObjectsInConfig);
```
MSNev, line 305: The new version of this is that each "field" can define an `isVal` which is a callback function that is past the user supplied "value" and if `false` the default will be assigned otherwise the default value will attempt to be merged with this value.

So we should probably use this for these cases, and because of the dynamic nature of the config now we don't want to "remove" the invalid values.

extensions/applicationinsights-clickanalytics-js/Tests/Unit/src/ClickEventTest.ts
```diff
@@ -29,6 +30,138 @@ export class ClickEventTest extends AITestClass {
...
+                Assert.equal(pageActionPageTags(), null, "pageActionPageTags should be set by default");
+                Assert.equal(contentName(), null, "contentName should be set by default");
+                
+                Assert.ok(extConfig.behaviorValidator,"behaviorValidator should be set by default");
+                let bhvVal = extConfig.behaviorValidator;
+                Assert.equal(bhvVal(), "", "behaviorValidator should return ''");
+                Assert.equal(bhvVal(1), 1, "behaviorValidator should return key");
+
+                //config(non-Object) changes
+                core.config["extensionConfig"][clickAnalyticsPlugin.identifier].urlCollectHash = true;
```
MSNev, line 97: Suggestion: Just to make this a little easier to read you can assign `core.config["extensionConfig"][clickAnalyticsPlugin.identifier]` to a local variable

extensions/applicationinsights-clickanalytics-js/src/common/Utils.ts
```diff
@@ -28,7 +28,7 @@ export type _ExtendedInternalMessageId = number | _eExtendedInternalMessageId |
  */
 export function removeNonObjectsAndInvalidElements(overrideConfig: IClickAnalyticsConfiguration, attributeNamesExpectedObjects: Array<string>): void {
     removeInvalidElements(overrideConfig);
-    for (var i in attributeNamesExpectedObjects) {
+    arrForEach(attributeNamesExpectedObjects, (i) => {
         if (objHasOwn(attributeNamesExpectedObjects, i)) {
```
MSNev, line 32: nit: I think we can remove the `objHasOwn` call when using `arrForEach`, but we can do this later

MSNev (approved):

PR #1968, opened 2023-01-07
[Beta][AI] Add tee channel tests

Karlie-777:


channels/tee-channel-js/Tests/Unit/src/TeeChannelCore.Tests.ts
```diff
@@ -17,6 +17,48 @@ export class TeeChannelCoreTests extends AITestClass {
...
+                const channelPlugin = new ChannelPlugin();
+                const teeChannel = new TeeChannel();
+                const appInsightsCore = new AppInsightsCore();
+                const channelPlugin1 = new ChannelPlugin();
+                channelPlugin1.priority = 1030;
+                const channelPlugin2 = new ChannelPlugin();
+                channelPlugin2.priority = 1031;
+
+                appInsightsCore.initialize(
+                    { instrumentationKey: "09465199-12AA-4124-817F-544738CC7C41", channels: [[channelPlugin, teeChannel],[channelPlugin1]]},[]
```
MSNev, line 33: We have a test iKey elsewhere, we should use the same one for consistency

channels/tee-channel-js/Tests/Unit/src/TeeChannelCore.Tests.ts
```diff
@@ -17,6 +17,48 @@ export class TeeChannelCoreTests extends AITestClass {
...
+                
+                appInsightsCore.config.extensionConfig = appInsightsCore.config.extensionConfig? appInsightsCore.config.extensionConfig : {};
+                let extConfig = appInsightsCore.config.extensionConfig;
+                let teeChanneConfig = extConfig[teeChannel.identifier];
+                Assert.deepEqual(teeChanneConfig, {ignoreCoreChannels: false, teeChannels: null}, "default config is set");
+                appInsightsCore.config.extensionConfig[teeChannel.identifier] = {ignoreCoreChannels: false, teeChannels: [[channelPlugin2]]};
+                this.clock.tick(1);
+                coreChannels = appInsightsCore.getChannels();
+                Assert.equal(2, coreChannels.length, "Total number of core channel queues");
+                let channel = teeChannel.getTeeChannels();
```
MSNev, line 50: Oh, this is interesting...
Question: Looking at the code this looks like this sequence would include itself as it's own TeeChannel -- that would not end well. We should fix this so that a TeeChannel doesn't add itself to it's set of channels so that we don't end up with a recursive issue.

channels/tee-channel-js/Tests/Unit/src/TeeChannelCore.Tests.ts
```diff
@@ -17,6 +17,59 @@ export class TeeChannelCoreTests extends AITestClass {
...
+
+                appInsightsCore.config.extensionConfig[extId] = {ignoreCoreChannels: false, teeChannels: [[channelPlugin2]]};
+                this.clock.tick(1);
+                coreChannels = appInsightsCore.getChannels();
+                Assert.equal(2, coreChannels.length, "Total number of core channel queues");
+                Assert.deepEqual(coreChannels[0], teeChannel, "core channel 1 in test2");
+                Assert.deepEqual(coreChannels[1], channelPlugin, "core channel 2 in test2");
+                teeChannels = teeChannel.getTeeChannels();
+                Assert.equal(2, teeChannels.length, "Total number of tee channel queues");
+                Assert.equal(channelPlugin1, teeChannels[0][0], "tee channel 1 in test2");
```
MSNev, line 58: Can we also check the length of each of these sub arrays as the getTeeChannels() returns and array of arrays.
And it should NOT contain itself, while in theory teeChannels[0][]0] should be the TeeChannel (because it has a lower priority) it would be nice to confirm that teeChannels[0].length is 1

MSNev (approved):

PR #1972, opened 2023-01-18
[Beta][Task]15982357: Dynamic config for Properties extension

Karlie-777:

PR #1976, opened 2023-01-27
Followup for [Beta][Task]15982357: Dynamic config for Properties extension #1972

MSNev:
- Add ability to tag Dynamic Config properties as referenced to support updating extensionConfig in-place with a new object instance
- Add IUnloadHookContainer to allow passing extension level unload hook container to sub-components

### DIFF
--- a/extensions/applicationinsights-properties-js/Tests/Unit/src/SessionManager.Tests.ts
+++ b/extensions/applicationinsights-properties-js/Tests/Unit/src/SessionManager.Tests.ts
@@ -1,6 +1,6 @@
 import { Assert, AITestClass } from "@microsoft/ai-test-framework";
 import { SinonStub } from 'sinon';
-import { AppInsightsCore, DiagnosticLogger, createCookieMgr, newId, dateNow } from "@microsoft/applicationinsights-core-js";
+import { AppInsightsCore, DiagnosticLogger, createCookieMgr, newId, dateNow, createDynamicConfig } from "@microsoft/applicationinsights-core-js";
 import PropertiesPlugin from "../../../src/PropertiesPlugin";
 import { _SessionManager } from "../../../src/Context/Session";
 
@@ -64,13 +64,12 @@ export class SessionManagerTests extends AITestClass {
             test: () => {
 
                 var sessionPrefix = newId();
-                var config = {
+                var config = createDynamicConfig({
                     namePrefix: sessionPrefix,
                     sessionExpirationMs: undefined,
                     sessionRenewalMs: undefined,
                     cookieDomain: undefined
-
-                };
+                }).cfg;
                 // Setup
                 let cookie = "";
                 const cookieStub: SinonStub = this.sandbox.stub(this.core.getCookieMgr(), 'set').callsFake((cookieName, value, maxAge, domain, path) => {
@@ -92,14 +91,15 @@ export class SessionManagerTests extends AITestClass {
             test: () => {
 
                 var sessionPrefix = newId();
-                var config = {
+                var config = createDynamicConfig({
                     namePrefix: sessionPrefix,
                     sessionCookiePostfix: "testSessionCookieNamePostfix",
                     sessionExpirationMs: undefined,
                     sessionRenewalMs: undefined,
                     cookieDomain: undefined
 
-                };
+                }).cfg;
+
                 // Setup
                 let cookie = "";
                 const cookieStub: SinonStub = this.sandbox.stub(this.core.getCookieMgr(), 'set').callsFake((cookieName, value, maxAge, domain, path) => {
@@ -121,12 +121,12 @@ export class SessionManagerTests extends AITestClass {
             useFakeTimers: true,
             test: () => {
                 var sessionPrefix = newId();
-                var config = {
+                var config = createDynamicConfig({
                     namePrefix: sessionPrefix,
                     sessionExpirationMs: 30 * 60 * 1000,
                     sessionRenewalMs: 24 * 60 * 60 * 1000,
                     cookieDomain: undefined
-                };
+                }).cfg;
 
                 // Simulate 100ms as when zero the cookie values are deemed to be invalid
                 this.clock.tick(100);
@@ -152,12 +152,12 @@ export class SessionManagerTests extends AITestClass {
             useFakeTimers: true,
             test: () => {
                 var sessionPrefix = newId();
-                var config = {
+                var config = createDynamicConfig({
                     namePrefix: sessionPrefix,
                     sessionExpirationMs: 5000,
                     sessionRenewalMs: 24 * 60 * 60 * 1000,
                     cookieDomain: undefined
-                };
+                }).cfg;
 
                 // Simulate 100ms as when zero the cookie values are deemed to be invalid
                 this.clock.tick(100);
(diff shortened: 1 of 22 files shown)


extensions/applicationinsights-properties-js/Tests/Unit/src/SessionManager.Tests.ts
```diff
@@ -64,13 +64,12 @@ export class SessionManagerTests extends AITestClass {
             test: () => {
 
                 var sessionPrefix = newId();
-                var config = {
+                var config = createDynamicConfig({
```
MSNev, line 67: As the Session Manager now supports `onConfigChange` the passed in configuration needs to already be a dynamic configuration instance (which it will be already when initialized via the Core).

So for testing we need to simulate the same situation.

extensions/applicationinsights-properties-js/Tests/Unit/src/properties.tests.ts
```diff
@@ -375,8 +372,229 @@ export class PropertiesTests extends AITestClass {
...
+                    sessionCookiePostfix: "postfix2",
+                    userCookiePostfix: "usercookie2",
+                    idLength: 26,
+                    getNewId: (idLength?: number) => {
+                        return "" + (idLength || 0) + 1;
+                    }
+                } as IPropertiesConfig;
+
+                core.config.extensionConfig =  core.config.extensionConfig?  core.config.extensionConfig : {};
+                core.config.extensionConfig[id] = newConfig;
```
MSNev, line 523: This is the "Key" change from the previous test, where this will cause the properties of `newConfig` to be copied to `core.config.extensionConfig[id]` rather than getting directly assigned (and overwriting) the previous value.
This occurs because the `extensionConfig` and its properties are marked as referenced (using the new `ref()`) as part of the dynamic configuration.

This is required because the PropertiesPlugin passes down it's configuration to its supporting components, and if this allowed the config to be "replaced" those support components would never be notified or receive the "updated/changed" configuration values (because they would be a different object instance)

extensions/applicationinsights-properties-js/Tests/Unit/src/properties.tests.ts
```diff
@@ -962,7 +1180,6 @@ export class PropertiesTests extends AITestClass {
 
     private getTelemetryConfig(): IPropertiesConfig {
         return {
-            instrumentationKey: "",
```
MSNev, line 965: This should never have been on the `IPropertiesConfig` as it's never used or referenced from here, it's only ever used from the root (`IConfiguration`) definition -- So the removal is just removing dead code (defaults) and definitions.

extensions/applicationinsights-properties-js/src/Context/Internal.ts
```diff
@@ -37,11 +37,13 @@ export class Internal implements IInternal {
...
-    constructor(config: IPropertiesConfig) {
+    constructor(config: IPropertiesConfig, unloadHookContainer?: IUnloadHookContainer) {
         
-        onConfigChange((config), () => {
+        let unloadHook = onConfigChange((config), () => {
             let prefix =  config.sdkExtension;
             this.sdkVersion = (prefix ? prefix + "_" : "") + "javascript:" + Version;
         });
+
+        unloadHookContainer && unloadHookContainer.add(unloadHook);
```
MSNev, line 47: We always need to "register" (`add()`) the returned value from the `onConfigChange()` because it has internal side-effects that indirectly cause the callback function (the listener) to be associated with and referenced configuration item. And if the `unloadHook` is never called all of these internal references would cause this class instance to never be garbage collected, which would result in a "memory leak" (because of the possibility of multiple instances getting left behind as referenced -- lots of assumptions in this statement and this edge case)

extensions/applicationinsights-properties-js/src/Context/Session.ts
```diff
@@ -51,18 +51,33 @@ export class _SessionManager {
...
+        let _sessionRenewalMs: number;
 
         dynamicProto(_SessionManager, self, (_self) => {
    
             if (!config) {
                 config = ({} as any);
             }
     
+            let unloadHook = onConfigChange(config, (details) => {
+                _sessionExpirationMs = config.sessionExpirationMs || ACQUISITION_SPAN;
```
MSNev, line 70: Just converting these back to local variables to avoid function calls on every usage (this was what the TODO was about)

shared/AppInsightsCore/Tests/Unit/src/DynamicConfig.Tests.ts
```diff
@@ -553,6 +561,196 @@ export class DynamicConfigTests extends AITestClass {
             }
         });
 
+        this.testCase({
```
MSNev, line 564: Additional tests to validate the new `ref()` and `rdOnly()` functionality for dynamic configuration items.

shared/AppInsightsCore/src/Config/DynamicProperty.ts
```diff
@@ -124,23 +144,35 @@ function _makeDynamicProperty<T, C, V = any>(state: _IDynamicConfigHandlerState<
...
+            if (inPlace || rdOnly) {
+                getter = _getOwnPropGetter(target, name);
+            }
         } else {
-            // Looks like it's already dynamic and a different value so just assign the new value
+            // Looks like it's already dynamic just assign the new value
             target[name] = value;
         }
+
+        // Assign the optional flags if true
```
MSNev, line 168: By-design that once "marked" as referenced or read-only there is no going back (ie. there is no `unref` or `read-write` option).

shared/AppInsightsCore/src/Config/DynamicState.ts
```diff
@@ -11,26 +11,10 @@ import { _IDynamicConfigHandlerState } from "./_IDynamicConfigHandlerState";
 const symPrefix = "[[ai_";
 const symPostfix = "]]";
 
-/**
- * @internal
- * @ignore
- */
-interface _IWaitingHandlers<T> {
```
MSNev, line 18: Not used anymore

shared/AppInsightsCore/src/JavaScriptSDK/BaseTelemetryPlugin.ts
```diff
@@ -215,13 +209,9 @@ export abstract class BaseTelemetryPlugin implements ITelemetryPlugin {
...
         
-            _self._addHook = (hooks: IUnloadHook | IUnloadHook[] | ILegacyUnloadHook | ILegacyUnloadHook[]) => {
-                if (hooks) {
-                    arrAppend(_hooks, hooks);
-                }
-            };
-
             proxyFunctionAs(_self, "_addUnloadCb", () => _unloadHandlerContainer, "add");
+            proxyFunctionAs(_self, "_addHook", () => _hookContainer, "add");
+            objDefine(_self, "_unloadHooks" as keyof BaseTelemetryPlugin, { g: () => _hookContainer });
```
MSNev, line 214: Using `objDefine` instead of just assigning for
- supporting unload / re-initializing a plugin which will change the value or `_hookContainer` without re-defining in `initialize`
- Using `objDefine` means we can explicitly enforce the `read-only` nature of this property so it can never be externally changed.

Karlie-777 (approved):

PR #1978, opened 2023-01-31
[Beta][Task]16961420: fix throttleMgr tests

Karlie-777:
https://github.com/microsoft/ApplicationInsights-JS/issues/1979


shared/AppInsightsCommon/Tests/Unit/src/ThrottleMgr.tests.ts
```diff
@@ -225,7 +225,7 @@ export class ThrottleMgrTest extends AITestClass {
                     } as IThrottleLimit,
                     interval: {
                         monthInterval: 1,
-                        dayInterval: 31,
+                        dayInterval: 32,
```
Karlie-777, line 228: This should be set to 32 to make sure that it is invalid to trigger throttle any day.

MSNev (approved):

Issue #1979, opened 2023-01-31
[BUG][ThrottleMgr] Incorrectly fires based on the number of days past


MSNev:
The ThrottleMgr tests is currently failing due to a bug in the canThrottle() method.

When the dayInterval is set to number (currently 31)

The canThrottle() is calculating the dayCheck based on the current date and is not taking the previous date into consideration (like it is doing for the month)

                let curDate = _getThrottleDate();
                let date = localStorageObj.date;
                let interval = config.interval;
                let monthExpand = (curDate.getUTCFullYear() - date.getUTCFullYear()) * 12 + curDate.getUTCMonth() - date.getUTCMonth();
                let monthCheck = _checkInterval(interval.monthInterval, 0, monthExpand);
                let dayCheck = _checkInterval(interval.dayInterval, 0, curDate.getUTCDate()) -1;
                return monthCheck >= 0 && dayCheck >= 0 && dayCheck <= config.interval.maxTimesPerMonth;

I believe it should be calculating the difference between the last day and the current day
curDate.getUTCDate() - date.getUTCDate()

PR #1981, opened 2023-02-01
[AI][Task]16961420: fix throttleMgr incorrectly fires based on the number of days past

Karlie-777:
https://github.com/microsoft/ApplicationInsights-JS/issues/1979

### DIFF
--- a/shared/AppInsightsCommon/src/ThrottleMgr.ts
+++ b/shared/AppInsightsCommon/src/ThrottleMgr.ts
@@ -199,7 +199,7 @@ export class ThrottleMgr {
                 let curDate = _getThrottleDate();
                 let date = localStorageObj.date;
                 let interval = config.interval;
-                let monthCheck = 1
+                let monthCheck = 1;
                 if (interval?.monthInterval) {
                     let monthExpand = (curDate.getUTCFullYear() - date.getUTCFullYear()) * 12 + curDate.getUTCMonth() - date.getUTCMonth();
                     monthCheck = _checkInterval(interval.monthInterval, 0, monthExpand);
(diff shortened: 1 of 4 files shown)


shared/AppInsightsCommon/src/ThrottleMgr.ts
```diff
@@ -181,8 +196,15 @@ export class ThrottleMgr {
...
                 let monthExpand = (curDate.getUTCFullYear() - date.getUTCFullYear()) * 12 + curDate.getUTCMonth() - date.getUTCMonth();
                 let monthCheck = _checkInterval(interval.monthInterval, 0, monthExpand);
-                let dayCheck = _checkInterval(interval.dayInterval, 0, curDate.getUTCDate()) -1;
-                return monthCheck >= 0 && dayCheck >= 0 && dayCheck <= config.interval.maxTimesPerMonth;
+                
+                let dayCheck = -1;
+                if (_isSpecificDaysGiven) {
+                    dayCheck = arrIndexOf(interval.daysOfMonth, curDate.getUTCDate());
+                } else {
+                    let dayInternal = _checkInterval(interval.dayInterval, date.getUTCDate(), curDate.getUTCDate());
```
MSNev, line 204: `getUTCDate()` returns the date of the month, so it's not really the "interval", for example if the previous date was the 31st and today is the 2nd with an interval of 2 should cause it to fire.

I think you need to calculate the number of days between the 2 dates (if there is a last date)
Something like.
let diffDays = Math.floor(curDate.getTime() - date.getTime() / 86400000);

Karlie-777: I have extra flag `_ isTriggered` to make sure that we only trigger throttle one time a day, but I don't know if this flag could resolve the issue😊

MSNev (approved):

PR #1993, opened 2023-02-16
[Beta][Task]16954171: Dynamic config for ThrottleMgr

Karlie-777:
use IConfiguration so that throttleMge can be configured from root


shared/AppInsightsCommon/src/ThrottleMgr.ts
```diff
@@ -37,6 +41,20 @@ export class ThrottleMgr {
         let _isSpecificDaysGiven: boolean = false;
 
         _initConfig();
+        objDefine(_self, "config", {
```
Karlie-777, line 44: do need this

MSNev (approved):

PR #1997, opened 2023-02-21
change readme

siyuniu-ms:
change instrumentation key to connection string


siyuniu-ms:
skip applicationinsights-web-snippet readme

siyuniu-ms:
in the root readme, there is a seperate section for connection string, as we use connection string as default now, that section could be removed.

MSNev (approved):

Issue #1996, opened 2023-02-17
[BUG] SharedWorker does not Instrument fetch correctly


MSNev:
@celluj34 has identified as part of issue #1995 that for a Shared worker the InstrumentFunc for the fetch is not working due to checks around `hasOwnProperty` failing for `globalThis.fetch` and the beta usage of `Object.getOwnPropertyDescriptor` also fails.

We need to investigate and add an additional fallback from the `_getOwner()` or equivalent to use either an existence check and/or `in`. We also need to validate that even with these checks that we can correctly hook the relevant function and/or prototype.

This should be fixed in v2.x (if possible) and v3.x.

MSNev:
Assigned to 2.8.11 so we can attempt to address in v2 which will them also get merged to v3 (beta)

PR #2000, opened 2023-02-23
[Beta] [BUG] SharedWorker does not Instrument fetch correctly #1996

MSNev:
- Also add a Shared Worker example

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -438,6 +438,13 @@ module.exports = function (grunt) {
                                         path: "./extensions/applicationinsights-properties-js",
                                         unitTestName: "prop.tests.js"
                                     },
+
+            // Examples
+            "example-shared-worker": {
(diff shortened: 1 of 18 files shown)


MSNev:
This will also need to be merged / cherry-picked into master for a 2.8.11 release, which should be possible.

gruntfile.js
```diff
@@ -438,6 +438,13 @@ module.exports = function (grunt) {
                                         path: "./extensions/applicationinsights-properties-js",
                                         unitTestName: "prop.tests.js"
                                     },
+
+            // Examples
+            "example-shared-worker": {
+                                        autoMinify: false,
+                                        path: "./examples/shared-worker",
+                                        unitTestName: "shared-worker.tests.js"
```
Karlie-777, line 446: maybe need to change the unitTestName here😊

PR #2026, opened 2023-03-29
[Master] Fix examples, Throttle tests and export dependency types from Sku

MSNev:


shared/AppInsightsCommon/Tests/Unit/src/ThrottleMgr.tests.ts
```diff
@@ -286,6 +304,10 @@ export class ThrottleMgrTest extends AITestClass {
                     date.setUTCFullYear(year-1);
                     date.setUTCMonth(11);
                 } else {
+                    let dayOfMonth = date.getDate();
+                    if (dayOfMonth > (daysInMonth[month - 1] + (month === 2 ? (isLeapYear(year - 1) ? 1 : 0 ) : 0 ))) {
+                        date.setDate(daysInMonth[month - 1] + (month === 2 ? (isLeapYear(year - 1) ? 1 : 0 ) : 0 ));
+                    }
                     date.setUTCMonth(month-1);
```
MSNev, line 311: These tests started failing today because Feb 29, 30, 31 don't exist 😄 . So just simple test fixes to set to the last day of the previous month as the same issue would occur for Oct, Dec, May, Jul 31st.

Karlie-777 (approved):

shared/AppInsightsCommon/Tests/Unit/src/ThrottleMgr.tests.ts
Karlie-777: Thanks Nev❤️

Issue #2034, opened 2023-04-06
[BUG] Failed XHR request after ever tracked item when gathered logs exceed maxBatchSizeInBytes while offline


peitschie:
**Description/Screenshot**
I'm using ApplicationInsights-JS from within a cordova-based application, which often operates offline for reasonable periods of time.

I've discovered that when the collected offline logs exceed the configured `maxBatchSizeInBytes` (default is ~10kb), the sender starts trying to post records via XHR after every new log item added. This isn't really expected behaviour, as the `Sender` knows it's currently offline, and should be using the backoff policies.

**Steps to Reproduce**

 - OS/Browser: Any. These profiles were captured with Android 12's WebView class
 - SDK Version [e.g. 22]: 2.8.3
 - How you initialized the SDK: All default settings
- Track ~200 items while being offline (need enough to exceed `maxBatchSizeInBytes`)

**Expected behavior**
When offline, no immediate send batches should be attempted, as they can't possibly succeed.

I suspect a check for the offline state should be added here: https://github.com/microsoft/ApplicationInsights-JS/blob/master/channels/applicationinsights-channel-js/src/Sender.ts#L372

MSNev:
I've just checked the code and the "offline" detection is only used to stop retries, it appears that any event will always be batched for sending and the issue your hitting here is because the `maxBatchSizeInBytes` is being exceeded which effective is causing a flush of the batched events.

To add support for full offline detection and the dropping of "old" batch events is not a small task. Tagging this as an enhancement. Not sure of the timeframe for scheduling something like this as the SDK is primarily designed as an Online SDK....

But this is not the first ask for this.

peitschie:
Would there be any disadvantage to adding an offline check before the forced send?

I'm happy to raise a PR for this... it seems like a fairly safe fix, that eliminates the unnecessary call failure?

MSNev:
That is one of the locations to put the check, but once we do that we start to violate the `maxBatchSizeInBytes` and this could easily lead to memory exhaustion. So I'm not sure it's that simply.

In terms of PR's yes, we are happy to take them.

Just be aware that we are about to release the `beta` (which has now been promoted to `main`) and we will be switching the default branch very shortly from `master` (v2.x) to `main` (v3.x)

peitschie:
Yeah... the `maxBatchSizeInBytes` is an issue, but fortunately (or unfortunately 😅 ) it's an existing issue unaffected by this fix 🙂 

[...]

PR #2036, opened 2023-04-06
[BUG] Don't attempt to send message batch when known to be offline #2034

peitschie:


channels/applicationinsights-channel-js/Tests/Unit/src/Sender.tests.ts
```diff
@@ -22,6 +22,7 @@ export class SenderTests extends AITestClass {
 
         if (this._sender && this._sender.isInitialized()) {
             this._sender.pause();
+            this._sender._buffer.clear();
```
peitschie, line 25: I realised while writing my test that session storage wasn't being cleared between tests 😅

MSNev (approved):

Issue #1419, opened 2020-11-03
Clarification around offline & retry support


jpalmowski:
Hi I'm just seeking some clarification around the behaviour I'm seeing when the browser is offline

1. I have a simple test app that logs an exception on load.  (Angular app, not using plugin)
1. When I'm offline AI stores the exception as expected in SessionStorage[AI_Buffer]
1. After maxBatchInterval AI attempts to send the batch which fails 
1. The failed batch is cleared out of SessionStorage[AI_sentBuffer] but is never added back to SessionStorage[AI_Buffer] for retry

I can see code for exponential back off on retries so I'm assuming this should be a supported scenario but wanted to verify before I debugged any further.

I'm also seeking any information on if/when localStorage is going to be supported.  It seems like it used to exist (https://github.com/microsoft/ApplicationInsights-JS/issues/602) and I can see some lower level util code to support it - Given your architecture is pluggable and the interfaces for session and local storage are the same I'm wondering why this hasn't been included?

My config if it helps:
```
config: {
        instrumentationKey: 'APP_KEY',
        enableAutoRouteTracking: false, // option to log all route changes,
        enableSessionStorageBuffer: true,
        isRetryDisabled: false,
        disableDataLossAnalysis: false
      }

```

MSNev:
Hi @jpalmowski, sorry for the delay. As you have already seen, looking into the code the retry backoff should be kicking in to delay the retry operation (based on your isRetryDisabled: false (which is also the default)). But this check occurs after this one
if ((xhr.status < 200 || xhr.status >= 300) && xhr.status !== 0) {

So if the XHR implementation is returning something that satisifes this check that it's likely why the request is getting dropped.
Do (or can) you get the XHR result code returned (assuming via the F12 browser), an alternative would be to load the debug plugin (https://github.com/microsoft/ApplicationInsights-JS/tree/master/extensions/applicationinsights-debugplugin-js) which listens to the events that might provide some additional information about why the event is getting dropped.

LocalStorage: We currently don't have any scheduled work to bring LocalStorage into the App Insights codebase, so I don't have a current timeframe for when this might occur.

You are correct though in that because of the general design it's possible, and we do have an internal set of extensions that supports both LocalStorage and IndexedDb but at this point it doesn't work with the App Insights Sender. There is a general plan to move these internal plugins into the public space, but again no committed timeframe.

Bort-777:
Hi, @MSNev 
[...]

PR #2037, opened 2023-04-06
[Master] Add local storage-based implementation #1419

peitschie:
I'm aiming for the smallest diff with this to allow local storage.

This can be used like:

**UPDATED**
```
const appInsights = new ApplicationInsights({
  config: {
    enableSessionStorageBuffer: true,
    bufferOverride: { 
      getItem: (logger, key) => localStorage.getItem(key),
      setItem: (logger, key, value) => localStorage.setItem(key, value),
    }
  }
});
```

I wasn't sure if I should expose this on config just yet (e.g., `enableLocalStorageBuffer`). Ideally this eventually becomes a pluggable component, so that a hypothetical `IndexedDb` or completely custom storage mechanism can be inserted trivially. I thought about adding a simple factory method to `IConfig`, but the `ISendBuffer` interface is not available to that package currently.

Open to any feedback on this, including rejection if it's not taking this mechanism in a direction you're comfortable with.

Longer term, I'm starting to poke at how to nicely make the `ISendBuffer` async to allow more flexible storage mechanisms to be built, but enabling local storage seemed a very useful first step regardless.


channels/applicationinsights-channel-js/src/SendBuffer.ts
```diff
@@ -345,3 +348,23 @@ export class SessionStorageSendBuffer extends BaseSendBuffer implements ISendBuf
         // @DynamicProtoStub -- DO NOT add any code as this will be removed during packaging
     }
 }
+
+
+
+/*
+ * Session storage buffer holds a copy of all unsent items in the browser session storage.
+ */
+export class SessionStorageSendBuffer extends StringStorageSendBuffer implements ISendBuffer {
```
peitschie, line 357: If desired, these could be turned into simple factory functions instead. E.g.,
```
export const SessionStorageSendBuffer = (logger: IDiagnosticLogger, config: ISenderConfig) =>
    new StringStorageSendBuffer(logger, config, utlSetSessionStorage, utlGetSessionStorage);
```

This would mean the `setBuffer` argument can be simplified a bit:
```
_self.setBuffer = (bufferFactory: (logger: IDiagnosticLogger, config: ISenderConfig) => ISendBuffer) => {
    _self._buffer = bufferFactory(_self.diagLog(), _self._senderConfig);
}
```

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -283,6 +283,10 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControlsAI {
                     _syncUnloadSender = _fallbackSender;
                 }
             };
+
+            _self.setBuffer = (buffer: new (logger: IDiagnosticLogger, config: ISenderConfig) => ISendBuffer) => {
```
MSNev, line 287: Yeh, we don't (currently) really want to expose this as a public API.

Longer term (hopefully by the end of '23) we have a different approach that we want to take in relation to the buffers.
ie. As part of the internal code that we will be bringing over we abstract the storage out completely so it's not even included in the Sender but rather managed by a pluggable component and using the notifications...

Currently the Sender doesn't send notifications for all types -- so there is an amount of refactoring before we can bring that over.

So as part of this we probably want to consider am internal config type approach.

We had a team meeting about this one as well and we don't want to include this PR into the next release (2.8.12) as we are trying to released that late today / tomorrow. But we are silling for 2.8.13 (which would be next month)

channels/applicationinsights-channel-js/src/SendBuffer.ts
```diff
@@ -185,36 +185,39 @@ export class ArraySendBuffer extends BaseSendBuffer implements ISendBuffer {
...
-export class SessionStorageSendBuffer extends BaseSendBuffer implements ISendBuffer {
+export class StringStorageSendBuffer extends BaseSendBuffer implements ISendBuffer {
     static BUFFER_KEY = "AI_buffer";
     static SENT_BUFFER_KEY = "AI_sentBuffer";
 
     // Maximum number of payloads stored in the buffer. If the buffer is full, new elements will be dropped.
     static MAX_BUFFER_SIZE = 2000;
 
-    constructor(logger: IDiagnosticLogger, config: ISenderConfig) {
+    constructor(logger: IDiagnosticLogger, config: ISenderConfig, setStorage: StorageSetter, getStorage: StorageGetter) {
```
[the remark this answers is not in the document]
MSNev: To support v3 and dynamic configuration support, I believe we change these classes (or at least their usage) quite a bit.

So If we could keep the name the same it would simplify the cherry-pick process at this point

channels/applicationinsights-channel-js/src/Sender.ts
MSNev: The rationale about delaying is that we are also trying to get v3.0.0 out as well and this would require quite a bit of churn for the first release.

In terms of a config approach defining as a `bufferOverride` with a get/set would probably work nicely now and as part of future refactoring...

channels/applicationinsights-channel-js/src/SendBuffer.ts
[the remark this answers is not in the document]
peitschie: Done. I've leveraged the `SessionStorageSendBuffer` as a base class for `LocalStorageSendBuffer`, but tried to keep the impacts minimal there, so hopefully it's trivial to adjust to the new approaches.

channels/applicationinsights-channel-js/src/SendBuffer.ts
peitschie: Have removed this anyways now.

channels/applicationinsights-channel-js/src/Sender.ts
peitschie: I've removed this and gone with your suggestion.

To enable local storage, you now must set `enableSessionStorageBuffer` to `true` AND then specify the `bufferOverride`.

`bufferOverride` is just a simple string for now, but should be easy to extend for the future.

Example usage might be:
```
const appInsights = new ApplicationInsights({
  config: {
    enableSessionStorageBuffer: true,
    bufferOverride: "localStorage"
  }
});
```

peitschie:
I've pushed a full rewrite of this PR @MSNev. Once you're happy with the prod code, I'll add some tests proving the configuration options all behave nicely 🙂 

I can appreciate major releases are many moving parts, so am happy to keep tweaking this until it fits nicely into your release plans!

channels/applicationinsights-channel-js/src/Interfaces.ts
```diff
@@ -29,6 +29,11 @@ export interface ISenderConfig {
      */
     enableSessionStorageBuffer: () => boolean;
 
+    /**
+     * Specify the storage buffer type implementation.
+     */
+    bufferOverride: () => "sessionStorage" | "localStorage" | undefined;
```
MSNev, line 35: I was actually thinking that this would be an object with the getStorage / setStorage functions that would be called (if supplied). This would allow anyone to replace with whatever implementation they wanted.

Like some sort of cached / lazy in-memory storage mechanism etc.

MSNev: As there is "consent" issues with storing in LocalStorage that if we supplied we would need to document and follow.

peitschie: > I was actually thinking that this would be an object with the getStorage / setStorage functions that would be called (if supplied)

Yeah... I definitely agree. I started poking at it, but it felt like a much bigger patch, as ideally the exposed interface needs to support async. 

This simple interface seemed like a reasonable first step as a way to support a few more useful first-class storage mechanisms. In a future PR, I'd imagine putting an interface into the type here as well so this becomes:

```
    bufferOverride: () => "sessionStorage" | "localStorage" | ISendBuffer | undefined;
```

> As there is "consent" issues with storing in LocalStorage that if we supplied we would need to document and follow.

Interesting, this hadn't occurred to me.

So, for the specifics here, is your preference this type becomes:
```
    bufferOverride: () => { getItems: StorageGetter: setItems: StorageSetter } | undefined;
```

It'd be trivial to make that happen 🙂 

In future, I'd like to consider exposing the `ISendBuffer` directly, as various other storage solutions (e.g., indexeddb, or a sqlite implementation) probably want the full interface.

MSNev: > `bufferOverride: () => { getItems: StorageGetter: setItems: StorageSetter } | undefined;`

Yes

> In future, I'd like to consider exposing the ISendBuffer directly, as various other storage solutions (e.g., indexeddb, or a sqlite implementation) probably want the full interface.

Assuming that we bring over the same internal batching that we have, it actually won't be necessary as the ISendBuffer will go away entirely as we would be updating Sender to send notifications instead using the existing [`INotificationListener`](https://github.com/microsoft/ApplicationInsights-JS/blob/master/shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/INotificationListener.ts) with the eventsSent, eventsDiscarded, eventsSentRequest.

Part of this is already exposed in the [@microsoft/1ds-post-js](https://www.npmjs.com/package/@microsoft/1ds-post-js?activeTab=code) package, deep in the HttpManager.ts

peitschie: New usage looks something like:
```
const appInsights = new ApplicationInsights({
  config: {
    enableSessionStorageBuffer: true,
    bufferOverride: { 
      getItem: (logger, key) => localStorage.getItem(key),
      setItem: (logger, key, value) => localStorage.setItem(key, value),
    }
  }
});
```

MSNev: Yep, looks good.

peitschie:
I've added tests exercising the new config paths now.

MSNev (approved):

Issue #2055, opened 2023-04-19
Add ability to disable the pollInternalLogs via config and change to stop using setInterval


MSNev:
When maxMessageLimit to zero don't start the interval timer.

Change this from an interval timer to a timeout which is only started when there is something to be logged this is to better support applications that sleep to reduce power usage

PR #2056, opened 2023-04-20
[Master] Add ability to disable the pollInternalLogs via config and change to stop using setInterval #2055

MSNev:


siyuniu-ms (approved):

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewManager.ts
```diff
@@ -227,9 +231,9 @@ export class PageViewManager {
             };
 
             _self.teardown = (unloadCtx?: IProcessTelemetryUnloadContext, unloadState?: ITelemetryUnloadState) => {
-                if (intervalHandle) {
-                    clearInterval(intervalHandle);
-                    intervalHandle = null;
+                if (queueTimer) {
+                    clearInterval(queueTimer);
```
siyuniu-ms, line 235: Do we need to change this to clearTimeout here?

MSNev: Oh, nice catch, its actually the other way around this should stay as clearInterval but line 68 should be setInterval

MSNev: Correction, as this is the PageViewManager yes this should be clearTimeout.

PR #2064, opened 2023-05-05
[Main] Task 20788238: [AI] Add ApplicationInsights namespace and workaround to support v2 and v3 loaded from the CDN

MSNev:


AISKU/src/Init.ts
```diff
@@ -95,15 +95,6 @@ try {
     //         ApplicationInsights, Telemetry
     //     };
     // }
-
```
MSNev, line 98: This code is no longer needed as it's auto injected as part of the rollup build step for the browser bundles

AISKU/rollup.config.js
```diff
@@ -23,7 +23,7 @@ updateDistEsmFiles(replaceValues, banner, true, true, "dist-es5");
 
 export default createConfig(banner, 
   {
-    namespace: "Microsoft.ApplicationInsights3",
+    namespace: "Microsoft.ApplicationInsights",
```
MSNev, line 26: This is now the setting for the "global" namespace, previously we had some code during the module initialization that would "reassign" (copy a reference) to the original global namespace for backward compatibility.

AISKULight/rollup.config.js
```diff
@@ -23,7 +23,7 @@ updateDistEsmFiles(replaceValues, banner, true, true, "dist-es5");
 
 export default createConfig(banner, 
   {
-    namespace: "Microsoft.ApplicationInsights3",
+    namespace: "Microsoft.ApplicationInsights",
```
MSNev, line 26: Resetting all of the global namespaces back to the standard default "global" name as apart from the AISKU, none of the other browser bundles included the code to "assign" back to the global namespace (which was an unintended bug)

rollup.base.config.js
```diff
@@ -24,26 +24,124 @@ function doCleanup() {
...
+    }
+
+    return {
+        common: common,
+        browser: brTokens.slice(idx),
+        gbl: gblTokens.slice(idx)
+    };
+};
+
+const getIntro = (format, theNameSpace, moduleName, theVersion) => {
```
MSNev, line 61: Simplistically, this uses the rollup `intro` and `outro` config options to "take control" of the prefix and postfix (the code rollup wraps around the generated files) so that we can more efficiently create multiple namespaces (Microsoft.ApplicationInsights and Microsoft.ApplicationInsights__X__ (where X is the major version).

This also allow us to add additional debugging information to assist with supporting environment that "load" multiple major versions.

rollup.base.config.js
```diff
@@ -24,26 +24,124 @@ function doCleanup() {
...
+    let theIntro = "";
+    if (format === "iife" || format === "umd") {
+        let nsTokens = getCommonNamespace(theNameSpace.browser, theNameSpace.gbl);
+        theIntro += "(function (global, factory) {\n";
+        let prefix = "    ";
+        theIntro += prefix + "var undef = \"undefined\";\n";
+        if (format === "umd") {
+            // UMD supports loading via requirejs and 
+            theIntro += prefix + "typeof exports === \"object\" && typeof module !== undef ? factory(exports) :\n";
```
MSNev, line 70: This is the main code that is injected to the top of every browser bundle (we don't run/change this for the main node (package.json) entrypoints -- currently as I don't believe we need this because of the way npm will wrap everything in a closure).

Once minified the new code is only around 200 bytes more, compared with "adding" the code and creating a new entrypoint (and adding code like this to every module) would add around 1k.

The previous UMD format generated by rollup for UMD format look like

```
(function (global, factory) {
    typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
    typeof define === 'function' && define.amd ? define(['exports'], factory) :
    (global = typeof globalThis !== 'undefined' ? globalThis : global || self, factory((global.Microsoft = global.Microsoft || {}, global.Microsoft.ApplicationInsights3 = global.Microsoft.ApplicationInsights3 || {})));
})(this, (function (exports) { 'use strict';
```
This new code changes this to look like

```
(function (global, factory) {
    var undef = "undefined";
    typeof exports === "object" && typeof module !== undef ? factory(exports) :
    typeof define === "function" && define.amd ? define(["exports"], factory) :
    (function(global){
        var nsKey, key, nm, theExports = {}, modName = "es5_ai_3_0_1", msMod="__ms$mod__";
        var mods={}, modDetail=mods[modName]={}, ver="3.0.1";
        var baseNs=global, nsKey="Microsoft", baseNs=baseNs[nsKey]=(baseNs[nsKey]||{});
        // Versioned namespace "Microsoft.ApplicationInsights3"
        var exportNs=baseNs, nsKey="ApplicationInsights3", exportNs=exportNs[nsKey]=(exportNs[nsKey]||{});
        // Global namespace "Microsoft.ApplicationInsights"
        var destNs=baseNs, nsKey="ApplicationInsights", destNs=destNs[nsKey]=(destNs[nsKey]||{});
        var expNsDetail=(exportNs[msMod]=(exportNs[msMod] || {})), expNameVer=(expNsDetail["v"]=(expNsDetail["v"] || []));
        var destNsDetail=(destNs[msMod]=(destNs[msMod] || {})), destNameVer=(destNsDetail["v"]=(destNsDetail["v"] || []));
        (destNsDetail["o"]=(destNsDetail["o"] || [])).push(mods);
        factory(theExports);
        for(var key in theExports) {
            // Always set the imported value into the "export" versioned namespace (last-write wins)
            nm="x", exportNs[key]=theExports[key], expNameVer[key]=ver;
            // Copy over any named element that is not already present (first-write wins)
            typeof destNs[key]===undef ? (nm="n", destNs[key]=theExports[key]) && (destNameVer[key]=ver) : !destNameVer[key] && (destNameVer[key]="---");
            (modDetail[nm] = (modDetail[nm] || [])).push(key);
        }
    })(typeof globalThis !== undef ? globalThis : global || self);
})(this, (function (exports) {
'use strict';
```

The "support" / "debug" additions added as part of this populate a `__ms$mod___` property to both namespaces, where the "global" one (Microsoft.ApplicationInsights in this case) will include a `v` object which enumerates every property name with the version that was applied to the namespace, for "unknown" (if an earlier version of the SDK was already loaded) this is set to `---`. It also contains an `o` which is an array of objects that includes the module name, `n` array with the property keys applied from this named module as well as an optional `x`  array with exported names from this module that where not applied to the global namespace

MSNev: image

rollup.base.config.js
```diff
@@ -157,40 +255,122 @@ export function createConfig(banner, cfg, importCheckNames) {
...
-        browserRollupConfigFactory(banner, importCheckNames, targetType, cfg.namespace, cfg.browser.entryPoint, cfg.browser.outputName, cfg.version, false, 'cjs', '.cjs'),
-        browserRollupConfigFactory(banner, importCheckNames, targetType, cfg.namespace, cfg.browser.entryPoint, cfg.browser.outputName, majorVersion, true, 'iife', '.gbl'),
-        browserRollupConfigFactory(banner, importCheckNames, targetType, cfg.namespace, cfg.browser.entryPoint, cfg.browser.outputName, majorVersion, false, 'iife', '.gbl'),
-        browserRollupConfigFactory(banner, importCheckNames, targetType, cfg.namespace, cfg.browser.entryPoint, cfg.browser.outputName, cfg.version, true, 'iife', '.gbl'),
-        browserRollupConfigFactory(banner, importCheckNames, targetType, cfg.namespace, cfg.browser.entryPoint, cfg.browser.outputName, cfg.version, false, 'iife', '.gbl')
+        nodeUmdRollupConfigFactory(banner, importCheckNames, targetType, cfg.namespace, cfg.node.entryPoint, cfg.node.outputName, false)
     );
 
+    let browserFormats = cfg.browser.formats || [ 
+        { format: 'umd', postfix: '' },
```
MSNev, line 262: Oops!, need to add back the `cjs` option

siyuniu-ms (approved):

Issue #2075, opened 2023-05-15
[BUG] ClickAnalytics throwing errors in console


cadriel:
Having installed the latest `applicationinsights-web` and `applicationinsights-clickanalytics-js` packages, errors are occurring in the console.

<img width="1273" alt="Screenshot 2023-05-15 at 2 30 10 PM" src="https://github.com/microsoft/ApplicationInsights-JS/assets/205520/43a433a1-f4c6-49b8-8c0f-1548f4abc06a">

`caught TypeError: Cannot read properties of undefined (reading 'core')`

Packages and versions installed are;
```
    "@microsoft/applicationinsights-clickanalytics-js": "^3.0.1",
    "@microsoft/applicationinsights-web": "^3.0.1",
```

And a snippet of how I'm installing is;

```ts
export const appInsightsPlugin = {
  install: (app: App, opts: AppInsightsOptions) => {
    const clickPluginInstance = new ClickAnalyticsPlugin()
    const clickPluginConfig = {
      autoCapture: true
    }

    const appInsightsConfig = {
      ...opts.config,
      extensions: [clickPluginInstance],
      extensionConfig: {
        [clickPluginInstance.identifier]: clickPluginConfig
      }
    }

    const appInsights = new ApplicationInsights({
      config: appInsightsConfig
    })

    appInsights.loadAppInsights()
    appInsightsPlugin.setupPageTracking(opts, app)
...
```
 
This is a Vue3 project - and Application Insights seems to run ok when loading without the Click Plugin.

If this is a mis-configuration - I'd expect an error indicating so, or something else to go on. :)

MSNev:
The `trackPageView` action of the PageAction class is referencing a non-existent `_self._clickAnalyticsPlugin.core.track(event);`

MSNev:
Ok, I believe I found the issue and it's caused by the new `onConfigChange` callback when a configuration items is changed which indirectly then causes a new `_pageAction` instance to be created which is passed `this`, however, the `this` won't necessarily be the `ClickAnalyticsPlugin` instace -- it should be `_self`

antondrozd:
@MSNev Is it known when the release with the fix will happen?

MSNev:
I don't have a date yet, but it will be in the next few weeks (if not before), once released to `npm` we remove the `fixed - waiting release` label

MSNev:
[...]

PR #2077, opened 2023-05-16
[BUG] ClickAnalytics throwing errors in console #2075

MSNev:


extensions/applicationinsights-clickanalytics-js/src/ClickAnalyticsPlugin.ts
```diff
@@ -139,7 +139,7 @@ export class ClickAnalyticsPlugin extends BaseTelemetryPlugin {
                     let logger = _self.diagLog();
                     _contentHandler = new DomContentHandler(_config, logger);
                     let metaTags = _contentHandler.getMetadata();
-                    _pageAction = new PageAction(this, _config, _contentHandler, _config.callback.pageActionPageTags, metaTags, logger);
+                    _pageAction = new PageAction(_self, _config, _contentHandler, _config.callback.pageActionPageTags, metaTags, logger);
```
MSNev, line 142: This is the source of the issue in that when the `onConfigChange` is re-called the `this` is undefined which then causes the creation of a new `PageAction` with `undefined` as the ClickAnalytics instance.

I've also looked for a "fixed" other potential issues of the same in the other classes. This doesn't replace "all" of the `this` usages as there are some where the `this` does need to represent the current execution context (ie. `this`)

JacksonWeber (approved):

Issue #2065, opened 2023-05-07
Type 'T' does not satisfy the constraint 'IConfiguration'


JakobMadsen717:
I have angular app that uses "@microsoft/applicationinsights-web": "3.0.0"

When I upgrade to "@microsoft/applicationinsights-web": "3.0.1", I get compile errors (on ng serve/build) of this type:
Error: node_modules/@microsoft/applicationinsights-core-js/types/applicationinsights-core-js.d.ts:905:85 - error TS2344: Type 'T' does not satisfy the constraint 'IConfiguration'.

So I downgrade to 3.0.0 for now where all works fine, but I am unsure if some bug was introduced in applicationinsights-core-js.d.ts in version 3.0.1?

thomasfroehle:
Same here

MSNev:
We have not yet released an AI v3.x compatibly version of the angular plugin
as per https://github.com/microsoft/applicationinsights-angularplugin-js#compatibility-matrix

I'm actually surprised that it is working with v3.0.0 [die to these](https://microsoft.github.io/ApplicationInsights-JS/upgrade/v3_BreakingChanges.html) as when using the angular plugin you are merging them and the angular plugin is not self contained)

We are planning on releasing an update version of the angular (and react) plugins this month (probably towards the end of the month). Which will use the version numbering [discussed here](https://github.com/microsoft/applicationinsights-angularplugin-js/issues/57)

JakobMadsen717:
Just wanted to mention that I am not actually using @microsoft/applicationinsights-angularplugin-js in my angular app.
I am only using @microsoft/applicationinsights-web when I see the compilation error. Only wanted to mention it in case the compilation error when using @microsoft/applicationinsights-web in angular/typescript was unintentional. Feel free to ignore and I can look at things again at a later point in time.

MSNev:
Umm, I have no idea what's going on with the following....

```ts
"@microsoft/applicationinsights-common": "^3.0.1",
"@microsoft/applicationinsights-web": "^3.0.1",
```

Version 3.0.1 now no longer includes references within themselves so they always reference the source packages.
`@microsoft/applicationinsights-web` is re-exporting `IConfiguration`  from `@microsoft/applicationinsights-core-js` and common doesn't seem to reference `IConfiguration`... And all references that I'm finding in the web package are importing the type from core-js (and not the re-exported version)...

MSNev:
[...]

PR #2078, opened 2023-05-17
Type 'T' does not satisfy the constraint 'IConfiguration' #2065

MSNev:


AISKU/Tests/es6-module-type-check/src/main.ts
```diff
@@ -0,0 +1,2 @@
+import { ContextTagKeys } from "@microsoft/applicationinsights-common";
```
MSNev, line 1: Simple example provide by the community which is consuming the generated type(s) and identifying that there was an issue with the type constraints

Karlie-777 (approved):

Karlie-777 (approved):

PR #2091, opened 2023-06-21
rework web snippet

siyuniu-ms:


siyuniu-ms:
The difference between the generated js file vs original js file
1. some comments disappeared
2. func like _handleError is converted to inline func, would be renamed to _handleError_1 in js
3. if (snipConfig.ld < 0) is not supported, so I change to if (snipConfig.ld && snipConfig.ld < 0) 
4. there is an extra line of export{} in the end of the generated js file

tools/applicationinsights-web-snippet/src/type.ts
```diff
@@ -0,0 +1,63 @@
+
+import { IConfig } from "@microsoft/applicationinsights-common";
```
MSNev, line 2: This should be

`import { IConfiguration } from "@microsoft/applicationinsights-web";

tools/applicationinsights-web-snippet/package.json
```diff
@@ -11,20 +11,22 @@
...
-
+    "npm-pack": "npm pack",
+    "ai-min": "grunt snippet-min",
+    "ai-restore": "grunt snippet-restore"
   },
   "license": "MIT",
   "devDependencies": {
     "@types/qunit": "^2.19.3",
     "@microsoft/ai-test-framework": "0.0.1",
+    "@microsoft/applicationinsights-common": "3.0.2",
```
MSNev, line 29: You will need to include "@microsoft/applicationinsights-web" here

tools/applicationinsights-web-snippet/rollup.config.js
```diff
@@ -46,6 +52,39 @@ const nodeUmdRollupConfigFactory = () => {
   return nodeRollupConfig;
 };
 
+const snippetNodeUmdRollupConfigFactory = () => {
+  const nodeRollupConfig = {
+    input: `${snippetInputName}.js`,
+    output: {
+      file: `${distPath}/${snippetOutputName}.js`,
+      banner: banner,
+      format: "umd",
```
MSNev, line 61: We don't want to generate a "snippet.min.js" using the "umd" module, we will probably want to use either "cjs" or "esm"...

cjs adds a "use strict" (which we don't really want)

tools/applicationinsights-web-snippet/src/copy-generated.js
```diff
@@ -0,0 +1,301 @@
+(function (win, doc, snipConfig) {
```
MSNev, line 1: Remove the "copy" files before committing the PR

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -0,0 +1,331 @@
+import { AIConfig, AppInsights, Envelope, Fields, ISnippetConfig, ScriptElement } from "./type";
```
MSNev, line 1: AIConfig should not exist, it should be importing IConfiguration from applicationinsights-web

tools/applicationinsights-web-snippet/src/type.ts
```diff
@@ -0,0 +1,60 @@
...
+    useXhr?: boolean;
+    crossOrigin?: string;
+    onInit?: any;
+    cfg: {
+        connectionString: any;
+    };
+}
+
+export interface AIConfig extends IConfiguration {
+    autoExceptionInstrumented?: boolean;
```
MSNev, line 15: Oh, interesting....

These are not "really" configuration items, but "hacks" to store some information indirectly into the configuration...

tools/applicationinsights-web-snippet/src/type.ts
```diff
@@ -0,0 +1,60 @@
+import { IConfiguration } from "@microsoft/applicationinsights-web";
+export interface ISnippetConfig {
+    src: string;
+    name?: string;
+    ld?: number;
+    useXhr?: boolean;
+    crossOrigin?: string;
+    onInit?: any;
+    cfg: {
```
MSNev, line 9: This should be

`cfg: IConfiguration`

tools/applicationinsights-web-snippet/src/type.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export interface Fields {
+    strIngestionendpoint?: any;
+    endpointsuffix?: any;
+    location?: any;
+    // Add other properties as needed
+  }
+
+export interface Envelope{
```
MSNev, line 26: We should be able to import these from the core instance rather than redefining here -- I don't remember which file.

tools/applicationinsights-web-snippet/src/type.ts
```diff
@@ -0,0 +1,60 @@
...
+            exceptions?: any[];
+            message?: string;
+            properties?: any;
+        },
+        baseType?: any,
+
+    }
+  }
+
+export interface ScriptElement extends HTMLElement{
```
MSNev, line 44: This "should" be defined by TypeScript somewhere. We should not need to define it...

tools/applicationinsights-web-snippet/src/type.ts
```diff
@@ -0,0 +1,60 @@
...
+    src?: any;
+    onload: any;
+    onreadystatechange?: any;
+    readyState?: any;
+  }
+
+export interface AppInsights{
```
MSNev, line 51: again, lets inport the real defined interface rather than redefining here -- we already have way too many `AppInsights` interfaces / classes.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -0,0 +1,331 @@
...
+                hasFullStack: false,
+                stack: message + "\nSnippet failed to load [" + targetSrc + "] -- Telemetry is disabled\nHelp Link: " + helpLink + "\nHost: " + (locn && locn.pathname || "_unknown_") + "\nEndpoint: " + endpointUrl,
+                parsedStack: []
+            }];
+
+            return envelope;
+        }
+    
+        // Assigning these to local variables allows them to be minified to save space:
+        let targetSrc = aiConfig.url || snipConfig.src;
```
MSNev, line 165: aiConfig.url here is really here for backwards compatibility.

Lets use by casting rather than defining a type with this.

`let targetSrc = (aiConfig as any)["url"] || snipConfig.src;

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -0,0 +1,331 @@
...
+                        lineNumber: lineNumber,
+                        columnNumber: columnNumber,
+                        error: error,
+                        evt: win.event
+                    });
+                }
+    
+                return handled;
+            };
+            aiConfig.autoExceptionInstrumented = true;
```
MSNev, line 298: Lets define this in the IConfiguration as it is used via the AnalyticsPlugin.

tools/applicationinsights-web-snippet/src/type.ts
MSNev: autoExceptionInstrumented - Lets define this in the IConfiguration as it is used via the AnalyticsPlugin.
url - Lets remove (I have another comment where it's used)

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -0,0 +1,338 @@
...
+        // Only set if supplied or another name is defined to avoid polluting the global namespace
+        win[sdkInstanceName] = aiName;
+    }
+    let aiSdk = win[aiName] || (function (aiConfig: IConfiguration) {
+        let loadFailed = false;
+        let handled = false;
+        let appInsights: (Snippet & {initialize:boolean, cookie?:any, core?:any})= {
+            initialize: true,   // initialize sdk on download
+            queue: [],
+            sv: "6",            // Track the actual snippet version for reporting.
```
MSNev, line 34: One last thing -- change to "7"

MSNev (approved):

PR #2096, opened 2023-06-28
rollup rewrite

siyuniu-ms:


rollup.base.config.js
```diff
@@ -0,0 +1,472 @@
...
+
+    if (treeshakeSet){
+        treeshakeConfig = treeshakeCfg;
+    } 
+let tasks = [];
+    if (specialV){
+        tasks.push(
+            nodeUmdRollupConfigFactory2(banner, importCheckName, targetType, cfg.namespace, cfg.node.entryPoint, cfg.node.outputName, true, replaceValues, treeshakeConfig),
+            nodeUmdRollupConfigFactory2(banner, importCheckName, targetType, cfg.namespace, cfg.node.entryPoint, cfg.node.outputName, false, replaceValues, treeshakeConfig)
+        )
```
siyuniu-ms, line 417:
```suggestion
        );
```

rollup.base.config.js
```diff
@@ -0,0 +1,472 @@
...
+            es3Check()
+        ]
+    };
+
+
+    if (isProduction) {
+        browserRollupConfig.output.file = prodOutputPath;
+        browserRollupConfig.plugins.push(
+            uglify({
+                ie8: false,
```
Karlie-777, line 71: nit: I am just curious that do we set ie8 to false for this new change?

siyuniu-ms: oh right, that may cause a problem. Thanks for pointing that out. Let's wait for Nev's suggestion.

extensions/applicationinsights-analytics-js/rollup.config.js
```diff
@@ -20,126 +15,20 @@ const replaceValues = {
...
-    input: `dist-esm/${outputName}.js`,
-    output: {
-      file: `browser/${outputName}.js`,
-      banner: banner,
-      format: "umd",
-      name: "Microsoft.ApplicationInsights",
-      extend: true,
-      freeze: false,
-      sourcemap: true
+export default simpleConfig(banner,
```
MSNev, line 20: I don't remember why we have this "special" case (topLevel: true), but lets remove it so that all of the main plugins are the same.

rollup.base.config.js
```diff
@@ -392,81 +347,69 @@ export function createUnVersionedConfig(banner, cfg, importCheckName, replaceVal
...
-                browserRollupConfigFactory(banner, importCheckName, targetType, browserNamespace, cfg.browser.entryPoint, cfg.browser.outputName, noVersion, true, browserFmt, browserPostfix, browserTeam, replaceValues, treeshakeConfig),
-                browserRollupConfigFactory(banner, importCheckName, targetType, browserNamespace, cfg.browser.entryPoint, cfg.browser.outputName, noVersion, false, browserFmt, browserPostfix, browserTeam, replaceValues, treeshakeConfig)
-            );
-        }
-    }
-
-    return tasks;
-}
\ No newline at end of file
+// export function simpleConfig(banner, cfg, importCheckName, replaceValues, treeshakeSet = false) {
```
MSNev, line 358: Lets remove this (after you validate everything) :-)

siyuniu-ms:
the only difference would be in shared/appinsightscore and shared/appinsightsCommon, which is due to the difference of whether to set toplevel and module to be true or false.

MSNev (approved):

PR #2099, opened 2023-07-07
change loader src when detect IE

siyuniu-ms:


tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -171,6 +180,13 @@ import { IConfiguration, Snippet } from "@microsoft/applicationinsights-web";
         // Assigning these to local variables allows them to be minified to save space:
         let targetSrc = (aiConfig as any)["url"] || snipConfig.src
         if (targetSrc) {
+            if (isIE() && targetSrc.indexOf("2") == -1) {
+                // Redrict to version 2 which support IE
+                targetSrc = "https://js.monitor.azure.com/scripts/b/ai.2.min.js";
```
MSNev, line 185: We don't want to completely replace the URL as they "may" also be hosting the SDK on the own endpoint and they also could be using one of the other formats as you have identified....
Another this is that they "could" be using a fixed version and not just ai.2.min.js... or the non minified version

So we would probably want to use a regex for a bit of flexibility with minimal code to replace the version
```ts
if ((isIE() && targetSrc.indexOf("ai.3") !== -1) {
    // This regex matches any URL which contains "\ai.3." but not any full versions like "\ai.3.1" etc
    targetSrc = targetSrc.replace(/(/*\/)(ai\.3\.)([^\d]*)$/, function(all, g1, g2) {
        return g1 + "ai.2" + g2;
    });
}
```

MSNev (approved):

PR #2102, opened 2023-07-14
Internal Task 15757413: [VSCode] [1DS] Move 1DS Core and 1DS Post to GitHub

MSNev:
- Update publishing and version
- Add typedocs


Karlie-777 (approved):

AmeyaLokre:
hey team, where can we find the source code for `microsoft/1ds-post-js@3.2.13` ?

Karlie-777:
hi @AmeyaLokre all previous source code for 1ds post channels are inside our internal repo and they are not public in GitHub.

MSNev:
And for reference the code "mostly" moved as is (ie. the order and functionality of the flow is the same), there were some type refactoring / moving to merge the internal components into this repo, but this did not affect the functionality.

The major difference between v3.x and v4.x (full open sourced) is the support for the dynamic configuration.

If you "really" want to dig into the source code (to validate what I've said above), the npm packages actually "include" the full source code, which was the approach before we took before moving the code here.

See https://www.npmjs.com/package/@microsoft/1ds-post-js/v/3.2.13?activeTab=code and then in the `src` folder.

AmeyaLokre:
thank you for the explanation @Karlie-777 and @MSNev . Will use the npmjs' "code" tab if we end up needing the source code.

PR #2104, opened 2023-07-18
[main] assign LoadEventStart time to Time 

siyuniu-ms:
test implementing ...


channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
```diff
@@ -148,7 +148,7 @@ function _createEnvelope<T>(logger: IDiagnosticLogger, envelopeType: string, tel
     _setValueIf(envelope, "sampleRate", telemetryItem[SampleRate]);
     if ((telemetryItem[strBaseData] || {}).startTime) {
         envelope.time = toISOString(telemetryItem[strBaseData].startTime);
-    }
+    } // siyu 3.0.3 time = loadEvent time, use startTime to record it
```
MSNev, line 151: This comment is specifically for the page view event, however, this code runs for every event and is the mechanism used for the dependencies (ajax) requests.

So qualifying that since v3.0.3 the pageview event ..(etc).. would be better.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -366,7 +366,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                     _maxBatchInterval = senderConfig.maxBatchInterval;
                 }));
             };
-        
+            // siyu what is this doing?
```
MSNev, line 369: ??

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
```diff
@@ -284,6 +284,13 @@ export class AnalyticsPlugin extends BaseTelemetryPlugin implements IAppInsights
                 if (doc) {
                     pageView.refUri = pageView.refUri === undefined ? doc.referrer : pageView.refUri;
                 }
+
+                // Access the performance timing object
+                const navigationEntries = window.performance.getEntriesByType("navigation") as PerformanceNavigationTiming[];
```
MSNev, line 289: Note: not every runtime has `window`, `window.performance`, the `getEntriesByType()` etc.
So this should use the helpers and check for their existence -- otherwise we will have runtime javascript errors.

eg.
```ts
let perf = getPerformance();
if (perf && perf.getEntriesByType) {
    const navigationEntries = perf.getEntriesByType("navigation");
   ...
} else {
    // We will need to handle this case as well and subtract the `duration` that we would normally have calculated from the `time`

File: extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts [hunk 4]
```diff
@@ -284,17 +284,29 @@ export class AnalyticsPlugin extends BaseTelemetryPlugin implements IAppInsights
                 if (doc) {
                     pageView.refUri = pageView.refUri === undefined ? doc.referrer : pageView.refUri;
                 }
-        
+
+                let perf = getPerformance();
+                if (perf && perf.getEntriesByType && perf.getEntriesByType("navigation")) {
```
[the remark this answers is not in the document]
MSNev: actually you can probably do this

```ts
const navigationEntries = (perf && perf.getEntriesByType && perf.getEntriesByType("navigation"));
if (navigationEntries) {
} else {
...
}

**Review by MSNev — APPROVED** [review]

**Comment by siyuniu-ms** [comment]
01:25.791Z   - current time when calling send page view internal
01:25.332Z   - current time - duration
01:24.552Z   - timeOrigin + loadEventStart
01:24.092Z  - timeOrigin
loadEventStart = duration = 460

File: extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts [hunk 5]
```diff
@@ -284,17 +284,29 @@ export class AnalyticsPlugin extends BaseTelemetryPlugin implements IAppInsights
...
+                 // Access the performance timing object
+                const navigationEntries = (perf && perf.getEntriesByType && perf.getEntriesByType("navigation")); 
+                if (navigationEntries) {
+                    // Get the value of loadEventStart
+                    const navigationEntry = navigationEntries[0] as PerformanceNavigationTiming;;
+                    const loadEventStart = navigationEntry.loadEventStart;
+                    pageView.startTime =  new Date(perf.timeOrigin + loadEventStart);
+                } else {
+                    // calculate the start time manually
+                    pageView.startTime = new Date(new Date().getTime() - pageView.properties.duration);
```
siyuniu-ms, line 298:
```suggestion
                    pageView.startTime = new Date(new Date().getTime() - properties.duration);
```

MSNev: Looking at all of the usages of the sendPageInternal and trackPageView, while generally (when there is a `properties`) the `properties.duration` and `pageView.properties.duration` are the same, it is not always the case.

But reviewing the code I did notice that it is very possible and likely that both `pageView.properties` and `properties` may be undefined. So we should protect against that and also use a fallback, keeping the current loging of the pageView.properties.duration and the primary.

siyuniu-ms: ((properties || paveView.prperties || {}.duration || 0)

MSNev: The tests are failing because of dependencies `rush update --recheck --purge --full` should fix

MSNev:
The AppInsights-JS-SDK-PR (1DS) Validation test is failing because of the size test

```
>> AnalyticsExtensionSizeCheck > Test applicationinsights-analytics-extension deflate size
>> Message: max 23 KB, current deflate size is: 23.01 KB
```

This is because the PR tests "change" the build extension which causes the full version to be embedded into the file (eg. "3.0.3-dev-123456.123" etc which then causes the file size to get exceede.

PR #2107, opened 2023-07-24
[Master][Task]24316375: add support for cfgSync plugin

Karlie-777:


extensions/applicationinsights-cfgsync-js/package.json
```diff
@@ -0,0 +1,64 @@
...
+        "mintest": "grunt cfgsync-mintests",
+        "lint": "tslint -p tsconfig.json",
+        "dtsgen": "api-extractor run --local && node ../../scripts/dtsgen.js \"Microsoft Application Insights idsync plugin\"",
+        "sri": "node ../../tools/subResourceIntegrity/generateIntegrityFile.js",
+        "ai-min": "grunt cfgsync-min",
+        "ai-restore": "grunt cfgsync-restore",
+        "npm-pack": "npm pack"
+    },
+    "devDependencies": {
+        "@nevware21/ts-async": ">= 0.1.0 < 2.x",
```
MSNev, line 31: We can't use this either as it's ES5+ and master needs to support ES3, especially for this case.

Karlie-777: I added it for testing only! it won't be used in any non-testing files!

Karlie-777: Here is another one for dependency plugin and it is also used for testing https://github.com/microsoft/ApplicationInsights-JS/blob/3a9fc30c73c99f394af3817b9c8c88fa937bed68/extensions/applicationinsights-dependencies-js/package.json#L30

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -0,0 +1,419 @@
...
+                            if (response.ok) {
+                                response.text().then((text) => {
+                                    _doOnComplete(oncomplete, response.status, text, isAutoSync);
+                                })
+                            } else {
+                                _doOnComplete(oncomplete, response.status, null, isAutoSync);
+                            }
+
+                        }).catch((error) => {
+                            // eslint-disable-next-line no-empty
```
MSNev, line 250: This will still need to call the on complete otherwise the timer won't get re-started

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -0,0 +1,419 @@
...
+                        // eslint-disable-next-line no-empty
+                    }
+                }
+            }
+
+            function _xhrSender(url: string, oncomplete: OnCompleteCallback, isAutoSync?: boolean) {
+                try {
+                    let xhr = new XMLHttpRequest();
+                    xhr.open(STR_GET_METHOD, url);
+                    xhr.onreadystatechange = () => {
```
MSNev, line 263: This would also need the onerror for the same reason as the fetch (probably the same for main as well)

PR #2113, opened 2023-07-26
[Main][Task]16238553: Provide an override option for the Sender 

Karlie-777:


shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IXHROverride.ts
```diff
@@ -0,0 +1,47 @@
+import { IPromise } from "@nevware21/ts-async";
+
+// /** IPayloadData describes interface of payload sent via POST channel */
```
MSNev, line 3: It would be ideal if we can keep this so that Post channel and Sender can use the same definitions.

This will mean refactoring the Sender to construct and pass this updated format.

Karlie-777: I will do it! Thanks Nev!

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -325,41 +342,55 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
-                        sendPostFunc = _beaconSender;
+                        httpInterface = _getSenderInterface([TransportType.Beacon], false);
                     }
+
+                    httpInterface = _alwaysUseCustomSend? customInterface : (httpInterface || customInterface);
     
-                    _self._sender = sendPostFunc || _xhrSender;
-    
-                    if (!senderConfig.onunloadDisableFetch && isFetchSupported(true)) {
+                    _self._sender = (payload: string[], isAsync: boolean) => {
```
Karlie-777, line 373: add extra config and let users to choose transports type

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IXHROverride.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/** IPayloadData describes interface of payload sent via POST channel */
+export interface IPayloadData {
+    urlString: string;
+    data: Uint8Array | string;
+    headers?: { [name: string]: string };
+    timeout?: number;
+    disableXhrSync?: boolean;
+    disableFetchKeepAlive?: boolean;
+}
```
Karlie-777, line 11: add send reason to IPayloadData

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -801,44 +922,51 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                 });
         
                 xhr.onreadystatechange = () => {
-                    _self._xhrReadyStateChange(xhr, payload, payload.length);
```
Karlie-777, line 804: tag xhrReadystate and xdr function as deprecated

channels/applicationinsights-channel-js/src/Sender.ts
Karlie-777: not included in this PR, create a task 25251035 for this

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -1266,6 +1394,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
 
     /**
      * xhr state changes
+     * @deprecated
```
Karlie-777, line 1397: when marked as deprecated, is it safe to remove `self._xhrReadyStateChange` and put everything inside  `self._xhrReadyStateChange` to a private function?

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IXHROverride.ts
```diff
@@ -0,0 +1,36 @@
+import { IPromise } from "@nevware21/ts-async";
+import { SendRequestReason } from "../applicationinsights-core-js";
+
+/** IPayloadData describes interface of payload sent via POST channel */
+export interface IPayloadData {
```
MSNev, line 5: Post Channel should now also be updated to reference this along with the enums above.

We may need to "re-export" these from the PostChannel extension to ensure that anyone attempt to reference from there picks up this common instance.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -325,41 +344,48 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
-                        sendPostFunc = _fetchSender;
-                    }
+                    let customInterface = senderConfig.httpXHROverride;
+                    let httpInterface: IXHROverride = null;
+                    let fallbackInterface: IXHROverride = null;
+                    let syncInterface: IXHROverride = null;
+                    httpInterface = _getSenderInterface([TransportType.Xhr, TransportType.Fetch], false);
+                  
+                    // *****************************************************************************************************************
+                    //NOTE: shoud we remove fallback sender?
```
MSNev, line 356: 👍

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -613,12 +639,99 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
+
+            function _getOnComplete(payload: string[], status: number, headers: {[headerName: string]: string;}, response?: string) {
+
+                // ***********************************************************************************************
+                //Note: status code should be discussed
+                if (status === 200) {
+                    _self._onSuccess(payload, payload.length);
+                }
+                if (status === 400) {
+                    _self._onError(payload, response);
```
MSNev, line 688: I think we always need to call one of these??? Not "just" on 200 and 400?

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -613,12 +639,99 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
+                }
+            }
+
+            function _getSender(sendInterface: IXHROverride, payload: string[], isAsync: boolean): void | IPromise<boolean> {
```
MSNev, line 692: This looks like it's actually "sending" (calling the sendPost) the data, so `_getSender` feels like the wrong name, perhaps `_doSend` or something like that?

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -613,12 +639,99 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
+                _onComplete = (status: number, headers: {[headerName: string]: string;}, response?: string) => {
+                    return _getOnComplete(payload, status, headers, response);
+                }
+                let payloadData = _getPayload(payload);
+                
+                let sendPostFunc:  SendPOSTFunction = sendInterface && sendInterface.sendPOST;
+                if (sendPostFunc) {
```
MSNev, line 699: Maybe? also check if the payload data is `null` as the `_getPayload()` may return null.

channels/applicationinsights-channel-js/src/applicationinsights-channel-js.ts
```diff
@@ -1 +1,2 @@
-export { Sender } from "./Sender";
\ No newline at end of file
+export { Sender } from "./Sender";
+export { SendRequestReason, IPayloadData, IXHROverride, OnCompleteCallback, SendPOSTFunction, TransportType } from "@microsoft/applicationinsights-core-js";
```
Karlie-777, line 2: need update import for 1ds post and re-export

channels/applicationinsights-channel-js/src/Sender.ts
Karlie-777: don't need to remove the function. will remove it when we completely disable the function.

channels/applicationinsights-channel-js/src/Sender.ts
Karlie-777: removed `fallbackSender` and use `xhrSender` directly.

**Previous**: 
when using `beaconSender`, if the payload is too large, it will try

- fetch(when disableXhr is true && fetch is supported)
- xhr (other scenarios)

**Current:**

- if payload is too large, use xhr direcltly.

**1DS:**

-  will not use fallback sender, will only send notification.

(so maybe, we should keep `fallbackSender`?)

channels/applicationinsights-channel-js/src/Interfaces.ts
```diff
@@ -94,6 +95,24 @@ export interface ISenderConfig {
...
+
+    /**
+     * [Optional] The HTTP override that should be used to send requests, as an IXHROverride object.
+     * By default during the unload of a page or if the event specifies that it wants to use sendBeacon() or sync fetch (with keep-alive),
+     * this override will NOT be called.
+     * If onunloadDisableFetch and onunloadDisableBeacon are true, You can now change this behavior by enabling the 'alwaysUseXhrOverride' configuration value.
+     * The payload data (first argument) now also includes any configured 'timeout' (defaults to undefined) and whether you should avoid
+     * creating any synchronous XHR requests 'disableXhr' (defaults to false/undefined)
+     */
+    httpXHROverride?: IXHROverride;
```
MSNev, line 107: Please also add the `@since` so that the generated typedoc will have the correct version (when we fully publish the docs)

MSNev (approved):

PR #2123, opened 2023-08-02
[main] test version conflict and wirte documentation

siyuniu-ms:


MSNev:
Lets move this to tests / Manual as it's not really an example that we want people to use.

The code scanning issues look like they are all from ai.test.js (v1 of the SDK), so lets just mark those as won't fix with a description that this is the previously published version that we are using for testing only.

MSNev (approved):

PR #2125, opened 2023-08-03
[Main][Task]24677395:add configurations for CDN to enable specific users opt-in implementation 

Karlie-777:


extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,76 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
     // if errors happen, do nothing
     return cfg;
 }
+
+//                                cdn Mode (cdn Value = B, SDK Config Defaults = C)
+//                   |--------------------------------------------------------------------------|
+//                   |                    | none        | disabled    | enabled     | force     |
+//                   | ------------------ | ----------- | ----------- | ----------- | --------- |
+// | User Mode       | none               | A || C      | A || C      | B || A || C | B || C    |
```
Karlie-777, line 50: user=none, cdn=false: B|| A|| C

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,82 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+//                   | ------------------ | ----------- | ----------- | ----------- | --------- |
+// | User Mode       | none               | A || C      | B || A || C | B || A || C | B || C    |
+// (user Value = A)  | disabled           | A || C      | B || A || C | A || C      | B || C    |
+//                   | enabled            | A || C      | B || A || C | A || B || C | B || C    |
+//                   | none(blockCdn)     | A || C      | A || C      | A || C      | A || C    |
+//                   | disabled(blockCdn) | A || C      | A || C      | A || C      | A || C    |
+//                   | enabled(blockCdn)  | A || C      | A || C      | A || C      | A || C    |
+export function shouldOptInFeature(field: string, cdnCfg?: ICfgSyncCdnConfig, customOptInDetails?: IFeatureOptIn) {
+    
+    if (!cdnCfg || !cdnCfg.enabled || !cdnCfg.featureOptIn || !cdnCfg.featureOptIn[field] || !customOptInDetails || !customOptInDetails[field]) {
```
MSNev, line 58: I think this isn't quite correct...

for example, if the user (or SDK defaults) doesn't provide any value (undefined) for `customOptInDetails[field]` then this wouldn't allow any CDN value to be validated (I think)

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,82 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+        return customFeatureVal;
+    }
+
+    if (cdnMode === CdnFeatureMode.force) {
+        return cdnFeatureVal;
+    }
+    let featureVal = customFeatureVal;
+    if (cdnMode === CdnFeatureMode.disable) {
+        if (!isNullOrUndefined(cdnFeatureVal)) {
+            featureVal = cdnFeatureVal;
```
MSNev, line 80: We could probably just return here and avoid the extra check for the `cdnMode === CdnFeatureMode.enable` on 84 and 90

extensions/applicationinsights-cfgsync-js/src/Interfaces/ICfgSyncCdnConfig.ts
```diff
@@ -0,0 +1,44 @@
...
+export interface ICdnFeatureOptIn {
+    /**
+     * Identifies current cdn opt in mode
+     */
+    mode: CdnFeatureMode,
+    /**
+     * Identifies value should be set for the feature
+     * Default: undefined
+     */
+    value?: any
```
Karlie-777, line 43: featureName: {mode: 4, value: {["throttleMgr.limit.***"]: true}}
change value to object {[field: string]: {[value: string]: any}};

Karlie-777: and have a general throttle config as well

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,76 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+
+    if (blockCdn) {
+        return customFeatureVal;
+    }
+
+    if (cdnMode === CdnFeatureMode.force) {
+        return cdnFeatureVal;
+    }
+
+    if (cdnMode === CdnFeatureMode.disable || (customMode === FeatureOptInMode.none && cdnMode === CdnFeatureMode.enable)) {
```
MSNev, line 80: I know we have this is the matrix above, but... based on the excel spreadsheet we created do / should we just return null when the `cdnMode` is `disabled`

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/newylie/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/newylie/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">



  |   |   | 1 | 2 | 3 | 4
-- | -- | -- | -- | -- | -- | --
  |   |   |   | CDN |   |  
  |   |   | none | disabled / force off | enabled / allow opt in | force on
  | Not Opted In | none | none | disabled | disabled | enabled
A \|\| C | Opt Out | disabled | disabled | disabled | disabled | enabled
  | Opt In | enabled | enabled | disabled | enabled | enabed
  | blocked | none | none | none | none | none
  |   | disabled | disabled | disabled | disabled | disabled
  |   | enabled | enabled | enabled | enabled | enabled



</body>

</html>

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -77,7 +75,7 @@ export class CfgSyncPlugin extends BaseTelemetryPlugin implements ICfgSyncPlugin
 
             // used for V2 to manaully trigger config udpate
             _self.setCfg = (config?: IConfiguration & IConfig) => {
```
Karlie-777, line 77: add config/var to block fetching cdn

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,114 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+//                                | ------------------ | ----------- | ----------- | ----------- | --------- |
+// | User Mode, value = A || C    | none               | none        | disabled    | disabled    | enabled   |
+// (user Value = A)               | disabled           | disabled    | disabled    | disabled    | enabled   |
+// (SDK Defaults = C)             | enabled            | enabled     | disabled    | enabled     | enabled   |
+//                                | none(blockCdn)     | none        | none        | none        | none      |
+//                                | disabled(blockCdn) | disabled    | disabled    | disabled    | disabled  |
+//                                | enabled(blockCdn)  | enabled     | enabled     | enabled     | enabled   |
+
+//                                cdn Mode (cdn feature Value = B, SDK Config Defaults = C (value from cdn config/defaults))
+//                   |--------------------------------------------------------------------------|
```
Karlie-777, line 59: remove C

extensions/applicationinsights-cfgsync-js/src/Interfaces/ICfgSyncConfig.ts
```diff
@@ -18,6 +18,12 @@ export interface ICfgSyncConfig {
      * If set, any core config changes will be ignored.
      */
     cfgUrl?: string;
+
+    /**
+     * Identifies if fetching cdn endpoint should be blocked
+     * Default to false.
+     */
+    blockCdn?: boolean;
```
Karlie-777, line 26: added `blockCdn` config here. Default: `false`. 
If it is set to `true`, fetch/xhr will not make fetching calls

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IFeatureOptIn.ts
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * sets feature opt-in mode
+     * @default undefined
+     */
+    mode?: FeatureOptInMode;
+    /**
+     * set values of current feature
+     * @default undefined
+     */
+    cfgValue?: any;
```
MSNev, line 13: Based on how we are using this now we need to define this as a map now.

`{ [key: string]: any }`

extensions/applicationinsights-cfgsync-js/src/Interfaces/ICfgSyncConfig.ts
Karlie-777: TODO:
currently `blockCdn` is inside `onConfigChange` so it will listen to `core.config` changes.
need to do: (haven't added)
1. true -> false: clear timer
2. false -> true trigger fetch and timer

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -121,6 +122,7 @@ export class CfgSyncPlugin extends BaseTelemetryPlugin implements ICfgSyncPlugin
                 _self._addHook(onConfigChange(config, () => {
                     let ctx = createProcessTelemetryContext(null, config, core);
                     _extensionConfig = ctx.getExtCfg(identifier, _defaultConfig);
+                    _blockCdn = !!_extensionConfig.blockCdn;
```
Karlie-777, line 125: maybe it would be better:
```
let preBlockCdn =  _blockCdn;
 _blockCdn = !!_extensionConfig.blockCdn;
if (preBlockCdn !== _blockCdn) {
    if (!_blockCdn) {
        do _fetch....
    } else {
       do clearTimer()
    }
}


```

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,114 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+    return {
+        [mField]: FeatureOptInMode.disable
+    };
+
+}
+
+
+
+// helper function to get cdn config with opt-in features
+export function applyCdnfeatureCfg(cdnCfg: ICfgSyncCdnConfig, core: IAppInsightsCore) {
```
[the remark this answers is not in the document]
MSNev: This not only allows programmatic enabling / disabling of features we define, but it also lets users define their own with whatever overrides they want.

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,114 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+        if (!cdnCfg.featureOptIn) {
+            return cdnCfg.config;
+        }
+        let optInMap = cdnCfg.featureOptIn;
+        let cdnConfig = cdnCfg.config || {};
+        objForEachKey(optInMap, (key) => {
+            let featureVal = resolveCdnFeatureCfg(key, cdnCfg, core.config.featureOptIn);
+            if (!isNullOrUndefined(featureVal)) {
+                objForEachKey(featureVal, (config, val) => {
+                    setValueByKey(cdnConfig, config, val);
```
Karlie-777, line 146: only `featureOptIn` under `IConfig & IConfiguration` will be overwritten

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,114 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+//                                |                    | none        | disabled    | enabled     | force     |
+//                                | ------------------ | ----------- | ----------- | ----------- | --------- |
+// | User Mode, value = A || C    | none               | none        | disabled    | disabled    | enabled   |
+// (user Value = A)               | disabled           | disabled    | disabled    | disabled    | enabled   |
+// (SDK Defaults = C)             | enabled            | enabled     | disabled    | enabled     | enabled   |
+//                                | none(blockCdn)     | none        | none        | none        | none      |
+//                                | disabled(blockCdn) | disabled    | disabled    | disabled    | disabled  |
+//                                | enabled(blockCdn)  | enabled     | enabled     | enabled     | enabled   |
+
+//                                cdn Mode (cdn Value = B)
```
MSNev, line 58: I went back to the first instance of the table to make sure that we didn't miss anything

```
// This matrix identifies how feature based overrides are selected (if present),
// feature overrides are defined as a map of encoded keys with the value like
// featureOptIn: {
//     "disableUnloadEvent" : {
//         cfg: {
//            "disablePageUnloadEvents": [ "unload" ]
//         }
//     }
// }
//
// ---------------------------------------------------------------------------------
//                                cdn Mode (cdn override value = B)
//                    |------------------------------------------------------------|
//                    | Value mapping      | none    | disabled | enabled | force  |
//                    | ------------------ | --------| ---------| --------| ------ |
// | User Mode        | none               | A       | A        | A || B  | B || A |
// (user or SDK       | disabled           | A       | A        | A       | B || A |
// default Value = A) | enabled            | A       | A        | A || B  | B || A |
//                    | none(blockCdn)     | A       | A        | A       | A      |
//                    | disabled(blockCdn) | A       | A        | A       | A      |
//                    | enabled(blockCdn)  | A       | A        | A       | A      |
//
// Where the "value" is an object (map) of encoded key/values which will be used to override the real configuration
// A = either the user supplied value (via the `config.featureOptIn[name]`) or if not defined by the user the SDK defaults of the same.
// B = The value obtained from the CDN for the named feature
// These are evaluated based on the above matrix to either
// - A (Always the user/sdk value)
// - B (Always the value from the CDN)
// - A || B (If A is null or undefined use the value from the CDN (B) otherwise A)
// - B || A (If B is null or undefined use the user/SDK value otherwise use the value from the CDN)
//
// The result of the value may also be null / undefined, which means there are no overrides to apply when the feature is enabled
```

If we allow the programmatic "enable/disable" feature to provide some additional configuration (ie. for `B` overrides) then we may have to revisit this again. "But" as they will be able to "change" the `A` value in the configuration themselves (including the `featureOptIn` config we can probably avoid it (at least for now) -- although it might make testing easier and maybe even the code...

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,114 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+        return {
+            [mField]: customMode,
+            [vField]: customFeatureVal
+        };
+    }
+
+    if (cdnMode === CdnFeatureMode.force) {
+        return {
+            [mField]: FeatureOptInMode.enable,
+            [vField]: cdnFeatureVal
```
MSNev, line 98: This is the first bit that caused me to create the new table, and looking at the `setValueByKey` if the value used in undefined it "will" replace any existing value with undefined, and while we are "forcing" the feature on we may not be supplying any cdn override values so this probably should be B || A.
`[vField]: cdnFeatureVal || customFeatureVal` // We should be able to use this shortcut once we define this as always an object map.

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
MSNev: Went to add some additional comments about the `disabled` state returning `A`, if we rephrase this matrix as the "overrides" that will be applied to when the feature is "enabled" then for the `disabled` states should this return anything???

Additional thought, should we have 2 sets of overrides available in the `featureOptIn` so that when enabled it applies the "enabled" set and when disabled it applies the "disable" set??? With the `none` state just not applying any overrides....
Then it should always be
- `A[enabled] || B[enabled]`
- `A[disabled] || B[disabled]`
- null

And for the force options
- `B[enabled] || A[enabled]`
- `B[disabled] || A[disabled]`

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
MSNev, line 125: I know that I like to keep the names short, but we probably should identify that this is for fetching the "config" from the cdn and not anything else (like loading the SDK from the Cdn). So maybe `blkCdnCfg` or `disableCdnCfg`

extensions/applicationinsights-cfgsync-js/Tests/Unit/src/cfgsynchelper.tests.ts
```diff
@@ -430,6 +431,539 @@ export class CfgSyncHelperTests extends AITestClass {
...
+                        [field5]: {mode: CdnFeatureMode.force, value: true},
+                        [field6]: {mode: CdnFeatureMode.enable}
+                    },
+                    config: {
+                        maxMessageLimit: 10
+                    }
+                } as ICfgSyncCdnConfig;
+
+                let customFeatureOptIn = {
+                    [field]: {mode: FeatureOptInMode.enable} as IFeatureOptInDetails,
```
Karlie-777, line 921: call cfgValue -> details

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
Karlie-777: `enableCfgVal`, `disableCfgVal` inside map inside `featureOptIn`

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,168 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+// These are evaluated based on the above matrix to either
+// - A (Always the user/sdk value)
+// - B (Always the value from the CDN)
+// - A || B (If A is null or undefined use the value from the CDN (B) otherwise A)
+// - B || A (If B is null or undefined use the user/SDK value otherwise use the value from the CDN)
+//
+// The result of the value may also be null / undefined, which means there are no overrides to apply when the feature is enabled
+const FLD = "featureOptIn.";
+const MODE = ".mode";
+const EFLD = ".cfgValue";
```
Karlie-777, line 82: onCfg offCfg

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,168 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
...
+ 
+    let featureOptIn= cdnCfg.featureOptIn || {};
+    let cdnFeature = featureOptIn[field] || {mode: CdnFeatureMode.none};
+    let cdnMode = cdnFeature.mode;
+    let cdnFeatureVal = cdnFeature.value;
+    let cdnFeatureDisVal = cdnFeature.disableValue;
+    let customOptIn = customOptInDetails || {};
+    let customFeature = customOptIn[field] || {mode: FeatureOptInMode.disable}; // default custom mode is disable
+    let customMode = customFeature.mode;
+    let customFeatureVal = customFeature.cfgValue;
```
Karlie-777, line 99: userFeatureVal

extensions/applicationinsights-cfgsync-js/src/CfgSyncHelperFuncs.ts
```diff
@@ -39,3 +42,168 @@ export function replaceByNonOverrideCfg<T=IConfiguration & IConfig, T1=NonOverri
     // if errors happen, do nothing
     return cfg;
 }
+
+
+//                                                     CDN Mode, value = B (CDN value = B)
+//                                |--------------------------------------------------------------------------|-----------|
+//                                |                    | none        | disabled    | enabled     | forceOn   | forceOff  |
+//                                | ------------------ | ----------- | ----------- | ----------- | --------- | --------- |
+// | User Mode, value = A         | none               | none        | disabled    | disabled    | enabled   | disabled  |
```
MSNev, line 51: CDN "enabled" should be enabled

extensions/applicationinsights-cfgsync-js/src/applicationinsights-cfgsync-js.ts
```diff
@@ -1,8 +1,8 @@
-import { replaceByNonOverrideCfg } from "./CfgSyncHelperFuncs";
+import { applyCdnfeatureCfg, replaceByNonOverrideCfg, resolveCdnFeatureCfg } from "./CfgSyncHelperFuncs";
```
Karlie-777, line 1: remove the helper exports

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -121,6 +122,16 @@ export class CfgSyncPlugin extends BaseTelemetryPlugin implements ICfgSyncPlugin
                 _self._addHook(onConfigChange(config, () => {
                     let ctx = createProcessTelemetryContext(null, config, core);
```
Karlie-777, line 123: listen to sync mode ?

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -77,7 +77,7 @@ export class CfgSyncPlugin extends BaseTelemetryPlugin implements ICfgSyncPlugin
 
             // used for V2 to manaully trigger config udpate
             _self.setCfg = (config?: IConfiguration & IConfig) => {
```
Karlie-777, line 79: add pause(), resume() functions

extensions/applicationinsights-cfgsync-js/Tests/Unit/src/cfgsynchelper.tests.ts
```diff
@@ -430,6 +431,477 @@ export class CfgSyncHelperTests extends AITestClass {
 
         });
```
Karlie-777, line 433: helper functions and plugin is updated.
TODO: should add more tests

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -75,6 +76,14 @@ export class CfgSyncPlugin extends BaseTelemetryPlugin implements ICfgSyncPlugin
                 return _mainConfig;
             }
 
+            _self.pause = () => {
+                _paused = true;
+            }
+
+            _self.resume = () => {
+                _paused = false;
```
MSNev, line 84: These probably should also stop / resume any timer for the CDN lookup (if configured that way)

MSNev (approved):

PR #2131, opened 2023-08-10
[main] Adding round robin retry for snippet script src loading

siyuniu-ms:


tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -183,6 +183,18 @@ declare var cfg:ISnippetConfig;
 
             return envelope;
         }
+
+
+        let cdnEndpoints = [
```
MSNev, line 188: This is ok for initial testing, but eventually we will want to abstract this out (to reduce the size) to just be the domains, so then we would just replace the previous domain with the new one...

This is also required so we 't have to list every format (eg. scripts/b/ai.2.gbl.min.js etc).

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -235,7 +247,18 @@ declare var cfg:ISnippetConfig;
                 return scriptElement;
             }
 
-            let theScript = _createScript();
+            let theScript = _createScript(targetSrc);
+
+            if (appInsights.cr && loadFailed && domains.some(endpoint => targetSrc.indexOf(endpoint) !== -1)){
```
MSNev, line 252: We should not use `some` as it's not supported on IE8, so this would cause the snippet to throw an error and fail.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -235,7 +247,18 @@ declare var cfg:ISnippetConfig;
...
+                for (const currentEndpoint of domains) {
```
MSNev, line 253: We should also use for (lp = 0 ; lp < ...) rather than "of" as some pages frameworks have extensions installed which causes this to not do what we expect (like iterate into other "manually" added functions via the Array.prototype -- again causing this to throw)

tools/applicationinsights-web-snippet/src/snippet.ts
MSNev, line 252: The loadFailed flag is designed to mark the SDK script fetching as 100% failed and the snippet should stop doing everything, so we should not be using this flag.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -235,7 +247,18 @@ declare var cfg:ISnippetConfig;
...
+                    loadFailed = false;
+                    theScript = _createScript("https://" + currentEndpoint + "/scripts/b/ai.2.min.js");
+                    if (!loadFailed) {
```
MSNev, line 256: I'm not sure what is going on here, the `_createScript` will return synchronously, so checking the `loadFailed` here is not going to work properly.

I'm also not following how this would "create" a new script for the next CDN entry, only AFTER the previous one failed. I was expecting that we would need to add some logic to the `_handleError` function so that instead of reporting a failure it would trigger a new script tag to be created at that point and added to the page and only after running out of domains to "try" would it then report the failure.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -183,6 +183,18 @@ declare var cfg:ISnippetConfig;
 
             return envelope;
         }
+
+
+        let domains = [
+            "js.monitor.azure.com",
```
MSNev, line 189: I like this order.

While we don't recommend that they use the az416426 domain anymore we should still have some detection here so that if that domain is used we will still retry with these.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -206,12 +207,24 @@ declare var cfg:ISnippetConfig;
                 });
                 // let message = "Load Version 2 SDK instead to support IE"; // where to report this error?
             }
+            if (appInsights.cr){
+                for (var i = 0; i < domains.length; i++){
+                    if (targetSrc.indexOf(domains[i])){
```
MSNev, line 212: This will always be true even if not match ;-)  as returning -1 (no match) will still be true.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -206,12 +207,24 @@ declare var cfg:ISnippetConfig;
...
+                    }
+             }
             const _handleError = (evt?: any) => {
-                loadFailed = true;
                 appInsights.queue = []; // Clear the queue
                 if (!handled) {
-                    handled = true;
-                    _reportFailure(targetSrc);
+                    // start retry
+                    if (domainRetryIndex > 0 && domainRetryIndex < domains.length){
```
MSNev, line 221: `domainRetryIndex >= 0` perhaps?

But we also want to not retry `js.monitor.azure.com` if that is what they had. Likewise, if they used `az416426...` we don' really want to try that again...

Perhaps, we split this into a retry count and "startIndex" (matching the first domain) and then the "next" script is 
`nextIdx = (startIndex + domainRetryCount) % 6` (where 6 === `domains.length`)

AISKU/src/Snippet.ts
```diff
@@ -15,4 +15,5 @@ export interface Snippet {
     queue?: Array<() => void>;
     sv?: string;
     version?: number;
+    cr?: boolean; // cdn retry would be proceed if ture
```
MSNev, line 18: Once we have tested this, we will (I think) want this to be the default, so then if someone doesn't want to retry they would be required to set this to `false` and then we just change `if (appInsights.cr){` to `if (appInsights.cr !== false){` so that anything other then `false` would cause the retry to occur.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -194,12 +206,33 @@ declare var cfg:ISnippetConfig;
                 });
                 // let message = "Load Version 2 SDK instead to support IE"; // where to report this error?
             }
+
+            if (cfg.cr){
```
siyuniu-ms, line 210: using appinsights.cr here does not work, it does not read from the 
`    src: "https://js3.monitor.azure.com/scripts/b/ai.2.min.js",
    // name: "appInsights", // Global SDK Instance name defaults to "appInsights" when not supplied
    // ld: 0, // Defines the load delay (in ms) before attempting to load the sdk. -1 = block page load and add to head. (default) = 0ms load after timeout,
    // useXhr: 1, // Use XHR instead of fetch to report failures (if available),
    crossOrigin: "anonymous", // When supplied this will add the provided value as the cross origin attribute on the script tag
    // onInit: null, // Once the application insights instance has loaded and initialized this callback function will be called with 1 argument -- the sdk instance (DO NOT ADD anything to the sdk.queue -- As they won't get called)
    cfg: { // Application Insights Configuration
        connectionString: "YOUR_CONNECTION_STRING"
    },
    cr:true
});`

tools/applicationinsights-web-snippet/src/type.ts
```diff
@@ -8,6 +8,7 @@ export interface ISnippetConfig {
     crossOrigin?: string;
     onInit?: any;
     cfg: IConfiguration;
+    cr?: boolean;
```
siyuniu-ms, line 11: add cr here could enable getting cfg.cr in snippet.ts

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -98,7 +98,7 @@ declare var cfg:ISnippetConfig;
             let ingest = conString[strIngestionendpoint];
             let endpointUrl = ingest ? ingest + "/v2/track" : aiConfig.endpointUrl; // only add /v2/track when from connectionstring
 
-            let message = "SDK LOAD Failure: Failed to load Application Insights SDK script (See stack for details)";
+            let message = "SDK LOAD Failure: Failed to load Application Insights SDK script via " + targetSrc + " (See stack for details)";
```
MSNev, line 101: I don't think we need to add the targetSrc here as it gets included in the message details via the `_createException` and `_createInternal`, so this becomes a duplication

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -194,12 +207,29 @@ declare var cfg:ISnippetConfig;
                 });
                 // let message = "Load Version 2 SDK instead to support IE"; // where to report this error?
             }
+
+            if (cfg.cr){
```
MSNev, line 211: I think we want this to be `if (cfg.cr !== false) {` so that when anything else (undefined, true, etc) it will perform the retry

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -194,12 +207,29 @@ declare var cfg:ISnippetConfig;
...
+            }
+
             const _handleError = (evt?: any) => {
-                loadFailed = true;
                 appInsights.queue = []; // Clear the queue
                 if (!handled) {
-                    handled = true;
-                    _reportFailure(targetSrc);
+                    // start retry
+                    if (domainRetryIndex >= 0 && domainRetryCount < domains.length){
```
MSNev, line 224: Nice handling of when they don't use one of our supported domains (I almost missed that), maybe add a comment here so that someone in the future doesn't wonder (too much) why we are checking whether `domainRetryIndex` < 0.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -194,12 +207,29 @@ declare var cfg:ISnippetConfig;
...
+                        let nextIdx = (domainRetryIndex + domainRetryCount + 1) % domains.length;
+                        _createScript("https://" + domains[nextIdx] + "/scripts/b/ai.2.min.js");
```
MSNev, line 226: But this bit should not always default to `/scripts/b/ai.2.min.js` it should always be whatever they supplied as it could also be
- version 3
- one of the other formats (like `ai.2.gbl.min.js`)

MSNev: So maybe a regex like the IE fallback

```
_createScript(targetSrc.replace(/^(.*\/\/)([\w\.]*)(\/.*)$/, function (_all, http, domain, qs) {
    return http + domains[nextIdx] + qs;
});
```

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -194,12 +207,31 @@ declare var cfg:ISnippetConfig;
                 });
                 // let message = "Load Version 2 SDK instead to support IE"; // where to report this error?
             }
+
+            if (cfg.cr != false){
```
MSNev, line 211: This should be `!==` so that it's an exact match, otherwise the value of `cfg.cr` will get coerced to get compared against `false`

MSNev (approved):

PR #2147, opened 2023-09-06
[main] throttle manager in aisku sending message

siyuniu-ms:


AISKU/src/AISku.ts
```diff
@@ -259,8 +258,33 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                        if (!_throttleMgr){ //&& cfgHandler.cfg.disabled
+                            _throttleMgr = new ThrottleMgr(_core);
+                        }
+                        //  && !_config.throttleMgrCfg.disabled
+                        if (_throttleMgr && !_throttleMgr.isReady() && _config.extensionConfig && _config.extensionConfig[_cfgSyncPlugin.identifier]) {
+                            // set ready state to true will automaatically trigger flush()
+                            _throttleMgr.onReadyState(true);
+                        }
+
+                        if (!_iKeySentMessage){ //should not send again
```
MSNev, line 272: This needs to use the new `isFeatureEnabled()` helper to determine if it should check whether it should true to execute this code

AISKU/src/AISku.ts
```diff
@@ -259,8 +258,33 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                            if (!_config.connectionString && _config.messageSwitch?.disableIkeyDeprecationMessage === false) {
+                                _throttleMgr.sendMessage( _eInternalMessageId.InstrumentationKeyDeprecation, "Instrumentation key support will end soon, see aka.ms/IkeyMigrate");
+                                _iKeySentMessage = true;
+                            }
+                        }
+                       
+                        if (sdkSrc && sdkSrc.indexOf("az416426") != -1 && _config.messageSwitch?.disableCdnDeprecationMessage === false) {
+                            _throttleMgr.sendMessage( _eInternalMessageId.CdnDeprecation, "Support for domain az41626 will end soon, use js.monitor.azure.com instead");
+                        }
```
MSNev, line 281: Needs to have a "sent" flag

AISKU/src/AISku.ts
```diff
@@ -463,6 +487,9 @@ export class AppInsightsSku implements IApplicationInsights {
                 properties = null;
                 _sender = null;
                 _snippetVersion = null;
+                _cfgSyncPlugin = null;
+                _throttleMgr = null;
+                _iKeySentMessage = false;
```
MSNev, line 492: Clear the sent flags

AISKU/Tests/Unit/src/ThrottleSentMessage.tests.ts
```diff
@@ -0,0 +1,217 @@
...
+            enableRequestHeaderTracking: true,
+            enableResponseHeaderTracking: true,
+            maxBatchInterval: 2500,
+            disableExceptionTracking: false,
+            enableCorsCorrelation: true,
+            samplingPercentage: 50,
+            convertUndefined: "test-value",
+            disablePageUnloadEvents: [ "beforeunload" ],
+            throttleMgrCfg: {
+                [_eInternalMessageId.InstrumentationKeyDeprecation as number]:tconfig,
```
MSNev, line 52: try using _InternalMessageId to avoid the TS error???

AISKU/Tests/Unit/src/ThrottleSentMessage.tests.ts
```diff
@@ -0,0 +1,244 @@
...
+            disablePageUnloadEvents: [ "beforeunload" ],
+            throttleMgrCfg: {
+                [_eInternalMessageId.InstrumentationKeyDeprecation]:tconfig,
+                [_eInternalMessageId.SnippetUpdate]:tconfig,
+                [_eInternalMessageId.CdnDeprecation]:tconfig
+            },
+            featureOptIn:{
+                ["disableIkeyDeprecationMessage"]: {mode: FeatureOptInMode.enable},
+                ["disableCdnDeprecationMessage"]: {mode: FeatureOptInMode.enable},
+                ["disableSnippetVersionUpdateMessage"]: {mode: FeatureOptInMode.enable}
```
MSNev, line 54: As we have changes the public name to SdkLoader we should update these new names to reflect.

AISKU/Tests/Unit/src/ThrottleSentMessage.tests.ts
```diff
@@ -0,0 +1,244 @@
...
+            samplingPercentage: 50,
+            convertUndefined: "test-value",
+            disablePageUnloadEvents: [ "beforeunload" ],
+            throttleMgrCfg: {
+                [_eInternalMessageId.InstrumentationKeyDeprecation]:tconfig,
+                [_eInternalMessageId.SnippetUpdate]:tconfig,
+                [_eInternalMessageId.CdnDeprecation]:tconfig
+            },
+            featureOptIn:{
+                ["disableIkeyDeprecationMessage"]: {mode: FeatureOptInMode.enable},
```
MSNev, line 52: Lets also make the names smaller so that we don't have to check these really long strings.

I'll also need to read the PR more completely as I was expecting that this would be the other way around.
Where the feature would "enable" the check not have it disable it.

Name suggestions
- iKeyUsage
- CdnUsage
- SdkVerUpdate

shared/AppInsightsCore/src/JavaScriptSDK.Enums/LoggingEnums.ts
```diff
@@ -123,7 +123,9 @@ export const enum _eInternalMessageId {
     InstrumentationKeyDeprecation = 106,
     ConfigWatcherException = 107,
     DynamicConfigException = 108,
-    DefaultThrottleMsgKey = 109
+    DefaultThrottleMsgKey = 109,
+    CdnDeprecation = 110,
+    SnippetUpdate = 111
```
MSNev, line 128: SdkVerUpdate

AISKU/src/AISku.ts
```diff
@@ -152,6 +159,9 @@ export class AppInsightsSku implements IApplicationInsights {
 
             // Will get recalled if any referenced values are changed
             _addUnloadHook(onConfigChange(cfgHandler, () => {
+                if (!_cfgSyncPlugin){
```
MSNev, line 162: nit: As we are basically "setting" (creating a new one every 1st instance) then we can probably just move this into the `_initDefaults()` (no if required) just always create a new instance `_cfgSyncPlugin = new CfgSyncPlugin()` in there.

AISKU/src/AISku.ts
```diff
@@ -224,11 +227,11 @@ export class AppInsightsSku implements IApplicationInsights {
                         if (!isNullOrUndefined(_snippetVersion)) {
                             snippetVer += _snippetVersion;
                         }
-        
                         if (_self.context && _self.context.internal) {
                             _self.context.internal.snippetVer = snippetVer || "-";
                         }
-        
+                        console.log("_self.context.internal.snippetVer", _self.context.internal.snippetVer)
```
MSNev, line 233: Remove any and all `console.log` references from the production code

AISKU/src/AISku.ts
```diff
@@ -245,8 +248,7 @@ export class AppInsightsSku implements IApplicationInsights {
 
                 doPerf(_self.core, () => "AISKU.loadAppInsights", () => {
                     // initialize core
-                    _core.initialize(_config, [ _sender, properties, dependencies, _analyticsPlugin ], logger, notificationManager);
-                    objDefine(_self, "context", {
+                    _core.initialize(_config, [ _sender, properties, dependencies, _analyticsPlugin, _cfgSyncPlugin], logger, notificationManager);                    objDefine(_self, "context", {
```
MSNev, line 251: nit: newline

AISKU/src/AISku.ts
```diff
@@ -259,8 +261,40 @@ export class AppInsightsSku implements IApplicationInsights {
                     _self.emptyQueue();
                     _self.pollInternalLogs();
                     _self.addHousekeepingBeforeUnload(_self);
+
+                    _addUnloadHook(onConfigChange(cfgHandler, () => {
+                        if (!_throttleMgr){
+                            _throttleMgr = new ThrottleMgr(_core);
```
MSNev, line 267: As with the config sync, maybe just move this to line 264 (before the onConfigChange) so that we don't need the if check.

AISKU/src/AISku.ts
```diff
@@ -259,8 +257,31 @@
                     _self.emptyQueue();
                     _self.pollInternalLogs();
                     _self.addHousekeepingBeforeUnload(_self);
+
+                    _addUnloadHook(onConfigChange(_config, () => {
+                        if (!_throttleMgr){ //&& cfgHandler.cfg.disabled
+                            _throttleMgr = new ThrottleMgr(_core);
+                        }
+                        //  && !_config.throttleMgrCfg.disabled
+                        if (_throttleMgr && !_throttleMgr.isReady() && _config.extensionConfig && _config.extensionConfig[_cfgSyncPlugin.identifier]) {
```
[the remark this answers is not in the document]
MSNev: Once you move the `_throttleMgr = new ThrottleMgr(_core)`;  outside of the `onConfigChange` this will either self resolve  or we can just dismiss the alert as it would then be possible for the `_throttleMgr` to be null when the called is called.

AISKU/src/AISku.ts
```diff
@@ -259,8 +261,40 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                            // set ready state to true will automaatically trigger flush()
+                            _throttleMgr.onReadyState(true);
+                        }
+
+                        if (!_iKeySentMessage){ //should not send again
+                            if (!_config.connectionString && !isFeatureEnabled("disableIkeyDeprecationMessage", _config)) {
+                                _throttleMgr.sendMessage( _eInternalMessageId.InstrumentationKeyDeprecation, "Instrumentation key support will end soon, see aka.ms/IkeyMigrate");
```
MSNev, line 277: As the `aka.ms/xxxx` are Microsoft Global, we should make these more unique to us, a bit like the `JsActiveCdn`. Maybe that should be even more unique... like
`JsSdkActiveCdn`, `JsSdkIKeyMigrate`, `JsSdkLdrVer` etc
As other people / teams in the company may also want to call things "Snippet" and in the case of the `IkeyMigrate` this is an Azure Monitor team wide change so it actually affects all SDK's and all users. So claiming might get tricky when other teams also want to do this, by keeping more specific to our team we have a greater level of justification for note being kicked out of our usage 😄

AISKU/src/AISku.ts
```diff
@@ -259,8 +261,40 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                        if (!_throttleMgr){
+                            _throttleMgr = new ThrottleMgr(_core);
+                        }
+                        //  && !_config.throttleMgrCfg.disabled
+                        if (_throttleMgr && !_throttleMgr.isReady() && _config.extensionConfig && _config.extensionConfig[_cfgSyncPlugin.identifier]) {
+                            // set ready state to true will automaatically trigger flush()
+                            _throttleMgr.onReadyState(true);
+                        }
+
+                        if (!_iKeySentMessage){ //should not send again
```
MSNev, line 275: nit: it (looks like) these 2 if's could be combined into a single check rather then the 2 levels..., Likewise with the other ones

AISKU/src/AISku.ts
```diff
@@ -259,8 +261,40 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                            if (!_config.connectionString && !isFeatureEnabled("disableIkeyDeprecationMessage", _config)) {
+                                _throttleMgr.sendMessage( _eInternalMessageId.InstrumentationKeyDeprecation, "Instrumentation key support will end soon, see aka.ms/IkeyMigrate");
+                                _iKeySentMessage = true;
+                            }
+                        }
+                       
+                        if (!_cdnSentMessage){
+                            if (sdkSrc && sdkSrc.indexOf("az416426") != -1 && !isFeatureEnabled("disableCdnDeprecationMessage", _config)) {
+                                _throttleMgr.sendMessage( _eInternalMessageId.CdnDeprecation, "Support for this domain will end soon, https://aka.ms/JsActiveCdn");
```
MSNev, line 284: Using terms like "will end soon" is problematic in the SDK as "when" is "soon" if someone starts using this version of the SDK in 5 years time (in theory) "soon" was several years ago. It's better to be a little bit more generic and just redirect them to the `aka.ms` link which can include dates as the resulting page can include specific timeframes and additional wording.

As an example (for this and the others) it could be phrased with something like
"Cdn Usage Notice, see https://aka.ms/xxxxx"

AISKU/src/AISku.ts
```diff
@@ -259,8 +261,40 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                                _cdnSentMessage = true;
+                            }
+                        }
+                       
+                        if (!_sdkVerSentMessage){
+                            if (parseInt(_snippetVersion) < 6 && !isFeatureEnabled("disableSnippetVersionUpdateMessage", _config)) {
+                                _throttleMgr.sendMessage( _eInternalMessageId.SnippetUpdate, "Snippet ver is updated, see https://aka.ms/SnippetVer");
```
MSNev, line 291: "A newer Sdk Loader is available, see https://aka.ms/xxxxx" or
"An updated Sdk Loader is available, see ...."

AISKU/src/AISku.ts
MSNev: Actually the message on this one I think is consistent with the general support message, we could shorten slightly dropping the "will end soon"

"Instrumentation key support, see aka.ms/....." or maybe
"See Instrumentation key support at aka.ms/....."

AISKU/src/AISku.ts
```diff
@@ -259,8 +261,40 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                            _throttleMgr = new ThrottleMgr(_core);
+                        }
+                        //  && !_config.throttleMgrCfg.disabled
+                        if (_throttleMgr && !_throttleMgr.isReady() && _config.extensionConfig && _config.extensionConfig[_cfgSyncPlugin.identifier]) {
+                            // set ready state to true will automaatically trigger flush()
+                            _throttleMgr.onReadyState(true);
+                        }
+
+                        if (!_iKeySentMessage){ //should not send again
+                            if (!_config.connectionString && !isFeatureEnabled("disableIkeyDeprecationMessage", _config)) {
```
MSNev, line 276: Ok, been through the PR now, lets not call these `disableXXXX` as we have a double negative.

Lets rename to something shorter (because the string won't be compressible" and then change this so it reads

`if(.... && isFeatureEnabled("<theFeatureName>", _config)....`  so it's more of a positive rather than getting into mental gymnastics of "is not feature enabled (disable feature name)"

MSNev: Important!
And the `defaultConfigValues` (up at line 57) should set this version of the SDK defaults, which would have this as disabled by default for this release or at the very least it MUST define the feature names otherwise they will NOT be dynamic and all of the `onConfigChange` settings would never get called if we forced them on at the CDN level and the user did not supply any defaults.

AISKU/src/AISku.ts
```diff
@@ -52,7 +54,12 @@ const defaultConfigValues: IConfigDefaults<IConfiguration> = {
     connectionString: UNDEFINED_VALUE,
     endpointUrl: UNDEFINED_VALUE,
     instrumentationKey: UNDEFINED_VALUE,
-    diagnosticLogInterval: cfgDfValidate(_chkDiagLevel, 10000)
+    diagnosticLogInterval: cfgDfValidate(_chkDiagLevel, 10000),
+    featureOptIn:{
+        ["iKeyUsage"]: {mode: FeatureOptInMode.disable},
+        ["CdnUsage"]: {mode: FeatureOptInMode.disable},
+        ["SdkLoaderVer"]: {mode: FeatureOptInMode.disable}
```
MSNev, line 61: extract to const at top of file like

`const SDK_LOADER_VER = "SdkLoaderVer";`

AISKU/Tests/Unit/src/ThrottleSentMessage.tests.ts
```diff
@@ -0,0 +1,240 @@
...
+            disableFetchTracking: false,
+            enableRequestHeaderTracking: true,
+            enableResponseHeaderTracking: true,
+            maxBatchInterval: 2500,
+            disableExceptionTracking: false,
+            enableCorsCorrelation: true,
+            samplingPercentage: 50,
+            convertUndefined: "test-value",
+            disablePageUnloadEvents: [ "beforeunload" ],
+            throttleMgrCfg: {
```
MSNev, line 46: Add some defaults in AISku for these, so they become dynamic.

MSNev (approved):

AISKU/Tests/Unit/src/ThrottleSentMessage.tests.ts
```diff
@@ -0,0 +1,246 @@
...
+                let loggingSpy = this.sandbox.stub(this._logger, 'throwInternal');
+
+                let config = this.getAi.config;
+
+                // test throttleCfg has controll on message sending
+
+                config.throttleMgrCfg= {[_eInternalMessageId.InstrumentationKeyDeprecation]:tconfig};
+                this.clock.tick(1);
+
+                // TODO: the sequence of these two changes cannot be reversed
```
siyuniu-ms, line 137: if we set featureoptin, then throttleMgrCfg, the message would not be sent out.

Karlie-777 (approved):

PR #2148, opened 2023-09-08
[Main][Task]25113121: Add isFeatureEnable helper function to core

Karlie-777:


shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
```diff
@@ -347,3 +348,14 @@ export function objExtend<T1, T2, T3, T4, T5, T6>(obj1?: T1 | any, obj2?: T2, ob
 }
 
 export const asString = asString21;
+
+export function isFeatureEnable(feature?: string, optInMap?: IFeatureOptIn): boolean {
```
MSNev, line 352: nit: probably should be called `isFeatureEnabled` as this is not "enabling", just checking if it is 😃

MSNev: It would also be "better" if we just passed the `config` (IConfiguration) to this, otherwise everyone how calls will need to have
`isFeatureEnabled("xxx", core.config.featureOptIn)` where the `featureOptIn` cant be minified, and we will have multiple references.

MSNev (approved):

PR #2165, opened 2023-09-25
[main] use proper URL for tracking when fetch is passed an empty string

### LINKED ISSUE #2164: [BUG] Fetch with empty string as first parameter does not include traceparent and does not successfully save dependency to Application Insights
**Description/Screenshot**
Now that #2149 is fixed, fetch calls with an empty string (`""`) succeed.  However, they do not add a `Traceparent` request header, and Application Insights rejects the dependency:
image

**Steps to Reproduce**

- OS/Browser: Windows 10/Chrome 117.0.5938.89
- SDK Version [e.g. 22]: Current snippet from https://learn.microsoft.com/en-us/azure/azure-monitor/app/javascript-sdk?tabs=javascriptwebsdkloaderscript.
- How you initialized the SDK: Javascript snippet

**Expected behavior**
Repro: https://jorupp.github.io/ai-post-test/with-ai.html and open dev tools, network tab.
Click "POST with URL" - as expected, it does a POST (which gets a 405), but it sends `Traceparent`.
Click "POST with empty string" - as expected, it does a POST (which gets a 405), but does not send `Traceparent`.

**Additional context**
I believe the issue is with https://github.com/microsoft/ApplicationInsights-JS/blob/3c8318bfad6cdd59b92ce87b28e86cc281c7cc2a/extensions/applicationinsights-dependencies-js/src/ajax.ts#L1187-L1191

When `input === ""`, it needs to set requestUrl to `window.location.href.split("#")[0]` (I can't find specific documentation that it should work this way, but browsers seem to include query string parameters on the request when the first arg is empty, though they do _not_ include hash parameters).

jorupp:
This allows AI to track the dependency and sets the Traceparent header on the request to the server (if otherwise allowed).

`Traceparent` is now set
image

Fully-populated dependency is sent to ApplicationInsights
image
image

Data is displayed in Application Insights
image

Fixes #2164

### DIFF
--- a/extensions/applicationinsights-dependencies-js/Tests/Unit/src/ajax.tests.ts
+++ b/extensions/applicationinsights-dependencies-js/Tests/Unit/src/ajax.tests.ts
@@ -1670,6 +1670,7 @@ export class AjaxTests extends AITestClass {
                     Assert.equal(1, dependencyFields.length, "trackDependencyDataInternal was called");
                     Assert.ok(dependencyFields[0].dependency.startTime, "startTime was specified before trackDependencyDataInternal was called");
                     Assert.equal(undefined, dependencyFields[0].sysProperties, "no system properties");
+                    Assert.equal(window.location.href.split("#")[0], dependencyFields[0].dependency.target, "Target is captured.");
 
                     // Assert that the HTTP method was preserved
                     Assert.equal(1, fetchCalls.length);
@@ -1732,11 +1733,16 @@ export class AjaxTests extends AITestClass {
                     Assert.ok(dependencyFields[0].dependency.startTime, "startTime was specified before trackDependencyDataInternal was called");
                     Assert.equal(expectedTraceId, dependencyFields[0].sysProperties!.trace.traceID, "system properties traceId");
                     Assert.equal(expectedSpanId, dependencyFields[0].sysProperties!.trace.parentID, "system properties spanId");
+                    Assert.equal(window.location.href.split("#")[0], dependencyFields[0].dependency.target, "Target is captured.");
 
                     // Assert that the HTTP method was preserved
                     Assert.equal(1, fetchCalls.length);
                     Assert.notEqual(undefined, fetchCalls[0].init, "Has init param");
                     Assert.equal("post", fetchCalls[0].init?.method, "Has post method");
+                    let headers:Headers = fetchCalls[0].init.headers as Headers;
+                    Assert.notEqual(undefined, headers, "has headers");
+                    Assert.equal(true, headers.has(RequestHeaders.requestIdHeader), "AI header should be present"); // AI
+                    Assert.equal(true, headers.has(RequestHeaders.traceParentHeader), "W3c header should be present"); // W3C
 
                     testContext.testDone();
                 }, () => {
--- a/extensions/applicationinsights-dependencies-js/src/ajax.ts
+++ b/extensions/applicationinsights-dependencies-js/src/ajax.ts
@@ -15,7 +15,7 @@ import {
     dumpObj, eLoggingSeverity, eventOn, generateW3CId, getExceptionName, getGlobal, getIEVersion, getLocation, getPerformance, isFunction,
     isNullOrUndefined, isString, isXhrSupported, mergeEvtNamespace, onConfigChange, strPrototype, strTrim
 } from "@microsoft/applicationinsights-core-js";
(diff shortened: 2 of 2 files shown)


extensions/applicationinsights-dependencies-js/Tests/Unit/src/ajax.tests.ts
```diff
@@ -1,4 +1,4 @@
-﻿import { SinonStub } from "sinon";
+import { SinonStub } from "sinon";
```
jorupp, line 1: hmm.. this change wasn't in what I staged, and my editor is telling me it's not actually changed.  I'm guessing this is a CR/LF issue I need to sort out.

jorupp: ok, got it fixed by using `git` instead of VSCode to do the interactive add to ignore that hunk.

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -1186,6 +1186,9 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
 
                 if (input instanceof Request) {
                     ajaxData.requestUrl = input ? input.url : "";
+                } else if (input === "" ) {
+                    const location = getLocation();
```
jorupp, line 1190: this seemed to be the way other code was accessing `window.location`

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -1186,6 +1186,9 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
+                    ajaxData.requestUrl = location?.href?.split("#")[0] || input;
```
MSNev, line 1191: We tend to not use the TypeScript `?.` operator as this causes a lot of JavaScript to get generated and we are always attempting to keep the size of the generated JS code to a minimum.

Suggest maybe something like the following so that we handle the case where input.url === ""
```ts
let requestUrl: string;
if (input instanceof Request) {
    requestUrl = (input||{}).url || "";
} else {
    requestUrl = input;
}

// Handle case where either the input (as a string) or the input.url is an empty string
if (requestUrl === "") {
    const location = getLocation();
    if (location && location.href) {
        requestUrl = strSplit(location.href, "#")[0];
    }
}

ajaxData.requestUrl = requestUrl;
```

Where `strSplit` is imported from `@nevware21/ts-utils`

extensions/applicationinsights-dependencies-js/src/ajax.ts
jorupp: well, if i wrote the code, I wouldn't have used `""`, but that's how NextJS 13's app router handles server-actions - I brought it up in [a discussions there](https://github.com/vercel/next.js/discussions/55405#discussioncomment-7006185), but after digging further into the code that appeared to be causing the issue (<https://github.com/vercel/next.js/blob/82eb6e6f4955dc90752a4875ccf86b7ac0c8d03a/packages/next/src/client/components/router-reducer/reducers/server-action-reducer.ts#L58>), it didn't seem that what the Next code was doing was actually contrary to how `fetch` is documented (or actually) works, so I figured the fix was better on this side in case there were other apps/libraries taking advantage of that behavior for `""` with `fetch`.

extensions/applicationinsights-dependencies-js/src/ajax.ts
jorupp: ah, so not a "does it work" thing, but "how much JS gets generated" - got it.

jorupp: oh, that's a good point too re: "what if `input.url === ""`"

jorupp: `new Request("")` appears to be setting `url` to the current URL itself, but it doesn't hurt to handle that here.

jorupp: update pushed

MSNev: > ah, so not a "does it work" thing, but "how much JS gets generated" - got it.

Will need to create a work item to track those down and remove ☹️

MSNev (approved):

MSNev:
@microsoft-github-policy-service

MSNev:
@microsoft-github-policy-service rerun

MSNev:
Going to close and re-open to force the CLA to rerun

Issue #2163, opened 2023-09-25
[BUG] Using App Insights connection string leads to double slash


vincent-fortum:
**Description/Screenshot**

When using an Application Insights connection string, the tracking url contains a double forward slash. This is an issue because it conflicts with our CSP headers:

![Screenshot 2023-09-25 at 12 47 19](https://github.com/microsoft/ApplicationInsights-JS/assets/100686345/1835405d-0903-4579-991c-e4c4eb32a17b)

> Content-Security-Policy: The page’s settings blocked the loading of a resource at https://westeurope-5.in.applicationinsights.azure.com//v2/track (“connect-src”).

Of course we can whitelist the url with the double forward slash but I think this should rather be fixed at the source.

**Steps to Reproduce**

 - OS/Browser: MacOS / Firefox 117.0 (shouldn't matter)
 - SDK Version [e.g. 22]: @microsoft/applicationinsights-web 2.8.9
 - How you initialized the SDK:

```sh
AI_CONNECTION_STRING=$(az monitor app-insights component show --resource-group $RG --app $AI --query connectionString)
# connection string contains IngestionEndpoint=https://westeurope-5.in.applicationinsights.azure.com/
```

```ts
new ApplicationInsights({
  config: {
    connectionString: process.env.AI_CONNECTION_STRING,
    enableAutoRouteTracking: true,
    extensions: [new ReactPlugin()],
  },
})
```

**Expected behavior**

Requests are made with a single forward slash

```diff
- https://westeurope-5.in.applicationinsights.azure.com//v2/track
+ https://westeurope-5.in.applicationinsights.azure.com/v2/track
```

**Additional context**

MSNev:
@siyuniu-ms this seems vaguely familiar can you please check if this has been addressed in later v2.x and v3.x releases and if not we will schedule work to sanitize the inbound connection string.

siyuniu-ms:
The problem could be reproduced in current v3.x. A task is made to fix this bug.

MSNev:
Fixed and released with v3.0.4, waiting for backporting to 2.8.x

PR #2174, opened 2023-10-09
[main] remove double slash for endPointUrl

siyuniu-ms:
for issue #2163

### DIFF
--- a/shared/AppInsightsCommon/src/ConnectionStringParser.ts
+++ b/shared/AppInsightsCommon/src/ConnectionStringParser.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { arrReduce, objKeys } from "@microsoft/applicationinsights-core-js";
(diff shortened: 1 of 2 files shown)


shared/AppInsightsCommon/src/ConnectionStringParser.ts
```diff
@@ -37,6 +37,10 @@ export function parseConnectionString(connectionString?: string): ConnectionStri
 
         // apply the default endpoints
         result.ingestionendpoint = result.ingestionendpoint || DEFAULT_BREEZE_ENDPOINT;
+        
+        if (strEndsWith(result.ingestionendpoint, "/")) {
+            result.ingestionendpoint = result.ingestionendpoint.slice(0,-1);
```
MSNev, line 42: We probably should also do this for the snippet as well, although we won't be able to use `strEndsWith` in the snippet or the string `endsWith` as it's not supported on IE and the snippet needs to work for IE :-(.

nit: I don't normally use `slice` on strings, but it works for this case and it's available on all browsers.

MSNev (approved):

Issue #2162, opened 2023-09-21
[BUG] `window is not defined` at _getStackFromErrorObj


EvgenyWas:
**Description/Screenshot**
I connected the Application Insights JavaScript SDK to my Nuxt 3 app on both sides - server and client, and in my Application Insights events I see this log with the following message for the tracked exception value:
`ReferenceError: window is not defined
    at _getStackFromErrorObj (/app/.output/server/node_modules/@microsoft/applicationinsights-web/dist/es5/applicationinsights-web.js:6331:22)`
    
And in your code I found this line that causes this issue - [link](https://github.com/microsoft/ApplicationInsights-JS/blob/1e369eb7aeebded11a00540df16262e2331ce241/shared/AppInsightsCommon/src/Telemetry/Exception.ts#L159).

I understand it might be happening because I use the Application Insights package for the web and some methods might be executed during SSR, but it's a little bit annoying because this check for the existing window is not correct.

 - OS/Browser: windows x64
 - SDK Version [e.g. 22]: `"@microsoft/applicationinsights-web": "^3.0.2"`
 - How you initialized the SDK:
```typescript
const appInsights = new ApplicationInsightsClient({
    config: {
      connectionString,
      enableAutoRouteTracking: true,
      autoTrackPageVisitTime: true,
      enableCorsCorrelation: true,
      enableUnhandledPromiseRejectionTracking: true,
      enableAjaxPerfTracking: true,
    },
  });
``` 

**Expected behavior**
There is no such a log in Application Insights tracked exceptions. The usage should be isomorphic or at least fault-tolerant for such cases.

**Additional context**
Thanks for your work and I will be glad to update this issue with more information for a speedy solution.

MSNev:
While the code was created to "check" for the existence of `window` before using it, this is occurring because the runtime is complaining about the non-existence of the "global"

```ts
} else if (window && window["opera"] && errorObj[strMessage])
```

So we will need to change this to use the `getWindow()` helper which handles this scenario.

PR #2178, opened 2023-10-13
[main] use helper func to get window

siyuniu-ms:
for issue #2162 
did not have any manual test for it yet


MSNev (approved):

MSNev:
It's probably not going to be possible to create a test for this one (at least for now) as we only have "browser" based testing so `getWindow()` will always return something.

PR #2183, opened 2023-10-19
[main] fix startTime timestamp

siyuniu-ms:
only modify the startTime for first page view telemetry

### DIFF
--- a/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
+++ b/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
@@ -285,22 +285,6 @@ export class AnalyticsPlugin extends BaseTelemetryPlugin implements IAppInsights
                 if (doc) {
                     pageView.refUri = pageView.refUri === undefined ? doc.referrer : pageView.refUri;
                 }
-
-                let perf = getPerformance();
-                // Access the performance timing object
-                const navigationEntries = (perf && perf.getEntriesByType && perf.getEntriesByType("navigation"));
-
-                // Edge Case the navigation Entries may return an empty array and the timeOrigin is not supported on IE
-                if (navigationEntries && navigationEntries[0] && !isUndefined(perf.timeOrigin)) {
-                    // Get the value of loadEventStart
-                    const loadEventStart = (navigationEntries[0] as PerformanceNavigationTiming).loadEventStart;
-                    pageView.startTime =  new Date(perf.timeOrigin + loadEventStart);
-                } else {
-                    // calculate the start time manually
-                    let duration = ((properties || pageView.properties || {}).duration || 0);
-                    pageView.startTime = new Date(new Date().getTime() - duration);
-                }
-
                 let telemetryItem = createTelemetryItem<IPageViewTelemetryInternal>(
                     pageView,
                     PageView.dataType,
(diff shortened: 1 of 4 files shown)


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewManager.ts
```diff
@@ -95,6 +96,24 @@ export class PageViewManager {
...
+    
+                    // Edge Case the navigation Entries may return an empty array and the timeOrigin is not supported on IE
+                    if (navigationEntries && navigationEntries[0] && !isUndefined(perf.timeOrigin)) {
+                        // Get the value of loadEventStart
+                        const loadEventStart = (navigationEntries[0] as PerformanceNavigationTiming).loadEventStart;
+                        pageView.startTime =  new Date(perf.timeOrigin + loadEventStart);
+                    } else {
+                        // calculate the start time manually
+                        let duration = ((customProperties || pageView.properties || {}).duration || 0);
+                        pageView.startTime = new Date(new Date().getTime() - duration);
```
MSNev, line 113: This still needs to be done of ALL other cases, otherwise the reported "time" will be wrong.

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
```diff
@@ -285,22 +285,6 @@ export class AnalyticsPlugin extends BaseTelemetryPlugin implements IAppInsights
...
-                const navigationEntries = (perf && perf.getEntriesByType && perf.getEntriesByType("navigation"));
-
-                // Edge Case the navigation Entries may return an empty array and the timeOrigin is not supported on IE
-                if (navigationEntries && navigationEntries[0] && !isUndefined(perf.timeOrigin)) {
-                    // Get the value of loadEventStart
-                    const loadEventStart = (navigationEntries[0] as PerformanceNavigationTiming).loadEventStart;
-                    pageView.startTime =  new Date(perf.timeOrigin + loadEventStart);
-                } else {
-                    // calculate the start time manually
-                    let duration = ((properties || pageView.properties || {}).duration || 0);
```
MSNev, line 300: Looking at all of the "usages" of the `sentPageViewInternal`, it looks like (once we remove this from here), that we should do this duration update in the `AngularPlugin.ts Line 602` (before the `_pageTracking` calls the `_self.sendPageViewInternal` so that the "time" is still shifted back by the `duration` otherwise the Portal will show the wrong "startTime"

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewManager.ts
MSNev: Otherwise the Portal will show the wrong "startTime"

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
MSNev: Maybe?? (testing required)
We can just keep something here like

```
if (isNullOrUndefined(pageView.startTime)) {
    // calculate the start time manually
    let duration = ((properties || pageView.properties || {}).duration || 0);
    if (duration) {
        pageView.startTime = new Date(new Date().getTime() - duration);
    }
}
```

So that if the previous code (like the line 602 above) doesn't specify any `startTime` then we keep "using" this default, this should also work for the subsequent page views, even with your code below for the `if (!firstPageViewSent){`

siyuniu-ms:
manual test time result:
first page view:
2023-10-20T00:15:01.881Z
page view performance time:
2023-10-20T00:15:04.612Z (this time is not adjusted with duration)

second page view triggered around 3 seconds after:
2023-10-20T00:15:04.888Z
third page view triggered around 4 seconds after:
2023-10-20T00:15:05.787Z

Karlie-777 (approved):

PR #2185, opened 2023-10-24
[main] retrieve reponseText when fetch in HttpManager

siyuniu-ms:


channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -470,6 +470,23 @@ export class HttpManager {
                     requestInit.headers = payload.headers;
                 }
 
+                const handleResponse = (status, headerMap, responseText) => {
```
MSNev, line 473: Do these need type qualifiers?
ie. Is TypeScript happy with these or is just assuming that they are `any`, if it thinks they are `any` then we (should) try and give them specific types. Although the `headerMap` is an object (`any`) always anyway.

MSNev: If TypeScript (if VSCode) inferes the type's then don't worry about defining them.

MSNev (approved):

PR #2186, opened 2023-10-25
[main] offlineListener preparation

siyuniu-ms:


shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -21,6 +37,7 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
     var _navigator = getNavigator();        // Gets the window.navigator or workerNavigator depending on the global
     let _isListening: boolean = false;
     let _onlineStatus: boolean = true;
+    let listenerList = new Set<OfflineCallback>();
```
MSNev, line 40: This should be an array not a Set

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -1,11 +1,27 @@
...
+    readonly isOnline: boolean;
+    readonly rState: eOfflineValue; // runtime state
+    readonly uState: eOfflineValue;// user state
+}
+export declare function OfflineCallback(state?: IOfflineState): void;
 export interface IOfflineListener {
     isOnline: () => boolean;
     isListening: () => boolean;
     unload: () => void;
+    _setOnlineState: (isOnline: boolean) => void;
```
MSNev, line 22: This should not have a "_" prefix as it's a public main function

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -1,11 +1,27 @@
...
+    Unknown = 0,
+    Online = 1,
+    Offline = 2
+}
+export interface IOfflineState {
+    readonly isOnline: boolean;
+    readonly rState: eOfflineValue; // runtime state
+    readonly uState: eOfflineValue;// user state
+}
+export declare function OfflineCallback(state?: IOfflineState): void;
```
MSNev, line 17: Please add comments above all of these to identify what they are / will be used for.

Likewise for all of the functions.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -53,21 +70,20 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
     function _enableEvents(target: any): boolean {
         let enabled = false;
         if (target) {
-            enabled = eventOn(target, "online", _setOnline, _evtNamespace);
+            enabled = eventOn(target, "online", _setOnlineState(true), _evtNamespace);
```
MSNev, line 73: No, there are 2 different states that this class is going to monitor
- runtime state (which is what these events are setting)
- user state (which is what will be set via the `setOnlineState`

Also you are "calling" the `_setOnlineState` here and not "passing a function.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -53,21 +70,20 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
...
 
-    function _setOnline() {
-        _onlineStatus = true;
-    }
-
-    function _setOffline() {
-        _onlineStatus = false;
+    function _setOnlineState(isOnline: boolean){
+        _onlineStatus = isOnline;
+        listenerList.forEach((callback: OfflineCallback) => {
```
MSNev, line 84: This "listenerList" should only be processed "IF" the status changes, not if the state is still considered to be the same.

MSNev: And as `listenerList` should be an array this should be using `arrForEach()`

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -98,9 +114,20 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
         }
     }
 
+    function addListener(callback: OfflineCallback) {
+        listenerList.add(callback);
+        return {
+            rm: () => {
```
MSNev, line 120: As above, the list should be an array and an edge case that has to be handled is that someone "could" call this `rm` function multiple times, so the first time will remove the entry (if present) and subsequent times will do nothing.

On way to do this is to use a temporary variable and then when `rm` is called you just reassign the `rm` function on that variable (that you return) to a function that does nothing.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -1,11 +1,48 @@
...
+    Offline = 2
+}
+/**
+ * This is the interface for the Offline state
+ * runtime state is retrieved from the browser state
+ * user state is set by the user
+*/
+// export interface IOfflineState {
+//     readonly isOnline: boolean;
+//     readonly rState: eOfflineValue; // runtime state
```
siyuniu-ms, line 28: question here: if it is readonly here, how to enable user changing the state?

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -58,26 +101,55 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
...
     }
 
     function _setOffline() {
-        _onlineStatus = false;
+        offlineState.rState = eOfflineValue.Offline;
+        listnerNoticeCheck();
     }
 
-    function _isOnline(): boolean {
+    function _isNavOnline(): boolean {
```
siyuniu-ms, line 146: when should we call this func again except initialize?

shared/AppInsightsCommon/src/Offline.ts
siyuniu-ms: function addListener(callback: OfflineCallback) {
        listenerList.push(callback);
        const rm = (callback: OfflineCallback) => {
            let removed = false;
            return () => {
                if (!removed) {
                    const index = listenerList.indexOf(callback);
                    if (index !== -1) {
                        // Remove the callback from the list
                        listenerList.splice(index, 1);
                        removed = true; // Mark it as removed
                    }
                }
            };
        };
        return rm;
   does this works?

siyuniu-ms: seems like it does not work, vs is complaining the type error

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -1,11 +1,40 @@
...
+    readonly rState: eOfflineValue; // runtime state
+    readonly uState: eOfflineValue; // user state
+}
+
 export interface IOfflineListener {
     isOnline: () => boolean;
     isListening: () => boolean;
     unload: () => void;
+    addListener(callback: OfflineCallback): IUnloadHook;
+    userSetOnlineState: (uState: eOfflineValue) => void;
```
MSNev, line 36: Lets call this `setOnlineState` as we are not letting them set the runtime state.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -20,7 +49,13 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
     let _document = getDocument();
     var _navigator = getNavigator();        // Gets the window.navigator or workerNavigator depending on the global
     let _isListening: boolean = false;
-    let _onlineStatus: boolean = true;
+    let listenerList: OfflineCallback[] = [];
+
+    // Set the initial state
+    let _onlineStatus: boolean = _isNavOnline();
```
MSNev, line 55: the `_onlineStatus` IS the runtime state.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -20,7 +49,13 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
...
+    let rState: eOfflineValue = eOfflineValue.Online;
```
MSNev, line 56: So we don't need (and should not be using) this variable.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -58,26 +93,58 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
...
+    }
+
+    function listnerNoticeCheck(){
+        // we were offline and are now online or we were online and now offline
+        if (_isOnline() !== currentState()) {
+            _onlineStatus = currentState(); // use the resolved state to update
+            // send all the callbacks with the current state
+            listenerList.forEach((callback: OfflineCallback) => {
+                let offlineState: IOfflineState = {
+                    isOnline: _onlineStatus,
```
MSNev, line 117: This should be the result of the `currentState()`

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -58,26 +93,58 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
...
+        }
+        return true; // if both unknown, then we assume the network is good
+    }
+
+    function listnerNoticeCheck(){
+        // we were offline and are now online or we were online and now offline
+        if (_isOnline() !== currentState()) {
+            _onlineStatus = currentState(); // use the resolved state to update
+            // send all the callbacks with the current state
+            listenerList.forEach((callback: OfflineCallback) => {
```
MSNev, line 115: Please use `arrForEach` as we have had issue reported previously when we used it as some 3rd party library included some additional properties to the "array" prototype which broken this pattern.

shared/AppInsightsCommon/src/Offline.ts
MSNev: Looking at the reset of the code, it looks like you can change this to be called `rState` and remove the one below.

shared/AppInsightsCommon/src/Offline.ts
siyuniu-ms: it's updated at line 113 _onlineStatus = currentState();

shared/AppInsightsCommon/src/Offline.ts
MSNev: Ok, finally finished reading through the PR.

This should be called `_currentState` as you are recording / updating this in the current check.
It should also be "set" (by calling the `currentState()`) after you set the `rState` and `uState`

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -58,26 +93,58 @@ export function createOfflineListener(parentEvtNamespace?: string | string[]): I
...
+    function currentState(){
+        if (uState === eOfflineValue.Offline || rState === eOfflineValue.Offline){
+            return false;
+        }
+        return true; // if both unknown, then we assume the network is good
+    }
+
+    function listnerNoticeCheck(){
+        // we were offline and are now online or we were online and now offline
+        if (_isOnline() !== currentState()) {
```
MSNev, line 112: Rather than call the `_isOnline()` method just reference the variable (will be slightly faster) -- but won't change operation.

maybe refactor to
```ts
    let newState = currentState();
    if (_onlineState !== newState) {
       _onlineState = newState;
...
```

shared/AppInsightsCommon/src/Offline.ts
siyuniu-ms: Will it be a boolean or eOfflineValue? As rState is a eOfflineValue while _onlineStatus is a boolean.

siyuniu-ms: got it~ it will be set in listnerNoticeCheck

shared/AppInsightsCommon/src/Offline.ts
siyuniu-ms: I seperate the rstate and onlinestatus, so that they could be send back in the callback without any need of type changing.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,195 @@
...
+    readonly isOnline: boolean;
+    readonly rState: eOfflineValue; // runtime state
+    readonly uState: eOfflineValue; // user state
+}
+
+export interface IOfflineListener {
+    isOnline: () => boolean;
+    isListening: () => boolean;
+    unload: () => void;
+    addListener(callback: OfflineCallback): IUnloadHook;
```
MSNev, line 35: nit: Lets be consistent with the definition, so to be the same as the others
`addListener: (callback: OfflineCallback) => IUnloadHook;`

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,195 @@
...
+    setOnlineState : (uState: eOfflineValue) => void;
+    namespace?: string;
```
MSNev, line 37: We should not be exposing this??
If it's for testing purposes, then don't include in the interface, but we can still include as part of the instance, at which point we should prefix it with an underscore (only in the `createxxx` methods)

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,195 @@
...
+    let _document = getDocument();
+    var _navigator = getNavigator();        // Gets the window.navigator or workerNavigator depending on the global
+    let _isListening: boolean = false;
+    let listenerList: OfflineCallback[] = [];
+
+    // Set the initial state
+
+    // current state would be updated each time rState or uState is changed
+    // it is a resolved value of rState and uState
+    let _currentState: boolean = _isNavOnline();
```
MSNev, line 58: This should be initialized from the calculated version `currentState()` and only after assigning the `uState` and `rState`

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,195 @@
...
+    // rState is changed by the browser, both via events and when we check the navigator.onLine property
+    let rState: eOfflineValue = eOfflineValue.Online;
+    // ustate is changed by the user calling setOnlineState
+    let uState: eOfflineValue = eOfflineValue.Offline;
```
MSNev, line 62: This should be "Unknown", otherwise this will initialize as "Offline"

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,195 @@
...
+    let _isListening: boolean = false;
+    let listenerList: OfflineCallback[] = [];
+
+    // Set the initial state
+
+    // current state would be updated each time rState or uState is changed
+    // it is a resolved value of rState and uState
+    let _currentState: boolean = _isNavOnline();
+    // rState is changed by the browser, both via events and when we check the navigator.onLine property
+    let rState: eOfflineValue = eOfflineValue.Online;
```
MSNev, line 60: This should be using the _isNavOnline() state, which yes also means that it should be returning the new enum instead of just `true/false`

shared/AppInsightsCommon/src/Offline.ts
siyuniu-ms: when should the _isNavOnline be called? Seems like it is not being called anywhere else.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,195 @@
...
+        rState = eOfflineValue.Online;
+        listnerNoticeCheck();
+    }
+
+    function _setOffline() {
+        rState = eOfflineValue.Offline;
+        listnerNoticeCheck();
+    }
+
+    function _isNavOnline(): boolean {
```
MSNev, line 147: This looks like it was the old "isOnline()" which is only handling the now `rState`

shared/AppInsightsCommon/src/Offline.ts
siyuniu-ms: seems like we have _evtNamespace, so I removed namespace from the interface

shared/AppInsightsCommon/src/Offline.ts
siyuniu-ms: yes, it is the old isOnline, and now it is for rState init use only

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,195 @@
...
+                if (_enableEvents(target)) {
+                    _isListening = true;
+                }
+            }
+        }
+
+        if (_isListening) {
+            // We are listening to events so lets set the current status rather than assuming we are online #1538
+            if (_navigator && !isNullOrUndefined(_navigator.onLine)) { // navigator.onLine is undefined in react-native
+                _currentState = _navigator.onLine;
```
MSNev, line 85: This should set the "runtime" state

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    function _setOffline() {
+        rState = eOfflineValue.Offline;
+        listnerNoticeCheck();
+    }
+
+    function _isNavOnline(): eOfflineValue {
+        let result = true;
+        if (_isListening) {
+            result = _currentState;
```
MSNev, line 150: rState

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,187 @@
...
+            let target:any = _document.body || _document;
+
+            if (target.ononline) {
+                if (_enableEvents(target)) {
+                    _isListening = true;
+                }
+            }
+        }
+
+        if (_isListening) {
```
MSNev, line 84: I think we can get rid of this now as your setting the `rState` above and we don't want to "reset" the `_currentState` here.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,187 @@
...
+        if (_currentState !== newState) {
+            _currentState = newState; // use the resolved state to update
+            // send all the callbacks with the current state
+            arrForEach(listenerList, (callback: OfflineCallback) => {
+                let offlineState: IOfflineState = {
+                    isOnline: _currentState,
+                    rState: rState,
+                    uState: uState
+                };
+                callback(offlineState);
```
MSNev, line 129: Lets "protect" this from a faulty callback by wrapping this call in a try / catch so that if one of the callbacks fail then we still call all of the remaining.

```ts
try {
    callback(offlineState);
} catch (e) {
    // Do nothing, just making sure we run all of the callbacks
}
```

MSNev (approved):

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,184 @@
...
+                    callback(offlineState);
+                } catch (e) {
+                    // Do nothing, just making sure we run all of the callbacks
+                }
+            });
+        }
+    }
+
+    function setOnlineState (uState: eOfflineValue){
+        this.uState = uState;
```
MSNev, line 132: Actually the `this` is not required.

shared/AppInsightsCommon/src/Offline.ts
```diff
@@ -0,0 +1,184 @@
...
+                try {
+                    callback(offlineState);
+                } catch (e) {
+                    // Do nothing, just making sure we run all of the callbacks
+                }
+            });
+        }
+    }
+
+    function setOnlineState (uState: eOfflineValue){
```
MSNev, line 131: So the argument should be called `newState` as you can't assign `uState = uState`

shared/AppInsightsCommon/src/Offline.ts
MSNev: Do we have any tests for this as any tests should be failing.

siyuniu-ms: we do not have any test on this new code yet, maybe we could merge this into karlie's offline branch instead of main branch?

MSNev (approved):

shared/AppInsightsCommon/src/Offline.ts
siyuniu-ms: the current test is failing because of dep, I will rerun the rush update

MSNev: Lets just push into `main` as we are now not planning any more releases from this branch until this feature set is complete.

If we need to release a version 3.0.5 we will create a branch for that release.

MSNev (approved):

MSNev (changes requested):
Lets wait until we do the patch release before committing this, so that we don't have to branch.

MSNev (approved):
ready to go

PR #2190, opened 2023-10-31
[Main][Task]25648004: Export ISenderConfig from Sender

Karlie-777:


channels/applicationinsights-channel-js/src/applicationinsights-channel-js.ts
```diff
@@ -1 +1,2 @@
 export { Sender } from "./Sender";
+export { ISenderConfig } from "./Interfaces";
```
MSNev, line 2: We may also need to export from the 2 SKU's (main AISKU and the AISKULight) based on Logans' comment about using the basic SKU

MSNev (approved):

PR #2191, opened 2023-10-31
[Main][Task]25649658: Update AISKU Light to add default baseType and baseData when calling track function

Karlie-777:


AISKULight/src/index.ts
```diff
@@ -54,6 +54,7 @@ export class ApplicationInsights {
             _initialize();
           
             _self.initialize = _initialize;
+            _self.track = _track;
         
             proxyFunctions(_self, core, [
                 "track",
```
MSNev, line 60: You should remove this as this (might) "replace" the track you assigned on line 57.

Karlie-777: just about to move the proxy!

MSNev (approved):

PR #2193, opened 2023-11-02
[main] Add extra config in sender to let users define transports type

siyuniu-ms:


channels/applicationinsights-channel-js/src/Interfaces.ts
```diff
@@ -115,6 +115,12 @@ export interface ISenderConfig {
      * @since 3.0.4
      */
     alwaysUseXhrOverride?: boolean;
+
+      /**
+     * [Optional] Either an array or single value identifying the requested TransportType type that should be used.
+     * This is used during initialization to identify the requested send transport, it will be ignored if a httpXHROverride is provided.
+     */
+      transports?: number | number[];
```
MSNev, line 123: Actually, looking at the PostChannel we also have an `unloadTransports` that we should add here as well.

channels/applicationinsights-channel-js/Tests/Unit/src/Sender.tests.ts
```diff
@@ -1091,63 +1091,496 @@ export class SenderTests extends AITestClass {
             }
         });
 
+        
         this.testCase({
-            name: 'BeaconAPI is not used when isBeaconApiDisabled flag is false but payload size is over 64k, fall off to xhr sender',
-            useFakeTimers: true,
+            name: 'Unload Transport Type: User provide fetch in unloadtransports, but it is disabled, so we should use beacon',
             test: () => {
+                let window = getGlobalInst("window");
```
MSNev, line 1098: FYI: There is also a `getWindow()` function that does this internally.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -348,21 +360,30 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                     let customInterface = senderConfig.httpXHROverride;
                     let httpInterface: IXHROverride = null;
                     let syncInterface: IXHROverride = null;
-                    httpInterface = _getSenderInterface([TransportType.Xhr, TransportType.Fetch], false);
+
+                    // Prefix any user requested transport(s) values
+                    let theTransports: TransportType[] = _prependTransports([TransportType.Xhr, TransportType.Fetch], senderConfig.transports);
+
+                    if (senderConfig.isBeaconApiDisabled || !isBeaconsSupported()){
```
MSNev, line 367: Lets not do all of this here as it's a lot of code for no real  gain.

Please add an additional if check `senderConfig.isBeaconApiDisabled` to the `_getSenderInterface` for it to select the `beacon` transport. 
This will also help with the bundle size (one of the reasons the tests are failing), the above will be a lot smaller.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -348,21 +360,30 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
                     let xhrInterface = { sendPOST: _xhrSender} as IXHROverride;
                     _xhrSend = (payload: string[], isAsync: boolean) => {
                         return _doSend(xhrInterface, payload, isAsync);
                     };
     
-    
-                    if (!senderConfig.isBeaconApiDisabled && isBeaconsSupported()) {
-                        // Config is set to always used beacon sending
+                    // prioritize beacon over fetch and xhr
+                    if (!senderConfig.isBeaconApiDisabled && isBeaconsSupported() && !senderConfig.transports) {
```
MSNev, line 380: Once you add the check to the `_getSenderInterface` for the `isBeaconApiDisabled` then we can probably simplify this as well. 
We should also check whether they have anything in the `transports` as well as they "could" pass an empty transports...

Maybe just move all of this to above line 365, to "prepend" the beacon if they have not set the `isBeaconApiDisabled` so that if they have provided any transports they will be added first.

AISKULight/Tests/Unit/src/AISKULightSize.Tests.ts
```diff
@@ -2,8 +2,8 @@ import { AITestClass, Assert } from "@microsoft/ai-test-framework";
 import * as pako from "pako";
 
 export class AISKULightSizeCheck extends AITestClass {
-    private readonly MAX_RAW_SIZE = 81;
-    private readonly MAX_BUNDLE_SIZE = 81;
+    private readonly MAX_RAW_SIZE = 82;
```
MSNev, line 5: Is this change still needed, looking back the previous test pass seem to pass without this??
After you refactored the code to make it smaller.

If so then it's fine.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -655,7 +679,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                         }
                     } else if (transportType === TransportType.Fetch && isFetchSupported(syncSupport)) {
                         sendPostFunc = _fetchSender;
-                    } else if (isBeaconsSupported() && transportType === TransportType.Beacon) {
+                    } else if (transportType === TransportType.Beacon && (syncSupport && _beaconOnUnloadSupported || !syncSupport && _beaconNormalSupported)) {
```
MSNev, line 682: operator precedence may not do what you expect here, lets add some additional brackets /  rework

`((syncSupport && _beaconOnUnloadSupported) || (!syncSupport && _beaconNormalSupported))`
or
`(syncSupport ? _beaconOnUnloadSupported : _beaconNormalSupported)`

AISKULight/Tests/Unit/src/AISKULightSize.Tests.ts
siyuniu-ms: thanks! It was 81.05 in the cloud test. And after refactoring, it goes down.

MSNev (approved):

AISKULight/Tests/Unit/src/AISKULightSize.Tests.ts
siyuniu-ms: Turns out that locally it does not exceed 81, but on the cloud it does.

PR #2200, opened 2023-11-17
[main] [doc] Highlight an issue with the default UMD module format when loading from the CDN into an environment that may have require.js #14493816

siyuniu-ms:


README.md
```diff
@@ -561,9 +561,11 @@ eg. ```var aiSdk = require("@microsoft/applicationinsights-web");```
 
 This situation can also occur when the scripts are loaded lazily, late or dynamically (__and__ RequireJs is present) as this can cause a race condition between the SDK and RequireJS, which will cause the same issue if RequireJS is loaded first.
 
+If user is loading the Application Insights from the CDN via a script tag with require js running by other scripts, error may occur. A typical error could be "Error: Mismatched anonymous define() module". The root reason is explained [here](https://requirejs.org/docs/errors.html#mismatch).
```
Karlie-777, line 564: nit: **if users load Application Insights**  
**with [require.js](https://requirejs.org/)**
**errors** may occur.

README.md
```diff
@@ -561,9 +561,11 @@ eg. ```var aiSdk = require("@microsoft/applicationinsights-web");```
...
+
 To support this usage pattern we also produce and publish to the CDN endpoints an [iife (Immediately Invoked Function Expression)](https://www.codeproject.com/Articles/5265230/Understanding-all-JavaScript-Module-Formats-and-To#iife-module-javascript-module-pattern) module so that the SDK is always executed and initialized.
 
-To use these modules instead of using the default script name simply add ```.gbl``` before the ```.min.js``` eg. use ```.gbl.min.js``` instead of ```.min.js``` at the end of the script name. (Note: Since version 7, the gbl modules is set to be default.)
+To use these modules instead of using the default script name simply add ```.gbl``` before the ```.min.js``` eg. use ```.gbl.min.js``` instead of ```.min.js``` at the end of the script name. (Note: Since version 7, the gbl modules is set to be default to solve the potential [problem](#module-formats) caused by require.js)
```
Karlie-777, line 568: nit: `gbl` is set as default module

Karlie-777 (approved):

Issue #1672, opened 2021-09-14
Provide an option to disable "SDK LOAD Failure errors"


Cisien:
Please provide a feature that lets me opt-out of logging "SDK LOAD Failure: ..." errors as browser exceptions. This error frequently happens due to ad blockers and other conditions outside of my control, and I would rather not see the noise along side actual problems.

**Describe the solution you'd like**
I would like to see an option to disable SDK load failures in the javascript snippet.

**Describe alternatives you've considered**
As a workaround, I have removed this code from the snippet, but that's error prone and (for obvious reasons) not supportable by azure.

kryalama:
To my understanding you dont want the snippet to send these failures to Application Insights whenever the SDK failed to load.  Is my understanding correct? We can provide a config to opt-out of logging the exceptions.  But IMHO you might want to know when the SDK failure happened.

Cisien:
That's correct.

In my scenario, this failure has always been caused by ad blockers, or other methods that users are leveraging to block trackers. The end result is an error that I am unable to resolve as it's not a bug, leaving me with only one option, which is to tolerate it. For this reason, I'd like the option to disable this error from making it into the telemetry.

PR #2207, opened 2023-12-06
[main] disable sdk load error send #1672

siyuniu-ms:
related to #1672


README.md
```diff
@@ -85,6 +85,7 @@ The current version of the snippet is version 7, the version is identified by th
     // name: "appInsights", // Global SDK Instance name defaults to "appInsights" when not supplied
     // ld: 0, // Defines the load delay (in ms) before attempting to load the sdk. -1 = block page load and add to head. (default) = 0ms load after timeout,
     // useXhr: 1, // Use XHR instead of fetch to report failures (if available),
+    // disableSdkLoadError: true, // Prevent the SDK from reporting load failure,
```
MSNev, line 88: Can we make this smaller, as the snippet needs to be able to be as small as possible.

something like `dle`, `dlf`, `dler` etc -- as long as the documentation has a good description this should be as small as possible (but not just `d` for example) as most consumers wont' use this as we don't want to have losts of additional code thats rarely used.

PR #2219, opened 2023-12-12
[main] fix rush build error

siyuniu-ms:


MSNev:
I think I've found a workaround for this, if you add the following to the ci.yml (just above the `npm install`
https://github.com/nevware21/ts-async/blob/4e9c83c790fff13a25409938fbc9cc5ff9536807/.github/workflows/ci.yml#L30-L31

I managed to repo the issue in reverse (having node 20 locally and the node 16 build in github failed), because the generated hash for the internal rush packages got generated differently between the versions.

MSNev:
I'm trying this here https://github.com/microsoft/ApplicationInsights-JS/pull/2221

siyuniu-ms:
solved by @MSNev  in #2221

PR #2224, opened 2023-12-14
[release-3.0][Release] Increase version to 3.0.7

MSNev:


version.json
```diff
@@ -1,88 +1,88 @@
...
             "package": "extensions/applicationinsights-cfgsync-js/package.json",
-            "release": "3.0.6"
+            "release": "3.0.7"
         },
         "@microsoft/applicationinsights-example-cfgsync": {
             "package": "examples/cfgSync/package.json",
-            "release": "3.0.6"
+            "release": "3.0.7"
         },
         "@microsoft/1ds-post-js": {
```
MSNev, line 87: Note: explicitly not updating the 1ds packages as we won't be publishing them and this bug is specific to AI only

shared/1ds-core-js/package.json
```diff
@@ -37,7 +37,7 @@
     ],
     "dependencies": {
         "@microsoft/applicationinsights-shims": "3.0.1",
-        "@microsoft/applicationinsights-core-js": "3.0.6",
+        "@microsoft/applicationinsights-core-js": "3.0.7",
```
MSNev, line 40: Not updating the "version" for this package as it won't be published

version.json
MSNev: It's also not included in the "list" of changes as its package.json only references 1ds-core-js

Karlie-777 (approved):

Issue #2132, opened 2023-08-11
[BUG] Submitting `NaN` via `trackEvent` results in HTTP 400 Error Code


fahey252:
**Description/Screenshot**

Submitting various JavaScript data types via track event works except for `NaN` values. <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN>.  When when `trackEvent` submits the HTTP requests to `https://dc.services.visualstudio.com/v2/track` and the properties contains values of  `NaN`, you get an HTTP 400 with the response of `message": "200: Child value 'NaN' for key 'notANumber' is of incorrect type. Expected: number, Actual: string"`. 

All other JavaScript types seem to be supported or removed from the payload without error. Notice that `NaN` was moved from `properties` to `measurements` in the request payload with the other Number typed values. See Steps to Reproduce below for sample payload and how they there transformed in the request payload or stripped out.

HTTP Request Payload:
<img width="976" alt="request-payload" src="https://github.com/microsoft/ApplicationInsights-JS/assets/757838/c6cc4ad1-8520-4831-8b1f-8c8ee42e2a2f">

HTTP Response Payload:
<img width="1303" alt="response-payload" src="https://github.com/microsoft/ApplicationInsights-JS/assets/757838/03abdbad-9fad-42bb-9f19-a5016f3f3714">

**Steps to Reproduce**

```javascript
const properties = {
    bigInt: 9999999999999999,
    booleanValue: true,
    callback: () => { },
    emoji: '🤷‍♀️',
    largeValue: Infinity,
    listOfValues: [1, 3],
    nullValue: null,
    number: 5,
    object: {
      property1: {
        nestedProperty: 'nested object'
      }
    },
    notANumber: NaN,
    regex: /test/g,
    someClass: Alert,
    someDate: new Date(),
    symbol: Symbol,
    undefinedValue: undefined
  }
const eventName = 'DataTypeTesting'

appInsights.trackEvent({ name: eventName }, properties);
```

 - OS/Browser: React Native Debugger, Chrome 114
 - SDK Version [e.g. 22]:  `"@microsoft/applicationinsights-web": "^3.0.2",` and `"@microsoft/applicationinsights-react-native": "^4.0.0",`
 - How you initialized the SDK: `yarn`

**Expected behavior**

When a data type such as a regex is submitted via `trackEvent`, its value is converted to `undefined`. `NaN` values should be treated the same or converted to `'NaN'` string values.

[...]

PR #2232, opened 2024-01-04
[main] [BUG] Submitting NaN via trackEvent results in HTTP 400 Error Code #2132

siyuniu-ms:
check NaN while create Telemetry


shared/AppInsightsCommon/src/TelemetryItemCreator.ts
```diff
@@ -53,6 +53,9 @@ export function createTelemetryItem<T>(item: T,
     // Part C
     if (!isNullOrUndefined(customProperties)) {
         objForEachKey(customProperties, (prop, value) => {
+            if (Number.isNaN(value)) {
```
MSNev, line 56: While this would "work" for the example provided in issue, there are a couple of problems with this
- It's only going to support values directly assigned in the custom Properties (not any other field)
- If there is an "embedded" object then this won't detect / convert that either
- Calling the `Number.isNaN` function here will be expensive as function calls are expensive, it think that it would be "faster" to just use `if (value === NaN)` (we would need to do some perf testing).

We should probably review and do this within the channel Serializer.ts to better handle this, there is already some code in there for the measurements for handling `NaN` but the properties are expected to be strings...

channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
```diff
@@ -123,7 +123,11 @@ function _extractPropsAndMeasurements(data: { [key: string]: any }, properties:
     if (!isNullOrUndefined(data)) {
         objForEachKey(data, (key, value) => {
             if (isNumber(value)) {
-                measurements[key] = value;
+                if (isNaN(value)) {
+                    properties[key] = "NaN";
```
siyuniu-ms, line 127: this is a sample code

channels/applicationinsights-channel-js/src/EnvelopeCreator.ts
```diff
@@ -123,7 +123,11 @@ function _extractPropsAndMeasurements(data: { [key: string]: any }, properties:
...
+                } else {
+                    measurements[key] = value;
+                }
```
siyuniu-ms, line 130: this is the sample code

channels/applicationinsights-channel-js/src/Serializer.ts
```diff
@@ -149,11 +149,7 @@ export class Serializer {
                                 output[field] = "null";
                             } else {
                                 const num = parseFloat(value);
-                                if (isNaN(num)) {
-                                    output[field] = "NaN";
-                                } else {
-                                    output[field] = num;
-                                }
+                                    output[field] = num;
```
MSNev, line 152: nit: indenting is wrong here

MSNev (approved):

PR #2237, opened 2024-01-11
[main] instrument the typedoc github pages with ApplicationInsights

siyuniu-ms:


package.json
```diff
@@ -31,7 +31,8 @@
         "npm-pack": "node common/scripts/install-run-rush.js npm-pack --verbose",
         "npm-publish": "node ./tools/release-tools/npm_publish.js",
         "npm-set-latest": "node ./tools/release-tools/npm_set_latest.js",
-        "gh-status": "node ./tools/status-tools/github-status.js"
+        "gh-status": "node ./tools/status-tools/github-status.js",
+        "script-inject": "node tools/github-page-script-injection/injectScript.js"
```
siyuniu-ms, line 35: I do not have access to change github action currently, later could add  steps:
      - name: Checkout
        uses: actions/checkout@v4
      - name: Inject Script
        run:  node ./injectScript.js in github action for page deployment so that the inject could happen right before the doc website is published.

MSNev: I don't think there is any current github action to do this. It (should) just be a case that once we "generate" the docs we then just run this new task.

tools/github-page-script-injection/script.js
```diff
@@ -0,0 +1,18 @@
...
+    // useXhr: 1, // Use XHR instead of fetch to report failures (if available),
+    // dle: true, // Prevent the SDK from reporting load failure log
+    crossOrigin: "anonymous", // When supplied this will add the provided value as the cross origin attribute on the script tag
+    // onInit: null, // Once the application insights instance has loaded and initialized this callback function will be called with 1 argument -- the sdk instance (DO NOT ADD anything to the sdk.queue -- As they won't get called)
+    cfg: { // Application Insights Configuration
+      connectionString: "InstrumentationKey=814a172a-92fd-4950-9023-9cf13bb65696;IngestionEndpoint=https://eastus-8.in.applicationinsights.azure.com/;LiveEndpoint=https://eastus.livediagnostics.monitor.azure.com/"
+    }
+});
+let appInsights = new Microsoft.ApplicationInsights.ApplicationInsights({ config: {connectionString: "InstrumentationKey=814a172a-92fd-4950-9023-9cf13bb65696;IngestionEndpoint=https://eastus-8.in.applicationinsights.azure.com/;LiveEndpoint=https://eastus.livediagnostics.monitor.azure.com/", disableInstrumentationKeyValidation: true}});
```
MSNev, line 14: This bit doesn't look right... The above snippet should be all we require...

Is it also possible to "read" the generated snipped from the "tools / snippet" and just update with the connection String, rather than effectively hard-coding this version of the snippet here?

package.json
siyuniu-ms: I am thinking about add it to github action because the run takes after the file is uploaded and before the web is published, the page file is not changed. which means subsequent runs are not influenced by the changes made in previous runs. Presently, we execute 'npm run api-docs' in each folder for documentation updates. If we transition to running this command at the root level, taking into account the script injection at the root folder, it might simplify tracking changes without impacting subsequent runs.

MSNev: Well actually, we just run "rush api-docs" an it runs the api-docs in each project. We could just as easily ass a "root" npm script in the root to 
`"api-docs": "rush api-docs && npm run script-inject"`

Then we just run `npm run api-docs` from the root and everything is done. I also don't believe that we can "update" the uploaded "docs" via a github action because the `main` branch is protected so it always needs a PR to push changes.

MSNev (approved):

PR #2241, opened 2024-01-16
[Main][Task]26451789: Add Offline Support

Karlie-777:
TODO: add security doc to let users know that data would be stored


channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -1490,4 +1500,23 @@ export class HttpManager {
     public sendSynchronousBatch(batch: EventBatch, sendType?: EventSendType, sendReason?: SendRequestReason) {
         // @DynamicProtoStub - DO NOT add any code as this will be removed during packaging
     }
+
+    
+    /**
+     * Get Offline Serializer support
+     * @returns internal Offline Serializer object
+     */
+    public _serializeOfflineEvt(evt: ITelemetryItem | IPostTransmissionTelemetryItem): string {
```
MSNev, line 1509: minor nit: public functions should not "really" begin with an "_"

We can (and do) do this for internal only functions. But as the `HttpManager` is all internal anyway we don't need to to that here.

channels/1ds-post-js/src/PostChannel.ts
```diff
@@ -259,6 +274,41 @@ export class PostChannel extends BaseTelemetryPlugin implements IChannelControls
...
+                return rlt;
+
+            }
+
+            function _serialize(event: ITelemetryItem) {
+
+                let rlt = EMPTY_STR;
+                try {
+                    _cleanEvent(event);
+                    if (_overrideInstrumentationKey) {
```
MSNev, line 298: Question for review: when is this called (if I haven't worked it out already) as do we really want / need to override the instrumentation key.

ie. We don't want to override any previously save batches (because they may actually belong to someone else)

channels/1ds-post-js/src/PostChannel.ts
```diff
@@ -306,6 +346,38 @@ export class PostChannel extends BaseTelemetryPlugin implements IChannelControls
...
+                // // Remove extra AI properties if present
+                // if (event.ext && event.ext[STR_TRACE]) {
+                //     delete (event.ext[STR_TRACE]);
+                // }
+                // if (event.ext && event.ext[STR_USER] && event.ext[STR_USER]["id"]) {
+                //     delete (event.ext[STR_USER]["id"]);
+                // }
+
+                // // v8 performance optimization for iterating over the keys
+                // if (_optimizeObject) {
```
MSNev, line 371: We actually, will still need this as it dramatically affects the performance for serializing large modified objects on chromium

channels/offline-channel-js/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Description
+
+The Offline Channel supports the saving of events when your application is offline and resending those events when the application is online.
+
+### Note
+
+- A post or sender channel is required for processing online events.
+- Request header details will be stored in local/session storage or IndexedDB based on your configuration.
+- If you are using the default Breeze endpoint `https://dc.services.visualstudio.com` for Application Insights, partial success is currently considered as success and events not sent in partial success will be dropped.
```
MSNev, line 11: Breeze is the internal name this should be either Applicatin Insights or Azure Monitor

channels/offline-channel-js/rollup.config.js
```diff
@@ -0,0 +1,38 @@
+import { createConfig } from "../../rollup.base.config";
+import { updateDistEsmFiles } from "../../tools/updateDistEsm/updateDistEsm";
+
+const version = require("./package.json").version;
+const outputName = "applicationinsights-offlinechannel-js";
+
+const banner = [
+    "/*!",
+    ` * Application Insights JavaScript SDK - Channel, ${version}`,
```
MSNev, line 9: Offline Channel

channels/offline-channel-js/src/Helpers/Utils.ts
```diff
@@ -0,0 +1,206 @@
...
+            output = output + String.fromCharCode(chr3);
+        }
+
+    }
+    let arr = output.split(",").map(c => Number(c));
+    return new Uint8Array(arr);
+
+}
+
+export function forEachMap<T>(map: { [key: string]: T }, callback: (value: T, key: string) => boolean): void {
```
MSNev, line 114: This looks like the same thing the `objForEachKey` function does

channels/offline-channel-js/src/Helpers/Utils.ts
```diff
@@ -0,0 +1,206 @@
...
+        for (let lp = 0; lp < keys.length; lp++) {
+            let key = keys[lp];
+            if (!callback(map[key], key)) {
+                break;
+            }
+        }
+    }
+}
+
+export function isMapEmpty<T>(map: { [key: string]: T }) {
```
MSNev, line 126: Probably just `obKeys(xxx).length === 0`

channels/offline-channel-js/src/InMemoryBatch.ts
```diff
@@ -0,0 +1,122 @@
+import dynamicProto from "@microsoft/dynamicproto-js";
```
MSNev, line 1: nit: each file should have the copyright header (we probably don't have this on every file -- but we should)

channels/offline-channel-js/src/InMemoryBatch.ts
```diff
@@ -0,0 +1,122 @@
...
+                return [_bufferFullMessageSent];
+            };
+
+            
+            _self.addEvent = (payload: IPostTransmissionTelemetryItem | ITelemetryItem) => {
+                //TODO: handle space free up, handle drop here
+                if (!isNullOrUndefined(evtsLimitInMem) && _self.count() >= evtsLimitInMem) {
+                    // sent internal log only once
+                    if (!_bufferFullMessageSent) {
+                        _throwInternal(logger,
```
MSNev, line 31: We probably don't need to log this as we are "adding" events and once we exceed the limit we "should" be attempting to flush it....

Soo... "in theory" this should never occur....

channels/offline-channel-js/src/InMemoryBatch.ts
```diff
@@ -0,0 +1,122 @@
...
+                            eLoggingSeverity.WARNING,
+                            _eInternalMessageId.InMemoryStorageBufferFull,
+                            "Maximum offline in-memory buffer count reached: " + _self.count(),
+                            true);
+                        _bufferFullMessageSent = true;
+                    }
+
+                    return;
```
MSNev, line 39: We should probably return true/false so the caller knows whether an event was added (or not)...

So then perhaps the caller can do something (like flush / start a new buffer) when this returns false....

channels/offline-channel-js/src/OfflineBatchHandler.ts
```diff
@@ -0,0 +1,387 @@
...
+
+            }
+
+            _self.sendNextBatch = (cb?: OfflineBatchCallback, sync?: boolean, xhrOverride?: IXHROverride, cnt: number = 1) => {
+                return createAsyncPromise((resolve, reject) => {
+                    let provider = _provider;
+                    try {
+                       
+                        if (_unloadProvider) {
+                            doAwaitResponse(_unloadProvider.getNextBatch(), (res) => {
```
MSNev, line 97: Maybe this should return here and after the promise has resolve / rejected it either returns the "unload" batch, or simply calls the code below (which would need to be extracted out to it's own function so that it can be called below or from within this async handler.

channels/offline-channel-js/src/OfflineBatchHandler.ts
```diff
@@ -0,0 +1,387 @@
...
+                    if (!provider) {
+                        reject(new Error(NoProviderErrMsg));
+                    }
+
+                    return doAwaitResponse(provider.getNextBatch(), (response: AwaitResponse<IStorageTelemetryItem[]>) => {
+                        try {
+                            if(!response || response.rejected) {
+                                let res = {state: eBatchSendStatus.Failure, data: response.reason};
+                                storeResolve(res);
+                                return;
```
MSNev, line 122: This should be calling either the resolve / reject from the outer promise -- otherwise it will never be fulfilled, so anyone waiting on the original to complete will wait forever.

channels/offline-channel-js/src/OfflineBatchHandler.ts
```diff
@@ -0,0 +1,387 @@
...
+
+                                                            } else {
+                                                                result.state = eBatchSendStatus.Drop;
+                                                            }
+                                                        }
+                                                        storeResolve(result);
+                                                        
+                                                    };
+                                                    return doAwaitResponse(sender(evt, onCompleteCallback, sync) as any, (res) => {
+                                                        // do nothing, because it should be resolved in callback
```
MSNev, line 161: This also looks like it should call the outer resolve / reject as this looks like it would cause the outer promise to never be marked as fullfilled.

MSNev: Oh, I see you have the `storeResult` outer function which calls the `resolve` function

channels/offline-channel-js/src/OfflineBatchHandler.ts
MSNev: As below the `storeResolve` is calling the `resolve`

channels/offline-channel-js/src/OfflineBatchHandler.ts
```diff
@@ -0,0 +1,387 @@
...
+                                });
+                            }
+                            // don't have any evts returned by getNext, resolve
+                            storeResolve(result);
+                            
+                        } catch(e) {
+                            reject(e);
+                        }
+                    });
+                    function storeResolve(result) {
```
MSNev, line 192: We should move this to the top of this block (line 93) so that when someone is reading this they have the context that this will be "resolving" the outer promise -- which I missed originally

channels/offline-channel-js/src/OfflineBatchHandler.ts
```diff
@@ -0,0 +1,387 @@
...
+                                                
+                                            } catch (e) {
+                                                reject(e)
+                                            }
+                                        }
+                                        // this case should never happen because this function is called after we confirm that evt is available
+                                        // add resolve here in case
+                                        result.state = eBatchSendStatus.Complete;
+                                        storeResolve(result);
+
```
MSNev, line 176: Probably should return here, so that the `result.state` (if this code ever executes) doesn't get replaced with the `Failure` below.

While you "can" call the `resolve` (the `storeResolve`) of a promise more than once, and the code handles only really calling once -- we shouldn't rely on this behavior in case a future browser / Javascript framework starts treating that as an exception and throws.

channels/offline-channel-js/src/OfflineBatchHandler.ts
```diff
@@ -0,0 +1,387 @@
...
+                            }
+                            // don't have any evts returned by getNext, resolve
+                            storeResolve(result);
+                            
+                        } catch(e) {
+                            reject(e);
+                        }
+                    });
+                    function storeResolve(result) {
+                        cb && cb(result);
```
MSNev, line 193: Just some edge case issue, we should wrap any callback in a try / catch, so that we still call (at least) the `resolve` "just in case" any callback fails for some uexpected reason.

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,604 @@
...
+
+const version = "#version#";
+const DefaultOfflineIdentifier = "OfflineChannel";
+const DefaultBatchInterval = 15000;
+const DefaultInMemoMaxTime = 15000;
+const PostChannelIdentifier = "PostChannel";
+
+
+interface IUrlLocalStorageConfig {
+    iKey: string;
```
MSNev, line 36: We shouldn't need this (unless we want to support different persistence cache level for different iKeys -- which would be possible for 1DS with the Shared Manager, but it's part of the general design

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,604 @@
...
+
+                return createProcessTelemetryContext(pluginChain, coreConfig, core, nextPlugin);
+            }
+
+
+            function _getPayload(evt: IPostTransmissionTelemetryItem | ITelemetryItem) {
+                try {
+
+                    if (evt) {
+                        evt.iKey = evt.iKey || _urlCfg.iKey;
```
MSNev, line 421: If the iKey is not set by the time we get here then there is a potential issue -- but we also talked about the "overrideIkey"  (but only within the `processTelemetry` -- not here (I don't think)

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,604 @@
...
+            }
+
+            function _constructPayloadData(payloadArr: string[], criticalCnt?: number) {
+
+                try {
+                    if (!_offineSupport) {
+                        return null;
+                    }
+                    let cnt = criticalCnt || 0;
+                    let payload = _offineSupport.batch(payloadArr);
```
MSNev, line 439: We should probably change this to return the IPayloadData as the PostChannel / HttpManager may be configured to append additional values to the query string and / or create headers from them -- so we don't want to duplicate that code here as Sender doesn't have that level of complexity.

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,604 @@
...
+                    let ctx = createProcessTelemetryContext(null, theConfig, core);
+                    storageConfig = ctx.getExtCfg<ILocalStorageConfiguration>(_self.identifier, defaultLocalStorageConfig);
+                    let channelIds = storageConfig.primaryOnlineChannelId;
+                    if (channelIds && channelIds.length) {
+                        arrForEach(channelIds, (id) => {
+                            let plugin = _self.core.getPlugin<IChannelControls>(id);
+                            let channel = plugin && plugin.plugin;
+                            if (channel) {
+                                _primaryChannelId = id;
+                                _offineSupport = channel.getOfflineSupport();
```
MSNev, line 480: Someone (might) try and use this with an older version of Application Insights at which point the `getOfflineSupport` function won't exist and will cause this to throw an exception, so we need to check if the function is there before calling.

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,604 @@
...
+                                return;
+                            }
+                            
+                        })
+                    }
+                
+                    
+    
+                    let urlConfig: IUrlLocalStorageConfig = _urlCfg;
+                    let curUrl = coreConfig.endpointUrl || DEFAULT_BREEZE_ENDPOINT + DEFAULT_BREEZE_PATH;
```
MSNev, line 490: Option: maybe we get the endpoint from the `_offlineSupport` object instead?? and not have any default or `_urlCfg`???

Potential issue however, is that we have an upcoming request to "handle" a Promise for the connection string, so we might not have resolved the endpoint and if it changes it could be problematic) -- although because the `channel.getOfflineSupport()` is inside of an `onConfigChange` the caller could support returning an updated value if it's referenced.

channels/offline-channel-js/src/Sender.ts
```diff
@@ -0,0 +1,714 @@
...
+import { isFunction, isNumber } from "@nevware21/ts-utils";
+import { ILocalStorageConfiguration, IOfflineSenderConfig } from "./Interfaces/IOfflineProvider";
+import { IBackendResponse, XDomainRequest as IXDomainRequest } from "./Interfaces/ISender";
+
+const DefaultOfflineIdentifier = "OfflineChannel";
+//const FetchSyncRequestSizeLimitBytes = 65000; // approx 64kb (the current Edge, Firefox and Chrome max limit)
+const PostChannelId = "PostChannel";
+
+
+declare var XDomainRequest: {
```
MSNev, line 22: Is this still needed now?

channels/1ds-post-js/src/PostChannel.ts
```diff
@@ -1187,4 +1246,13 @@ export class PostChannel extends BaseTelemetryPlugin implements IChannelControls
     public _clearBackOff() {
         // @DynamicProtoStub - DO NOT add any code as this will be removed during packaging
     }
+
+    /**
+     * Get Offline Serializer support
+     * @returns internal Offline Serializer object
+     */
+    public getOfflineSupport() {
```
MSNev, line 1254: nit: this should specify that it returns the interface so that the typescript definition is reported correctly.

channels/offline-channel-js/src/Helpers/Utils.ts
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+    }
+    let arr = output.split(",").map(c => Number(c));
+    return new Uint8Array(arr);
+
+}
+
+
+export function getTimeId(): string {
```
MSNev, line 117: add todo's: to provide comments for each function

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,614 @@
...
+
+                return true;
+            }
+
+
+            function _setupInMemoTimer() {
+                if (!_inMemoFlushTimer) {
+                    _inMemoFlushTimer = scheduleTimeout(() => {
+                        _flushInMemoItems();
+                        _inMemoFlushTimer.refresh();
```
MSNev, line 281: We should only "refresh" (restart) the timer if there are items waiting to be flushed.

I think (not finished reading all of this yet) that you would just need to remove this from here and add it as an else clause.

```ts
if (!_inMemoFlushTimer) {
    _inMemoFlushTimer = scheduleTimeout(() => {
        _flushInMemoItems();
    }, _inMemoTimerOut);
    _inMemoFlushTimer.unref();
} else {
    // Restart the timer if not already running https://nevware21.github.io/ts-utils/typedoc/interfaces/ITimerHandler.html#enabled
    _inMemoFlushTimer.enabled = true;
}                        
```

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,614 @@
...
+                            }
+                            let added = _inMemoBatch.addEvent(evt);
+                            // inMemo is full
+                            if (!added) {
+                                _flushInMemoItems();
+                                // if event can't be added again, simply drop it
+                                // TODO: better handle clean in memo events
+                                // TODO: add logs/notification
+                                _inMemoBatch.addEvent(evt);
+                                _inMemoFlushTimer && _inMemoFlushTimer.refresh();
```
MSNev, line 176: If we are always calling the `_setupInMemoTimer` (where it "resets" if not already running -- based on my comment I added to it), then I don't think this would be needed.

channels/offline-channel-js/src/OfflineChannel.ts
MSNev: Note: Teams are VERY sensitive (and have tests) which detect times that keep running for no reason, because it stops devices from going to (and staying) asleep, so any timers we create we need to make sure that they are not always "polling" (running) unless we have something that we are waiting for.

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,614 @@
...
+                                    _sendNextBatchTimer && _sendNextBatchTimer.refresh();
+                                }
+                                let promise = _urlCfg.batchHandler.sendNextBatch(callback, false, _senderInst);
+                                _queueStorageEvent("sendNextBatch", promise);
+                            }
+                           
+                        } else {
+                            _sendNextBatchTimer.cancel();
+                        }
+                        // if offline, do nothing;
```
MSNev, line 367: We should also "stop" (cancel) the timer if we don't have anything in the "next" queue (ie. we need should check the `hasStoredBatch` and only "restart" / "start" the timer if it returns true. Which we would also do after "adding" an in memory batch (because we know the answer will be true)  -- except of course like you have here that we believe we are offline.

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,614 @@
...
+                    _inMemoFlushTimer = scheduleTimeout(() => {
+                        _flushInMemoItems();
+                        _inMemoFlushTimer.refresh();
+                        
+                    }, _inMemoTimerOut);
+                    _inMemoFlushTimer.unref();
+                }
+            }
+
+            function _flushInMemoItems(unload?: boolean) {
```
MSNev, line 288: Future: as part of the TODO you have on line 173 etc in the `processTelemetry` this should probably return the promise so that it can react (after the batch is stored) that the inMemory batch was added (or not)

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -0,0 +1,614 @@
...
+                        _inMemoBatch = new InMemoryBatch(_self.diagLog(),curUrl, null, evtsLimit);
+                        _inMemoTimerOut = storageConfig.inMemoMaxTime;
+                        let onlineConfig = ctx.getExtCfg<any>(_primaryChannelId, {}) || {};
+                        _convertUndefined = onlineConfig.convertUndefined;
+                        _endpoint = curUrl;
+                        _setRetryTime();
+                        _maxBatchInterval = storageConfig.maxSentBatchInterval;
+                        _maxBatchSize = storageConfig.maxBatchsize;
+                    }
+                    _urlCfg = urlConfig;
```
MSNev, line 531: Add todo: this will overwrite any "previous" if the endpoint is changed as you construct a "new" one. Which means all of the previous stuff will either get lost or left behind in memory -- because it will never be cleaned up on teardown.

channels/offline-channel-js/src/Sender.ts
```diff
@@ -0,0 +1,717 @@
...
+}
+
+
+export type SenderFunction = (payload: string[], isAsync: boolean) => void | IPromise<boolean>;
+
+export class Sender {
+
+    public _appId: string; //TODO: set id
+
+    constructor(endpointUrl?: string) {
```
MSNev, line 63: Add a TODO: to remove the passing (and internal usages) of the endpointURL, as your already always creating without an endpoint anyway.

If there is no other endpoint (the URL) lets just not send anything

channels/offline-channel-js/src/Sender.ts
```diff
@@ -0,0 +1,717 @@
...
+                    let offlineCfg = ctx.getExtCfg(DefaultOfflineIdentifier) as ILocalStorageConfiguration;
+                    _onlineChannelId = channelId || BreezeChannelIdentifier;
+                    let senderConfig = ctx.getExtCfg(_onlineChannelId, {}) as any;
+                    let offlineSenderCfg = offlineCfg.senderCfg || {} as IOfflineSenderConfig;
+                  
+                    if (_onlineChannelId == PostChannelId) {
+                        _isOneDs = true;
+                    }
+                    
+                    _endpointUrl = endpointUrl || senderConfig.endpointUrl || config.endpointUrl || DEFAULT_BREEZE_ENDPOINT + DEFAULT_BREEZE_PATH;
```
MSNev, line 134: As above lets remove this as a fallback if the payload doesn't provide an endpoint we just don't send / return an error.

channels/offline-channel-js/src/Sender.ts
MSNev: Actually, lets add a TODO to remove this class completely so we can delegate the "sending" to the real channels via a new "sendbatch" (or something like that) on IOfflineSupport, that way we don't need to worry about all of the transports etc)

MSNev (approved):
Lets push this change in and we can address at least the timer stuff in a follow up PR (before release)

channels/offline-channel-js/src/Sender.ts
Karlie-777: Endpoint is removed, and if no url is provided, `doOncomplete` will be called with status code 400 and will throw internal as well.
created Task 26681160 for this to add more tests

channels/offline-channel-js/src/OfflineChannel.ts
Karlie-777: created task 26681188 for this

channels/offline-channel-js/src/OfflineChannel.ts
Karlie-777: created task 26681220 for this

channels/offline-channel-js/src/OfflineChannel.ts
Karlie-777: created task 26681249 for this

PR #2245, opened 2024-01-18
[main] [snippet update] add functions to set configs of snippet

siyuniu-ms:


siyuniu-ms:
Will it make things more clear if we remove const webSnippetCs = "##replaceConnStringSnippet##"? Currently, the only difference between webSnippet and webSnippetCs is:
webSnippet  ends with    connectionString: \"InstrumentationKey=INSTRUMENTATION_KEY
webSnippetcs ends with  connectionString: \"YOUR_CONNECTION_STRING\"\r\n 
btw, maybe webSnippet  should end with instrumentationKey: \"YOUR_INSTRUMENTATION_KEY, this maybe a mistake when I wrote the rush json months before

tools/applicationinsights-web-snippet/src/applicationinsights-web-snippet.ts
```diff
@@ -10,4 +10,20 @@ function webSnippetVersion() {
     return "";
 }
 
-export { webSnippet, webSnippetCs, webSnippetVersion }
\ No newline at end of file
+function substituteInstrumentationKey(key: string) {
```
MSNev, line 13: Lets call these something like

`getSdkLoaderScript(config: SdkLoaderConfig )`
And define a specific config for this that

```ts
export interface SdkLoaderConfig {
<include the instrumentation key and the connection string>
}
```

Where this configuration could (or will eventually) support all of the snippet configurations -- so we could even support the default snippet configuration.

tools/applicationinsights-web-snippet/src/applicationinsights-web-snippet.ts
```diff
@@ -10,20 +12,17 @@ function webSnippetVersion() {
...
-    if (key && key.trim() !== "") {
-        userSnippet = webSnippet.replace("InstrumentationKey=INSTRUMENTATION_KEY", key);
+function getSdkLoaderScript(config: SdkLoaderConfig) {
+    let snippet: string = webSnippet;
+    if (config && config.connectionString) {
+        snippet = webSnippetCs.replace("YOUR_CONNECTION_STRING", config.connectionString);
+    } else if (config && config.instrumentationKey) {
+        snippet = webSnippet.replace("InstrumentationKey=INSTRUMENTATION_KEY", config.instrumentationKey);
+    } else {
+        console.log("No instrumentationKey or connectionString was provided to the init function");
```
MSNev, line 22: We should not be using `console.log`, which is a little hard with this component as we don't have any other way to get "error" information out.

Additionally, this will often not running on a browser... So there will be no console. So in this specific case I think we should just return the default snippet..

MSNev (approved):

PR #2247, opened 2024-01-19
[main] [1ds-post] export add header func for auth-plugin to consume

siyuniu-ms:


channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -675,6 +675,10 @@ export class HttpManager {
                 _headers[name] = value;
             };
 
+            _self.removeHeader = (name: string) => {
+                delete _headers[name];
```
hectorhdzg, line 679: Is it possible _headers is null? this could throw an exception I believe

channels/1ds-post-js/src/InternalConstants.ts
```diff
@@ -32,6 +32,7 @@ export const STR_TIME_DELTA_TO_APPLY = "time-delta-to-apply-millis";
 export const STR_UPLOAD_TIME = "upload-time";
 export const STR_API_KEY = "apikey";
 export const STR_MSA_DEVICE_TICKET = "AuthMsaDeviceTicket";
+export const STR_AUTH_PLUGIN = "WebAuthToken";
```
hectorhdzg, line 35:
```suggestion
export const STR_AUTH_WEB_TOKEN = "WebAuthToken";
```

hectorhdzg (approved):
Apart from some minor suggestions LGTM

channels/1ds-post-js/src/HttpManager.ts
MSNev: it shouldn't ever be :-) as it's initialized in the `_initDefaults` to an empty object.

Issue #2258, opened 2024-02-04
[BUG] `ITelemetryItem` uses intersection type instead of union type for `tags` property


devm33:
The `ITelemetryItem` interface uses an intersection type (`&`) between `Tags` and `Tags[]` instead of a union type (`|`). There's a note on the line that `Tags[]` will be deprecated:

https://github.com/microsoft/ApplicationInsights-JS/blame/f9d5ec1e2045f6206c8d7dc9b0cb3e6509f45473/shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/ITelemetryItem.ts#L37

However using an intersection type forces passing an array in order to satisfy the constraints of `Tags & Tags[]`:

```
Type '{ [key: string]: string; }' is not assignable to type 'Tags & Tags[]'.
  Type '{ [key: string]: string; }' is missing the following properties from type 'Tags[]': length, pop, push, concat, and 29 more.
```

MSNev:
Just had a quick look at the code and I think we can just remove the `Tags[]` as `Tags` is defined (and used -- based on quick look) as an object with dynamic keys -- so the only "real" usage would be if this is set as a tag with a numeric key.

As we are about to release a `3.1.0` this would seem a good point to remove that later support.

MSNev:
Also, because of backward compatibility support, this might be problematic -- so as suggested it might just be easier to change the definition to a union type as suggested (rather than removing the `Tags[]` definition completely.

devm33:
@MSNev I'm running `5.1.3`. The point I was trying to raise was the current typing is preventing passing `Tags`. It only accepts an array. Which based on the comment seemed like the opposite of the intention.

MSNev:
Yeah, it looks like someone in the past tried to convert this to an object (and when we construct the outbound serialized form this is what we do), and while this definition "works" (from a previous version and raw JS perspective), it's problematic -- especially if the latest versions of TS are now enforcing this more.

PR #2269, opened 2024-02-09
[main] [Possible Break] Removed Tags[] from ITelemetryItem as this was breaking later versions of TypeScript by using the intersection type instead of union type for tags property #2258

siyuniu-ms:
for #2258

### DIFF
--- a/extensions/applicationinsights-properties-js/src/TelemetryContext.ts
+++ b/extensions/applicationinsights-properties-js/src/TelemetryContext.ts
@@ -9,7 +9,7 @@ import {
     IUserContext, IWeb, PageView, dataSanitizeString
 } from "@microsoft/applicationinsights-common";
 import {
-    IAppInsightsCore, IDistributedTraceContext, IProcessTelemetryContext, ITelemetryItem, IUnloadHookContainer, _InternalLogMessage,
+    IAppInsightsCore, IDistributedTraceContext, IProcessTelemetryContext, ITelemetryItem, IUnloadHookContainer, Tags, _InternalLogMessage,
     getSetValue, hasWindow, isNullOrUndefined, isString, objKeys, setValue
 } from "@microsoft/applicationinsights-core-js";
 import { Application } from "./Context/Application";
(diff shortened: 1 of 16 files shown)


extensions/applicationinsights-properties-js/src/TelemetryContext.ts
```diff
@@ -35,6 +35,17 @@ function _nullResult(): string {
     return null;
 }
 
+function setTageValue(tags: Tags | Tags[], field: string, value: string): void {
+    // Function body
+    if (Array.isArray(tags)) {
+        tags.forEach(tag => setValue(tag, field, value, isString)
```
MSNev, line 41: This doesn't look correct... Let me try and repo locally to see the errors.

extensions/applicationinsights-properties-js/src/TelemetryContext.ts
```diff
@@ -35,6 +35,17 @@ function _nullResult(): string {
     return null;
 }
 
+function setTageValue(tags: Tags | Tags[], field: string, value: string): void {
+    // Function body
+    if (Array.isArray(tags)) {
```
MSNev, line 40: A couple of side notes for minification.

You can use the `isArray` implementation instead of `Array.isArray` (the `isArray` is actually just an alias (constant variable) for the `Array.isArray` (so it's still calling the same thing but it can be minified a LOT smaller.

For the `tags.forEach` we also have an `arrForEach` which also will become smaller and effectively do the same thing (it has a bit more functionality than the `xxx.forEach` but as we use it elsewhere  it's already included so it's "better" (for minification) to use `arrForEach` instead. If this was a performance bottleneck then because of the additional capabilities the `arrForEach` is not as performant (fast) as the `xxx.forEach`

extensions/applicationinsights-properties-js/src/TelemetryContext.ts
MSNev: Ok, looking at the usage a bit deeper and the returned error message, I think we do the following
- Remove the `Tags[]` completely so it's just defined as `tags?: Tags;` as during the serialization process in `Serializer.ts` we process this field with the `_serializeStringMap` and if that encounters an "error" it will actually report it as an error (which is what would happen if they do actually pass an incomplete array).
- Add some release notes about this possible breaking change, at the very least lets call this out in the "name" of this PR, so if someone looks at the history it will be obvious

As this change will only affect `npm` users and the correct "typing" is really Tags anyway along with the next release being `4.1.0` its a good time to break anyone how might be trying to pass an array (which they shouldn't).

So

MSNev (commented):
There are a couple of references the tags[] in the API.md

And can you also put a placeholder in the RELEASES.md so that we don't forget about this potential breaking change. I'll update the title of this PR

MSNev (approved):

PR #2275, opened 2024-02-15
[main] [Web-Snippet Package] Prepare for release by update readme with new functions

siyuniu-ms:


tools/applicationinsights-web-snippet/README.md
```diff
@@ -15,13 +15,25 @@ Add the Application Insights Web Snippet to your app's dependencies and package.
 npm i @microsoft/applicationinsights-web-snippet
 ```
 
+### Pass in Key via Function (Preferred Method)
+Import the helper function from the package to easily get the snippet
+```
+import { getSdkLoaderScript } from "@microsoft/applicationinsights-web-snippet";
+```
+
+Configure the snippet with your key.
```
MSNev, line 24: It should mention with you connection string and lets "remove" the `replace` version as we want to effectively mark it as deprecated 😄 so lets just not included it in the getting started anymore.

Which also means we should add that this is the approach since v1.1.0

MSNev (approved):

Issue #2277, opened 2024-02-16
[BUG] [Web-Snippet] dependency chain issues


MSNev:
With the new release of the web-snippet v1.1.0 this is causing all of the common and it's dependencies to get included when not required for runtime.

The dependency chain of the web-snippet is including the 
- applicatoninsights-common which includes
- application insights core which includes
- ts-async which includes
- ts-utils

As the web-snippet doesn't "need" any of these to actually run, we need to review and break this dependency as it's causing a LOT of code to be dragged down into the node_modules folder for anyone consuming this package

PR #2278, opened 2024-02-16
[Main] [BUG] [Web-Snippet] dependency chain issues #2277

siyuniu-ms:


tools/applicationinsights-web-snippet/package.json
```diff
@@ -49,7 +50,6 @@
         "@nevware21/grunt-eslint-ts": "^0.2.2"
     },
     "dependencies": {
-        "@microsoft/dynamicproto-js": "^2.0.3",
-        "@microsoft/applicationinsights-common": "3.1.0"
+        "@microsoft/dynamicproto-js": "^2.0.3"
```
MSNev, line 53: It looks like this should move to a devDependency as well as this now has a dependency on ts-utils and I think this is only really used for rollup

MSNev (approved):

PR #2290, opened 2024-03-05
chore: remove extraneous console log from debug plugin

### LINKED ISSUE #2289: `console.log` unnecessarily noisy
Hi there!

I have noticed that this particular `console.log` is unnecessarily printing `null` in unit tests or a server (SSR) environment. In each of these cases, the application insights object is instantiated, but is not actually initialized - it is only initialized on the browser.

https://github.com/microsoft/ApplicationInsights-JS/blob/724d5204e3cdc50d9402d415f021d816a4c11b45/extensions/applicationinsights-debugplugin-js/src/components/styleNodeSrc.ts#L18C1-L18C29

However, in these environments, it always prints `null` to console and really clutters the test output.

image

I'm not sure what purpose this `console.log` serves; but I am happy to contribute a PR in case you have suggestions for what can be done to reduce the noise.

develohpanda:
Resolves #2289


MSNev (approved):

MSNev (changes requested):
You need to keep the import as it's used on line 56

develohpanda:
Whoops, missed that! 🤦‍♂️

MSNev (approved):

PR #2294, opened 2024-03-07
[main] implement Osplugin, extracting windows 11 info and put into telemetry

siyuniu-ms:


siyuniu-ms:
Comment from Nev:
Now that you have this mostly working (the console.log will need to be removed), I have a couple of extensions.

Lets not limit this to just windows, lets "ask" for the platform as well and when it's windows we will handle as required (as you seem to be already doing on line 155
Lets also "cache" the result(s) in Session Storage and look up those values first before attempting to do the call. So that on subsequent page loads if the values are in Session Storage we can avoid "blocking" the pipeline and just reuse the existing values

extensions/applicationinsights-osplugin-js/src/OsPlugin.ts
```diff
@@ -0,0 +1,253 @@
...
+    platformVersion: platformVersionInterface
+}
+interface ModernNavigator {
+    userAgentData?: {
+      getHighEntropyValues?: (fields: ["platformVersion"]) => Promise<UserAgentHighEntropyData>;
+    };
+  }
+const defaultOSConfig: IConfigDefaults<IOSPluginConfiguration> = objDeepFreeze({
+    getOSTimeoutMs: defaultgetOSTimeoutMs,
+    endpointIsBreeze: true
```
MSNev, line 37: Lets not call this the "internal" endpoint name.

How about something like mergeOsNameVersion, also lets have it default to `undefined` and we can add some code to automatically "detect" whether the Post channel or Sender is installed

`core.getPlugin("<sender or post channel name>").plugin !== null`

This way users don't really need to set this.

extensions/applicationinsights-osplugin-js/src/OsPlugin.ts
```diff
@@ -0,0 +1,253 @@
...
+            };
+        
+            _self.processTelemetry = (event: ITelemetryItem, itemCtx?: IProcessTelemetryContext) => {
+                itemCtx = _self._getTelCtx(itemCtx);
+
+                if (!_retrieveFullVersion && !_getOSInProgress && _retryTime < maxRetry) {
+                    // Start Requesting OS version process
+                    _getOSInProgress = true;
+                    // Timeout request if it takes more than 5 seconds (by default)
+                    _getOSTimeout = setTimeout(() => {
```
MSNev, line 120: Lets move the creation of the timer into the `startRetrieveOsVersion` so it doesn't try to create a timer unless it's needed

extensions/applicationinsights-osplugin-js/src/OsPlugin.ts
```diff
@@ -0,0 +1,253 @@
...
+                }
+        
+                if (_getOSInProgress) {
+                    _eventQueue.push({
+                        ctx: itemCtx,
+                        item: event
+                    });
+                } else {
+                    if (_retrieveFullVersion){
+                        if (_endpointIsBreeze){
```
MSNev, line 134: For the "setting" of the OS values, you have this here twice, lets extract into a single function

extensions/applicationinsights-osplugin-js/src/OsPlugin.ts
```diff
@@ -0,0 +1,253 @@
...
+                _ocConfig = null;
+                _getOSInProgress = false;
+                _getOSTimeout = null;
+                _getOSTimeoutMs = null;
+                _retrieveFullVersion = false;
+                _eventQueue = [];
+                _retryTime = 0;
+            }
+
+            // For backward compatibility
```
MSNev, line 232: There is no backward compatibility required here.

If you need them for testing then use the debug hook rather than exposing the internal values

siyuniu-ms:
[like]  Siyu Niu reacted to your message:
________________________________

extensions/applicationinsights-osplugin-js/src/OsPlugin.ts
```diff
@@ -0,0 +1,255 @@
...
+                            _completeOsRetrieve();
+                        });
+                    }
+                }
+            }
+
+            function updateTeleItemWithOs(event: ITelemetryItem) {
+                if (_retrieveFullVersion){
+                    if (_mergeOsNameVersion){
+                        setValue(getSetValue(event, strExt), Extensions.OSExt, _os + _osVer);
```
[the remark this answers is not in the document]
siyuniu-ms: have a manual test, it works good without space

channels/applicationinsights-channel-js/Tests/Unit/src/Sender.tests.ts
```diff
@@ -2533,8 +2533,10 @@ export class SenderTests extends AITestClass {
                         device: {
                             deviceClass: "Browser",
                             localId: "browser"
+                        },
+                        os: {
+                            osVer: "Windows11"
```
MSNev, line 2538: Is it really `Windows11`, I'm pretty sure it's supposed to be `Windows 11`

extensions/applicationinsights-osplugin-js/src/OsPlugin.ts
```diff
@@ -0,0 +1,257 @@
...
+            _initDefaults();
+
+            _self.initialize = (coreConfig: IConfiguration & IConfig, core: IAppInsightsCore, extensions: IPlugin[]) => {
+                let _self = this;
+                _core = core;
+                super.initialize(coreConfig, core, extensions);
+                let identifier = _self.identifier;
+                _evtNamespace = mergeEvtNamespace(createUniqueNamespace(identifier), core.evtNamespace && core.evtNamespace());
+                if (utlCanUseSessionStorage){
+                    _platformVersionResponse = JSON.parse(utlGetSessionStorage(core.logger, "ai_osplugin"));
```
MSNev, line 80: We should wrap this into a try / catch as "someone" could put garbage into the field that causes an exception,

extensions/applicationinsights-osplugin-js/src/OsPlugin.ts
```diff
@@ -0,0 +1,257 @@
...
+                _initDefaults();
+            };
+
+        
+            /**
+             * Wait for the response from the browser for the OS version and store info in the session storage
+             */
+            function startRetrieveOsVersion() {
+                // Timeout request if it takes more than 5 seconds (by default)
+                _getOSTimeout = setTimeout(() => {
```
MSNev, line 155: For this we can now use the `scheduleTimeout` which returns an `ITimerHandler`

MSNev: It then has a `cancel` function (instead of calling `clearTimeout` and you can just reset to null when done.

extensions/applicationinsights-osplugin-js/src/OsPlugin.ts
```diff
@@ -0,0 +1,257 @@
...
+                        setValue(extOS, "os", _os, isString);
+                    }
+                }
+            }
+        
+            /**
+            * Complete retrieving operating system info process, cleanup and release the event queue
+            */
+            function _completeOsRetrieve() {
+                if (_getOSTimeout) {
```
MSNev, line 208: So this will become 
`_getOSTimeout && _getOSTimeout.cancel()`

extensions/applicationinsights-osplugin-js/src/OsPlugin.ts
```diff
@@ -0,0 +1,257 @@
...
+                _releaseEventQueue();
+            }
+        
+            /**
+            * Release internal event queue
+            */
+            function _releaseEventQueue() {
+                arrForEach(_eventQueue, (evt) => {
+                    updateTeleItemWithOs(evt.item);
+                    evt.ctx.processNext(evt.item);
```
MSNev, line 221: This should really be `_self.processNext(event, itemCtx)`, while "technically" this will call the `itemCtx.processNext` as here. It's actually possible that the `ctx` is undefined or null (because of backward compatability), this is because depending on the extension that is "loaded" before this one it "may" pre-date the introduction of the `processNext` and instead just be the old "chained" `processTelemetry` calls.

extensions/applicationinsights-osplugin-js/src/DataModels.ts
```diff
@@ -0,0 +1,11 @@
+/**
+* DataModels.ts
+* @author Siyu Niu (siyuniu)
+* @copyright Microsoft 2024
+* File containing the interfaces for OS Plugin SDK.
+*/
+
+export interface IOSPluginConfiguration {
```
MSNev, line 8: Please also add the typedoc for this, as new components we should start by adding the documentation. Rather than requiring that we defer and add it later as part of the bigger typedoc improvements.

extensions/applicationinsights-osplugin-js/README.md
```diff
@@ -0,0 +1,54 @@
...
+npm install
+npm run build --silent
+```
+
+## Run unit tests:
+```
+npm run test
+```
+
+## Contributing
```
MSNev, line 27: This should include a Quick start guide on how to "use" this plugin and add it to a running system.

While we will probably "add" this by default in a future AISku version, we will still need to document it separately as some people (including 1DS users) won't get it as part of AISku

extensions/applicationinsights-osplugin-js/tslint.json
```diff
@@ -0,0 +1,8 @@
+{
```
MSNev, line 1: I think we can delete this file now as we (should) be using eslint by default. This is a left over of the previous linting code we used.

extensions/applicationinsights-osplugin-js/typedoc.json
```diff
@@ -0,0 +1,62 @@
...
+    "readme": "none",
+    "githubPages": true,
+    "gitRevision": "main",
+    "compilerOptions": {
+        "stripInternal": true
+    },
+    "sidebarLinks": {
+        "Changelog": "https://github.com/microsoft/ApplicationInsights-JS/blob/main/RELEASES.md",
+        "Examples": "https://github.com/microsoft/ApplicationInsights-JS/blob/main/examples/README.md",
+        "Readme": "https://github.com/Microsoft/ApplicationInsights-JS"
```
MSNev, line 50: This (I think) should be the Readme of this plugin not the root one (this is one of the "fixes" we will need for the other components as well)

shared/AppInsightsCommon/Tests/Unit/src/ThrottleMgr.tests.ts
```diff
@@ -797,11 +797,11 @@ export class ThrottleMgrTest extends AITestClass {
                 this._core.initialize(coreCfg, [this._channel]);
                 let throttleMgr = new ThrottleMgr(this._core);
 
-                let canThrottle = throttleMgr.canThrottle(this._msgId);
-                Assert.equal(canThrottle, false);
+                // let canThrottle = throttleMgr.canThrottle(this._msgId);
```
MSNev, line 800: ?

siyuniu-ms: The build pipeline was failed last night, so I am trying to narrow down the root casue which points to this test.

MSNev: Ah, this might be another "timing issue" based on "dates" as last month (Feb normally only has 28 days) -- leap years are fun.

MSNev (approved):

Issue #2311, opened 2024-03-20
[BUG] Excessive memory usage for SPA where unload hooks keep accumulating


MSNev:
An issue has been identified with the tracking the removal of all events, timers and config tracking functions where an excessive number of `unload` hooks can accumulate in long running applications.

The primary identified culprit is that the `getPerfMgr()` function will ALWAYS add a new unload hook whenever no performance manager currently exists (either explicitly or via create perf manager config function) -- which is the default situation, and when called multiple times without any of the configuration they keep accumulating. The object created and stored is not large, so it takes quite a while (or a lot of events) for the problem to become visible.

And as part of our performance testing (to track for excessive memory usage), we actually "install" a performance manager -- which doesn't catch this edge case...

The identified `getPerfMgr()` case only exists in v3.x and was introduced as part of the dynamic configuration support as the watcher being accumulated is the configuration watcher to "handle" configuration changes.

This Issue affects the following versions
- 3.0.0 -> 3.0.8
- 3.1.0 -> 3.1.1

We will be publishing both a 3.0.9 and a 3.1.2 with a fix for this issue.

siyuniu-ms:
related to https://github.com/microsoft/applicationinsights-react-js/issues/59

MSNev:
As a temporary workaround for this you can "add" a dummy performance manager the same was as VS Code has in their PR above. To avoid the issue, until we can release a fix for this (a PR of which is available)

PR #2312, opened 2024-03-20
[BUG] Excessive memory usage for SPA where unload hooks keep accumulating #2311

MSNev:
- Rework Asynchronous notification handling

### DIFF
--- a/AISKU/Tests/Unit/src/GlobalTestHooks.Test.ts
+++ b/AISKU/Tests/Unit/src/GlobalTestHooks.Test.ts
@@ -1,0 +1,13 @@
+import { Assert } from "@microsoft/ai-test-framework";
+import { _testHookMaxUnloadHooksCb } from "@microsoft/applicationinsights-core-js";
+import { dumpObj } from "@nevware21/ts-utils";
+
+export class GlobalTestHooks {
+
+    public registerTests() {
+        // Set a global maximum
+        _testHookMaxUnloadHooksCb(20, (state: string, hooks: Array<any>) => {
+            Assert.ok(false, "Max unload hooks exceeded [" + hooks.length + "] - " + state + " - " + dumpObj(hooks));
+        });
+    }
+}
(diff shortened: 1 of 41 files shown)


shared/AppInsightsCore/Tests/Unit/src/ApplicationInsightsCore.Tests.ts
```diff
@@ -999,6 +998,30 @@ export class ApplicationInsightsCoreTests extends AITestClass {
             }
         });
 
+        // this.testCase({
```
MSNev, line 1001: This is how I initially confirmed that this was the issue (after adding the test hooks).

Let me uncomment this to validate it again.

AISKU/Tests/Unit/src/GlobalTestHooks.Test.ts
```diff
@@ -0,0 +1,13 @@
+import { Assert } from "@microsoft/ai-test-framework";
+import { _testHookMaxUnloadHooksCb } from "@microsoft/applicationinsights-core-js";
+import { dumpObj } from "@nevware21/ts-utils";
+
+export class GlobalTestHooks {
+
+    public registerTests() {
```
MSNev, line 7: This is using a "hack" of our current test infrastructure to "set" some global configuration that will affect "ALL" tests that follow the initialization. (which is why this was added as the "first" item of every tests bundle.

Of note, is that IF any test also changes / sets / calls the test hook it will "replace" and invalidate this setting (as this is currently a single value, not a list of callbacks)

MSNev: They also needed to be created / added here as the "common" test infrastructure does not have access to the main "Application Insights" code to be able to add code specific "hooks"

Karlie-777 (approved):

siyuniu-ms (approved):

PR #2325, opened 2024-04-10
[Main][Task]25716927: Change default CfgSync values to turn on the ikey deprecation message 

Karlie-777:


AISKU/src/AISku.ts
```diff
@@ -89,7 +89,12 @@ const defaultConfigValues: IConfigDefaults<IConfiguration|IConfig> = {
             [_eInternalMessageId.SdkLdrUpdate]:cfgDfMerge<IThrottleMgrConfig>(default_throttle_config),
             [_eInternalMessageId.CdnDeprecation]:cfgDfMerge<IThrottleMgrConfig>(default_throttle_config)
         }
-    )
+    ),
+    extensionConfig: cfgDfMerge<{[key: string]: any}>({
+        ["AppInsightsCfgSyncPlugin"]: cfgDfMerge<ICfgSyncConfig>({
+            cfgUrl: CONFIG_ENDPOINT_URL
```
Karlie-777, line 95: This is our settings, the endpointURL set in CfgSyncPlugin should be the "user settings"

Karlie-777: and version for this should be 3.2.0

PR #2333, opened 2024-04-21
[Main][Task]27749889: Change AISKU Sync Mode to Receive

Karlie-777:


AISKU/src/AISku.ts
```diff
@@ -92,7 +92,8 @@ const defaultConfigValues: IConfigDefaults<IConfiguration & IConfig> = {
     ),
     extensionConfig: cfgDfMerge<{[key: string]: any}>({
         ["AppInsightsCfgSyncPlugin"]: cfgDfMerge<ICfgSyncConfig>({
-            cfgUrl: CONFIG_ENDPOINT_URL
+            cfgUrl: CONFIG_ENDPOINT_URL,
+            syncMode: ICfgSyncMode.Receive
```
Karlie-777, line 96: change the AISKU default receive mode to "receive" (will call "updateCfg" to update current core.config), previous one is "broadcast" (which means only send events out but not receiving any incoming changes)

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -317,7 +317,9 @@ export class CfgSyncPlugin extends BaseTelemetryPlugin implements ICfgSyncPlugin
                         if (JSON) {
                             let cdnCfg = JSON.parse(response); //comments are not allowed
                             let cfg = applyCdnfeatureCfg(cdnCfg, _self.core);
-                            cfg && _setCfg(cfg, isAutoSync);
+                            let newCfg = cfg && isPlainObject(cfg) && _replaceTartgetByKeys(cfg);
+                            newCfg && _setCfg(newCfg, isAutoSync);
```
Karlie-777, line 321: previous design: any configs from cdn, they will be directly applied to current `core.config` (it was designed to use `blkCdnCfg` to control if we need to apply cdn changes or not).
Current:
So add an extra layer, even if `blkCdnCfg` is set to `false`, any config values defined in `nonOverrideConfigs`  will not be replaced by cdn configs.

MSNev (approved):

PR #2337, opened 2024-05-01
[draft][main] image upload process

siyuniu-ms:


MSNev (changes requested):
We don't want to check all of these files into the repo, we may needs to "generate" them locally (like in a build or dist folder), but not check them in.

PR #2338, opened 2024-05-01
[Main][Task]27923018: Post Channel getOfflineSupport should set correct headers and url based on payload data

Karlie-777:


channels/1ds-post-js/src/PostChannel.ts
```diff
@@ -247,9 +247,11 @@ export class PostChannel extends BaseTelemetryPlugin implements IChannelControls
                                 return !_disableTelemetry;
                             },
                             createPayload: (evt) => {
+                                // should get new url headers based on payload directly
+                                let curDetails = _httpManager && _httpManager.getOfflineRequestDetails();
```
MSNev, line 251: This is probably badly named I think `getOfflineRequestDetails` should really be called `createPayload` or `getRequestDetails` to make it a little clearer what it's doing.

It's probably also a case that it should be (maybe) passing down and using the provided `evt` rather than just relying in the HttpManager returning the payload that it has batched?

It's probably a little messier than that, as the sender (AI) version does seem to create the IPayloadData with the provided data (evt).

MSNev (approved):

PR #2339, opened 2024-05-02
[main] create snippet mini loader

siyuniu-ms:


tools/applicationinsights-web-snippet/src/miniLoader.ts
```diff
@@ -0,0 +1,401 @@
...
+            "js.cdn.applicationinsights.io",
+            "js.cdn.monitor.azure.com",
+            "js0.cdn.applicationinsights.io",
+            "js0.cdn.monitor.azure.com",
+            "js2.cdn.applicationinsights.io",
+            "js2.cdn.monitor.azure.com",
+            "az416426.vo.msecnd.net" // this domain is supported but not recommended
+        ];
+
+        fetch("https://js.monitor.azure.com/beta/ai.3.integrity.json")
```
MSNev, line 207: `fetch` doesn't exist in all runtimes :-(

So we will need to abstract this out to use either
`fetch` or `XMLHttpRequest`, lets not add support for IE (which means we would also have to support `XDomainRequest` for this loader. Although, if we would want this to replace the auto injection snippet for the languages it would need to support that also.

MSNev: This is also why the `_sendEvents` in this script checks for the existence of either -- it also doesn't support sending events via `XDomainRequest`, but that is just for the "unable to load the SDK" error event.

It would probably be easier to just fallback to the `ai.3.min.xxx.js` rather than the full version script in that case as it would be less code.

tools/applicationinsights-web-snippet/src/miniLoader.ts
```diff
@@ -0,0 +1,401 @@
...
+            .then(response => response.json())
+            .then(json => {
+                let currentVersion = json.version;
```
MSNev, line 210: Rather than use the "version" it should use the value from the `ext.@<type>.file` and we will need a new config for the `<type>`  which will be one of the values in the file `@cjs.js`, `@cjs.min.js`, `@gbj.js`, `@gbl.min.js`, `@js` or `@min.js`
It will also need to grab the `integrity` value and include that as part of the `script` tag, as we are providing 2 assertions with this change
1) We will load the "full" version of the file for better page load (because it will be cached)
2) We will be adding the "integrity" to allow the browser to validate the script for better security.

For the second option, we will also need a configuration to allow the user to opt-out of adding the integrity

I've removed all of the hashes from the below copy of the current integrity file.

```json
{
    "name": "ai",
    "version": "3.2.0",
    "ext": {
        "@cjs.js": {
            "file": "ai.3.2.0.cjs.js",
            "type": "text/javascript; charset=utf-8",
            "integrity": "sha256-...==",
...
        },
        "@cjs.min.js": {
            "file": "ai.3.2.0.cjs.min.js",
            "type": "text/javascript; charset=utf-8",
            "integrity": "sha256-...==",
...
        },
        "@gbl.js": {
            "file": "ai.3.2.0.gbl.js",
            "type": "text/javascript; charset=utf-8",
            "integrity": "sha256-...==",
...
        },
        "@gbl.min.js": {
            "file": "ai.3.2.0.gbl.min.js",
            "type": "text/javascript; charset=utf-8",
            "integrity": "sha256-...==",
...
        },
        "@js": {
            "file": "ai.3.2.0.js",
            "type": "text/javascript; charset=utf-8",
            "integrity": "sha256-...==",
...
        },
        "@min.js": {
            "file": "ai.3.2.0.min.js",
            "type": "text/javascript; charset=utf-8",
            "integrity": "sha256-...==",
...
        }
    }
}

File: tools/applicationinsights-web-snippet/src/miniLoader.ts [hunk 3]
```diff
@@ -0,0 +1,434 @@
...
+            "js.cdn.monitor.azure.com",
+            "js0.cdn.applicationinsights.io",
+            "js0.cdn.monitor.azure.com",
+            "js2.cdn.applicationinsights.io",
+            "js2.cdn.monitor.azure.com",
+            "az416426.vo.msecnd.net" // this domain is supported but not recommended
+        ];
+
+        var sender = window.fetch;
+        var integrityUrl = "https://js.monitor.azure.com/beta/ai.3.integrity.json";
```
siyuniu-ms, line 209: load the integrity file, use this to get the latest version number, load the latest.min.js, and also check its hash number

tools/applicationinsights-web-snippet/src/type.ts
```diff
@@ -16,7 +16,7 @@ export interface ISnippetConfig {
     cfg: IConfiguration;
     cr?: boolean; // cdn retry would be proceed if ture
     dle?: boolean; // Custom optional value to disable sdk load error to be sent
-    type?: string; // Custom optional value to specify the type of the snippet, e.g. "cjs.js"
+    integrity?: boolean; // Custom optional value to specify whether add the integrity attribute to the script tag
```
MSNev, line 19: Lets keep the size smaller perhaps `intg` or `fi` (fetch integrity)or `ui` (user integrity) or

maybe better use the security name
- `sri` for (sub resource integrity)  https://www.w3schools.com/Tags/att_script_integrity.asp

MSNev: We should also document that when you use this the load order and script execution is affected, which means the usage of `ld` will be ignored (or have no affect) and if the page "navigates" away before we can load the integrity file etc that they may lose the events.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -202,11 +203,54 @@ declare var cfg:ISnippetConfig;
...
             "az416426.vo.msecnd.net" // this domain is supported but not recommended
-        ]
-    
-        // Assigning these to local variables allows them to be minified to save space:
-        let targetSrc : string = (aiConfig as any)["url"] || cfg.src
-        if (targetSrc) {
+        ];
+
+        let targetSrc : string = (aiConfig as any)["url"] || cfg.src;
+        var match = targetSrc.match(/^(https?:\/\/[^/]+)\/.*?\/ai\.(\d+(\.\d+){0,2})\.(.*?)$/);
```
MSNev, line 209: From investigations
^http[s]?:\/\/.*\/ai\.(\d+(\.\d+){0,2})\.(([\w\.]*\.){0,1}js)$

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -202,11 +203,52 @@ declare var cfg:ISnippetConfig;
...
+                    });
+            } else if (XMLHttpRequest) {
+                var xhr = new XMLHttpRequest();
+                xhr.open(strGetMethod, integrityUrl);
+                xhr.onreadystatechange = function () {
+                    if (xhr.readyState === XMLHttpRequest.DONE) {
+                        if (xhr.status === 200) {
+                            var json = JSON.parse(xhr.responseText);
+                            integrity = json.ext[targetType].integrity;
+                            targetSrc = match[1] + json.ext[targetType].file.substring(3);
```
siyuniu-ms, line 237: do we have any better solution instead of calling substring? I am thinking about using [2:] also. Not quite sure which one is better.

MSNev: Lets stick with substring for now as it is supported on all browsers, so it should be supported on the runtimes.

MSNev: Actually, why use substring at all, we should be able to get the regex to give us the domain portion and then this is just a domain + file -- this will also support case (in the future)  where the prefix might not be "ai.3" for example there is already the basic version which is aib.x.min.js (for AISKULight), so we can't guarantee that it will always be that way. And as this is the SDK Loader that gets embedded into applications we won't have the option to "update" it if/when we change it.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -234,7 +233,7 @@ declare var cfg:ISnippetConfig;
                         if (xhr.status === 200) {
                             var json = JSON.parse(xhr.responseText);
                             integrity = json.ext[targetType].integrity;
-                            targetSrc = match[1] + json.ext[targetType].file.substring(3);
+                            targetSrc = match[1] + json.ext[targetType].file;
```
MSNev, line 236: Actually, we should add some additional checks here so we can correctly handle any error cases like
- Unable to convert the response to JSON (this will either throw or return an empty object)
- Check for the existence of the `json.ext` `json.ext[targetType]` as these would case an exception to be thrown

A try/catch might be useful for these rather than testing for all cases, so that if for some reason we fail to fetch/process the json response we still fall back to the error case of either sending an "SDK Load failure" message (not a console log) or just trying to load the default provided version without the integrity attribute

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -202,11 +204,60 @@ declare var cfg:ISnippetConfig;
...
-    
-        // Assigning these to local variables allows them to be minified to save space:
-        let targetSrc : string = (aiConfig as any)["url"] || cfg.src
-        if (targetSrc) {
+        ];
+
+        let targetSrc : string = (aiConfig as any)["url"] || cfg.src;
+        var match = targetSrc.match(/^(http[s]?:\/\/.*\/)([\w]+\.)?(\d+(\.\d+){0,2})\.(([\w]+\.){0,2}js)$/);
+        if (match.length === 7 && cfg.sri) {
+            var integrityUrl = match[1] + "ai." + match[3] + ".integrity.json";
```
MSNev, line 212: We don't really want to depend on the "name" being `ai.` here either...

MSNev: if you add an extra group around your regex you can get another group match which includes everything except the the end bit like `.min.js` and then use that to just append the `.integrity.json`

((http[s]?:\/\/.*\/)([\w]+\.)?(\d+(\.\d+){0,2}))\.(([\w]+\.){0,2}js)$

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -202,11 +204,60 @@ declare var cfg:ISnippetConfig;
...
+                    .then(json => {
+                        if (!json.ext || !json.ext[targetType]) {
+                            throw new Error(strJsonResponseError);
+                        }
+                        integrity = json.ext[targetType].integrity;
+                        targetSrc = match[1] + json.ext[targetType].file;
+                        setScript(targetSrc, integrity);
+                    })
+                    .catch(error => {
+                        console.error(strJsonResponseError, error);
```
MSNev, line 229: We don't really want to use the console for anything as the console object itself doesn't always exist in all runtimes.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -202,11 +204,60 @@ declare var cfg:ISnippetConfig;
...
+                xhr.open(strGetMethod, integrityUrl);
+                xhr.onreadystatechange = function () {
+                    if (xhr.readyState === XMLHttpRequest.DONE && xhr.status === 200) {
+                        try {
+                            var json = JSON.parse(xhr.responseText);
+                            if (!json.ext || !json.ext[targetType]) {
+                                throw new Error(strJsonResponseError);
+                            }
+                            integrity = json.ext[targetType].integrity;
+                            targetSrc = match[1] + json.ext[targetType].file;
```
MSNev, line 243: We probably should also check if `integrity` and `targetSrc` contain values (as someone could have a json file that doesn't include these) and if we pass null/undefined onto the `setScript` it won't be happy either and will throw on the first line.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -202,11 +204,60 @@ declare var cfg:ISnippetConfig;
...
-        ]
-    
-        // Assigning these to local variables allows them to be minified to save space:
-        let targetSrc : string = (aiConfig as any)["url"] || cfg.src
-        if (targetSrc) {
+        ];
+
+        let targetSrc : string = (aiConfig as any)["url"] || cfg.src;
+        var match = targetSrc.match(/^(http[s]?:\/\/.*\/)([\w]+\.)?(\d+(\.\d+){0,2})\.(([\w]+\.){0,2}js)$/);
+        if (match.length === 7 && cfg.sri) {
```
MSNev, line 211: If there is no "match" because the URL is something unexpected, I think it will be null -- which will then cause an exception here...

We should check if there was a match and maybe even defer the regex until after we check if they have requested (enabled) the `cfg.sri` so that we don't try to do anything until they actually ask us.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -202,11 +204,59 @@ declare var cfg:ISnippetConfig;
...
-        if (targetSrc) {
+        ];
+
+        let targetSrc : string = (aiConfig as any)["url"] || cfg.src;
+        const fallback = () => setScript(targetSrc, null);
+        if (cfg.sri) {
+            const match = targetSrc.match(/^((http[s]?:\/\/.*\/)([\w]+\.))?(\d+(\.\d+){0,2})\.(([\w]+\.){0,2}js)$/);
+            
+            if (match && match.length === 8) {
+                const integrityUrl = `${match[1]}${match[4]}.integrity.json`;
```
MSNev, line 215: This doesn't look correct...
for http://js.monitor.azure.com/scripts/b/ai.3.min.js
match[1] looks like `http://js.monitor.azure.com/scripts/b/ai.` and
match[4] looks like ""

So while this "works" (because match[4] is empty) I think it should just be `match[1] + ".integrity.json"` (note not using the string replacement "`${match[1]....." for generated code sizing.

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -202,11 +204,59 @@ declare var cfg:ISnippetConfig;
...
+                    }
+                    const integrity = json.ext[targetType].integrity || null;
+                    targetSrc = match[2] + json.ext[targetType].file;
+                    setScript(targetSrc, integrity);
+                };
+        
+                if (sender && !cfg.useXhr) {
+                    sender(integrityUrl, { method: strGetMethod, mode: "cors" })
+                        .then(response => response.json().catch(() => ({})))
+                        .then(handleResponse)
```
MSNev, line 230: 👍

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -202,11 +204,59 @@ declare var cfg:ISnippetConfig;
...
+                        .catch(fallback);
+                } else if (XMLHttpRequest) {
+                    const xhr = new XMLHttpRequest();
+                    xhr.open(strGetMethod, integrityUrl);
+                    xhr.onreadystatechange = () => {
+                        if (xhr.readyState === XMLHttpRequest.DONE) {
+                            if (xhr.status === 200) {
+                                try {
+                                    handleResponse(JSON.parse(xhr.responseText));
```
MSNev, line 239: 👍

tools/applicationinsights-web-snippet/src/snippet.ts
siyuniu-ms: work for local test: 
http://js.monitor.azure.com/scripts/b/ai.3.min.js
http://js.monitor.azu/sdfsd2344f.3.js
http://azure.com/script/b/ai.3.2.1.min.js
http://azure.com/script/b/ai.3.js
http://azure.com/script/b/ai444.3.js
http://azure.com/script/b/ai444ddf.3.js
http://js.monitor.azure.com/scripts/b/ai.3.2.1.gbl.min.js

README.md
```diff
@@ -135,6 +136,7 @@ The available configuration options are: -
...
 | cfg | object **[required]** | The configuration passed to the Application Insights SDK during initialization.
+| sri | boolean *[optional]* | Custom optional value to specify whether to fetch the snippet from an integrity file and perform an integrity check. When this option is used, the integrity file is loaded first, affecting the load order and script execution. Hence the ld option will be ignored. Additionally, if the page navigates away before the integrity file is loaded, some events may be lost.
```
MSNev, line 139: 👍

README.md
```diff
@@ -80,14 +80,15 @@ The current version of the snippet is version 7, the version is identified by th
...
...("GET",d),a.onreadystatechange=function(){var e;a.readyState===XMLHttpRequest.DONE&&(200===a.status?(e=JSON.parse(a.responseText),i=e.version,n=e.ext[t].integrity,r(o=o.replace(/(?<=\/ai\.)\d+(\.\d+){0,2}/,i),n)):(console.error("Error loading JSON:",a.statusText),r(o,n)))},a.send())):o&&r(o,n);try{h.cookie=w.cookie}catch(p){}function s(e){for(;e.length;)!function(t){h[t]=function(){var e=arguments;f||h.queue.push(function(){h[t].apply(h,e)})}}(e.pop())}var c,u,e="track",d="TrackPage",l="TrackEvent",e=(s([e+"Event",e+"PageView",e+"Exception",e+"Trace",e+"DependencyData",e+"Metric",e+"PageViewPerformance","start"+d,"stop"+d,"start"+l,"stop"+l,"addTelemetryInitializer","setAuthenticatedUserContext","clearAuthenticatedUserContext","flush"]),h.SeverityLevel={Verbose:0,Information:1,Warning:2,Error:3,Critical:4},(g.extensionConfig||{}).ApplicationInsightsAnalytics||{});return!0!==g[q]&&!0!==e[q]&&(s(["_"+(c="onerror")]),u=E[c],E[c]=function(e,t,n,i,a){var o=u&&u(e,t,n,i,a);return!0!==o&&h["_"+c]({message:e,url:t,lineNumber:n,columnNumber:i,error:a,evt:E.event}),o},g.autoExceptionInstrumented=!0),h}(cfg.cfg),(E[n]=i).queue&&0===i.queue.length?(i.queue.push(e),i.trackPageView({})):e();})({
```
MSNev, line 83: I'm not seeing the updated regex here...

AISKU/Tests/Manual/testVersionConflict.html
```diff
@@ -7,10 +7,10 @@
...
+    src: "http://js.monitor.azu/sdfsd2344f.3.js",
```
MSNev, line 13: This should still be the recommended default 😄

MSNev (approved):
Just one comment on the Manual Test file

PR #2340, opened 2024-05-02
[Main][Task]27939476: Initialization with iKey and endpoint to be promises

Karlie-777:


AISKU/src/AISku.ts
```diff
@@ -194,8 +194,47 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                let configCs =  _config.connectionString;
+                
+                if (isPromiseLike(configCs)) {
+                    let ikeyPromise = createAsyncPromise<string>((resolve, reject) => {
+                        doAwaitResponse(configCs, (res) => {
+                            let curCs = res.value;
+                            let ikey = _config.instrumentationKey;
+                            if (!res.rejected && curCs) {
+                                // replace cs with resolved values in case of circular promises
+                                _config.connectionString = curCs;
```
Karlie-777, line 206: need to set connection string to be string to avoid future promise calls

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -268,6 +270,12 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                         }
                     });
 
+                    if (core.activeStatus() === ActiveStatus.PENDING) {
+                        // waiting for core promises to be resolved
+                        return;
```
Karlie-777, line 275: if core active status is pending, do nothing

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -330,19 +343,104 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
...
+                    let isPending = _activeStatus === eActiveStatus.PENDING;
+                    
+                    if (isPending){
+                        // mean waiting for previous promises to be resolved, won't apply new changes
+                        return;
+                    }
+
+                    if (isNullOrUndefined(ikey)) {
+                        _activeStatus = ActiveStatus.INACTIVE;
+                        if (!_isInitialized) {
```
Karlie-777, line 363: only throw error during first initialization, not during config changes

AISKU/src/AISku.ts
```diff
@@ -200,8 +200,47 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                                let resolvedCs = parseConnectionString(curCs);
+                                ikey = resolvedCs.instrumentationkey || ikey;
+                            }
+                            resolve(ikey);
+                        });
+
+                    });
+
+                    let urlPromise = createAsyncPromise<string>((resolve, reject) => {
```
MSNev, line 221: There are some optimizations we could do here, but this works because promises support having multiple resolve / reject listeners -- but probably not necessary for v1. 

As it's probably messier, where you would extract out the `parseConnectionString()` so it's only called once and as part of this you would cause the iKey and endpoint Promises to get resolved, which means they need to be created before the doAwaitResponse for the connection string and their resolve / reject functions put into external variables that could be called from within the doAwaitResponse...

Karlie-777: Thanks Nev!  I will do a minifications and optimizations(on other parts as well)

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -598,11 +598,15 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
 
                     if (_enableAjaxPerfTracking) {
                         let iKey = config.instrumentationKey || "unkwn";
-                        if (iKey.length > 5) {
-                            _markPrefix = AJAX_MONITOR_PREFIX + strSubstring(iKey, iKey.length - 5) + ".";
-                        } else {
-                            _markPrefix = AJAX_MONITOR_PREFIX + iKey + ".";
+                        // only change the ikey if it is string
+                        if (isString(iKey)) {
```
MSNev, line 602: This is probably going to be an interesting race condition at some point. We might need to change this at some point so it's using the core instance id rather than the iKey.

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -330,19 +347,131 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
...
+                        // string or null
+                        _endpoint = endpointUrl;
+                    }
+
+                    // at least have one promise
+                    if (promises.length) {
+                        // reset to false for new dynamic changes
+                        _isStatusSet = false;
+                        _activeStatus = eActiveStatus.PENDING;
+                        let initTimeout = isNullOrUndefined(rootCfg.initTimeOut)?  rootCfg.initTimeOut : maxInitTimeout; // rootCfg.initTimeOut could be 0
```
MSNev, line 403: This looks like it's the wrong way around...

extensions/applicationinsights-dependencies-js/src/ajax.ts
Karlie-777: add todo and a task

shared/AppInsightsCommon/src/applicationinsights-common.ts
```diff
@@ -7,6 +7,7 @@ export {
 } from "./Util";
 export { ThrottleMgr } from "./ThrottleMgr";
 export { parseConnectionString, ConnectionStringParser } from "./ConnectionStringParser";
+export { ConnectionString } from "./Interfaces/ConnectionString";
```
Karlie-777, line 10: export connection string type from common

AISKU/src/AISku.ts
```diff
@@ -200,8 +200,62 @@ export class AppInsightsSku implements IApplicationInsights {
...
+                        _parseCs().then((cs) => {
+                            let url = _config.endpointUrl;
+                            let ingest = cs && cs.ingestionendpoint;
+                            url = ingest? ingest + DEFAULT_BREEZE_PATH : url;
+                            resolve(url);
```
MSNev, line 241: This should also return the `_config.userOverrideEndpointUrl` as if someone passes this and a promise connection string  this would end up overriding their override with the connection string.

Actually, if they provide a user override, we probably just should worry about parsing the connection string here. So avoid all of the lines from 236-248

extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -598,11 +598,16 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
 
                     if (_enableAjaxPerfTracking) {
                         let iKey = config.instrumentationKey || "unkwn";
-                        if (iKey.length > 5) {
-                            _markPrefix = AJAX_MONITOR_PREFIX + strSubstring(iKey, iKey.length - 5) + ".";
-                        } else {
-                            _markPrefix = AJAX_MONITOR_PREFIX + iKey + ".";
+                        // only change the ikey if it is string
```
MSNev, line 601: Lets drop this code (for now) as this is the bit that Siyu is now looking at. If you need to make it compile just cast the instrumentationKey to a string (to minimize the amount of merge conflicts this is going to cause).

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -330,19 +347,131 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
...
+                        _activeStatus = eActiveStatus.PENDING;
+                        let initTimeout = isNotNullOrUndefined(rootCfg.initTimeOut)?  rootCfg.initTimeOut : maxInitTimeout; // rootCfg.initTimeOut could be 0
+                        let allPromises = createAllSettledPromise<string>(promises);
+                        _initTimer = scheduleTimeout(() => {
+                            // set _isStatusSet to true
+                            // set active status
+                            // release queues
+                            _initTimer = null;
+                            if (!_isStatusSet) {
+                                _setStatus();
```
MSNev, line 411: Lets create a task (for potential later implementation) to add support to "clear" all of the _eventQueue items when we fail to initialize... I can see that some people might want to clear (to save memory) while others might want to "try" again to get the connection string

extensions/applicationinsights-dependencies-js/Tests/Unit/src/ajax.tests.ts
```diff
@@ -208,6 +209,59 @@ export class AjaxTests extends AITestClass {
             }
         });
 
+        this.testCaseAsync({
+            name: "Dependencies Configuration: init with cs promise ikey promise",
```
MSNev, line 213: This test is probably going to fail if we roll back the mark changes :-( -- @siyuniu-ms we might need for you to add this tests as part of your changes

Karlie-777: I will comment out this test! and once the change is done, we can uncomment it!

siyuniu-ms: I've added this in the task description for reminder.

MSNev (approved):

PR #2358, opened 2024-05-31
[Main][Task]27080650: Initialization Should Handle Offline Support Dependency

Karlie-777:


shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -913,7 +919,7 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
                 
                 // Initializing the channels first
                 if (_channels && _channels.length > 0) {
-                    initializePlugins(rootCtx.createNew(_channels), allExtensions);
+                    initializePlugins(rootCtx.createNew(_channels), allExtensions, delayInit);
                 }
```
Karlie-777, line 923: offline support will not be initialized with other channels

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -885,6 +886,11 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
             function _initPluginChain(updateState: ITelemetryUpdateState | null) {
                 // Extension validation
                 let theExtensions = _validateExtensions(_self.logger, ChannelControllerPriority, _configExtensions);
+                let delayInit: IPlugin[] = [];
```
MSNev, line 889: Not exactly what I was thinking as the offline channel still needs to init before the post channel....

I was thinking that within the Offline channel it's dependencies on the Post Channel (Offline support) would be lazy so it doesn't try and "get" it until after initialization of itself.

channels/offline-channel-js/Tests/Unit/src/channel.tests.ts
```diff
@@ -79,6 +81,8 @@ export class ChannelTests extends AITestClass {
                     }
                 });
 
+                this.clock.tick(1);
```
MSNev, line 84: Hmm, this triggered an interesting thought (not sure if it true)...

When the SDK is initialized from the snippet it "initializes" and then sends out a "trackPageView" (synchronously) which if we are online is not a problem, but if the offlineListener says we are offline this will try and add the event to the offline batch which its won't be able to do because the "timer" has not yet been triggered (because everything from initialization -> trackPageview is synchronous) so this could actually cause the lose of this event....

Let me go back and look at the offline channel, we should be able to rework fairly simply to get rid of the timer (I think) and only perform the "lazy" initialization (if) required.....

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -119,31 +123,40 @@ export class OfflineChannel extends BaseTelemetryPlugin implements IChannelContr
...
+                scheduleTimeout(() => {
+                    try {
+                        let _dependencyPlugin = _getDependencyPlugin(coreConfig, core);
+                        // make sure that online sender is initialized
+                        if (!_isLazyInit  && _dependencyPlugin && _dependencyPlugin.isInitialized()) {
+                            _isLazyInit = true;
+                            _createUrlConfig(coreConfig, core, extensions, pluginChain);
+                            let ctx = _getCoreItemCtx(coreConfig, core, extensions, pluginChain);
+                            _sender.initialize(coreConfig, core, ctx, _diagLogger, _primaryChannelId, _self._unloadHooks);
+                            if (_sender) {
```
MSNev, line 136: nit: _sender will always be true otherwise the above line would have thrown an exception, so don't really need this if check.

channels/offline-channel-js/src/OfflineChannel.ts
```diff
@@ -119,31 +123,40 @@ export class OfflineChannel extends BaseTelemetryPlugin implements IChannelContr
...
-                                _sendNextBatchTimer && _sendNextBatchTimer.cancel();
-                            } else {
+                // TODO: add refresh inside timer until dependency plugin is initialized
+                scheduleTimeout(() => {
+                    try {
+                        let _dependencyPlugin = _getDependencyPlugin(coreConfig, core);
+                        // make sure that online sender is initialized
+                        if (!_isLazyInit  && _dependencyPlugin && _dependencyPlugin.isInitialized()) {
+                            _isLazyInit = true;
+                            _createUrlConfig(coreConfig, core, extensions, pluginChain);
```
MSNev, line 133: As this is where the `_offlineSupport` gets populated the issue I mentioned in the test could occur as the `processTelmetry` is not currently checking if `_offlineSupport` is null / undefined (so it would throw)...

What we could do is rather than define the function here (inline), we could extract it out and then.
- Still keep this in a timer so it's gets initiated as it does not
- During the "processTelemetry" (or really anywhere where `_offlineSupport` is used) if `_offlineSupport` does not yet have a value we could then call this function their (also cancelling this timer) so it should then cause the lazy initialization to occur (synchronously).

MSNev: As discussed there is already the check for _offlineSupport in processtelemetry that will handle this case

channels/offline-channel-js/Tests/Unit/src/channel.tests.ts
MSNev: As discussed there is already the check for _offlineSupport in processtelemetry that will handle this case

channels/offline-channel-js/Tests/Unit/src/offlinetimer.tests.ts
```diff
@@ -43,6 +43,8 @@ export class Offlinetimer extends AITestClass {
                 let channel = new OfflineChannel();
                 let onlineChannel = new TestChannel();
                 this.core.initialize(this.coreConfig,[channel, onlineChannel]);
+
+                this.clock.tick(1);
```
MSNev, line 47: I assume that there is already a test which doesn't do this somewhere, so that the PostChannel -> isOffline -> OfflineListener sequence is tested, if not we should probably add that (it will be a very asynchronous test -- and may already be (or be better located) in the e2e tests.

shared/AppInsightsCore/Tests/Unit/src/ApplicationInsightsCore.Tests.ts
```diff
@@ -338,6 +338,49 @@ export class ApplicationInsightsCoreTests extends AITestClass {
...
+            useFakeTimers: true,
+            test: () => {
+                let offlineChannelPlugin = new TestOfflineChannelPlugin();
+
+                let channelPlugin = new TestChannelPlugin();
+              
+
+                const appInsightsCore = new AppInsightsCore();
+                appInsightsCore.initialize(
+                    { instrumentationKey: "09465199-12AA-4124-817F-544738CC7C41", channels: [[offlineChannelPlugin, channelPlugin]] },
```
MSNev, line 352: Do we really need a real looking key? using a dummy key for testing is probably better to avoid any possible clash or exposure of internal testing keys

MSNev (approved):
Just a couple of suggestions -- no real blockers

channels/offline-channel-js/src/OfflineChannel.ts
Karlie-777: removed if (_sender)

channels/offline-channel-js/Tests/Unit/src/offlinetimer.tests.ts
Karlie-777: added additional test in AISKU with the scenario

shared/AppInsightsCore/Tests/Unit/src/ApplicationInsightsCore.Tests.ts
Karlie-777: changed the ikey to "testIkey"

PR #2365, opened 2024-06-20
[main][snippet] prepare new snippet release, add support for more config

siyuniu-ms:


tools/applicationinsights-web-snippet/src/type.ts
```diff
@@ -1,9 +1,18 @@
...
-// in the future, we can add more fields here
 export interface SdkLoaderConfig {
     instrumentationKey?: string;
     connectionString?: string;
+    src?: string;
+    name?: string;
+    ld?: number;
+    useXhr?: boolean;
+    crossOrigin?: string;
+    onInit?: any;
```
MSNev, line 11: onInit: should be defined as a function -- although I'm not sure that the JSON.stringify() is going to do what we want

tools/applicationinsights-web-snippet/src/applicationinsights-web-snippet.ts
```diff
@@ -13,13 +14,28 @@ function webSnippetVersion() {
...
-    } else if (config && config.instrumentationKey) {
-        snippet = webSnippet.replace("InstrumentationKey=INSTRUMENTATION_KEY", config.instrumentationKey);
+    let snippetConfig: ISnippetConfig = {
+        src: config.src? config.src : "https://js.monitor.azure.com/scripts/b/ai.3.gbl.min.js",
+        crossOrigin: config.crossOrigin ? config.crossOrigin : "anonymous",
+        cfg: {},
+        name: config.name ? config.name : "appInsights",
+        ld: config.ld,
+        useXhr: config.useXhr,
+        onInit: config.onInit,
```
MSNev, line 24: Can you write a quick test to see if you pass in a function whether it will actually get called?

I don't think it will, as I think this process is just going to populate with a text version of the function....

If it doesn't work, lets just remove it from this version as we are going to have to do something tricky to handle / pass a Promise through this for the same reason as why I think the function isn't going to work.

siyuniu-ms: it doesn't work, I will remove onInit for this version. Thanks!

MSNev (approved):

PR #2373, opened 2024-07-08
[main] add withCredentials config

siyuniu-ms:


shared/AppInsightsCore/src/JavaScriptSDK/SenderPostManager.ts
```diff
@@ -93,16 +93,17 @@ export class SenderPostManager {
...
-                    if (!_isOneDs) {
-                        _sendCredentials = false; // for appInsights, set it to false always
+                    // remove since 3.3.1
+                    // if (!_isOneDs) {
+                    //     _sendCredentials = false; // for appInsights, set it to false always
+                    // }
+                    let location = getLocation();
+                    if (location && location.protocol && location.protocol.toLowerCase() === "file:") {
+                        // Special case where a local html file fails with a CORS error on Chromium browsers
+                        _sendCredentials = false;
```
siyuniu-ms, line 103: not quite sure about this part.   Previously, when disableCredentials is true, and the code is not being run from a local HTML file, then sendCredentials will still be ture. Is this by design?

shared/AppInsightsCore/src/JavaScriptSDK/SenderPostManager.ts
```diff
@@ -379,8 +380,7 @@ export class SenderPostManager {
                 }
 
 
-                if (_sendCredentials && _isOneDs) {
-                    // for 1ds, Don't send credentials when URL is file://
+                if (_sendCredentials) {  // fetch request credentials
```
siyuniu-ms, line 383: this is also related to credentials. also not sure we should change it or not as this is for fetch request instead of xhr.

shared/AppInsightsCore/src/JavaScriptSDK/SenderPostManager.ts
```diff
@@ -57,7 +57,7 @@ export class SenderPostManager {
 
         dynamicProto(SenderPostManager, this, (_self, _base) => {
 
-            let _sendCredentials = true; // for 1ds
+            let _sendCredentials = true; // default
```
MSNev, line 60: I believe that this class is used by both AI and 1DS, so I don't think its going to be a simple true/false

shared/AppInsightsCore/src/JavaScriptSDK/SenderPostManager.ts
```diff
@@ -379,8 +380,7 @@ export class SenderPostManager {
                 }
 
 
-                if (_sendCredentials && _isOneDs) {
-                    // for 1ds, Don't send credentials when URL is file://
+                if (_sendCredentials) {  // fetch request credentials
                     init.credentials = "include";
```
MSNev, line 384: This is also not exactly what is being asked.
the `credentials` value for fetch has 3 diffrent values https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch, the ask is to set this to `omit`

siyuniu-ms: yes, fetch has three and xhr only have true and false. I will combine them into one config.

MSNev (approved):

channels/1ds-post-js/src/DataModels.ts
```diff
@@ -245,6 +245,22 @@ export interface IChannelConfiguration {
      * @since 4.1.0
      */
     excludeCsMetaData?: boolean;
+
+    /**
```
MSNev, line 249: Lets also add a @since tag stating that this was added in v4.3.1

PR #2378, opened 2024-07-16
[main] image generator (temp fix for pipeline)

siyuniu-ms:


AISKU/scripts/publishAzImgToCdn.ps1
```diff
@@ -107,18 +117,23 @@ if ([string]::IsNullOrWhiteSpace($global:connectDetails.sasToken) -eq $true) {
...
 Write-Log "Release Files : $($releaseFiles.Count)"
-
+Write-Log "Release Night Files : $($releaseNightFiles.Count)"
 Write-Log "----------------------------------------------------------------------"
 
 $contentDisposition = "inline"
 
 # Publish the img to the folder that is same to the script folder.
 PublishFiles $releaseFiles "scripts/b" $cacheControl $contentType $overwrite $contentDisposition
+PublishFiles $releaseNightFiles "nightly" $cacheControl $contentType $overwrite $contentDisposition
```
MSNev, line 137: Why call out nightly specifically???

This can be abstracted out as the "container" which can be "dev", "nightly", "next", "beta" or "tst" which is what this portion of the main publishing script is doing https://github.com/microsoft/ApplicationInsights-JS/blob/db897fe314e0fc8d09f6b2768979b2e0d33c8c99/AISKU/scripts/publishReleaseToCdn.ps1#L125-L148...

This script "should" be able to handle any of these without needing to call out and add specific collections like lines 73-77

siyuniu-ms: because when I passed in the versionlist, the version is a string and does not have type

siyuniu-ms: I see. I thought the type was passed in. Turns out that Get-VersionDetails could assign the type.

tools/sizeImageGenerator/size-image-generator.js
```diff
@@ -1,29 +1,34 @@
...
+            label = 'GZip size';
+        } else if (type == "min") {
+            label = "minified size";
+            sizecolor = "darkorchid";
+        } 
+        const format = {
+            label: label,
+            message: fileSize + ' KB',
+            color: sizecolor,
+          }
```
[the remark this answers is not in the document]
MSNev: We should fix this to stop any future issue being created because of the CodeQL issue.

MSNev (approved):

siyuniu-ms:
implemented in internal pipeline instead

PR #2388, opened 2024-08-02
[main] enhance exception telemetry with optional script Information

siyuniu-ms:


shared/AppInsightsCommon/src/Interfaces/IConfig.ts
```diff
@@ -389,6 +389,12 @@ export interface IConfig {
      * This URL takes precedence over the 'config.endpointUrl' and any endpoint in the connection string.
      */
     userOverrideEndpointUrl?: string;
+
+    /**
+     * [Optional] If set to true, when exception is sent out, the SDK will also send out all scripts basic info that are loaded on the page.
+     * Notice: This would increase the size of the exception telemetry.
+     */
+    getExceptionScriptsInfo?: boolean;
```
MSNev, line 397: Rather than "get" (which to me sounds like a function) we want to call out that this is an option, which is where we have in the past called these "enableXXXX" or "disableXXXX" in this case maybe something like "includeXXX" and as we are also adding the debug logs and maybe some other stuff for the internal project, perhaps we call these out into an exception config (like the Cookie manager config)

So something like

```ts
expCfg: {
    inclScripts: true/false,
    getDbgLogs: () => string[]
}
```

We can call the interface name for this something meaningful like `IExceptionConfig` as I suspect that we are going to need additional optional "values" that partners will need to include.

shared/AppInsightsCommon/src/Interfaces/IConfig.ts
```diff
@@ -391,10 +392,9 @@ export interface IConfig {
     userOverrideEndpointUrl?: string;
 
     /**
-     * [Optional] If set to true, when exception is sent out, the SDK will also send out all scripts basic info that are loaded on the page.
-     * Notice: This would increase the size of the exception telemetry.
+     * [Optional] Set additional configuration for exceptions, such as more scripts to include in the exception telemetry.
      */
-    getExceptionScriptsInfo?: boolean;
+    exceptionConfig?: IExceptionConfig;
```
MSNev, line 397: Lets keep this small `expCfg`

shared/AppInsightsCommon/src/Interfaces/IExceptionTelemetry.ts
```diff
@@ -4,6 +4,14 @@
 import { SeverityLevel } from "./Contracts/SeverityLevel";
 import { IPartC } from "./IPartC";
 
+export interface IExceptionConfig{
+    /**
+     * If set to true, when exception is sent out, the SDK will also send out all scripts basic info that are loaded on the page.
+     * Notice: This would increase the size of the exception telemetry.
+     */
+    includeScripts?: boolean;
```
MSNev, line 12: likewise, keep the name small

shared/AppInsightsCore/src/JavaScriptSDK/W3cTraceParent.ts
```diff
@@ -202,3 +202,46 @@ export function findW3cTraceParent(selectIdx?: number): ITraceParent {
...
+ */
+export function findAllScripts(doc: any) {
+    let scripts = doc.getElementsByTagName("script");
+    let result: scriptsInfo[] = [];
+    for (let i = 0; i < scripts.length; i++) {
+        let script = scripts[i];
+        let src = script.getAttribute("src");
+        if (src) {
+            let crossOrigin = script.getAttribute("crossorigin");
+            let async = script.getAttribute("async") === "true";
```
MSNev, line 227: Depending on the browser, these may actually have no value as you only "need" to have the attribute present like

https://www.w3schools.com/tags/att_script_async.asp and https://www.w3schools.com/tags/att_script_defer.asp

So they won't necessarily have a value of "true", so you should check with `hasAttribute` and even then it's possible that the value may be false (in older browsers). So its a little complicated

siyuniu-ms: Thanks. Didn't notice that part could fail.

shared/AppInsightsCore/src/JavaScriptSDK/W3cTraceParent.ts
```diff
@@ -202,3 +202,46 @@ export function findW3cTraceParent(selectIdx?: number): ITraceParent {
...
+
+/**
+ * Find all script tags in the provided document and return the information about them.
+ * @param doc
+ * @returns
+ */
+export function findAllScripts(doc: any) {
+    let scripts = doc.getElementsByTagName("script");
+    let result: scriptsInfo[] = [];
+    for (let i = 0; i < scripts.length; i++) {
```
MSNev, line 222: Lets use `arrForEach` to help with minification

shared/AppInsightsCore/src/JavaScriptSDK/W3cTraceParent.ts
```diff
@@ -202,3 +202,46 @@ export function findW3cTraceParent(selectIdx?: number): ITraceParent {
 
     return traceParent;
 }
+
+export interface scriptsInfo {
+    url: string;
+    crossOrigin?: string;
+    async?: boolean;
+    defer?: boolean;
+    referrePolicy?: string;
```
MSNev, line 211: typo, missing an 'r' 😄

MSNev (approved):

PR #2391, opened 2024-08-07
[main] (type change!) enhance exception telemetry with customer log Information 

siyuniu-ms:


docs/exceptionTelemetry.md
```diff
@@ -14,7 +14,10 @@ const appInsights = new ApplicationInsights({
     config: {
         connectionString: 'InstrumentationKey=YOUR_INSTRUMENTATION_KEY_GOES_HERE',
         expCfg: {
-            inclScripts: true
+            inclScripts: true,
+            expLog : () => {
+                return {message: ["log info 1", "log info 2"], maxLength: 100};
```
MSNev, line 19: The maxLength should be it's own config at the same level as the inclScripts and expLog, and I was thinking that it would just return a string []. ( `expLog: () => string[]` ), but you pose an interesting option....

If we make this return an "object" instead, then it provides for some level of extension without introducing a new function....

Maybe
```ts
interface IExceptionConfig: {
    inclScripts: boolean,
    expLog: () => { logs: string[] }
    maxLogs: number
}
```

Where the expLog function can return the logs wrapped in the object and we can add "new" optional properties based on feedback / requests.
```ts
function getExpLogs() {
    return {
        logs: ["log info 1", "log info 2"]
    };
}

extCfg: {
    inclScripts: true,
    expLog: getExpLogs,
    maxLogs: number
}

```

siyuniu-ms:
NOTICE
The expCfg is moved from IConfig to IConfiguration, which may cause a breaking change for types, note this in the next release note!

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
```diff
@@ -52,7 +52,7 @@ function _getReason(error: any) {
 
 const MinMilliSeconds = 60000;
 
-const defaultValues: IConfigDefaults<IConfig> = objDeepFreeze({
+const defaultValues: IConfigDefaults<IConfig&IConfiguration> = objDeepFreeze({
```
MSNev, line 55: Should the default `maxLogs` be 50?

MSNev (approved):

PR #2395, opened 2024-08-13
[main][doc] add document for how to fix type error in extensions

siyuniu-ms:
https://github.com/microsoft/applicationinsights-react-js/issues/95


docs/ExtensionErrorSteps.md
```diff
@@ -0,0 +1,38 @@
...
+
+```plaintext
+Type 'ReactPlugin' is not assignable to type 'ITelemetryPlugin'.
+```
+
+Plugin-type errors often arise due to version mismatches between applicationinsights-web and its extensions.
+
+## Reason 
+
+When a new version of applicationinsights-web (which includes applicationinsights-core as a dependency) is released, package management tools may automatically update applicationinsights-core to the new version. However, if the extensions (e.g., React, Angular) have not been updated, their dependencies on applicationinsights-core may not match, leading to type errors.
```
MSNev, line 17: It would be good to call out here that this is a TypeScript issue only as we work really hard to ensure that while the "types" (might) appear different to different versions of TypeScript at runtime the code should still interoperate as long as the "core" version is later than any sub-component as we support backward compatibility but not forward compatibility.

Which means we support older components working with "newer" dependencies (ie the newer dependency is backward compatible for older components) but not newer components working with "older" dependencies.

A good example of this is that we may from time to time "add" new public helper functions to core components that some "newer" extensions might use / need. As such that "newer" function is only available in the "newer" version of the core component and not in the older previous published components. (ie. we don't know now (or back then) what functions we "might" need in a future release -- why I call it forward compatibility)

docs/ExtensionErrorSteps.md
```diff
@@ -0,0 +1,38 @@
...
+
+### 2. Perform a Complete Update
+The easiest way to ensure all dependencies are updated is to delete the node_modules folder and reinstall the packages:
+```
+rm -rf node_modules
+npm install
+```
+### 3. Examples to follow
+Here is a discussion that provide deeper insights into resolving these issues:
+https://github.com/microsoft/applicationinsights-react-js/issues/95
```
MSNev, line 34: I think there is a new TypeScript flag (which most people wont use) -
- skipLibCheck
- noCheck
I thought I read recently (somewhere) that there was a new config but I can't find it currently

MSNev (approved):

docs/ExtensionErrorSteps.md
siyuniu-ms: Thanks. Will check that potential walk around before merge this pr.

PR #2400, opened 2024-08-30
[Main] [Release] Increase version to 3.3.2

Karlie-777:


tools/applicationinsights-web-snippet/package.json
```diff
@@ -32,8 +32,8 @@
     "devDependencies": {
         "@types/qunit": "^2.19.3",
         "@microsoft/ai-test-framework": "0.0.1",
-        "@microsoft/applicationinsights-web": "3.3.1",
-        "@microsoft/applicationinsights-common": "3.3.1",
+        "@microsoft/applicationinsights-web": "3.3.2",
+        "@microsoft/applicationinsights-common": "3.3.2",
```
Karlie-777, line 36: Only updated dependencies but not package version and it is still 3.3.1

RELEASES.md
```diff
@@ -24,6 +24,13 @@ Users can also enable Integrity Check and minimize snippet loading time by setti
 - #2360 [main] set script attribut to avoid race condition when multiple sdks are inited #2355 
 - #2339 [main] create snippet mini loader
 
+## 3.3.2 (Sep 3rd, 2024)
+
+### Changelog
+
+- #2396 [Main] Update generated typedoc documentation
+- #2397[Main] Change tslib peerDependency from "*" to open range
+
```
MSNev, line 33: I didn't notice this last time, but this release and the previous one should be moved to the top.

ie. this file is in date order not grouped by type, so this should be above the web snippet releases.

Also this is missing 
- https://github.com/microsoft/ApplicationInsights-JS/pull/2391, which should have a special note because we have moved some config from IConfig to IConfiguration -- so it IS going to cause the TypeScript type error.

MSNev (approved):

PR #2403, opened 2024-09-06
[Main][Task]28736784: Update the Config Used for Type Docs with Test Web Config Endpoint

Karlie-777:


siyuniu-ms (approved):

tools/github-page-script-injection/injectScript.js
```diff
@@ -31,7 +31,9 @@ let scriptContent = fs.readFileSync(scriptFilePath, 'utf8');
 // Replace the placeholder string with the actual connection string
 const connectionString = 'InstrumentationKey=1ae9e7ce-18f1-4e14-8fc0-acbf0ed28895;IngestionEndpoint=https://eastus-8.in.applicationinsights.azure.com/;LiveEndpoint=https://eastus.livediagnostics.monitor.azure.com/';
 let iKeyForCfgSync = "1ae9e7ce-18f1-4e14-8fc0-acbf0ed28895";
-scriptContent = scriptContent.replace(`connectionString: "YOUR_CONNECTION_STRING"`, `instrumentationKey: "${iKeyForCfgSync}",\n \t\textensionConfig: {\n\t\t\t"AppInsightsCfgSyncPlugin": {\n\t\t\t\tcfgUrl:"https://js.monitor.azure.com/beta/ai.config.1.cfg.json"\n\t\t\t}\n\t\t}`);
+//let cfgUrl = "https://js.monitor.azure.com/beta/ai.config.1.cfg.json";
+let cfgUrl = "https://tst2cdnstore.blob.core.windows.net/cdn/nevtest/ai.cfgSync.json";
```
MSNev, line 35: Lets not use this URL as it's pointing directly to the test storage account which won't do everything needed for every browser and won't really handle the scale correctly.

Let use the "test" folder of the standard `js.monitor.azure.com` (we will need to update the deployment pipeline for this -- or perhaps use the "nightly" folder and we just put in a "test" file (not called ai.cfgSync.json) perhaps ai.typedoc.cfgSunc.json) or our test CDN

MSNev: So current nightly is going to `https://[js.monitor.azure.com/nightly/ai.3-nightly3.min.js](https://js.monitor.azure.com/nightly/ai.3-nightly3.min.js)`

But I noticed that we are not currently publishing the cfgSync at all nightly, so we will need to add that to the pipeline and then also "add" the ability for it to publish a "test" version (so we don't accidently publish over the real one in the standard pipeline... (Which I was I suggested another file) so maybe either
- `ai_typedoc.cfgSunc.json` or
- `ai_test.cfgSync.json` (as the dot after the `ai` will cause issues parsing the name for promoting the version

>

gruntfile.js
```diff
@@ -803,6 +803,13 @@ module.exports = function (grunt) {
                         { src: "./tools/config/config.json", dest: `./tools/config/browser/es5/ai.config${configVer}.cfg.json` },
                         { src: "./tools/config/config.json", dest: `./tools/config/browser/es5/ai.config${configMajorVer}.cfg.json` }
                     ]
+                },
+
+                testConfig: {
+                    files: [
+                        { src: "./tools/config/config.json", dest: `./tools/config/browser/es5/ai_test.config${configVer}.cfg.json` },
```
MSNev, line 810: Do we "really" want to copy the single "default" one to the test, I was suggesting that we actually have 2 checked in, so this would copy the test one as the test

Karlie-777: Thanks Nev for pointing it out! I just found it as well and it should be "test-config", I already updated the PR!

MSNev (approved):

tools/github-page-script-injection/injectScript.js
```diff
@@ -32,7 +32,7 @@ let scriptContent = fs.readFileSync(scriptFilePath, 'utf8');
 const connectionString = 'InstrumentationKey=1ae9e7ce-18f1-4e14-8fc0-acbf0ed28895;IngestionEndpoint=https://eastus-8.in.applicationinsights.azure.com/;LiveEndpoint=https://eastus.livediagnostics.monitor.azure.com/';
 let iKeyForCfgSync = "1ae9e7ce-18f1-4e14-8fc0-acbf0ed28895";
 //let cfgUrl = "https://js.monitor.azure.com/beta/ai.config.1.cfg.json";
-let cfgUrl = "https://js.monitor.azure.com/nightly/ai_test.config.1.cfg.json";
+let cfgUrl = "https://js.monitor.azure.com/nightly/ai_test.config.1.0-nightly3.cfg.json";
```
MSNev, line 35: Right now it's probably `1.0-nightly3` (because the version gets flattened from 1.0.0 to 1.0 and 1) but it probably should be listed as `1-nightly3`

Karlie-777: just changed it! Thanks Nev!

PR #2404, opened 2024-09-06
[main] fix expCfg to be optional

siyuniu-ms:
related to pr: https://github.com/microsoft/ApplicationInsights-JS/pull/2391

### DIFF
--- a/shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IExceptionConfig.ts
+++ b/shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IExceptionConfig.ts
@@ -1,14 +1,37 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+/**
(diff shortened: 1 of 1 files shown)


shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IExceptionConfig.ts
```diff
@@ -20,6 +43,14 @@ export interface IExceptionConfig{
...
      */
-    expLog: () => { logs: string[] },
-    maxLogs: number
+    expLog?: () => { logs: string[] },
+
+    /**
+     * The maximum number of logs to include in the telemetry data.
+     * If not explicitly set, it defaults to 50.
+     * This is used in conjunction with the `expLog` callback.
+     * @type {number}
```
MSNev, line 52: We don't need either of these 2, typedoc takes care of identifying this as a member of this interface and it's type

MSNev (approved):

PR #2410, opened 2024-09-11
[main] enable auto doc generation

siyuniu-ms:


.github/workflows/jekyll-gh-pages.yml
```diff
@@ -41,6 +41,8 @@ jobs:
       - run: node common/scripts/install-run-rush.js install
       - run: npm run build --verbose
         timeout-minutes: 20
+      - name: doc generate
```
MSNev, line 44: Oh, I see...
So at the point that this action "runs" which should be triggered by the "checkin" (please correct me if I'm wrong) we not only "decorate" the pages here but actually "generate them -- nice!

siyuniu-ms: this seems to be the branches that could auto trigger the run, but didn't give the access of publish the doc

Issue #1851, opened 2022-06-08
Cross-Origin-Embedder-Policy support


Sora2455:
**Is your feature request related to a problem? Please describe.**
When trying to set the [Cross-Origin-Embedder-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy) header to "require-corp" on my site, requests to application insights (in particular https://australiaeast-0.in.applicationinsights.azure.com//v2/track) are blocked, as they do not have a [Cross-Origin-Resource-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Resource-Policy) header.

**Describe the solution you'd like**
The application insights endpoints to serve a Cross-Origin-Resource-Policy header of the value "cross-origin".

**Describe alternatives you've considered**
If this was an image or script request, I could add the crossOrigin attribute to the <img> or <script> tag to get around this problem. However, this is a fetch/XHR request, and one that requires cookies if I understand correctly.

**Additional context**
Cross-Origin-Embedder-Policy is needed to enable the security feature cross-site isolation.

Karlie-777:
Related: https://github.com/microsoft/ApplicationInsights-JS/issues/1399

Sora2455:
Any chance this could be looked at anytime soon? I appreciate the team behind the JS library may not have any involvement with the team controlling the report servers, but in that case is there a better place I should be making this request?

Karlie-777:
Hi @Sora2455,
If your are using a customized endpoint, config [customerHeaders](https://github.com/microsoft/ApplicationInsights-JS#:~:text=value%20to%205.-,customHeaders,-%5B%7Bheader%3A%20string%2C%20value) should be able to add extra headers.
 And another way of doing it is to provide a xhroverride in [sender config](https://github.com/microsoft/ApplicationInsights-JS/blob/caf4dd92c4104f948ef922203f6e9ac690f931c3/channels/applicationinsights-channel-js/src/Interfaces.ts#L108) where you can control sending methods including headers as well.

Sora2455:
Hi @Karlie-777 

If I've understood correctly, those config options set headers on the _request_ - but I need them set on the _response_.

Karlie-777:
@Sora2455 Correct, they are for request headers. For response headers, we've created a task internally, let me get back to you as soon as we have updates!

siyuniu-ms:
[...]

PR #2423, opened 2024-09-25
[main] support adding cross origin resource policy for #1851

siyuniu-ms:
related to #1851


channels/applicationinsights-channel-js/src/Interfaces.ts
```diff
@@ -166,6 +166,20 @@ export interface ISenderConfig {
...
+     * which helps control how resources can be shared across different origins.
+     *
+     * Possible values:
+     * - `same-site`: Allows access only from the same site.
+     * - `same-origin`: Allows access only from the same origin (protocol, host, and port).
+     * - `cross-origin`: Allows access from any origin.
+     *
+     * @since 3.3.3
+     */
+    crossOriginResourcePolicy?: string;
```
MSNev, line 182: Way too long, perhaps `corsPolicy`

shared/AppInsightsCommon/src/Interfaces/IConfig.ts
```diff
@@ -390,6 +390,19 @@ export interface IConfig {
...
+     * which helps control how resources can be shared across different origins.
+     *
+     * Possible values:
+     * - `same-site`: Allows access only from the same site.
+     * - `same-origin`: Allows access only from the same origin (protocol, host, and port).
+     * - `cross-origin`: Allows access from any origin.
+     *
+     * @since 3.3.3
+     */
+    crossOriginResourcePolicy?: string;
```
MSNev, line 405: Lets keep this in just the `ISenderConfig` so if you want to set it you set via the Sender Config as this is not a common config, so we should not have it for "everyone" (at least in TypeScript -- technically, the code will also pick it up from the root, but we are going to say -- don't configure that way use the sender config approach and we will only test that way as the "supported" method.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -268,12 +268,13 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
                     }
-                    if (config.crossOriginResourcePolicy){
-                        this.addHeader(CrossOriginResourcePolicyHeader, config.crossOriginResourcePolicy);
-                    }
+                   
                     let ctx = createProcessTelemetryContext(null, config, core);
                     // getExtCfg only finds undefined values from core
                     let senderConfig = ctx.getExtCfg(identifier, defaultAppInsightsChannelConfig);
+                    if (senderConfig.corsPolicy){
+                        this.addHeader(CrossOriginResourcePolicyHeader, senderConfig.corsPolicy);
```
MSNev, line 276: As this is inside of an onConfigChange(), this will get called (potentially) more than once so we will need to
- protect or handling setting the same header more than once (which should already be handled by the way addHeader works)
- Removing this header if the corsPolicy is removed afterwards (unlikely, but not impossible)

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -78,9 +78,12 @@ const defaultAppInsightsChannelConfig: IConfigDefaults<ISenderConfig> = objDeepF
     alwaysUseXhrOverride: cfgDfBoolean(),
     transports: UNDEFINED_VALUE,
     retryCodes: UNDEFINED_VALUE,
+    crossOriginResourcePolicy: UNDEFINED_VALUE,
     maxRetryCnt: {isVal: isNumber, v:10}
 });
 
+const CrossOriginResourcePolicyHeader: string = "AI-Cross-Origin-Resource-Policy";
```
MSNev, line 85: Per Osvaldo's comment, lets prefix this with `X-`

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -268,6 +271,13 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                     let ctx = createProcessTelemetryContext(null, config, core);
                     // getExtCfg only finds undefined values from core
                     let senderConfig = ctx.getExtCfg(identifier, defaultAppInsightsChannelConfig);
+                    if (senderConfig.corsPolicy){
+                        if (senderConfig.corsPolicy === "same-origin" || senderConfig.corsPolicy === "same-site" || senderConfig.corsPolicy === "cross-origin") {
```
MSNev, line 275: nice check.

Lets use a local variable for all of the `senderConfig.corsPolicy` usages as while the `senderConfig` will get minified all of the `.corsPolicy` references won't which has 2 side effects.
1) The minified size 
2) Performance, it's not obvious but every time you reference a property of an object the runtime needs to "lookup" the value so you end up spending a few cycles running code unnecessarily, it's not a "big" problem here as this code only runs when there is a config change, but every little bit adds up.
- Also as this is a dynamic config reference, each lookup will actually run all of the dynamic config "tracking" code (a couple of function calls) for each reference.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -78,9 +78,12 @@ const defaultAppInsightsChannelConfig: IConfigDefaults<ISenderConfig> = objDeepF
     alwaysUseXhrOverride: cfgDfBoolean(),
     transports: UNDEFINED_VALUE,
     retryCodes: UNDEFINED_VALUE,
+    crossOriginResourcePolicy: UNDEFINED_VALUE,
```
MSNev, line 81: needs to use the short name.

MSNev:
General comment on the timing of this change. Can you ask the backend people on the timeframe that they believe they will be able to implement this change as we should not commit this change until the backend supports it (and we validate that it's working), to avoid someone accidently enabling this and causing all requests to stop. As this will most likely cause an OPTIONS call  (CORS validation) requesting whether the required header can be sent and if the server doesn't know anything about it the browser may block the call completely.

MSNev (approved):

MSNev (approved):

PR #2428, opened 2024-10-07
[main] [snippet] Integrate 1DS with AI Snippet Generation

siyuniu-ms:


tools/applicationinsights-web-snippet/createAiSnippet.js
```diff
@@ -0,0 +1,39 @@
...
+const aiFilePath = path.join(__dirname, 'src/aiSnippet.ts');
+const oneDSFilePath = path.join(__dirname, 'src/oneDSSnippet.ts');
+// Read the snippet.js file
+fs.readFile(filePath, 'utf8', (err, data) => {
+    if (err) {
+        console.error('Error reading file:', err);
+        process.exit(1);
+    }
+
+    let remove = data.replace(/let isOneDS = false;/g, '');
```
MSNev, line 16: Lets add an additional check after this to make sure that remove != data -- so that if for some reason the regex misses (because of content changes) the build breaks.

it might also be useful to make the regex a little more permissive
`/let\s+isOneDS\s+=\s+\w+;/i

tools/applicationinsights-web-snippet/createAiSnippet.js
```diff
@@ -0,0 +1,39 @@
+// createFile.js
+const fs = require('fs');
+const path = require('path');
+
+// Define the path to the snippet.js file
+const filePath = path.join(__dirname, 'src/snippet.ts'); // Adjust the path if necessary
+const aiFilePath = path.join(__dirname, 'src/aiSnippet.ts');
```
MSNev, line 7: It might also be worth either
- Placing these into a different folder (so we know they are generated)
- Explicitly adding these to .gitignore (or a generated folder from above) so that we don't check in these files

Lets also "Add" a comment to the top of the file 
```ts
// DO NOT EDIT - Automatically Generated
```

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -138,35 +181,114 @@ declare var cfg:ISnippetConfig;
...
+                    data: {
+                        baseData: {
+                            ver: 2
+                        }
+                    },
+                    ext: {
+                        app: { sesId: "0000" },
+                        intweb: {},
+                        sdk: {
+                            ver: "javascript:" + OneDSstrSnippetVersion,
```
MSNev, line 241: As we are creating a new "unified" (ish) snippet lets also make the reported version the same for both AI and 1DS

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -138,35 +181,114 @@ declare var cfg:ISnippetConfig;
...
+            if (typeof navigator !== strUndefined) {
+                const nav: Navigator & { userLanguage?: string } = navigator;
+                envelope.ext = envelope.ext || {};
+                envelope.ext.user = {
+                    locale: nav.userLanguage || nav.language
+                };
+            }
+        }
+
+        function _createEnvelope(iKey:string, theType:string) {
```
MSNev, line 226: I know this suggestion is probably going to make the resulting file larger, but I'm thinking maybe we should extract the creation of the reporting "error" (envelope) to a new file
So
`_createAiEnvelope()` and `_createOneDsEnvelope()`

So if we keep the code on lines 306..310 with the generation stuff you have added this "should" still cause the minification to work nicely and should make this core file a little easier to read...

We "may" even just have the unique portions of AI / 1DS in each of those files (now that we know we have these differences we need to do ☹️ )

MSNev: A bit like the way you have the 1dsType.ts, maybe we have a 1dsSupport.ts type thing

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -297,9 +423,18 @@ declare var cfg:ISnippetConfig;
                     // IE10, Opera calls loaded before the script is processed.
                     // so delaying to give the script a chance to be processed
                     setTimeout(function() {
-                        if (isAbort || !appInsights.core) {
-                            _handleError();
+                        if (isAbort){
+                            if (isOneDS){
```
MSNev, line 427: extension of above comment
- This stuff probably needs to remain here though as its only a few lines and uses the current functions
- So we only need to "move" the large (mostly) stand alone support functions (if we can)

tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -412,22 +547,42 @@ declare var cfg:ISnippetConfig;
...
-            "clearAuthenticatedUserContext",
-            "flush"]);
-    
+            trackPage + "View",
+            trackPage + "ViewPerformance",
+            "addTelemetryInitializer"
+        ]);
+
+        if (isOneDS){
+            _createMethods([
```
MSNev, line 561: These arrays (might) be able to move to a support file as well.

MSNev (approved):

siyuniu-ms:
aisnippet size 5722->5755

PR #2434, opened 2024-10-15
[Main][Task]29884493: Add a Function to Export Offline Listener From Sender

Karlie-777:


MSNev (approved):
Note: This wasn't actually a "bug" as previously this was just all internal as it only listened to the browser "online" state.

Now that we have extended the IOfflineListener to have the "user" state this makes sense.

PR #2448, opened 2024-11-20
[main] fix tsdoc error for param and type tag

siyuniu-ms:


AISKU/src/AISku.ts
```diff
@@ -626,33 +626,29 @@ export class AppInsightsSku implements IApplicationInsights {
 
     /**
      * Log a user action or other occurrence.
-     * @param event
-     * @param [customProperties]
+     * @param customProperties - properties that would be included as part of the event
```
Karlie-777, line 629: maybe we should add back param event

shared/AppInsightsCore/consolelog.md
```diff
@@ -0,0 +1,187 @@
+
+> @microsoft/applicationinsights-core-js@3.3.4 dtsgen
```
Karlie-777, line 2: Is it generated by rush?

shared/AppInsightsCommon/src/Interfaces/ITraceTelemetry.ts
```diff
@@ -7,19 +7,16 @@ import { IPartC } from "./IPartC";
 export interface ITraceTelemetry extends IPartC {
     /**
      * @description A message string
-     * @type {string}
```
Karlie-777, line 10: maybe we should keep type as well?

shared/AppInsightsCore/consolelog.md
siyuniu-ms: deleted

Karlie-777 (approved):

Issue #2442, opened 2024-10-31
[BUG] [Snippet] Unhandled exceptions are reported twice


VitaliyPotapov:
**Description/Screenshot**
When using appInsights as a snippet, unhandled exceptions are reported twice.

Screenshot:
![Image](https://github.com/user-attachments/assets/1b906e44-01c7-4eac-812f-ca042cc10b13)

**Steps to Reproduce**
1. Open reproduction demo: https://7rslpk.csb.app/  ( https://codesandbox.io/p/sandbox/7rslpk )
2. Open devtools network tab
3. Wait for second request to `/track` and check the payload

**Expected behavior**
Only one unhandled exception is reported.

**Additional context**

MSNev:
As part of fixing this the core SDK needs to either
- add an additional check on the `core` config for the `autoExceptionInstrumented` on the following line, it should use both the `core` and `_extConfig` https://github.com/microsoft/ApplicationInsights-JS/blob/d4a4d00283cc5686e360189b1975e26d1bc362d4/extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts#L707
- or as an alternative we could add this key to the defaults so that the underlying config resolver handles picking this up from the root config, it may or may not require an additional `fb` definition (see the existing usage for the Cookie Cfg)
- and/or update the next snippet to also populate the above "workaround" location, so that all SDK's releases (prior to fixing) will automatically resolve the issue.

Or a combination of the above, there may also be other alternative solutions that I have not yet identified.

VitaliyPotapov:
Thanks for the reply and the workaround.
I confirm it works!

PR #2449, opened 2024-11-25
[main] [BUG] [Snippet] Fix Unhandled exceptions are reported twice #2442

siyuniu-ms:


siyuniu-ms:
for #2442

README.md
```diff
@@ -121,7 +121,13 @@ The current version of the snippet is version 8, the version is identified by th
     // onInit: null, // Once the application insights instance has loaded and initialized this callback function will be called with 1 argument -- the sdk instance (DO NOT ADD anything to the sdk.queue -- As they won't get called)
     // sri: false, // Custom optional value to specify whether fetching the snippet from integrity file and do integrity check 
     cfg: { // Application Insights Configuration
-        connectionString: "YOUR_CONNECTION_STRING"
+        connectionString: "YOUR_CONNECTION_STRING",
+        extensionConfig: { 
+          ApplicationInsightsAnalytics: {
+            autoExceptionInstrumented: true  // Workaround for older SDK versions to prevent duplicate hooking of window.onerror.
```
Karlie-777, line 127: Is this a new config?

siyuniu-ms: no, it's not. It's a walkaround bug fix.

README.md
```diff
@@ -121,7 +121,13 @@ The current version of the snippet is version 8, the version is identified by th
     // onInit: null, // Once the application insights instance has loaded and initialized this callback function will be called with 1 argument -- the sdk instance (DO NOT ADD anything to the sdk.queue -- As they won't get called)
     // sri: false, // Custom optional value to specify whether fetching the snippet from integrity file and do integrity check 
     cfg: { // Application Insights Configuration
-        connectionString: "YOUR_CONNECTION_STRING"
+        connectionString: "YOUR_CONNECTION_STRING",
+        extensionConfig: { 
+          ApplicationInsightsAnalytics: {
+            autoExceptionInstrumented: true  // Workaround for older SDK versions to prevent duplicate hooking of window.onerror. 
+                                             // Fixed in v3.3.5; recommended for versions prior to this.
+          }
```
Karlie-777, line 129: maybe we can set this to true after the hook completes? and also here is another one https://github.com/microsoft/ApplicationInsights-JS/blob/869990cd1438edcfd4fb08d2da6ce5a046a286d9/shared/AppInsightsCommon/src/Interfaces/IConfig.ts#L333C5-L333C39 might be related as well

siyuniu-ms: we have a local var set to true after the hook completes, but it doesn't help reduce the exception to one. And maybe autoUnhandledPromiseInstrumented is only for _addUnhandledPromiseRejectionTracking function instead of general exceptionTracking (not quite sure about this part)

PR #2451, opened 2024-12-04
[main] enable compress api in 1ds-post-channel and applicationinsights-channel

siyuniu-ms:


channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -969,16 +971,54 @@ export class HttpManager {
...
-                                        // Send the original payload to the listener
-                                        _sendListener(orgPayloadData, payload, isSync, thePayload.isBeacon);
-                                    }
-                                } catch (ex) {
-                                    _warnToConsole(_logger, "Unexpected exception sending payload. Ex:" + dumpObj(ex));
 
-                                    _doOnComplete(onComplete, 0, {});
+                                const CompressionStream = (window as any).CompressionStream;
+                                // If CompressionStream is available, use it
+                                if (!_disableZip && !isSync && CompressionStream && typeof CompressionStream !== "undefined") {
```
MSNev, line 977: Lets move this if test to another function (to avoid duplicating 1000-1009 and 987 - 997

So something like
```
function _preparePayload(cb: (payload: ...) => void, payload, <any other args we need)) {
    if (_disableZip || isSync || has preprocessor || !hasCompression) {
        cb(payload);
    }
... // Your compression code
    cb(payload); // <- gets called from within the Compression handling on success and failure
}
```
And then you just call it like this and it will either be the compressed code (called after it's compressed) or called immediately with the same payload
```
_preparePayload((payload) => {
    try {
         sendInterface.sendPOST(payload, onComplete, isSync);
         if (_sendListener) {
             // Send the original payload to the listener
             _sendListener(orgPayloadData, payload, isSync, thePayload.isBeacon);
        }
    } catch (ex) {
        _warnToConsole(_logger, "Unexpected exception sending payload. Ex:" + dumpObj(ex));
    }
|});

File: channels/1ds-post-js/src/HttpManager.ts [hunk 2]
```diff
@@ -222,6 +224,13 @@ export class HttpManager {
                         }
     
                         _xhrTimeout = channelConfig.xhrTimeout;
+                        const CompressionStream = (window as any).CompressionStream;
+
+                        _disableZip = !!channelConfig.disableZip;
+                        if (typeof CompressionStream !== "function" || _sendHook) {
```
MSNev, line 230: You can use the `isFunction()` helper to help with minification

channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -969,17 +978,17 @@ export class HttpManager {
...
+                                            // Send the original payload to the listener
+                                            _sendListener(orgPayloadData, processedPayload, isSync, thePayload.isBeacon);
+                                        }
+                                    } catch (ex) {
+                                        _warnToConsole(_logger, "Unexpected exception sending payload. Ex:" + dumpObj(ex));
                                     }
-                                } catch (ex) {
-                                    _warnToConsole(_logger, "Unexpected exception sending payload. Ex:" + dumpObj(ex));
-
-                                    _doOnComplete(onComplete, 0, {});
```
MSNev, line 981: The error case is not calling _doOnComplete(), for 1DS this is REALLY important as it will block any further requests because it tracks the number of "active" connections.

channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -1049,6 +1058,59 @@ export class HttpManager {
...
+                if (_disableZip || isSync || !payload.data) {
+                    // If body is null or undefined, we don't need to compress it
+                    callback(payload);
+                    return;
+                }
+
+                const CompressionStream = (window as any).CompressionStream;
+
+                // Compression logic
+                const uint8Data = typeof payload.data === "string"
```
MSNev, line 1071: You can use the `isString()` helper to aid in minification.

channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -1049,6 +1058,59 @@ export class HttpManager {
...
+                        controller.close();
+                    }
+                });
+    
+                const compressedStream = body.pipeThrough(new CompressionStream("gzip"));
+                const reader = (compressedStream.getReader() as ReadableStreamDefaultReader<Uint8Array>);
+                const chunks: Uint8Array[] = [];
+                let totalLength = 0;
+        
+                return reader.read().then(function processChunk({ done, value }: { done: boolean, value?: Uint8Array }): IPromise<void> {
```
MSNev, line 1087: For minification you can use [`doAwaitResponse(...)`](https://nevware21.github.io/ts-async/typedoc/functions/doAwaitResponse.html) as this only then needs 1 function to be defined.

Also as the response is never used, you can drop the `return` as that won't get minified either.

channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -1049,6 +1058,59 @@ export class HttpManager {
...
+                        const combined = new Uint8Array(totalLength);
+                        let offset = 0;
+                        for (const chunk of chunks) {
+                            combined.set(chunk, offset);
+                            offset += chunk.length;
+                        }
+                        payload.data = combined;
+                        payload.headers["Content-Encoding"] = "gzip";
+                        callback(payload);
+                        const resolvedPromise: IPromise<void> = createPromise((resolve, reject) => {
```
MSNev, line 1099: this doesn't look like it's needed either.

channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -1049,6 +1058,59 @@ export class HttpManager {
...
+                            resolve();
+                        });
+                        return resolvedPromise;
+                    } else {
+                        chunks.push(value);
+                        totalLength += value.length;
+                        return reader.read().then(processChunk);
```
MSNev, line 1106: You should be able to change this to `return doAwait(reader.read(), processChunk);`

channels/1ds-post-js/src/HttpManager.ts
MSNev, line 1099: This can be changed to `return createResolvedPromise<void>();`

channels/1ds-post-js/src/HttpManager.ts
MSNev: Might be easier to break this out into a local (not in-line) function or a helper function that is passed the reader that returns the function so it has access to the reader internally.

channels/1ds-post-js/src/HttpManager.ts
MSNev: Note: is "may" not be the `processChunk` depending on how it's constructed, I'll try and create an example (it won't be compiled) on reorgainzing this function to be smaller

MSNev: Here is a possible solution, it could potentially be generalized a little more and extracted out (so it could potentially be used by both 1DS and AI), also the first check for the presence of the CompressionStream could be extracted out to the constructor as it is (or should be) an invariant

You will need to add the additional imports for the ts-async functions.

```ts
function _preparePayload(callback: (processedPayload: IPayloadData) => void, payload: IPayloadData, isSync: boolean): ReadableStreamDefaultReader<Uint8Array> {
    const CompressionStream = (window as any).CompressionStream;
    if (isSync || !payload.data || !CompressionStream) {
        // If the request is synchronous, the body is null or undefined or Compression is not supported, we don't need to compress it
        callback(payload);
        return;
    }

    // Create a readable stream from the uint8 data
    let body = new ReadableStream<Uint8Array>({
        start(controller) {
            controller.enqueue(isString(payload.data) ? new TextEncoder().encode(payload.data) : payload.data);
            controller.close();
        }
    });

    // Compress the data using gzip
    const compressedStream = body.pipeThrough(new CompressionStream("gzip"));
    const reader = (compressedStream.getReader() as ReadableStreamDefaultReader<Uint8Array>);
    const chunks: Uint8Array[] = [];
    let totalLength = 0;
    let callbackCalled = false;

    // Process each chunk from the compressed stream reader
    doAwaitResponse(reader.read(), function processChunk(response: AwaitResponse<ReadableStreamReadResult<Uint8Array>>) {
        if (!callbackCalled && !response.rejected) {
            // Process the chunk and continue reading
            const result = response.value;
            if (!result.done) {
                // Add current chunk and continue reading
                chunks.push(result.value);
                totalLength += result.value.length;

                return doAwaitResponse(reader.read(), processChunk);
            }

            // We are complete so combine all chunks
            const combined = new Uint8Array(totalLength);
            let offset = 0;
            for (const chunk of chunks) {
                combined.set(chunk, offset);
                offset += chunk.length;
            }
            
            // Update payload with compressed data
            payload.data = combined;
            payload.headers["Content-Encoding"] = "gzip";
        }

        if (!callbackCalled) {
            // Send the processed payload to the callback, if not already called
            // If the response was rejected, we will call the callback with the original payload
            // As it only gets "replaced" if the compression was successful
            callbackCalled = true;
            callback(payload);
        }

        // We don't need to return anything as this will cause the calling chain to be resolved and closed
    });

    // returning the reader to allow the caller to cancel the stream if needed
    // This is not a requirement but allows for better control over the stream, like if we detect that we are unloading
    // we could use reader.cancel() to stop the stream and avoid sending the request, but this may still be an asynchronous operation
    // and may not be possible to cancel the stream in time
    return reader;
}
```

shared/AppInsightsCore/src/JavaScriptSDK/SenderPostManager.ts
```diff
@@ -135,6 +135,72 @@ export class SenderPostManager {
...
             };
 
+            _self.preparePayload = (callback: (processedPayload: IPayloadData) => void, disableZip: boolean, payload: IPayloadData, isSync: boolean) => {
+                if (disableZip || isSync || !payload.data) {
+                    // If the request is synchronous, the body is null or undefined or Compression is not supported, we don't need to compress it
+                    callback(payload);
+                    return;
+                }
+
+                const CompressionStream = (window as any).CompressionStream;
```
MSNev, line 145: We should also wrap all of this in a try / catch and call the callback(payload) on error to ensure that if something here fails we still send the payload.

channels/applicationinsights-channel-js/src/Interfaces.ts
```diff
@@ -180,6 +180,16 @@ export interface ISenderConfig {
...
+
+    /*
+     * [Optional] If set to false and the CompressionStream API is available, the payload will be compressed using the CompressionStream API.
+     * Compression will only occur if the event is asynchronous. For events like unloads, compression will not be applied.
+     * Note: if user set payloadPreprocessor, this zip compression will not be applied.
+     * @default true
+     * @since 4.3.7
+     * Starting from minor version 4.4, the default will be changed to false (compression enabled by default).
+    */
+    disableZip?: boolean;
```
MSNev, line 192: Please also create a 2nd follow-up task as we also want to add in the throttle manager task here to "enable" the compression after the fact. But we need to also ensure that we are remembering the previous cfg state.

Lets talk about this a little more during the team meeting (after creating the task)

shared/AppInsightsCore/src/JavaScriptSDK/SenderPostManager.ts
```diff
@@ -135,6 +135,85 @@ export class SenderPostManager {
...
+            _self.preparePayload = (callback: (processedPayload: IPayloadData) => void, disableZip: boolean, payload: IPayloadData, isSync: boolean) => {
+                if (disableZip || isSync || !payload.data) {
+                    // If the request is synchronous, the body is null or undefined or Compression is not supported, we don't need to compress it
+                    callback(payload);
+                    return;
+                }
+
+                let CompressionStream: any;
+                try {
+                    CompressionStream = (window as any).CompressionStream;
```
MSNev, line 147: getinst("");

channels/1ds-post-js/src/DataModels.ts
```diff
@@ -262,6 +262,16 @@ export interface IChannelConfiguration {
      * @since 3.3.1
      */
     fetchCredentials?: RequestCredentials;
+
+    /*
+     * [Optional] If set to false and the CompressionStream API is available, the payload will be compressed using the CompressionStream API.
```
MSNev, line 267: Lets change the wording a little.

[Optional] If not set to `true` and the CompressionStream API...

channels/1ds-post-js/src/DataModels.ts
```diff
@@ -262,6 +262,16 @@ export interface IChannelConfiguration {
...
+     * Compression will only occur if the event is asynchronous. For events like unloads, compression will not be applied.
+     * Note: if user set payloadPreprocessor, this zip compression will not be applied.
+     * @default true
```
MSNev, line 270: Lets discuss this in the team meeting

AISKULight/src/index.ts
```diff
@@ -5,19 +5,29 @@ import dynamicProto from "@microsoft/dynamicproto-js";
...
 const defaultConfigValues: IConfigDefaults<IConfiguration> = {
     diagnosticLogInterval: cfgDfValidate(_chkDiagLevel, 10000),
     connectionString: UNDEFINED_VALUE,
     endpointUrl: UNDEFINED_VALUE,
     instrumentationKey: UNDEFINED_VALUE,
+    featureOptIn:{
+        [IKEY_USAGE]: {mode: FeatureOptInMode.enable}, //for versions after 3.1.2 (>= 3.2.0)
+        [CDN_USAGE]: {mode: FeatureOptInMode.disable},
+        [SDK_LOADER_VER]: {mode: FeatureOptInMode.disable},
+        [ZIP_PAYLOAD]: {mode: FeatureOptInMode.none}
```
siyuniu-ms, line 29: cfySync plugin is not included in aiskulight, this will let isFeatureEnable pass, but user have to set config.zipPayload to make the compress work.

README.md
```diff
@@ -447,11 +447,41 @@ Most configuration fields are named such that they can be defaulted to falsey. A
...
 
+### Feature
+
+You can use the `featureOptIn` configuration to enable or customize specific SDK features.
+
+#### Available Feature Flags
+
+| Name        | Default | Description                                  | Note |
+|-------------|---------|----------------------------------------------|------------|
+| `zipPayload` | `none`  | Enables compression using the Compression API to zip telemetry payloads. |If this feature is turned on and the CompressionStream API is available, the payload will be compressed using the CompressionStream API. Compression will only occur if the event is asynchronous. For events like unloads, compression will not be applied. Note: if user set payloadPreprocessor, this zip compression will not be applied.|
```
MSNev, line 463: As we intend to enable via the config sync we should reference that here somehow.

As having the default as `none` is not strictly correct...

Maybe something including words about this version will have this feature enabled automatically by the SDK, and if you wish to disable this capability you should set this to `disabled`

channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -222,6 +223,15 @@ export class HttpManager {
                         }
     
                         _xhrTimeout = channelConfig.xhrTimeout;
+                        
+                        const csStream = getInst("CompressionStream");
+
+                        // Controls whether payload compression (gzip) is enabled.
+                        _zipPayload = isFeatureEnabled("zipPayload", coreConfig);
+                        if (!isFunction(csStream) || _sendHook) {
```
MSNev, line 231: Lets add an additional comment for the team to note what the `_sendHook` is and why we are disabling if there is a payload processor. 
i.e. as the payload processor may have been added to provide application level compression, we don't want to cause the payload to get corrupted.

channels/1ds-post-js/src/HttpManager.ts
```diff
@@ -1049,6 +1060,7 @@ export class HttpManager {
                 }
             }
 
+
```
JacksonWeber, line 1063: nit: accidental double whitespace

JacksonWeber (approved):
LGTM

Karlie-777 (approved):

MSNev (approved):

PR #2455, opened 2024-12-09
[main] fix rush warning message

siyuniu-ms:


.gitignore
```diff
@@ -270,4 +270,7 @@ ai.tests.d.ts
 /index.html
 
 # Generated Constant File
-**/__DynamicConstants.ts
\ No newline at end of file
+**/__DynamicConstants.ts
+
+# Generated Docs
+docs/
```
MSNev, line 276: This should be docs/webSdk as the `docs` folder has files that are and should be checked in.

MSNev (approved):

PR #2465, opened 2025-01-14
[main][debug tool] enable check compressed data

siyuniu-ms:


tools/chrome-debug-extension/src/dataSources/defaultDataSource.ts
```diff
@@ -51,27 +52,90 @@ export class DefaultDataSource implements IDataSource {
         _self.removeListener = (id: number): boolean => {
             return listeners.delete(id);
         };
+
+        function isGzip(data: ArrayBuffer): boolean {
+            const checkGzip = new Uint8Array(data);
+            if (checkGzip[0] === 0x1F && checkGzip[1] === 0x8B) {
+                console.log("This is gzipped.");
```
MSNev, line 59: We should not have any console.log

tools/chrome-debug-extension/src/dataSources/defaultDataSource.ts
```diff
@@ -51,27 +52,90 @@ export class DefaultDataSource implements IDataSource {
...
-                            if (event !== undefined) {
-                                if (Array.isArray(event)) {
-                                    for (const subEvent of event) {
-                                        _handleMessage(subEvent, details);
-                                    }
-                                } else {
-                                    _handleMessage(event, details);
-                                }
+                if (details.requestBody && details.requestBody.raw) {
+                    let gzipped = isGzip(details.requestBody.raw[0].bytes as ArrayBuffer);
```
MSNev, line 120: While its good to "validate" that it's really gzipped, we should be checking the Encoding header and only attempt this check if it was reported as encoded as gzipped.

siyuniu-ms: requestHeaders is only available under onBeforeSendHeaders func

MSNev (approved):

PR #2474, opened 2025-01-31
[Main][Task]31194802: Fix Duplicated Exception Telemetry Issue When Both Snippet and Analytics Plugin hook onerror

Karlie-777:


Karlie-777:
https://github.com/microsoft/ApplicationInsights-JS/issues/2442

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
```diff
@@ -630,6 +638,7 @@ export class AnalyticsPlugin extends BaseTelemetryPlugin implements IAppInsights
 
                     _expCfg = _extConfig.expCfg;
                     _autoTrackPageVisitTime = _extConfig.autoTrackPageVisitTime;
+                    _autoExceptionInstrumented = (config as any).autoExceptionInstrumented;
```
MSNev, line 641: Actually, because this is inside of a cfg change this might actually get reset to false on subsequent passes.
So there is the potential to flip flop :-(
We should either 
- have this as a default outside of this cfg change (in the pop defaults) setting to false and have this check that value (so once it's true it stays true)
- When we set this local var to true we should set this root one (and maybe the extcfg one) to true, so that we don't try to reset it.

MSNev (approved):

Issue #2480, opened 2025-02-11
[BUG] Unable to use telemetry initializer to filter calls to retrieve config from js.monitor.azure.com


timtucker-dte:
Trying to cut down on the # of 3rd party dependency calls that get logged from our app to keep log volumes down, but ran into difficulties suppressing App Insights' calls to load its own config.

We wind up with at least one dependency call logged per page load for the following URL:
GET https://js.monitor.azure.com/scripts/b/ai.config.1.cfg.json

I tried adding a telemetry initializer to filter dependencies based on hostname, but it looks like the call to fetch the config is made & logged as soon as the ApplicationInsights object is created.

Since addTelemetryInitializer isn't called until after that, there's nothing to prevent the SDK from logging that initial request.

**Steps to Reproduce**

 - OS/Browser: Windows / Edge
 - SDK Version: 3.3.4
 - How you initialized the SDK:

```
const config: IConfiguration & IConfig = {
	connectionString: connectionString,

	enableAutoRouteTracking: true,
	enableAjaxPerfTracking: true,
	enableAjaxErrorStatusText: true,
	enableCorsCorrelation: true,

	// Monitor all AJAX calls
	maxAjaxCallsPerView: -1,

	// Whether or not to disable telemetry
	disableTelemetry: disableTelemetry,
};

// This triggers a fetch that gets logged as a dependency
const appInsights = new ApplicationInsights({
	config: config,
});
// Filter to exclude traffic is here
appInsights.addTelemetryInitializer(filterDependencyTracking);

const reactPlugin = new ReactPlugin();
appInsights.addPlugin(reactPlugin);

appInsights.loadAppInsights();
```

**Expected behavior**

All telemetry events get processed via the telemetry initializer as long as it's registered before calling loadAppInsights.

MSNev:
Agree, the telemetry initializers added before initialization should be valid and therefore process every event.

MSNev:
@siyuniu-ms Can you please investigate on why the above example is not "adding" and calling the telemetry initializer for all events.

PR #2481, opened 2025-02-11
[Main][Task]31338239: Add Config to allow excluding the configuration endpoints from being reported

Karlie-777:
Added an internal config to allow excluding the configuration endpoints being logged as a dependency request.

### DIFF
--- a/channels/applicationinsights-channel-js/src/Sender.ts
+++ b/channels/applicationinsights-channel-js/src/Sender.ts
@@ -183,7 +183,6 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
         let _disableBeaconSplit: boolean;
         let _sendPostMgr: SenderPostManager;
         let _retryCodes: number[];
-        let _enableIntTracking: boolean;
 
         dynamicProto(Sender, this, (_self, _base) => {
 
@@ -391,7 +390,6 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                         _customHeaders = null;
                     }
                     _enableSendPromise = senderConfig.enableSendPromise;
-                    _enableIntTracking = !!config.enableIntEndpointsTracking;
                     let sendPostConfig = _getSendPostMgrConfig();
 
                     // only init it once
@@ -696,8 +694,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                         disableXhr: _disableXhr,
                         disableBeacon: !_beaconNormalSupported,
                         disableBeaconSync: !_beaconOnUnloadSupported,
-                        senderOnCompleteCallBack: onCompleteFuncs,
-                        enableIntEndpointsTracking: _enableIntTracking
+                        senderOnCompleteCallBack: onCompleteFuncs
                     } as _ISendPostMgrConfig;
                     return config;
                  
@@ -1322,7 +1319,6 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                 _xhrSend = null;
                 _fallbackSend = null;
                 _sendPostMgr = null;
-                _enableIntTracking = false;
 
                 objDefine(_self, "_senderConfig", {
                     g: function() {
--- a/channels/applicationinsights-channel-js/src/Sender.ts
+++ b/channels/applicationinsights-channel-js/src/Sender.ts
@@ -183,6 +183,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
         let _disableBeaconSplit: boolean;
         let _sendPostMgr: SenderPostManager;
         let _retryCodes: number[];
+        let _enableIntTracking: boolean;
 
         dynamicProto(Sender, this, (_self, _base) => {
 
@@ -390,6 +391,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                         _customHeaders = null;
                     }
                     _enableSendPromise = senderConfig.enableSendPromise;
+                    _enableIntTracking = !!config.enableIntEndpointsTracking;
                     let sendPostConfig = _getSendPostMgrConfig();
 
                     // only init it once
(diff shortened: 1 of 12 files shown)


shared/AppInsightsCore/Tests/Unit/src/HelperFunc.Tests.ts
```diff
@@ -77,6 +78,26 @@ export class HelperFuncTests extends AITestClass {
             }
         });
 
+        this.testCase({
```
Karlie-777, line 81: TODO: add more tests

shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
```diff
@@ -432,6 +437,10 @@ export function openXhr(method: string, urlString: string, withCredentials?: boo
         // cause the request to fail and we no telemetry would be sent
         _wrapSetXhrProp(xhr, strDisabledPropertyName, disabled);
     }
+    
+    if (!enableIntTracking && intExcludeTrackingEndpoints.indexOf(urlString) > -1) {
+        _wrapSetXhrProp(xhr, strDisabledPropertyName, true);
```
MSNev, line 442: This is already happening on line 438 above, so we would just need to pass in "disabled" from the caller rather than a new `enableIntTracking` parameter here.

shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
```diff
@@ -403,6 +403,11 @@ export function prependTransports(theTransports: TransportType[], newTransports:
 const strDisabledPropertyName: string = "Microsoft_ApplicationInsights_BypassAjaxInstrumentation";
 const strWithCredentials: string = "withCredentials";
 const strTimeout: string = "timeout";
+// used full url string here to avoid regex
+// or should we use regex here to include ...ai.config.1.cfg.json?dfkfhfhh as well
+const intExcludeTrackingEndpoints: string[] = [
+    "https://js.monitor.azure.com/scripts/b/ai.config.1.cfg.json"
```
MSNev, line 409: If we can avoid adding this would be better as this is an awful lot of uncompressible bytes and doesn't cover all domains

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -390,6 +391,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                         _customHeaders = null;
                     }
                     _enableSendPromise = senderConfig.enableSendPromise;
+                    _enableIntTracking = !!config.enableIntEndpointsTracking;
```
MSNev, line 394: I think this should actually be in the CfgSync plugin and not in the sender, likewise with the setting of the "disabled" it should be initiated from the cfgSync extension itself when in attempts to create the connection.

Karlie-777: Moved everything to cfgSync. Since cfgSync didn't use the openXhr from the helper function, so add the disabledProperty tag inside the plugin

shared/AppInsightsCore/src/JavaScriptSDK/HelperFuncs.ts
Karlie-777: remain the openxhr and sender untouched, all the changes are moved into cfgsync plugin

extensions/applicationinsights-cfgsync-js/src/Interfaces/ICfgSyncConfig.ts
```diff
@@ -50,6 +50,13 @@ export interface ICfgSyncConfig {
      * @default 30mins (30*60*1000ms)
      */
     scheduleFetchTimeout?: number;
+
+    /**
+     * [Optional] Flag to indicate whether the internal endpoints will be taged as "DisabledPropertyName" to the xhr/fetch requests
+     * @default false
+     * @since 3.3.6
+     */
+    enableIntEndpointsTracking?: boolean;
```
MSNev, line 59: nit: can we make this a little smaller, maybe something that just says track requests (as this will only at least for now) having the 1 request. Some suggestions that might trigger some better names (all default to false), I've also added some suggestions on what the description of the config would be like
- `enableAjax` // Enable Ajax dependency tracking
- `trkRequest` // Track requests as ajax dependency requests to Azure Monitor
- `enableDep` // Enable reporting requests as dependency requests
- `trkAjaxDep` // Track the requests as an ajax dependency request to Azure Monitor

etc

The adding of the "DisabledPropertyName" is not something that most users / consumers of this SDK will know / understand as it's mostly only used as an internal flag to stop us from reporting request with all of the events being sent to Azure Monitor from also being reported as a dependency request.

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -291,6 +299,9 @@ export class CfgSyncPlugin extends BaseTelemetryPlugin implements ICfgSyncPlugin
             function _xhrSender(url: string, oncomplete: OnCompleteCallback, isAutoSync?: boolean) {
                 try {
                     let xhr = new XMLHttpRequest();
+                    if (!_enableIntTracking) {
+                        xhr[DisabledPropertyName] = true;
```
MSNev, line 303: 👍

extensions/applicationinsights-cfgsync-js/src/CfgSyncPlugin.ts
```diff
@@ -265,6 +269,10 @@ export class CfgSyncPlugin extends BaseTelemetryPlugin implements ICfgSyncPlugin
                         const init: RequestInit = {
                             method: STR_GET_METHOD
                         };
+                        if (!_enableIntTracking) {
+                            init[DisabledPropertyName] = true;
```
MSNev, line 273: 👍

MSNev (approved):

Karlie-777:
https://github.com/microsoft/ApplicationInsights-JS/issues/2480

PR #2489, opened 2025-03-05
[main][stats beat] implement stats beat in application insights

siyuniu-ms:


siyuniu-ms:
Current StatsBeat Logic
Definition & Creation in Core: StatsBeat is defined and instantiated in the core without a channel, making it compatible with both AI and 1DS.
Initialization in Sender: The sender initializes StatsBeat and assigns the appropriate channel for sending telemetry.
Pros
✔ Prevents infinite loops: Eliminates the risk of recursive creation between StatsBeat and Sender.
✔ Unified implementation: Both AI and 1DS share the same StatsBeat class, reducing duplication.

Cons
❌ Always included in the codebase: Since StatsBeat is created in the core, it remains part of the code even if the user disables it.

StatsBeat Data Handling
  StatsBeat data is stored within the payload.
  When the final result is ready to be sent, the data is extracted before transmission.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,216 @@
...
+            function _trackRequestsCount() {
+                var currentCounter = _networkCounter;
+                if (currentCounter.totalSuccesfulRequestCount > 0) {
+                    _statsbeatMetrics.properties = _statsbeatMetrics.properties || {};
+                    _statsbeatMetrics.properties[StatsbeatCounter.REQUEST_SUCCESS] = currentCounter.totalSuccesfulRequestCount;
+                    currentCounter.totalSuccesfulRequestCount = 0; //Reset
+                }
+                if (currentCounter.totalFailedRequestCount > 0) {
+                    _statsbeatMetrics.properties = _statsbeatMetrics.properties || {};
+                    _statsbeatMetrics.properties[StatsbeatCounter.REQUEST_FAILURE] = currentCounter.totalFailedRequestCount;
```
JacksonWeber, line 165: It looks like as of this PR all failures are added to a generic "failure" counter. I don't see any accounting for specific status codes as separate counters or including status code as a custom dimension. Are you planning to add support for that attribute later?

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+
+            _self.count = (status: number, payloadData: IPayloadData, endpoint: string) => {
+                if (!_isEnabled || !_checkEndpoint(endpoint)) {
+                    return;
+                }
+                if (payloadData && payloadData["statsBeatData"] && payloadData["statsBeatData"]["startTime"]) {
+                    _networkCounter.totalRequest++;
+                    _networkCounter.requestDuration += utcNow() - payloadData["statsBeatData"]["startTime"];
+                }
+                let retryArray = [401, 403, 408, 429, 500, 503];
```
lzchen, line 47: Missing `504` in retry. From https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/blob/main/ApplicationInsights/sdk_behavior_breeze.md

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+                let throttleArray = [402, 439];
+                if (status === 200) {
+                    _networkCounter.success++;
+                } else if (retryArray.includes(status)) {
+                    _networkCounter.retry[status] = (_networkCounter.retry[status] || 0) + 1;
+                } else if (throttleArray.includes(status)) {
```
lzchen, line 53: Any reason why throttle isn't separated by `status` as well like `retry`?

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+                    _networkCounter.throttle++;
+                } else if (status !== 307 && status !== 308) {
+                    _networkCounter.failure[status] = (_networkCounter.failure[status] || 0) + 1;
+                }
+                _setupTimer();
+            };
+
+            
+            
+            _self.countException = (endpoint: string, exceptionType: string) => {
```
lzchen, line 63: Not super clear on this logic, should this be only incremented on exception?

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+                if (!_isEnabled || !_checkEndpoint(endpoint)) {
+                    return;
+                }
+                _networkCounter.exception[exceptionType] = (_networkCounter.exception[exceptionType] || 0) + 1;
+                _setupTimer();
+            }
+
+            function _setupTimer() {
+                if (!_timeoutHandle) {
+                    _timeoutHandle = scheduleTimeout(() => {
```
lzchen, line 73: @MSNev 

Confirm that `scheduleTimeout` returns `ms`?

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+                    },
+                    baseType: "MetricData"
+                };
+                _core.track(statsbeatEvent);
+            }
+
+            function _trackSendRequestDuration() {
+                var currentCounter = _networkCounter;
+                if (currentCounter.totalRequest > 0 ) {
+                    var averageRequestExecutionTime = currentCounter.requestDuration  / currentCounter.totalRequest;
```
lzchen, line 128: I don't think a cumulative duration across intervals is the duration we want. We want to get the average duration WITHIN an export interval, so you need to reset the duration and count after each export.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+
+            function _checkEndpoint(endpoint: string) {
+                return _networkCounter && _networkCounter.host === endpoint;
+            }
+
+            function _getCustomProperties(ikey: string) {
+                _cikey = ikey;
+                _language = STATSBEAT_LANGUAGE;
+                _os = STATSBEAT_TYPE;
+                _runTimeVersion = STATSBEAT_TYPE;
```
lzchen, line 96: Is there a more useful value we can track for runtime version for web?

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+            function _sendStatsbeats(name: string, val: number, properties?: {}) {
+                // Add extra properties
+                let baseProperties = {
+                    "rp": "unknown",
+                    "attach": "Manual",
+                    "cikey": _cikey,
+                    "runtimeVersion": _runTimeVersion,
+                    "os": _os,
+                    "language": _language,
+                    "version": _sdkVersion,
```
lzchen, line 108: For the other sdks, we do not populate this with `sdkVersion` but rather JUST the version of the SDK.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+                    "endpoint": "breeze",
+                    "host": _networkCounter.host
+                }
+                let statsbeatEvent: ITelemetryItem = {
+                    iKey: INSTRUMENTATION_KEY,
+                    name: name,
+                    baseData: {
+                        name: name,
```
lzchen, line 116: Is this the correct way to build the `TelemetryItem`? In Python we have `type(TelemetryItem.baseData) = MetricsData` and `type(MetricsData.metrics) = List[MetricDataPoint]`.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+
+            function _trackSendRequestsCount() {
+                var currentCounter = _networkCounter;
+                if (currentCounter.success > 0) {
+                    _sendStatsbeats("Request_Success_Count", currentCounter.success);
+                }
+                
+                for (const code in currentCounter.failure) {
+                    const count = currentCounter.failure[code];
+                    _sendStatsbeats("failure", count, { statusCode: code });
```
lzchen, line 141: Probably don't want to send any of these metrics if the count is 0.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
siyuniu-ms: I followed the status code from https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/blob/main/ApplicationInsights/statsbeat/statsbeat.md, I could update the code there and specify its for js use.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
JacksonWeber: Might be able to parse the userAgent to get this information. This might be good default information though if none can be retrieved from that header.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,182 @@
...
+                let retryArray = [401, 403, 408, 429, 500, 503];
+                let throttleArray = [402, 439];
+                if (status === 200) {
+                    _networkCounter.success++;
+                } else if (retryArray.includes(status)) {
+                    _networkCounter.retry[status] = (_networkCounter.retry[status] || 0) + 1;
+                } else if (throttleArray.includes(status)) {
+                    _networkCounter.throttle++;
+                } else if (status !== 307 && status !== 308) {
+                    _networkCounter.failure[status] = (_networkCounter.failure[status] || 0) + 1;
```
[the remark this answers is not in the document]
siyuniu-ms: in the new update, I am treating 206 neither a success or a failure (it would not be counted in any category except totalReuqeustCounter)

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
siyuniu-ms: https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/pull/450

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
siyuniu-ms: yes, as we will do some envelop transform before we sent out telemetry, I will post some screenshot of how the telemetry will look like when go out of door here later

siyuniu-ms:
image
here is the example of request_duration and request_success_count telemetry that send out

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
siyuniu-ms: yes, based on the comment in ts-utils 
`* @param timeout - The time, in milliseconds that the timer should wait before the specified`

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
lzchen: It's not specifically for js use, all languages should be doing so. Let's merge the  spec change and then you can make the change here too.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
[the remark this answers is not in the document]
lzchen: Interesting, Python is treating 206 as success, I guess we did not explicitly specify what this is. @JacksonWeber @TimothyMothra @jeanbisutti do we know how each language treats 206 in terms of success/failure?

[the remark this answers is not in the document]
JacksonWeber: @lzchen we treat 206 as retriable and persist the failed envelopes to disk.

JacksonWeber (approved):
LGTM

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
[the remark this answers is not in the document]
MSNev: I would expect a 206 to be treated as a success as it most likely means that breeze has either
- applied server side sampling
- As our data is going to be mixed in with theirs (the single transport) we may be getting throttled, so when we extend this to client side details, we would also want to treat this as success -- unless we get told which ones failed and we add code to "parse" out the dropped events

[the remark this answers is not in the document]
MSNev: To be a little more simplistic, I would expect any 200 (200-299) response to be treated as a success, as a 204 response (no content) is a valid response that we get from the 1DS Collector.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
MSNev: As this component is in the core we don't really have access to the existing code that parses out the user-agent, we do a bunch of stuff in the 1DS code and some additional in the OsPlugin.

We could look at enhancing this in a later release -- probably as part of adding the OsPlugin by default (which we are planning in a future release)

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,198 @@
...
+        let _core: IAppInsightsCore;
+        let _timeoutHandle: ITimerHandler;      // Handle to the timer for sending telemetry. This way, we would not send telemetry when system sleep.
+        // Custom dimensions
+        let _cikey: string;
+        let _language: string;
+        let _sdkVersion: string;
+        let _os: string;
+        let _runTimeVersion: string;
+        dynamicProto(Statsbeat, this, (_self, _base) => {
+            _self.initialize = (core: IAppInsightsCore, ikey: string, endpoint: string, version?: string) => {
```
MSNev, line 27: Lets update the signature to pass the values as a config object (iKey, endpoint, version) as I'm feeling that we might want to update this in the future.

We could also then make these dynamic, so we could "update" then for things like passing in the browser version etc. So basically follow the normal type of "onConfigChange()" with setting defaults -- the code "supports" defining this for internal types (not part of the global config object (IConfiguration)), but it may need some signature updates or we fudge it.

For now just create an interface with these 3 and pass them in (ie. don't treat as dynamic or use the onConfigChange)

tools/chrome-debug-extension/manifest.json
```diff
@@ -2,8 +2,8 @@
     "name": "Telemetry Viewer - M3",
     "short_name": "Telemetry Viewer M3",
     "description": "A browser extension that provides a real time view of what's happening in Application Insights including what telemetry is being logged by the web application",
-    "version": "0.7.5",
-    "version_name": "0.7.5",
+    "version": "0.7.6",
```
MSNev, line 5: Changes to this file really shouldn't be in this PR.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -283,6 +288,11 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                         }
                     }
 
+                    if (!config.disableStatsBeat && _statsBeat && !_statsBeat.isInitialized()) {
```
MSNev, line 291: You need to "set" a default otherwise `disableStatsBeat` will be undefined and not dynamic.

I'm also in 2 minds on this as we currently don't have any throttling in place, so once someone turns this one (by looking at the code or the tests), this will spam (overload) our statsbeat subscription with no option to stop it....

So we "might" want to check this code in with all of the "enable" commented out so there is zero chance of someone accidently enabling this. And then as part of the follow up PR to add the throttling we then add the config and support for "enabling" / "disabling".

Basically, we don't want someone turning this on a killing ALL of our statsbeat metrics. As the end-user (customers) subscription won't see any of this data.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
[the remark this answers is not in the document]
JacksonWeber: Yeah, it looks like Node.js is out of spec treating these as retriable. I'll create a work item.

channels/applicationinsights-channel-js/Tests/Unit/src/StatsBeat.tests.ts
```diff
@@ -19,15 +19,14 @@ export class StatsbeatTests extends AITestClass {
...
     }
 
     public registerTests() {
         this.testCase({
-            name: "Statsbeat initializes when disableStatsBeat is false",
+            name: "Statsbeat initializes when intStats is true",
             test: () => {
                 let config = {
-                    disableStatsBeat: false,
+                    _sdk: {intStats: true},
```
MSNev, line 29: as `_sdk` represents just the settings for the sdk (ie. internal), then we can set this name just to `stats`

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -955,12 +956,16 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
 
             function _doSend(sendInterface: IXHROverride, payload: IInternalStorageItem[], isAsync: boolean, markAsSent: boolean = true): void | IPromise<boolean> {
                 let onComplete = (status: number, headers: {[headerName: string]: string;}, response?: string) => {
+                    if (status >= 0){
+                        _getOnComplete(payload, status, headers, response);
+                    }
+                    status = Math.abs(status);
```
MSNev, line 962: There is a `mathAbs` alias which will be shorter than this.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -705,7 +706,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                                 return;
                             }
                             _xhrReadyStateChange(request, payloadArr, payloadArr.length);
-                            oncomplete(request.status, payload.headers, getResponseText(request));
+                            oncomplete(-Math.abs(request.status), payload.headers, getResponseText(request));
```
MSNev, line 709: Why -ve?

AISKU/Tests/Unit/src/AISKUSize.Tests.ts
```diff
@@ -54,10 +54,10 @@ function _checkSize(checkType: string, maxSize: number, size: number, isNightly:
 }    
 
 export class AISKUSizeCheck extends AITestClass {
-    private readonly MAX_RAW_SIZE = 145;
-    private readonly MAX_BUNDLE_SIZE = 145;
-    private readonly MAX_RAW_DEFLATE_SIZE = 58;
-    private readonly MAX_BUNDLE_DEFLATE_SIZE = 58;
+    private readonly MAX_RAW_SIZE = 147;
```
JacksonWeber, line 57: Just a general question, but where are these max sizes calculated from?

channels/applicationinsights-channel-js/Tests/Unit/src/aichannel.tests.ts
```diff
@@ -1,9 +1,11 @@
 ﻿import { SenderTests } from "./Sender.tests";
 import { SampleTests } from "./Sample.tests";
 import { GlobalTestHooks } from "./GlobalTestHooks.Test";
+import { StatsbeatTests } from "./StatsBeat.tests";
 
 export function runTests() {
     new GlobalTestHooks().registerTests();
     new SenderTests().registerTests();
     new SampleTests().registerTests();
+    // new StatsbeatTests().registerTests();
```
JacksonWeber, line 10: Did you intend to leave this commented? Or should this get created for the test?

shared/AppInsightsCore/Tests/Unit/src/aiunittests.ts
```diff
@@ -26,4 +27,6 @@ export function runTests() {
     new UpdateConfigTests().registerTests();
     new EventsDiscardedReasonTests().registerTests();
     new W3cTraceParentTests().registerTests();
+    // new StatsBeatTests(false).registerTests();
```
JacksonWeber, line 30: Seems like possibly some more stray comments

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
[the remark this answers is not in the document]
JacksonWeber: There was a spec update made regarding 206 here: https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/pull/461 if you're interested in integrating with the spec for this PR.

JacksonWeber (approved):

AISKU/Tests/Unit/src/AISKUSize.Tests.ts
siyuniu-ms: when the test fail, it will say, "the current size is 147.32, exceed 147", then I will change the size to 148 (no smart tech here haha)

channels/applicationinsights-channel-js/Tests/Unit/src/aichannel.tests.ts
siyuniu-ms: yes, that's because I manually set statsbeat to be null in the code to make sure user would never accidently turn it on for this release, and hence the test could not pass, so i comment it out to avoid github build from fail.

AISKU/Tests/Unit/src/AISKUSize.Tests.ts
MSNev: The sizes are based on the previously released version, because bundle size is REALLY important for browsers, we have these tests that ensure that we understand the impacts of any changes and that there is no unexpected size increases

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -261,7 +264,7 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                 let diagLog = _self.diagLog();
                 _evtNamespace = mergeEvtNamespace(createUniqueNamespace("Sender"), core.evtNamespace && core.evtNamespace());
                 _offlineListener = createOfflineListener(_evtNamespace);
-
+                _core = core;
```
MSNev, line 267: nit: We actually don't need to "save" this here as `_self.core` (part of the base class) gets initialized with the core instance, this in combination with another suggestion below would remove the need for this.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -272,6 +275,8 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                     // getExtCfg only finds undefined values from core
                     let senderConfig = ctx.getExtCfg(identifier, defaultAppInsightsChannelConfig);
                     let curExtUrl = senderConfig.endpointUrl;
+                    _statsBeat = core.getStatsBeat();
```
MSNev, line 278: You have several instances of `core.getStatsBeat()` where the `getStatsBeat` portion is not going to be able to be minified.

suggestion:
Create a local helper function to return this value.

```ts
function _getStatsBeat() {
    return _self.core.getStatsBeat();
}
```

And then just call this function where you currently are, this means that all of those usages will get minified into a couple of characters, you no longer need the `_core`.

Downside: there is 1 additional function call `_getStatsBeat()` -> `_self.core.getStatsBeat()`, so there is a slight perf impact of this, but as it's not called in any tight loops and only when we are dealing with network requests I think this would be ok.

MSNev: It's always a trade-off between bundle size and performance (as this is also performing 2 lookups `_self.core` and then `core.getStateBeat` and then the function call.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -670,29 +683,65 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
-                            if (!data) {
+                            let payloadArr = _getPayloadArr(payload);
+                            if (!payloadArr) {
                                 return;
                             }
-                            return _xdrOnLoad(xdr, data);
+                            const responseText = _getResponseText(xdr);
+                            let statsbeat = _core.getStatsBeat();
+                            var endpointHost = urlParseUrl(_self._senderConfig.endpointUrl).hostname;
+                            if (xdr && (responseText + "" === "200" || responseText === "")) {
```
MSNev, line 693: you could just have this as `responseText == 200` as JavaScript will coerce both sides of a double equals `==` to the same type before comparison. Which would then make this a little smaller -- although I see this is how it was done for the other cases as well -- so maybe leave this for now

shared/AppInsightsCore/Tests/Unit/src/StatsBeat.Tests.ts
```diff
@@ -0,0 +1,260 @@
...
+                
+                // Set up spies to check internal calls
+                const countSpy = this.sandbox.spy(this._statsbeat, "count");
+                
+                // Count metrics for a different endpoint
+                this._statsbeat.count(200, payloadData, "https://different.endpoint.com");
+                
+                // The count method was called, but it should return early
+                Assert.equal(1, countSpy.callCount, "count method should be called");
+                Assert.equal(0, this._trackSpy.callCount, "track should not be called for different endpoint");
```
MSNev, line 171: This looks wrong as you are not waiting (or simulating the passing of time) for the timer to fire, so this will always be true because the timer is always created, even with a timeout of zero.

Even if the code handled a timeout value the negative test (make sure nothing is sent), should always wait (or simulate) the passing of a decent amount of time -- just to validate someone doesn't change the functionality

shared/AppInsightsCore/Tests/Unit/src/StatsBeat.Tests.ts
```diff
@@ -0,0 +1,260 @@
...
+                        break;
+                    }
+                }
+                
+                Assert.ok(foundExceptionMetric, "Exception metrics should be tracked");
+            }
+        });
+
+        this.testCase({
+            name: "StatsBeat: does not send metrics for different endpoints",
```
MSNev, line 142: What exactly is this test supposed to be doing?

As I didn't see any specific logic to "not send metrics" when the endpoint was different, and with a feature created by CSS yesterday this is actually going to be needed.

shared/AppInsightsCore/Tests/Unit/src/StatsBeat.Tests.ts
```diff
@@ -0,0 +1,260 @@
...
+                // Setup core with statsbeat enabled
+                this._core.initialize(this._config, [new ChannelPlugin()]);
+                
+                // Verify that statsbeat is created
+                const statsbeat = this._core.getStatsBeat();
+                Assert.ok(statsbeat, "Statsbeat should be created");
+                
+                // Update configuration to disable statsbeat
+if (!this._core.config._sdk) {
+                this._core.config._sdk = {};
```
MSNev, line 188: Does this really "remove" the `_sdk` object, if so we should update the default setting of the `_sdk` configuration to be a reference so that any defaults and any merged config (like fetching from the CDN) should "merge" not replace all of the properties of this object.

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -76,7 +78,8 @@ const defaultConfig: IConfigDefaults<IConfiguration> = objDeepFreeze({
     [STR_EXTENSION_CONFIG]: { ref: true, v: {} },
     [STR_CREATE_PERF_MGR]: UNDEFINED_VALUE,
     loggingLevelConsole: eLoggingSeverity.DISABLED,
-    diagnosticLogInterval: UNDEFINED_VALUE
+    diagnosticLogInterval: UNDEFINED_VALUE,
+    _sdk: {stats:false}
```
MSNev, line 82: Lets make this a mergable reference (using the helper function)

```ts
_sdk: cfgDfMerge<IInternalSdkxxx>({
    stats: false
}),
```

shared/AppInsightsCore/Tests/Unit/src/StatsBeat.Tests.ts
MSNev: And then explicitly set the value as required.

As by "removing" the reference this will actually break the dynamic configuration as any subsequent changes after it's removed will cause the automatic tracking to also get removed, so any "changes" after removal will actually not be tracked.

MSNev: Once you make the change I've just added to the core, this code as is "should" work correctly, although, technically the `_sdk` object should now never not exist

shared/AppInsightsCore/Tests/Unit/src/StatsBeat.Tests.ts
```diff
@@ -0,0 +1,260 @@
...
+                }
+                this._core.config._sdk.stats = false;
+                this.clock.tick(1); // Allow time for config changes to propagate
+                
+                // Verify that statsbeat is removed
+                const updatedStatsbeat = this._core.getStatsBeat();
+                Assert.ok(!updatedStatsbeat, "Statsbeat should be removed when disabled");
+                
+                // Re-enable statsbeat
+                this._core.config._sdk.stats = true;
```
MSNev, line 198: Lets also test setting to a non-boolean (like `null` and `undefined`) to make sure we behave and don't throw.

channels/applicationinsights-channel-js/Tests/Unit/src/StatsBeat.tests.ts
```diff
@@ -0,0 +1,246 @@
...
+                    this.assertStatsbeatCall(400, "failure");
+                    return true;
+                }
+                return false;
+            }, "Waiting for fetch sender and Statsbeat count to be called") as any)
+        });
+
+    
+
+        // this.testCaseAsync({
```
MSNev, line 186: As this is actually the "default" sender mechanism we really need to test this.

You will most likely need to use the fakeServer and the helper `this.sendJsonResponse(request, {}, 200)` to simulate the responses examples are in several tests.

MSNev (changes requested):
Blocking until the changes are addressed.

AISKU/Tests/Unit/src/AISKUSize.Tests.ts
JacksonWeber: Purpose makes sense, I was just curious if there was some kind of formula to understanding the size increase in correlation to the new data being created.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -670,29 +683,65 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
+                                if (statsbeat) {
+                                    statsbeat.count(200, payload, endpointHost);
+                                }
+                            } else {
+                                const results = parseResponse(responseText);
+                    
+                                if (results && results.itemsReceived && results.itemsReceived > results.itemsAccepted
+                                    && !_isRetryDisabled) {
+                                    if (statsbeat) {
+                                        statsbeat.count(201, payload, endpointHost);
```
lzchen, line 704: Should this be 206?

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -670,29 +683,65 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
+                                    }
```
lzchen, line 705: For 206 (partial success) we will also be adding a design that will parse the actual failed/succeeded items in the batch. I believe the logic for receiving 206 is:

1. Increment success by 1 for success
2. Increment retry by 1 if there are ANY retryable envelopes in rejected items
3. Increment failure by 1 if there are ANY failed (non-retryable) envelopes in rejected items
4. Increment throttle by 1 if there are ANY throttled response codes in rejected items.

@JacksonWeber  please correct me if I missed anything above.

The logic above has not been added to the specs yet so I don't believe we need to include this for the first iteration but it would be good to create a tracking issue for it @siyuniu-ms @MSNev .

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -670,29 +683,65 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
+                                } else {
+                                    statsbeat.count(201, payload, endpointHost);
```
lzchen, line 707: What is a 201 status code? It doesn't seem to be a valid code from [breeze](https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/blob/main/ApplicationInsights/sdk_behavior_breeze.md#the-response-returned-is-valid)

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -670,29 +683,65 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
+                                var endpointHost = urlParseUrl(_self._senderConfig.endpointUrl).hostname;
+                                statsbeat.count(request.status, payload, endpointHost);
+                            }
+                            return _xhrReadyStateChange(request, payloadArr, payloadArr.length);
                         },
                         beaconOnRetry: (data: IPayloadData, onComplete: OnCompleteCallback, canSend: (payload: IPayloadData, oncomplete: OnCompleteCallback, sync?: boolean) => boolean) => {
+                            let statsbeat = _core.getStatsBeat();
+                            if (statsbeat) {
+                                var endpointHost = urlParseUrl(_self._senderConfig.endpointUrl).hostname;
+                                statsbeat.count(-1, data, endpointHost);
```
lzchen, line 742: What is the `-1` for? Is this some hacky thing?

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -670,29 +683,65 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
-                            let data = _getPayloadArr(payload);
-                            if (!data) {
+                            let payloadArr = _getPayloadArr(payload);
+                            if (!payloadArr) {
                                 return;
                             }
-                            return _xhrReadyStateChange(request, data, data.length);
-                            
+                            let statsbeat = _core.getStatsBeat();
+                            if (statsbeat && request.readyState === 4) {
```
lzchen, line 732: Just curious, what is `request.readyState`?

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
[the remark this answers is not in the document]
lzchen: @jeanbisutti 

No problem, the spec was not clear before on how to handle 206. @JacksonWeber  added [this pr](https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/pull/461) to make defining this behavior more explicit. @siyuniu-ms  @MSNev feel free to implement the new spec in a different PR or create a tracking issue for it if you don't want to include it in this one.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
```diff
@@ -0,0 +1,196 @@
...
+                    _networkCounter.throttle[status] = (_networkCounter.throttle[status] || 0) + 1;
+                } else if (status !== 307 && status !== 308) {
+                    _networkCounter.failure[status] = (_networkCounter.failure[status] || 0) + 1;
+                }
+                _setupTimer();
+            };
+
+            
+            
+            _self.countException = (endpoint: string, exceptionType: string) => {
```
lzchen, line 63: It is not clear to me where this is being used.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
lzchen: @siyuniu-ms 

I am not sure why this is being resolved? Seems like you are still using `sdkVersion` instead of just `version`.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
lzchen: @siyuniu-ms 

I'm not sure why this was resolved? Did you check for `0` count somewhere else in the code?

shared/AppInsightsCore/Tests/Unit/src/StatsBeat.Tests.ts
siyuniu-ms: it's defined under _checkEndpoint function under statsbeat class

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -670,29 +686,64 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
-                            if (!data) {
+                            let payloadArr = _getPayloadArr(payload);
+                            if (!payloadArr) {
                                 return;
                             }
-                            return _xdrOnLoad(xdr, data);
+                            const responseText = _getResponseText(xdr);
+                            let statsbeat = _getStatsBeat();
+                            if (statsbeat) {
+                                var endpointHost = urlParseUrl(_self._senderConfig.endpointUrl).hostname;
```
MSNev, line 696: This should probably be using the `_endpointUrl` so that when redirects are detected and updated the correct URL that is being used is used.  See the `_checkAndUpdateEndPointUrl` function which is where it "updates" this internal endpoint URL and how it detects that breeze sent us a redirect.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
siyuniu-ms: It's not currently being called, but it can be added later as a customizable option for browser-side defined exceptions. For now, the function and its test is added.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -313,6 +314,8 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                         _endpointUrl = _orgEndpointUrl = senderConfig.endpointUrl;
                     }
 
+                    _getStatsBeat(); // first call will initialize the statsbeat object
```
MSNev, line 317: This is going to be potentially problematic, while it does cause the configuration to get referenced so that it would get called again, because the result is not used it "may" actually get removed by the minification process (and therefore would never get referenced).

Looking at the implementation, however, my question is 
- do we really need to call this during initialization? As anytime it's first used it should get initialized?

MSNev: i.e. Is it "needed" for some reason that I'm not spotting atm.

shared/AppInsightsCore/src/JavaScriptSDK/StatsBeat.ts
siyuniu-ms: yes, it's checked here: function _sendStatsbeats(name: string, val: number, properties?: { [name: string]: any }) {
                if (!val || val <= 0){
                    return;
                }

MSNev (approved):

PR #2491, opened 2025-03-17
[Main][Task]31248254: Add Monitor Tests for CDN OPTIONS Calls

Karlie-777:


AISKU/Tests/Unit/src/applicationinsights.e2e.tests.ts
```diff
@@ -725,6 +726,82 @@ export class ApplicationInsightsTests extends AITestClass {
...
 
+    public addCdnMonitorTests(): void {
+        this.testCaseAsync({
+            name: "E2E.GenericTests: Fetch Current CDN V3",
+            stepDelay: 1,
+            useFakeServer: false,
+            useFakeFetch: false,
+            steps: [() => {
+                // Use beta endpoint to pre-test any changes before public cdn
+                fetch("https://js.monitor.azure.com/beta/ai.3.gbl.min.js?dkjfashffafj", {
```
MSNev, line 737: While this will work for 1 run, when several runs occur from the same system this may not do what you want.

change the "random" stuff to use the current milliseconds or something that is almost guaranteed to be different for multiple runs.

MSNev: like (cb is short for cache breaking)
```js
    url += "cb=" + (utcNow());
```

MSNev (approved):

MSNev (approved):

PR #2497, opened 2025-04-02
JakesKarlie/addcfgsyncdefaults

JakeDavila210:


siyuniu-ms:
hi, what's the purpose of adding these exports? II thought CfgSyncPlugin could be consumed via adding the extension.

MSNev (changes requested):
We don't want to set a default value for this (beyond the undefined value that is already set), as this component is consumed by several other components (public and internal) and therefore having this as a default here would potentially break other consumers of this extension.

PR #2500, opened 2025-04-04
[main] Minor fixes for handling ikey promises with dynamic changes

MSNev:
- Additional Minification improvements

### DIFF
--- a/AISKU/src/AISku.ts
+++ b/AISKU/src/AISku.ts
@@ -103,12 +103,27 @@ function _chkDiagLevel(value: number) {
     return value && value > 0;
 }
 
+function _parseCs(config: IConfiguration & IConfig, configCs: string | IPromise<string>) {
+    return createSyncPromise<ConnectionString>((resolve, reject) => {
+        doAwaitResponse(configCs, (res) => {
+            let curCs = res && res.value;
+            let parsedCs = null;
+            if (!res.rejected && curCs) {
+                // replace cs with resolved values in case of circular promises
+                config.connectionString = curCs;
+                parsedCs = parseConnectionString(curCs);
+            }
+            
+            // if can't resolve cs promise, null will be returned
+            resolve(parsedCs);
+        });
+    });
+}
+
 /**
  * Application Insights API
  * @group Entrypoint
  * @group Classes
- * @class Initialization
- * @implements {IApplicationInsights}
  */
 export class AppInsightsSku implements IApplicationInsights {
     public snippet: Snippet;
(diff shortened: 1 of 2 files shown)


AISKU/src/AISku.ts
```diff
@@ -103,12 +103,27 @@ function _chkDiagLevel(value: number) {
     return value && value > 0;
 }
 
+function _parseCs(config: IConfiguration & IConfig, configCs: string | IPromise<string>) {
```
MSNev, line 106: Moved this out of the class as it's basically invariant so it doesn't need access to any of the properties in the closure.

AISKU/src/AISku.ts
```diff
@@ -200,60 +215,47 @@ export class AppInsightsSku implements IApplicationInsights {
...
                     let ikeyPromise = createSyncPromise<string>((resolve, reject) => {
-                        _parseCs().then((cs) => {
-                            let ikey = _config.instrumentationKey;
-                            ikey = cs && cs.instrumentationkey || ikey;
-                            resolve(ikey);
-                        }).catch((e) => {
-                            // parseCs will always resolve(unless timeout)
-                            // return null in case any error happens
-                            resolve(null);
+                        doAwaitResponse(_parseCs(_config, configCs), (rsp) => {
```
MSNev, line 222: Changed to use `doAwaitResponse` as this helps with minification in several ways.
- no `then` or `catch` names (which can't be minified)
- Only 1 function (when using then/catch each one generates a function)

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -348,130 +348,18 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
                 // This will be "re-run" if the referenced config properties are changed
                 _addUnloadHook(_configHandler.watch((details) => {
                     let rootCfg = details.cfg;
-                  
-                    let isPending = _activeStatus === eActiveStatus.PENDING;
-                    
-                    if (isPending){
```
MSNev, line 354: Moved this logic around as by having this here this will cause subsequent "calls" for the onConfigChanges to not tag / reference any contained properties which would then result on the onConfigChange to never get called again (as the last run didn't reference anything)

Found this as part of adding new configuration options for the OTel beta as I was adding additional checks.

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -348,130 +348,18 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
...
-                            }
-
-                        });
-                    } else {
-                        // means no promises
-                        _setStatus();
-                       
-                    }
+                    
+                    _handleIKeyEndpointPromises(rootCfg);
```
MSNev, line 354: Also broke these out into smaller functions to contain the logic rather than just having this intline

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -348,130 +348,18 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
...
-                        promises.push(endpointUrl);
-                        _endpoint = null; // reset current local endpoint variable (otherwise it will always be the previous urls if timeout is called before promise cb)
-                    } else {
-                        // string or null
-                        _endpoint = endpointUrl;
-                    }
-
-                    // at least have one promise
-                    if (promises.length) {
-                        // reset to false for new dynamic changes
```
MSNev, line 400: Also broke if part into it's own function rather than inline

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -647,6 +535,124 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
                 return _startLogPoller(true);
             };
 
+            function _handleIKeyEndpointPromises(theConfig: IConfiguration) {
+                // app Insights core only handle ikey and endpointurl, aisku will handle cs
+                // But we want to reference these config values so that if any future changes are made
+                // this will trigger the re-run of the watch function
+                // and the ikey and endpointUrl will be set to the new values
+                let ikey = theConfig.instrumentationKey;
```
MSNev, line 543: These 2 lines are really the key part of the referencing change, as per the comment if these direct configuration properties are not referenced as part of the subsequent call to the onConfigChange (they where originally after the "pending" check) then the dynamic configuration change usage detection doesn't identify that this function should be called (indirectly by the calling onConfigChange (watch) function) again if / when either of these configuration options change.

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -647,6 +535,124 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
...
+                let endpointUrl = theConfig.endpointUrl; // do not need to validate endpoint url, if it is null, default one will be set by sender
+
+                // Check if we are waiting for previous promises to be resolved, won't apply new changes
+                if (_activeStatus !== eActiveStatus.PENDING) {
```
MSNev, line 547: We "probably" should really look at handling reassignment to different promises but that's a much larger (and risker) change.

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
MSNev: Apart from moving these 2 referenced variables to before the pending check (and not having a `isPending` variable) the reset of the logic of this function "should" (please check) be the same.

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
```diff
@@ -647,6 +535,124 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
...
+            }
+
+            function _waitForInitPromises(theConfig: IConfiguration, promises: IPromise<string>[]) {
+                // reset to false for new dynamic changes
+                _isStatusSet = false;
+                _activeStatus = eActiveStatus.PENDING;
+                let initTimeout = isNotNullOrUndefined(theConfig.initTimeOut)?  theConfig.initTimeOut : maxInitTimeout; // theConfig.initTimeOut could be 0
+                let allPromises = createSyncAllSettledPromise<string>(promises);
+
+                if (_initTimer) {
```
MSNev, line 600: This bit is new, it cancels any previously running timer (which should never occur based on the above pending check), but just preparing and covering all bases.

MSNev: The rest of this function logic should be the same as previously (please check)

tools/chrome-debug-extension/manifest.json
```diff
@@ -2,8 +2,8 @@
     "name": "Telemetry Viewer - M3",
     "short_name": "Telemetry Viewer M3",
     "description": "A browser extension that provides a real time view of what's happening in Application Insights including what telemetry is being logged by the web application",
-    "version": "0.7.5",
-    "version_name": "0.7.5",
+    "version": "0.7.6",
```
MSNev, line 5: Will revert this

shared/AppInsightsCore/src/JavaScriptSDK/AppInsightsCore.ts
Karlie-777: The initial design of setting ikey after pending check is to avoid the scenarios where user changes the ikey to a string value before the previous ikey promises are resolved. But this did block the following onConfigChange before all promises are resolved.

Karlie-777 (approved):

PR #2506, opened 2025-04-28
Drop correlation header to be passed on the dependency

### LINKED ISSUE #2505: Trace-Id race condition for multiple instances of sdk
**Is your feature request related to a problem? Please describe.**
When multiple instances of application insights are running a race condition is there to inject the traceId on the request sent to the backend.

**Describe the solution you'd like**
I would expect a way to discard the event from an instance point of view, like we have the dependencyListers and dependencyInitializers

**Describe alternatives you've considered**
An alternative is to send telemetry manually, so the correct trace-id can be propagated

aimbrenda:
I want to add the possibility to drop the enrichment of the correlation header using the dependencyListenerFunction like it is done for the dependencyInitializerFunction. In this way we can drop events that are not relevant for the current instance of the SDK especially in case of multiple instances with microservices.

This resolves #2505

### DIFF
--- a/extensions/applicationinsights-dependencies-js/src/ajax.ts
+++ b/extensions/applicationinsights-dependencies-js/src/ajax.ts
@@ -410,10 +410,8 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
             _self.includeCorrelationHeaders = (ajaxData: ajaxRecord, input?: Request | string, init?: RequestInit, xhr?: XMLHttpRequestInstrumented): any => {
                 // Test Hook to allow the overriding of the location host
                 let currentWindowHost = _self["_currentWindowHost"] || _currentWindowHost;
-
-                let canBeProcessed = _processDependencyListeners(_dependencyListeners, _self.core, ajaxData, xhr, input, init);
-
-                if (canBeProcessed === true) {
+                
+                if (_processDependencyListeners(_dependencyListeners, _self.core, ajaxData, xhr, input, init)) {
                     if (input || input === "") { // Fetch
                         if (correlationIdCanIncludeCorrelationHeader(_extensionConfig, ajaxData.getAbsoluteUrl(), currentWindowHost)) {
                             if (!init) {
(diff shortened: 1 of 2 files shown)


extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -408,95 +411,97 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IDependenciesPlu
...
-                            headers.set(RequestHeaders[eRequestHeaders.requestContextHeader], RequestHeaders[eRequestHeaders.requestContextAppIdFormat] + appId);
-                            if (_enableRequestHeaderTracking) {
-                                ajaxData.requestHeaders[RequestHeaders[eRequestHeaders.requestContextHeader]] = RequestHeaders[eRequestHeaders.requestContextAppIdFormat] + appId;
-                            }
-                        }
-                        if (_isUsingW3CHeaders) {
-                            let traceFlags = ajaxData.traceFlags;
-                            if (isNullOrUndefined(traceFlags)) {
-                                traceFlags = 0x01;
+                if (canBeProcessed === true) {
```
MSNev, line 416: nit: for minification purposes this can just be either
```ts
const canBeProcessed = _processDependencyListeners(_dependencyListeners, _self.core, ajaxData, xhr, input, init);
if (canBeProcessed) {
```
or slightly better
```ts
if (_processDependencyListeners(_dependencyListeners, _self.core, ajaxData, xhr, input, init)) {
```

aimbrenda: resolved in the next commit, thanks for the feedback

MSNev (approved):

aimbrenda:
Do I need to close the pull request myself?

MSNev:
> Do I need to close the pull request myself?

😄 No need, I'll merge it shortly as part of one of the test runs failing yesterday, it identified a race condition (which I've now fixed) during our unload operations. So I wanted to get that in before hand, update this branch and then re-run the tests.

Thank you for your contribution on this and identifying the gap that we had from the underlying helper to the actual caller, this is a very neat implementation.

PR #2510, opened 2025-04-29
[main][statsbeat] enable cfySync and throttle on statsbeat

siyuniu-ms:


AISKU/src/AISku.ts
```diff
@@ -391,6 +394,13 @@ export class AppInsightsSku implements IApplicationInsights {
                             _throttleMgr.sendMessage( _eInternalMessageId.SdkLdrUpdate, "An updated Sdk Loader is available, see aka.ms/SnippetVer");
                             _sdkVerSentMessage = true;
                         }
+
+                        if (isFeatureEnabled(STATSBEAT, _config)) {
+                            if (!_throttleMgr.canThrottle(_eInternalMessageId.StatsBeat)){ // use the message id as the key to get throttleMgr control
+                                // statsBeat should be disabled
+                                _config.featureOptIn[STATSBEAT] = {mode: FeatureOptInMode.disable};
```
MSNev, line 401: I'm not sure we want to update the config like this...

Once this is set to disabled any future config sync operations will never occur as it will be blocked, so for any loooong running page / application would never be able to be re-enabled.

PR #2518, opened 2025-05-05
Remove Generated docs from the repo

MSNev:
- These docs are generated and published by default to the typedoc site, they don't need to be checked int


MSNev:
Note: These folders are already listed as excluded in the .gitignore file

Karlie-777 (approved):

PR #2524, opened 2025-05-09
Update Components to address governance issues

MSNev:
- Remove grunt-contrib-qunit to address tar-fs versioning
- Update puppeteer also to address tar-fs versioning
- Improve performance and simplify Asynchronous test generation

### DIFF
--- a/common/Tests/Framework/src/AITestClass.ts
+++ b/common/Tests/Framework/src/AITestClass.ts
@@ -1,6 +1,6 @@
 /// <reference path="../../External/qunit.d.ts" />
 import dynamicProto from "@microsoft/dynamicproto-js";
-import { arrForEach, dumpObj, getGlobal, isArray, objDefineProp, objForEachKey, objGetPrototypeOf, setBypassLazyCache, strStartsWith, throwError, getNavigator, getPerformance, strSubstr, strContains } from "@nevware21/ts-utils";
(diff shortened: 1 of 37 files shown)


tools/grunt-tasks/qunit.js
```diff
@@ -0,0 +1,465 @@
+/*
+ * grunt-contrib-qunit
```
MSNev, line 2: Copied from grunt-contrib-quint to remove the dependencies to directly fix the governance issues from this version to limit the amount of required changes

tools/grunt-tasks/chrome/bridge.js
```diff
@@ -0,0 +1,153 @@
+/*
+ * grunt-contrib-qunit
```
MSNev, line 2: Copied from grunt-contrib-qunit, this is a support file that it uses to hook into puppeteer.

tools/grunt-tasks/qunit.js
MSNev: This file has some changes which include 
- updates location of the bridge.js file
- directly includes the imported `pEachSeries` rather than use as a dependency
- Some minor fixes around the usage of the error.stack
- Merge some minor fixes from later releases to this version (required as we are still building in node 16 and the version of qunit we are using)

common/Tests/Framework/src/AITestClass.ts
```diff
@@ -191,6 +232,21 @@ export class AITestClass {
...
+            return AITestClass._currentTestContext;
+        }
+
+        return {
+            name: AITestClass.currentTestInfo ? AITestClass.currentTestInfo.name : "<undefined>",
+            context: {},
+            pollDelay: 100,
+            testDone: () => void { },
+            clock: AITestClass.currentTestClass ? AITestClass.currentTestClass.clock : null,
+            finished: true
```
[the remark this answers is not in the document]
MSNev: Intentional as this case only occurs if there is no running tests

hectorhdzg (approved):

PR #2534, opened 2025-05-20
[Main][Task]32864048: Improve Offline Channel Test

Karlie-777:


channels/offline-channel-js/Tests/Unit/src/dbprovider.tests.ts
```diff
@@ -115,67 +116,65 @@ export class OfflineDbProviderTests extends AITestClass {
                 let evt = TestHelper.mockEvent(endpoint, 3, false);
                 doAwait(provider.initialize(providerCxt), (val) => {
                     this.ctx.isInit = val;
+                    doAwait(provider.addEvent("", evt, itemCtx), (item) => {
+                        this.ctx.evt = item;
+                        Assert.deepEqual(item, evt, "should add expected item");
+                        Assert.ok(evt.id, "should add id to the item");
+                        this.preEvts.push(evt);
+                        doAwait(provider.getNextBatch(), (item) => {
```
MSNev, line 124: FYI - rather than have all of these "embedded" doAwaits (which works), as doAwait returns a promise you could (or should) actually just have all of these as individual asyncQueue entries, that way the "poll" interval doesn't need to be set to 1 second and instead each task (promise) would just wait until the next one to complete.

And as the asyncQueue.add(...) doesn't run the new task until the previous one is complete, it all "should" work.

It would (possibly) look something like
```ts
return this._asyncQueue()
    .add(() => doAwait(......)),
    .add(() => doAwait(....))
```
- The reason for putting each `doAwait` after the `() => doAwait` is so that this creates a new function for each for the task, so that the `doAwait` is not run until the previous task is complete. If you just went `queue.add(doAwait(....))` then the doAwait would be evaluated immediately and the processing of the response is what then gets "deferred" until the previous task completes.
- `add` and `concat` are the same thing as far as the asyncQueue is concerned

MSNev (approved):

channels/offline-channel-js/Tests/Unit/src/dbprovider.tests.ts
```diff
@@ -65,44 +65,41 @@ export class OfflineDbProviderTests extends AITestClass {
...
-                return false;
-            }, "Wait for init response" + new Date().toISOString(), 30, 1000) as any).concat(PollingAssert.createPollingAssert(() => {
-                let isclosed = this.ctx.isclosed;
-                if (isclosed) {
-                    return true;
-                }
-                return false;
-            }, "Wait for close response" + new Date().toISOString(), 30, 1000) as any)
+                return this._asyncQueue().add(() =>
+                    doAwait(provider.initialize(providerCxt), (val) => {
```
MSNev, line 82: Unless you create a local promise, when using `doAwait` it "should" return the result as it returns a promise which will make the caller (the queue) "wait" for it to finish before calling the next task, otherwise the next task will just run on the next execution cycle

MSNev (approved):

PR #2545, opened 2025-05-21
Enchanced dataSanitizeURL method to redact input URLs. Unit tests added.

rads-1996:


shared/AppInsightsCore/Tests/Unit/src/ApplicationInsightsCore.Tests.ts
```diff
@@ -1908,6 +1908,116 @@ export class ApplicationInsightsCoreTests extends AITestClass {
             }
         });
 
+        this.testCase({
```
JacksonWeber, line 1911: Does it make sense to add tests for URLs that contain ports or fragments as well?

rads-1996: So, according to the spec, apart from the "username:password" and certain query strings, everything else remains the same. Technically I could add ports/fragments just to make sure only what needs to redacted is being redacted does not affect any other part of the URL.

rads-1996: Updated the test file with new test cases.

JacksonWeber (approved):
LGTM

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IConfiguration.ts
```diff
@@ -219,6 +219,13 @@ export interface IConfiguration {
     // * @internal
     // */
     //_sdk?: IInternalSdkConfiguration;
+
+    /**
+     * [Optional] A flag to enable or disable the use of the field redaction for urls.
+     * @defaultValue true
+     */
+    redactionEnabled?: boolean;
```
MSNev, line 227: I don't see where / how this is getting used?

rads-1996: I have split up the PRs to make it easier to review. In this piece of code, I have suggest declared and am using it in subsequent PRs.

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -354,3 +363,44 @@ export function sendCustomEvent(evtName: string, cfg?: any, customDetails?: any)
     }
     return false;
 }
+
+/**
+ * Redacts sensitive information from a URL string, including credentials and specific query parameters.
+ * @param input - The URL string to be redacted.
+ * @returns The redacted URL string or the original string if no redaction was needed or possible.
+ */
+export function fieldRedaction(input: string): string {
```
JacksonWeber, line 372: If `input` can be of type `any` earlier (in the dataSanitizeUrl method) and then gets to this function, I'm not sure if declaring `input` as a string in the method definition for `fieldRedact` will automatically cast it, or if this could be a type error.

JacksonWeber: My suggestion would just be to improve the type safety in the `dataSanitizeUrl` method for the `url` parameter.

rads-1996: This is a good point. I will add a check to see if url is of type string before passing it to the datasanitize url method.

rads-1996:
Merged this branch into rads-1996/page-view-telemetry

PR #2549, opened 2025-05-22
Redacting urls before sending to telemetry data.

rads-1996:


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
```diff
@@ -287,6 +290,10 @@ export class AnalyticsPlugin extends BaseTelemetryPlugin implements IAppInsights
                 let doc = getDocument();
                 if (doc) {
                     pageView.refUri = pageView.refUri === undefined ? doc.referrer : pageView.refUri;
+
+                    if (pageView.refUri && _extConfig.redactionEnabled) {
```
MSNev, line 294: This pattern is repeated a lot, which results in a lot of additional code, I suggest that we move the "if" check into the fieldRedation into the helper function so that all of the usages just become `X = fieldRedaction(X, flag)` or something like that -- especially as we are defaulting to true so almost always we are going to have the performance impact of additional function call.

We may even extend this further and rather than pass the "boolean", perhaps just pass the config object and it does the complete lookup of the config.XXX once in the code (this will also further help with the smaller bundle size).

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
```diff
@@ -910,7 +925,11 @@ export class AnalyticsPlugin extends BaseTelemetryPlugin implements IAppInsights
             
                 // array with max length of 2 that store current url and previous url for SPA page route change trackPageview use.
                 let location = getLocation(true);
-                _prevUri = location && location.href || "";
+                if (location && _extConfig?.redactionEnabled) {
```
MSNev, line 928: As a general comment, because the Web SDK is still targeting ES5 we don't use the `?.` operator as this causes TypeScript to always expand this out to `if(x) { x.y} else {...}` type causing a larger code size which we can actually handler smaller with something as simple as `(config || {}).xxx` so that the initial "config" is either the real config or an empty object

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
rads-1996: That makes sense. I will modify the code for it.

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
rads-1996: I discovered another such compatibility issue. I am using the `URL` interface to parse the URLs and then checking the different parts of it to apply redaction. I just saw that this interface is not ES5 compatible. There are two alternate options we can use instead, one is to use the standard regex comparison, and the other is to use the `url-parse` method which is similar in usage to URL. However, this would mean installing the `url-parse` package. This is a cleaner approach. What would be a better approach to apply?

.github/workflows/ci.yml
```diff
@@ -8,6 +8,8 @@ on:
     branches: [ main, master, beta, Release2.7, release-3.0 ]
   pull_request:
     branches: [ main, master, beta, Release2.7, release-3.0 ]
+  merge_group:
```
MSNev, line 11: these changes seem odd as they are from my PR's...

Can you remove all but your changes

.github/workflows/copilot-triage-assignment.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     steps:
       - name: Check if actor is authorized
         id: check_auth
-        uses: actions/github-script@v6
+        uses: actions/github-script@v7
```
MSNev, line 24: This file should be deleted now.

rads-1996:
Please note, this is not the complete set of changes, I am still working on adding fieldRedaction in places where the url needs to be redacted.

shared/AppInsightsCommon/src/Telemetry/Common/DataSanitizer.ts
```diff
@@ -98,7 +99,10 @@ export function dataSanitizeString(logger: IDiagnosticLogger, value: any, maxLen
     return valueTrunc || value;
 }
 
-export function dataSanitizeUrl(logger: IDiagnosticLogger, url: any) {
+export function dataSanitizeUrl(logger: IDiagnosticLogger, url: any, config?: IConfiguration) {
+    if (typeof url === "string"){
```
MSNev, line 103: We have a helper to reduce the code size for this

`if(isString(url)) {`

extensions/applicationinsights-analytics-js/src/JavaScriptSDK/Telemetry/PageViewManager.ts
```diff
@@ -96,7 +96,9 @@ export class PageViewManager {
                     let location = getLocation();
                     uri = pageView.uri = location && location.href || "";
                 }
-
+                if (core && core.config){
+                    uri = pageView.uri = fieldRedaction(pageView.uri, core.config);
```
MSNev, line 100: 👍

extensions/applicationinsights-clickanalytics-js/src/DataCollector.ts
```diff
@@ -126,7 +126,7 @@ export function getPageName(config: IClickAnalyticsConfiguration, overrideValues
  * @param location - window.location or document.location
  * @returns Flag indicating if an element is market PII.
  */
-export function sanitizeUrl(config: IClickAnalyticsConfiguration, location: Location): string {
+export function sanitizeUrl(config: IClickAnalyticsConfiguration, location: Location, extConfig?: IConfiguration): string {
```
MSNev, line 129: nit: based on the new config being part of the root config (`IConfiguration`) lets just call this `config`

shared/AppInsightsCommon/src/Telemetry/Common/DataSanitizer.ts
rads-1996: Modified the code.

extensions/applicationinsights-clickanalytics-js/src/DataCollector.ts
rads-1996: Re-named it rootConfig.

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -30,6 +32,15 @@ const strMsie = "msie";
 const strTrident = "trident/";
 const strXMLHttpRequest = "XMLHttpRequest";
 
+const SENSITIVE_QUERY_PARAMS = [
+    "sig",
+    "Signature",
+    "AWSAccessKeyId",
+    "X-Goog-Signature"
+] as const;
```
Karlie-777, line 40: I think this might be from openTelemtry directly? I am thinking maybe we need to cover more, for example https://login.microsoftonline.com/{tenant}/oauth2/v2.0/authorize?

rads-1996: The specification particularly mentions these 4 for now and do say that this list is not exhaustive and more can be added later when the spec is updated.

rads-1996: > I am thinking, maybe we should have a config for this? we will have some defaults one, maybe we should let users to provide their customized ones as well?

Sounds good. I will work on it.

examples/dependency/src/dependencies-example-index.ts
```diff
@@ -35,6 +36,12 @@ function addDependencyInitializerOnClick() {
         // Change properties of telemetry event "before" it's been processed
         details.item.name = "dependency-name";
         details.item.properties.url = details.item?.target;
+        if (isString(details.item.properties.url)) {
```
hectorhdzg, line 39: You don't need to add this code in example apps we have here, we should be redacting the URL internally so this is more confusing than helpful

extensions/applicationinsights-clickanalytics-js/src/DataCollector.ts
```diff
@@ -141,7 +143,9 @@ export function sanitizeUrl(config: IClickAnalyticsConfiguration, location: Loca
     if (!!config.urlCollectQuery) { // false by default
         url += (isValueAssigned(location.search)? location.search : "");
     }
-
+    if (rootConfig) {
```
hectorhdzg, line 146: Instead of checking config everywhere you call fieldRedaction, add a config validation inside that method if is not already there

extensions/applicationinsights-clickanalytics-js/src/events/PageAction.ts
```diff
@@ -126,6 +126,9 @@ export class PageAction extends WebEvent {
                 
                 pageActionEvent.timeToAction = _getTimeToClick();
                 pageActionEvent.refUri = isValueAssigned(overrideValues.refUri) ? overrideValues.refUri : _self._config.coreData.referrerUri;
+                if (_self._clickAnalyticsPlugin.core.config) {
+                    pageActionEvent.refUri = fieldRedaction(pageActionEvent.refUri, _self._clickAnalyticsPlugin.core.config);
```
hectorhdzg, line 130: Do we want to redact URLs that are overridden and not automatically tracked?, in this case customer is explicitly adding this value right?

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -354,3 +357,140 @@ export function sendCustomEvent(evtName: string, cfg?: any, customDetails?: any)
...
+    }
+    return url.substring(0, schemeEndIndex + 3) + "REDACTED:REDACTED" + url.substring(atIndex);
+}
+
+/**
+ * Redacts sensitive query parameters from a URL
+ * @param url - The URL string to redact
+ * @returns The URL with sensitive query parameters redacted
+ */
+function redactQueryParameters(url: string, config?: IConfiguration): string {
```
Karlie-777, line 403: maybe we can pass sensitive_params string[] instead of the config object itself here?

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -29,6 +31,7 @@ const strReactNative = "ReactNative";
 const strMsie = "msie";
 const strTrident = "trident/";
 const strXMLHttpRequest = "XMLHttpRequest";
+const STR_REDACTED = "REDACTED";
```
hectorhdzg, line 34: Constants are usually added in a constants file in the project instead, in this case InternalConstants file

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -354,3 +357,140 @@ export function sendCustomEvent(evtName: string, cfg?: any, customDetails?: any)
...
+ * @param input - The URL string to be redacted.
+ * @param config - Configuration object that contain redactionEnabled setting.
+ * @returns The redacted URL string or the original string if no redaction was needed or possible.
+ */
+export function fieldRedaction(input: string, config: IConfiguration): string {
+    if (!input){
+        return input;
+    }
+    if (input.indexOf(" ") !== -1) {
+        return input; // Checking for URLs with spaces
```
hectorhdzg, line 483: Why the redaction is cancelled if there is a space?

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -354,3 +357,140 @@ export function sendCustomEvent(evtName: string, cfg?: any, customDetails?: any)
...
     return false;
 }
+
+/**
+ * Redacts user information from a URL
+ * @param url - The URL string to redact
+ * @returns The URL with user information redacted
+ */
+function redactUserInfo(url: string): string {
+    const schemeEndIndex = url.indexOf(":");
```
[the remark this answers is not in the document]
hectorhdzg: That will handle a bunch of edge cases, I think this code should be in that class though

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IConfiguration.ts
```diff
@@ -232,6 +232,19 @@ export interface IConfiguration {
...
+     */
+    redactionEnabled?: boolean;
+
+    /**
+     * [Optional] Additional query parameters to redact beyond the default set.
+     * Use this to specify custom parameters that contain sensitive information.
+     * These will be combined with the default parameters that are redacted.
+     * @example ["auth_token", "api_key", "private_data"]
+     */
+    redactQueryParams?: string[];
```
hectorhdzg, line 247: Let's add the defaultValue here as well

extensions/applicationinsights-clickanalytics-js/src/DataCollector.ts
```diff
@@ -126,7 +128,7 @@ export function getPageName(config: IClickAnalyticsConfiguration, overrideValues
  * @param location - window.location or document.location
  * @returns Flag indicating if an element is market PII.
  */
-export function sanitizeUrl(config: IClickAnalyticsConfiguration, location: Location): string {
+export function sanitizeUrl(config: IClickAnalyticsConfiguration, location: Location, rootConfig?: IConfiguration): string {
```
hectorhdzg, line 131: I'm very confused about this method, documentation is wrong because it says it return a flag and is actually returning location.search, https://developer.mozilla.org/en-US/docs/Web/API/Location/search that looks like is only the query string parameters of the URL not the full URL, that would cause your fieldRedaction method to fail right? unit tests are missing here if that is the case

shared/AppInsightsCore/src/applicationinsights-core-js.ts
```diff
@@ -60,7 +60,7 @@ export {
     getCrypto, getMsCrypto, getLocation, hasJSON, getJSON,
     isReactNative, getConsole, isIE, getIEVersion, isSafari,
     setEnableEnvMocks, isBeaconsSupported, isFetchSupported, useXDomainRequest, isXhrSupported,
-    findMetaTag, findNamedServerTiming, sendCustomEvent, dispatchEvent, createCustomDomEvent
+    findMetaTag, findNamedServerTiming, sendCustomEvent, dispatchEvent, createCustomDomEvent, fieldRedaction
```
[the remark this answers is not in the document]
Karlie-777: Currently it is in core because it will be used for 1ds as well but it might be good to re-export from common as well.

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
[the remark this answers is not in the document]
rads-1996: > It is a great point, reuse this function will definitely help to cover a lot of corner cases. I was thinking of using regex to parse urls if not re-using this urlHelper function. But one thing is that the function is inside common instead of core, maybe we can move this function to core and also 1ds can use it as well.

I talked to Hector about it. Nev and I had a discussion about this, and he agreed regex would be too complicated, hence we went ahead with the current approach.

extensions/applicationinsights-clickanalytics-js/src/events/PageAction.ts
Karlie-777: I think it is a good point! Does the spec mention about any customer defined URLs? or should we add configs to allow disabling auto-redact urls?

shared/AppInsightsCommon/Tests/Unit/src/AppInsightsCommon.tests.ts
```diff
@@ -293,5 +293,85 @@ export class ApplicationInsightsTests extends AITestClass {
...
+            test: () => {
+                // URLs with sensitive query parameters
+                const urlWithSensitiveParams = "https://example.com/api?Signature=secret&normal=value";
+                const expectedRedactedUrl = "https://example.com/api?Signature=REDACTED&normal=value";
+
+                // Act & Assert
+                const result = dataSanitizeUrl(this.logger, urlWithSensitiveParams);
+                Assert.equal(expectedRedactedUrl, result);
+            }
+        });
```
JacksonWeber, line 375: Should we add unit tests for edge-cases such as empty string and non-string input?

extensions/applicationinsights-clickanalytics-js/src/DataCollector.ts
rads-1996: I think it is returning the entire url - 

```
var url = location.protocol + "//" + (location.hostname || location.host) +         // location.hostname is not supported on Opera and Opera for Android
        (isValueAssigned(location.port) ? ":" + location.port : "") +
        location.pathname;
---------------------
return url
```

And I am redacting the URL right before it is returned. However, I do agree the return is incorrectly documented.

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IConfiguration.ts
hectorhdzg: I meant the @defaultValue tag in the comments

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -354,3 +356,139 @@ export function sendCustomEvent(evtName: string, cfg?: any, customDetails?: any)
...
+                const paramValue = pair.substring(equalsIndex + 1);
+                params[paramName] = paramValue;
+            }
+        }
+    }
+    
+    // Check if any parameters need redaction
+    let anyParamRedacted = false;
+    if (config && config.redactQueryParams) {
+        sensitiveParams = DEFAULT_SENSITIVE_PARAMS.concat(config.redactQueryParams);
```
Karlie-777, line 445: nit: 
```
let queryParams =  config.redactQueryParams
if (config && queryParams && queryParams.length) {
    sensitiveParams = queryParams

}
```

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IConfiguration.ts
```diff
@@ -232,6 +232,20 @@ export interface IConfiguration {
      */
     expCfg?: IExceptionConfig;
 
+    /**
+     * [Optional] A flag to enable or disable the use of the field redaction for urls.
+     * @defaultValue true
+     */
+    redactionEnabled?: boolean;
```
MSNev, line 239: A better name would be to describe that this is for Url's (as calling just `redactionEnabled` could also mean that you can disable "ALL" redaction for every field.

Suggestion: call it something like `redactUrls`

MSNev: Well... The term Redaction generally means that you are obfuscating any "sensitive" information that a field may contain, so while yes it could mean that we are converting the entire URL to some other obfuscated form (like "***" etc), this is where documentation plays a part on the configs.

Also for the web we need to be cognizant of the size constraints where every byte counts, so we play a trade off between size and readability. Also as a URL consists of different "parts" one of which is called the "Params" the `redactSensitiveParams` could also just mean that portion (which is not what we are doing).

Perhaps another alternative name could be `scrubUrl` etc, but as OpenTelemetry and the content is calling it redaction, I'd perfer `redact`

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
rads-1996: But we want to include both if the user has provided custom values, right? In the else part it only takes the default values considering that the redactedQueryParams has not been provided.

shared/AppInsightsCore/src/JavaScriptSDK.Interfaces/IConfiguration.ts
rads-1996: Modified the config.

shared/AppInsightsCore/Tests/Unit/src/ApplicationInsightsCore.Tests.ts
```diff
@@ -1908,6 +1908,338 @@ export class ApplicationInsightsCoreTests extends AITestClass {
             }
         });
 
+        this.testCase({
+            name: "should redact basic auth credentials from URL",
+            test: () => {
+                let config = {} as IConfiguration;
+
+                const url = "https://user:password@example.com";
+                if (config.redactUrls){
```
MSNev, line 1917: nit: If we are assuming that the "default" is true, then we should be checking that it's true -- otherwise when we default to false the test with the if will never get executed....

So either we don't need the if or we should have an explicit assertion that the value is defaulting to true.

shared/AppInsightsCore/Tests/Unit/src/ApplicationInsightsCore.Tests.ts
```diff
@@ -1908,6 +1908,338 @@ export class ApplicationInsightsCoreTests extends AITestClass {
...
+                Assert.equal(redactedLocation, "AISKU/Tests/UnitTests.html?sig=REDACTED");
+            }
+        });
+
+        this.testCase({
+            name: "should preserve query parameters while redacting auth - AWSAccessKeyId",
+            test: () => {
+                let config = {redactUrls: false} as IConfiguration;
+                const url = "https://www.example.com/path?color=blue&AWSAccessKeyId=secret";
+                if (config.redactUrls){
```
MSNev, line 1961: Same comment as above about the if

shared/AppInsightsCore/Tests/Unit/src/ApplicationInsightsCore.Tests.ts
```diff
@@ -1908,6 +1908,338 @@ export class ApplicationInsightsCoreTests extends AITestClass {
...
+            }
+        });
+
+        this.testCase({
+            name: "should handle URLs with Unicode characters",
+            test: () => {
+                let config = {} as IConfiguration;
+                const url = "https://user:пароль@例子.测试/路径?参数=值&sig=秘密";
+                const redactedLocation = fieldRedaction(url, config);
+                Assert.equal(redactedLocation, "https://REDACTED:REDACTED@例子.测试/路径?参数=值&sig=REDACTED",
```
MSNev, line 2111: 👍

shared/AppInsightsCommon/Tests/Unit/src/AppInsightsCommon.tests.ts
```diff
@@ -293,5 +293,116 @@ export class ApplicationInsightsTests extends AITestClass {
...
+                const result = dataSanitizeUrl(this.logger, urlWithCredentials, config);
+                Assert.equal(expectedRedactedUrl, result);
+            }
+        });
+        this.testCase({
+            name: 'DataSanitizerTests: dataSanitizeUrl properly redacts credentials in URLs with config disabled',
+            test: () => {
+                // URLs with credentials
+                let config = {redactUrls: false} as IConfiguration;
+                const urlWithCredentials = "https://username:password@example.com/path";
```
MSNev, line 314: Lets add some additional tests where the username / password has encoded values or even if they are invalid, like someone tries to pass in an invalid url with the same url (for some reason)

eg something unexpected like, but it starts to "look" like a username / password
`https://file:///C/user....`

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -354,3 +356,139 @@ export function sendCustomEvent(evtName: string, cfg?: any, customDetails?: any)
...
 }
+
+/**
+ * Redacts user information from a URL
+ * @param url - The URL string to redact
+ * @returns The URL with user information redacted
+ */
+function redactUserInfo(url: string): string {
+    const schemeEndIndex = url.indexOf(":");
+    if (schemeEndIndex === -1) {
```
MSNev, line 367: I know we talked about a regex being super complicated, but I was actually thinking about a single regex doing everything... 

As you have split this out into separate tasks (username/password and then query string), I think it would be faster (from an execution perspective) to use a RegEx for the username / password.

I asked co-pilot to create a regex and it cam up with the following which looks like it would work and also be a lot smaller from a code perspective and "should" be faster during execution as well

```ts
function redactBasicAuth(url) {
  return url.replace(/^([a-zA-Z][a-zA-Z0-9+.-]*:\/\/)([^:@]+):([^@]+)@/g, '$1REDACTED:REDACTED@');
}

// Example usage:
const url1 = 'ftp://user:pass@host.com/resource';
const url2 = 'custom-scheme://admin:1234@myapp.local';
console.log(redactBasicAuth(url1)); // ftp://REDACTED:REDACTED@host.com/resource
console.log(redactBasicAuth(url2)); // custom-scheme://REDACTED:REDACTED@myapp.local
```

We should also have tests for invalid scheme's etc so that this doesn't cause infinite looping when processing the regex. This might require changing the + to set upper limits {1,200} not sure if we should "allow" the username or password to be zero length (so 0, 200 instead of 1, 200)

`/^([a-zA-Z][a-zA-Z0-9+.-]*:\/\/)([^:@]{1,200}):([^@]{1,200})@/g`

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -354,3 +356,139 @@ export function sendCustomEvent(evtName: string, cfg?: any, customDetails?: any)
...
+        const pairs = queryString.split("&");
+        for (let i = 0; i < pairs.length; i++) {
+            const pair = pairs[i];
+            if (!pair) {
+                continue;
+            }
+            
+            const equalsIndex = strIndexOf(pair, "=");
+            if (equalsIndex === -1) {
+                params[pair] = null;
```
MSNev, line 433: This actually has some "potential" security implications as your "assigning" user passed values to a generic object, so it's technically" possible to overwrite core properties of an object. (this problem may exist elsewhere as well -- but this one is a fairly easy way to overwrite / reassign object properties)

So to avoid this is to create an array or (just rebuild the param string directly) which 
- keeps the parameters in the same order as processed (using the objForEachKey doesn't guarantee that the order of the parameters is preserved
- If the same parameter is passed more than once then that would also be preserved (yes this is possible)
- Within this loop we should check if the identified name is within the sensitive list and just assign the REDACTED value directly to this new result


We should probably also "speed" up the process by "pre-checking" if the URL contains a potential value that needs to be  redacted before attempting to split it out into a new string or array as the creation of new strings / array / objects can be expensive. (so a "check" on line 420 for example).

This can be a simple "check" that will sometimes return true for a "partial" match (ie. using arrIndexOf would work with a "=" after the key name

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
```diff
@@ -354,3 +356,139 @@ export function sendCustomEvent(evtName: string, cfg?: any, customDetails?: any)
...
+        return input;
+    }
+    if (input.indexOf(" ") !== -1) {
+        return input; // Checking for URLs with spaces
+    }
+    const isRedactionDisabled = config && config.redactUrls === false;
+    if (isRedactionDisabled) {
+        return input;
+    }
+    try {
```
MSNev, line 487: Or maybe (extension of the comment above), add the check here to see if the URL contains "@" or any "recactionKey=" before calling the either function,

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
rads-1996: Added mentioned logic. Would appreciate feedback on the same.

MSNev (approved):
Minor nit comment

shared/AppInsightsCore/Tests/Unit/src/ApplicationInsightsCore.Tests.ts
```diff
@@ -1908,6 +1908,670 @@ export class ApplicationInsightsCoreTests extends AITestClass {
             }
         });
 
+        this.testCase({
+            name: "FieldRedaction: should redact basic auth credentials from URL",
+            test: () => {
+                let config = {} as IConfiguration;
+
+                const url = "https://user:password@example.com";
+                if (config.redactUrls === true){
```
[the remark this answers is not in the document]
rads-1996: Thanks for pointing that out. I have added the config and set it to False.

shared/AppInsightsCore/Tests/Unit/src/ApplicationInsightsCore.Tests.ts
```diff
@@ -1908,6 +1908,670 @@ export class ApplicationInsightsCoreTests extends AITestClass {
             }
         });
 
+        this.testCase({
+            name: "FieldRedaction: should redact basic auth credentials from URL",
+            test: () => {
+                let config = {redactUrls: false} as IConfiguration;
```
MSNev, line 1914: This still result in the if being false... And therefore lines 1918..1919 won't get executed...

If you drop the If check (line 1917) and then change 1919 to check that the `redactedLocation` is still the same as the passed in url would be a better test (assuming I'm understanding what this is trying to test)

shared/AppInsightsCore/src/JavaScriptSDK/EnvUtils.ts
rads-1996: I am merging this PR for now. If a change is needed, I will send it as a follow up PR.

PR #2577, opened 2025-05-29
[Main][Task]32909441: Increase Web Snippet to 1.2.2

Karlie-777:


tools/applicationinsights-web-snippet/src/snippet.ts
```diff
@@ -62,7 +62,7 @@ declare var cfg:ISnippetConfig;
         let appInsights: (Snippet & {cookie?:any, core?:any, extensions?:any, initialize?: boolean, isInitialized?: () => boolean;}) = {
             initialize: true,   // initialize sdk on download
             queue: [],
-            sv: "8",       // Track the actual snippet version for reporting.
+            sv: "9",       // Track the actual snippet version for reporting.
```
Karlie-777, line 65: changed sv version to 9

version.json
```diff
@@ -102,7 +102,7 @@
         },
         "@microsoft/applicationinsights-web-snippet": {
             "package": "tools/applicationinsights-web-snippet/package.json",
-            "release": "1.2.4"
```
rads-1996, line 105: A bit confused here. Why is the version greater here than what it needs to be?

tools/applicationinsights-web-snippet/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@microsoft/applicationinsights-web-snippet",
-    "version": "1.2.4",
```
rads-1996, line 3: Same question.

version.json
Karlie-777: see my comments above

tools/applicationinsights-web-snippet/package.json
Karlie-777: see my comments above

PR #2504, opened 2025-04-17
[main] [Click analytics] not logging no native html input elements  #2136

siyuniu-ms:

PR #2587, opened 2025-06-12
Click Analytics - Fix capturning of HTML events

Barry-RG:
Changes made in https://github.com/microsoft/ApplicationInsights-JS/pull/2504 (b5a24d524bc0cae3c48c7010d88cef5f982449c8) appear to have introduced an issue when evaluating if events have come from known HTML objects. This appears to be due to a new array which had previously been an object still being indexed by its values. As a result these events are not being processed.

This change is to fix that by looking for the value in the array.


Karlie-777:
https://github.com/microsoft/ApplicationInsights-JS/issues/2589

Karlie-777 (approved):

Karlie-777:
@microsoft-github-policy-service agree

PR #2605, opened 2025-07-08
Adding copilot instructions

hectorhdzg:


Karlie-777 (approved):

.github/copilot-instructions.md
```diff
@@ -0,0 +1,254 @@
...
+## Project Overview
+This is the **Microsoft Application Insights JavaScript SDK** - a browser-based telemetry library for monitoring web applications. The SDK tracks page views, user interactions, performance metrics, exceptions, and custom events.
+
+## Architecture & Structure
+
+### Monorepo Structure
+- **AISKU/**: Main Application Insights SDK package
+- **AISKULight/**: Lightweight version of the SDK
+- **shared/**: Core shared libraries (AppInsightsCore, AppInsightsCommon, 1ds-core-js)
+- **extensions/**: Plugin-based extensions (analytics, dependencies, React, Angular, etc.)
```
MSNev, line 12: The React and Angular extensions are now in their own repo, so we shouldn't include them here, we "should" be able to tell copilot that we have extensions and include their repo's -- likewise when we add this file to those repos

.github/copilot-instructions.md
```diff
@@ -0,0 +1,254 @@
...
+- Always use semicolons
+- Use 4-space indentation
+- Maximum line length: 140 characters
+
+### Naming Conventions
+- **Classes**: PascalCase (e.g., `PageViewManager`, `TelemetryContext`)
+- **Interfaces**: PascalCase with `I` prefix (e.g., `ITelemetryItem`, `IPageViewTelemetry`)
+- **Methods/Functions**: camelCase (e.g., `trackPageView`, `sendTelemetry`)
+- **Constants**: UPPER_SNAKE_CASE (e.g., `MAX_DURATION_ALLOWED`)
+- **Private variables**: underscore prefix (e.g., `_logger`, `_hasInitialized`)
```
MSNev, line 39: We should add an entry for enums stating that they should be constant and named as PascalCase with a leading `e` and if required to be exported it should use the `createEnumStyle` helper to create the export. And ALL usages within the code base should use the constant enum -- which should also be interers and not strings.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,254 @@
...
+- Minimize object allocations in hot paths
+- Use `doPerf()` wrapper for performance tracking
+- Avoid synchronous operations that could block the browser
+- Implement lazy initialization where possible
+- Use object pooling for frequently created objects
+
+## Browser Compatibility
+
+### Target Support
+- **Modern browsers**: Chrome, Firefox, Safari, Edge
```
MSNev, line 80: Maybe this should specifically include targeting ES5 for modern browsers, so it doesn't try and use `async` and `await` (yet)

MSNev: Actually, we should state how it should support async operations (async / await) via the ts-async helpers
- doAwait for `await`
- createPromise, createSyncPromise, createIdlePromise instead of just declaring that a function is `async` and having TypeScript generate a promise. And the return type of the function signature should use `IPromise` instead of `Promise` (for IE support -- until we stop supporting ES5)

.github/copilot-instructions.md
```diff
@@ -0,0 +1,254 @@
...
+- **Legacy support**: Internet Explorer 8+ (ES5 compatibility required)
+- **Mobile browsers**: iOS Safari, Android Chrome
+
+### Compatibility Patterns
+- Feature detection over browser detection
+- Graceful degradation for missing APIs
+- Polyfills for missing functionality (when necessary)
```
MSNev, line 87: Using existing polyfills rather than generating its own new polyfills (hopefully, we don't need to create too many more of these -- although as we move forward we might need additional functions)

.github/copilot-instructions.md
```diff
@@ -0,0 +1,254 @@
...
+## Code Style & Patterns
+
+### TypeScript/JavaScript Conventions
+- Use **ES5-compatible** syntax for browser support
+- Prefer `function` declarations over arrow functions for better IE compatibility
+- Use `var` instead of `let/const` in some contexts for ES5 compatibility
+- Always use semicolons
+- Use 4-space indentation
+- Maximum line length: 140 characters
+
```
MSNev, line 33: Need a section (or sections) about the following probably around coding patterns
- Each package should be side-effect free
- All code should be tree-shakable
- Any globals should be lazily initialized to enable tree-shaking using the ILazyValue interface or some simular pattern to determine the difference between the value has not yet been check / assigned vs the value resulting value is null / undefined.
-

MSNev (approved):

Issue #2634, opened 2025-07-30
[BUG] Headers removed when _processDependencyListeners returns false in the dependency extension


aimbrenda:
**Description/Screenshot**
When one of the dependency listeners returns false, the includeCorrelationHeaders function, from the ajax.ts file, is removing all the headers.

**Steps to Reproduce**
1. Add a dependency listener that return false
2. make an api call adding custom headers
3. check sent headers

**Expected behavior**
It should keep all the existing headers

Karlie-777:
@aimbrenda Next Release will be scheduled in Next Week and it will be the main release

PR #2635, opened 2025-07-30
Fix issue #2634 moving the conditional check for filtering

aimbrenda:


extensions/applicationinsights-dependencies-js/src/ajax.ts
```diff
@@ -425,96 +425,94 @@ export class AjaxMonitor extends BaseTelemetryPlugin implements IAjaxMonitorPlug
...
+                    }
 
-                            // init headers override original request headers
-                            // so, if they exist use only them, otherwise use request's because they should have been applied in the first place
-                            // not using original request headers will result in them being lost
-                            let headers = new Headers(init.headers || (input instanceof Request ? (input.headers || {}) : {}));
-                            if (_isUsingAIHeaders) {
+                    return init;
+                } else if (xhr) { // XHR
+                    if (processDependencyListenersResult && correlationIdCanIncludeCorrelationHeader(_extensionConfig, ajaxData.getAbsoluteUrl(), currentWindowHost)) {
```
MSNev, line 471: Actually a better solution would be to change the `return undefined;` at the end of the function to `return init || xhr;` (or `return xhr || init;` and remove the 2 return statements line lines 469 and 515 as this will result in smaller code and when the dependencyListener returns false it will also continue to shortcut the replacement of the init object.

The other option option which looks like the root cause of the issue is [here](https://github.com/microsoft/ApplicationInsights-JS/blob/9b8a82acc0fe44ff48181d6fd7e2a75071451e63/extensions/applicationinsights-dependencies-js/src/ajax.ts#L676) if we change this to `if (newInit && newInit !== init) {`, then I think that would also address the issue as it will leave the original init in place.

extensions/applicationinsights-dependencies-js/Tests/Unit/src/ajax.tests.ts
```diff
@@ -1909,6 +1909,55 @@ export class AjaxTests extends AITestClass {
...
+
+                    // Act
+                    Assert.ok(fetchSpy.notCalled);
+                    fetch(url, init);
+
+                    // Assert
+                    Assert.ok(fetchSpy.calledOnce);
+                    Assert.ok(headerSpy.calledOnce);
+                    Assert.deepEqual(init, headerSpy.returnValue || headerSpy.returnValues[0]);
+            } catch (e) {
```
MSNev, line 1954: We would prefer if this just threw and failed the test as it would provide an indication that something bad has occurred, we also generally don't log to the console (except for debugging) as it makes the logs messy

extensions/applicationinsights-dependencies-js/Tests/Unit/src/ajax.tests.ts
```diff
@@ -1909,6 +1909,55 @@ export class AjaxTests extends AITestClass {
...
+                    console && console.warn("Exception: " + e);
+                    Assert.ok(false, e);
```
MSNev, line 1956: So a better pattern (if you really want to keep the exception handling -- which is ok because you have this assert) would be

```ts
Assert.ok(false, "Unexpected failure: "  dumpObj(e));
```

And just loose the console warn, as the dumpObj has special case handling for Error objects to provide all of the content (message and stack) it a slightly more friendly format

extensions/applicationinsights-dependencies-js/src/ajax.ts
aimbrenda: I tried the second option but it doesn't work.. it seems checking for undefined doesn't work as expected. shall i proceed with the first approach?

aimbrenda: the first approach is for sure working

JacksonWeber (approved):

aimbrenda:
Any information about the next release time ?

Issue #2653, opened 2025-10-13
[BUG] "getSdkLoaderScript" results in "SyntaxError: Unexpected token '}' "


work933k:
**Description/Screenshot**

I'm doing a little POC to generate a page with Angular SSR on which i load Application Insights. It seems the generated script is invalid.
```
import {getSdkLoaderScript } from '@microsoft/applicationinsights-web-snippet';

function getAppInsightsSnippet() {
  return getSdkLoaderScript({
    connectionString: process.env.APPLICATIONINSIGHTS_CONNECTION_STRING,
  })
}

app.get('/', (req, res) => {        
    res.type('html');    
    res.send(`<html><head><script type="text/javascript">${getAppInsightsSnippet()}</script></head><body>Hello!</body></html>`);
  });
```

<img width="504" height="92" alt="Image" src="https://github.com/user-attachments/assets/104de4f3-9bda-416c-8737-84b656fb0323" />

```
[...]

PR #2654, opened 2025-10-14
Fix: #2653 Update applicationinsights-web-snippet.ts

work933k:
Fix for issue: https://github.com/microsoft/ApplicationInsights-JS/issues/2653


hectorhdzg:
@work933k thanks for your contribution and to file the issue, this will be fixed in #2659 as we believe is better to fix the issue in the creation of the snippet instead.

PR #2660, opened 2025-10-28
[OTEL Web SDK] Consolidate all OTel code in single package

hectorhdzg:


shared/AppInsightsCommon/src/Utils/TraceParent.ts
```diff
@@ -17,32 +17,6 @@ export const INVALID_TRACE_ID = "00000000000000000000000000000000";
 export const INVALID_SPAN_ID = "0000000000000000";
 const SAMPLED_FLAG = 0x01;
 
-/**
```
JacksonWeber: Just trying to understand why this function is being moved and made public. Is it no longer intended to just be a util function? Or just getting moved out of core for dependency reasons?

hectorhdzg: This was duplicated code, it was already there in different utils file

Issue #2664, opened 2025-11-13
[BUG] Failed to monitor Window.fetch


TheMatrixan:
**Description/Screenshot**

I have a Next.js app (v15 app router) with AI integrated but in the AI there're a lot of errors like this one:

```
AI (Internal): 15 message:"
Failed to monitor Window.fetch, monitoring data for this ajax call may be incorrect.
props: {
  ajaxDiagnosticsMessage:,
  exception:[object Error]: {
    stack: "TypeError: n[o.Sj] is not a function
     at H (https://example.com/_next/static/chunks/465-cd84ab02758a5705.js:1:34667)
     at https://example.com/_next/static/chunks/5339-7f30b31bf9ed8cea.js:1:117023
     at req (https://example.com/_next/static/chunks/5339-7f30b31bf9ed8cea.js:1:117255)
     at https://example.com/_next/static/chunks/5339-7f30b31bf9ed8cea.js:1:34067
     at t1 (https://example.com/_next/static/chunks/5339-7f30b31bf9ed8cea.js:1:33916)
     at t2 (https://example.com/_next/static/chunks/5339-7f30b31bf9ed8cea.js:1:33967)
     at https://example.com/_next/static/chunks/5339-7f30b31bf9ed8cea.js:1:34560
     at y (https://example.com/_next/static/chunks/7934-69bcf3dc80c2431b.js:1:47919)
     at _ (https://example.com/_next/static/chunks/7934-69bcf3dc80c2431b.js:1:46789)
     at https://example.com/_next/static/chunks/7934-69bcf3dc80c2431b.js:1:10999",
     message: "n[o.Sj] is not a function",
     name: "TypeError"
  }
}
```

**Steps to Reproduce**

Just open the page with such configuration.

 - OS/Browser: 

 - SDK Version [e.g. 22]:
````
"@microsoft/applicationinsights-react-js": "^19.3.8",
"@microsoft/applicationinsights-web": "^3.3.10",
````

 - How you initialized the SDK:
```ts
const APP_INSIGHTS_REACT_PLUGIN = new ReactPlugin()

const browserHistory = createBrowserHistory({})

appInsights = new ApplicationInsights({
  config: {
    connectionString: process.env.APPLICATIONINSIGHTS_CONNECTION_STRING,
    enableAutoRouteTracking: true,
    enableRequestHeaderTracking: true,
    enableResponseHeaderTracking: true,
    enableAjaxPerfTracking: true,
[...]

PR #2667, opened 2025-11-18
Add a check to prevent non-string URLs from being passed to fieldRedaction method

rads-1996:
Add a check to prevent non-string URLs from being passed to `fieldRedaction` method

Fixes - #2664


extensions/applicationinsights-analytics-js/src/JavaScriptSDK/AnalyticsPlugin.ts
```diff
@@ -260,7 +260,9 @@ export class AnalyticsPlugin extends BaseTelemetryPlugin implements IAppInsights
                 try {
                     let inPv = pageView || {};
                     if (_self.core && _self.core.config) {
-                        inPv.uri = fieldRedaction(inPv.uri, _self.core.config);
+                        if (isString(inPv.uri)) {
```
nev21: Rather than have this `isString` check in multiple locations (as this causes the bundle size to be larger than required), I think it would be better to put this check inside of the `fieldRedaction` function and have the first line just check there.

```ts
    if (!input || !isString(input) || strIndexOf(input, " ") !== -1) {
        return asString(input);
    }
...
```
Although I'm not actually sure why checking for a space would exclude the string from being redacted as "technically" it's possible that a field (of some kind) may contain a space (although normally encoded as `%20` but if it has been decoded it will be a space

rads-1996: Here we want to check for unencoded empty spaces. I remember running the entire suite and there were instances when the URL was populated with empty spaces (unencoded) and tests were failing due to that.

PR #2670, opened 2025-11-21
Fix static web CDN test assertion (Fixes failing CI tests)

rads-1996:


MSNev (approved):

MSNev (approved):
Note: I don't think (or more correctly I don't know) if the Assert.fail exists in the main branch -- you may have to add that (or use a different Assert (like `Assert.equal(false, true, "message" or Qunit... etc) if the tests fail

MSNev (approved):

PR #2680, opened 2025-12-09
Follow up PR: Address feedback on OTEL Logs SDK

rads-1996:


shared/OpenTelemetry/src/sdk/OTelLogRecord.ts
```diff
@@ -33,13 +35,28 @@ export function createLogRecord(
...
+    const logAttributes = (!hasAttributeContainer && attributes) ? attributes : {};
     const now = Date.now();
     const hrTime = timeInputToHrTime(timestamp || now);
     const hrTimeObserved = timeInputToHrTime(observedTimestamp || now);
     const resource = sharedState.resource;
     const logRecordLimits: Required<IOTelLogRecordLimits> = sharedState.logRecordLimits;
     const handlers: IOTelErrorHandlers = {};
+    const attributeContainer = createAttributeContainer<any>(
+        {
+            traceCfg: {
```
MSNev: The config should be injected from the main config and not applied here as it's own version (non-user configurable)

PR #2707, opened 2026-02-21
Enable Customer SDK Stats

JacksonWeber:
This pull request introduces SDK statistics telemetry collection and notification improvements, as well as enhanced event retry notification and error reporting in the Application Insights JavaScript SDK. The main changes add support for dynamically enabling/disabling SDK stats telemetry, improve notification to listeners about retried and discarded events (with status codes), and enhance test coverage for these new behaviors.

**SDK Statistics Telemetry and Dynamic Listener Management**
* Added support for SDK statistics telemetry (`SdkStats`) in the main configuration and enabled dynamic creation and cleanup of the SDK stats notification listener (`_sdkStatsListener`) based on feature flags. The listener is properly unloaded and flushed with other telemetry during shutdown. (`AISKU/src/AISku.ts`) [[1]](diffhunk://#diff-ca99b1c604c519634795cc92d7fb7d2f7f80fd7d6ff4aa50d97e9fdd0b3ed919L14-R21) [[2]](diffhunk://#diff-ca99b1c604c519634795cc92d7fb7d2f7f80fd7d6ff4aa50d97e9fdd0b3ed919R68-R70) [[3]](diffhunk://#diff-ca99b1c604c519634795cc92d7fb7d2f7f80fd7d6ff4aa50d97e9fdd0b3ed919L96-R101) [[4]](diffhunk://#diff-ca99b1c604c519634795cc92d7fb7d2f7f80fd7d6ff4aa50d97e9fdd0b3ed919R204) [[5]](diffhunk://#diff-ca99b1c604c519634795cc92d7fb7d2f7f80fd7d6ff4aa50d97e9fdd0b3ed919R446-R466) [[6]](diffhunk://#diff-ca99b1c604c519634795cc92d7fb7d2f7f80fd7d6ff4aa50d97e9fdd0b3ed919L595-R639) [[7]](diffhunk://#diff-ca99b1c604c519634795cc92d7fb7d2f7f80fd7d6ff4aa50d97e9fdd0b3ed919R754)

**Event Retry and Notification Enhancements**
* Implemented notification to listeners when events are retried (requeued) after a retriable failure, including passing HTTP status codes. This ensures better observability of event processing and retries. (`channels/1ds-post-js/src/PostChannel.ts`) [[1]](diffhunk://#diff-5330539db27b83f56726c24a02c603a83b68258c9d8cb3d77a529c8e2f46f6d6R1027) [[2]](diffhunk://#diff-5330539db27b83f56726c24a02c603a83b68258c9d8cb3d77a529c8e2f46f6d6R1049) [[3]](diffhunk://#diff-5330539db27b83f56726c24a02c603a83b68258c9d8cb3d77a529c8e2f46f6d6R1058-R1066)

**Testing Improvements**
* Added a comprehensive unit test to verify that the `eventsRetry` notification is fired when events are requeued after a retriable failure, ensuring the feature works as intended. (`channels/1ds-post-js/test/Unit/src/PostChannelTest.ts`) [[1]](diffhunk://#diff-7db0707319124dfb1a687f1466466cbfceb4daff1261e350a83fa2d290ab3b0cR36-R37) [[2]](diffhunk://#diff-7db0707319124dfb1a687f1466466cbfceb4daff1261e350a83fa2d290ab3b0cR69-R70) [[3]](diffhunk://#diff-7db0707319124dfb1a687f1466466cbfceb4daff1261e350a83fa2d290ab3b0cR1019-R1078)

**Channel Plugin Notification Improvements**
* Enhanced the classic channel plugin to:
  - Store the telemetry item's `baseType` for SDK stats mapping.
  - Notify listeners of discarded events (with status code) and successful sends.
  - Notify listeners of retries, including the HTTP status code. (`channels/applicationinsights-channel-js/src/Interfaces.ts`, `channels/applicationinsights-channel-js/src/Sender.ts`) [[1]](diffhunk://#diff-10c5975857ed91e7dc329d2eeb091ba8bed30ebbf98d31fee96a90f463ac7968R17-R21) [[2]](diffhunk://#diff-d1f29788d9390adbc9ff2f85e60bf370be120a7985298d0657fea9ff831e5322L514-R515) [[3]](diffhunk://#diff-d1f29788d9390adbc9ff2f85e60bf370be120a7985298d0657fea9ff831e5322L629-R635) [[4]](diffhunk://#diff-d1f29788d9390adbc9ff2f85e60bf370be120a7985298d0657fea9ff831e5322L802-R819) [[5]](diffhunk://#diff-d1f29788d9390adbc9ff2f85e60bf370be120a7985298d0657fea9ff831e5322R865-R873) [[6]](diffhunk://#diff-d1f29788d9390adbc9ff2f85e60bf370be120a7985298d0657fea9ff831e5322L1117-R1136) [[7]](diffhunk://#diff-d1f29788d9390adbc9ff2f85e60bf370be120a7985298d0657fea9ff831e5322R1147-R1149) [[8]](diffhunk://#diff-d1f29788d9390adbc9ff2f85e60bf370be120a7985298d0657fea9ff831e5322R1158-R1167)


shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,273 @@
...
+        for (var i = 0; i < items.length; i++) {
+            if (!_isSdkStatsMetric(items[i])) {
+                var t = _getTelType(items[i]);
+                if (_safeKey(t)) {
+                    _successCounts[t] = (_successCounts[t] || 0) + 1;
+                }
+            }
+        }
+        _ensureTimer();
+    }
```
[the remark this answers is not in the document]
JacksonWeber: Added check.

lzchen (approved):

AISKU/src/AISku.ts
```diff
@@ -390,6 +396,19 @@ export class AppInsightsSku implements IApplicationInsights<IConfiguration & ICo
                     // initialize core
                     _core.initialize(_config, [ _sender, properties, dependencies, _analyticsPlugin, _cfgSyncPlugin], logger, notificationManager);
 
+                    // Register SDK Stats notification listener (on by default)
+                    if (isFeatureEnabled(SDK_STATS, _config, true)) {
```
MSNev: This initialization should be inside the onConfigChange below so that if/when the configuration is changed (by the config feature fetch) this can be enabled / disabled on the fly without needing to re-initialize the SDK (a new page load)

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -852,6 +862,15 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
              */
             function _onSuccess(payload: IInternalStorageItem[], countOfItemsInPayload: number) {
                 _self._buffer && _self._buffer.clearSent(payload);
+
+                // Notify listeners of successful send
+                let mgr = _getNotifyMgr();
+                if (mgr) {
+                    let items = _extractTelemetryItems(payload);
```
MSNev, line 869: Follow up PR -- lets try and minimize the need to call this, it might be a large PR as it will most likely require changing the whole serialization process (like we do in 1ds-post-js) where we convert to a string as late as possible so at this level we are still playing with the objects -- there is a gotya around the persistence to the session storage / array storage today for this though.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,272 @@
...
+};
+
+/**
+ * Configuration interface for the SDK Stats notification callback.
+ */
+export interface ISdkStatsConfig {
+    /**
+     * The track function to call when flushing metrics. Typically core.track().
+     */
+    trk: (item: ITelemetryItem) => void;
```
MSNev: I don't think you actually need this as it you pass the IAppInsightsCore instance that has the track function that is shared across both AI and 1DS

AISKU/src/AISku.ts
```diff
@@ -390,6 +396,19 @@ export class AppInsightsSku implements IApplicationInsights<IConfiguration & ICo
...
+                        _sdkStatsListener = createSdkStatsNotifCbk({
+                            trk: function (item: ITelemetryItem) {
+                                _core.track(item);
+                            },
+                            lang: "JavaScript",
+                            ver: SDK_STATS_VERSION,
+                            int: SDK_STATS_FLUSH_INTERVAL
```
MSNev: As one of the items is the interval which should be dynamic, I think we should really be passing the common sdkStats config (from the shared Config) or pass down the full config (IConfiguration) to this function and if we want to cache it internally in a local variable (still reading the PR) then we would need to use the onCfgChange() helper to listen to changes.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,272 @@
...
+ * @param cfg - The SDK stats configuration
+ * @returns An INotificationListener with flush and unload methods
+ */
+/*#__NO_SIDE_EFFECTS__*/
+export function createSdkStatsNotifCbk(cfg: ISdkStatsConfig): ISdkStatsNotifCbk {
+    var _successCounts: { [telType: string]: number } = objCreate(null);
+    var _droppedCounts: { [code: string]: { [telType: string]: number } } = objCreate(null);
+    var _retryCounts: { [code: string]: { [telType: string]: number } } = objCreate(null);
+    var _timer: ITimerHandler;
+    var _interval = cfg.int || FLUSH_INTERVAL;
```
MSNev: This should be the common dynamic config so we can change via the CDN callback

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,272 @@
...
+        }
+        for (var i = 0; i < items.length; i++) {
+            if (!_isSdkStatsMetric(items[i])) {
+                var t = _getTelType(items[i]);
+                if (_safeKey(t)) {
+                    bucket[t] = (bucket[t] || 0) + 1;
+                }
+            }
+        }
+        _ensureTimer();
```
MSNev: unlikely, but minor. Only call if a new value was set

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    function _incDropped(items: ITelemetryItem[], code: string) {
+        if (!items || !items.length) return;
+        if (!_safeKey(code)) {
+            return;
+        }
+        var bucket: { [telType: string]: number };
```
MSNev: this could be simplified slightly and avoid the call to `objHasOwn`

```ts
let bucket = _droppedCount[code];
if (!bucket) {
    bucket = _droppedCount[code] = objCreate(null);
}
```

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,272 @@
...
+        _ensureTimer();
+    }
+
+    function _incRetry(items: ITelemetryItem[], code: string) {
+        if (!items || !items.length) return;
+        if (!_safeKey(code)) {
+            return;
+        }
+        var bucket: { [telType: string]: number };
+        if (objHasOwn(_retryCounts, code)) {
```
MSNev: same as above

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,272 @@
...
+                var t = _getTelType(items[i]);
+                if (_safeKey(t)) {
+                    bucket[t] = (bucket[t] || 0) + 1;
+                }
+            }
+        }
+        _ensureTimer();
+    }
+
+    function _incRetry(items: ITelemetryItem[], code: string) {
```
MSNev: minification optimization, this function and _incDropped look mostly the same, could put the bulk in a common function and then pass the differences as parameters.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,272 @@
...
+                }
+            }
+        }
+        _ensureTimer();
+    }
+
+    function _createMetric(name: string, value: number, props: { [key: string]: any }): ITelemetryItem {
+        // Merge standard dimensions
+        props[P_LANG] = cfg.lang;
+        props[P_VER] = cfg.ver;
```
MSNev: Minification optimization: As these are only actually used once, it's better to just use the names inline instead of creating constants.

The build system "counts" all objects that are used more than the number of bytes saved by using local variables and will automatically translate (using the DynamicConstants) as part of the build, reusing the same constant across the entire bundle / package.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,272 @@
...
+        props[P_COMPUTE] = "unknown"; // Browser SDK cannot reliably detect compute type
+
+        return {
+            name: name,
+            baseType: "MetricData",
```
MSNev: There is now an exported constant `MetricDataType` for the "MetricData" so could use that instead (Total bundle minification stuff)

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,272 @@
...
+                if (cnt > 0) {
+                    var successProps: { [key: string]: any } = {};
+                    successProps[P_TEL_TYPE] = telType;
+                    cfg.trk(_createMetric(MET_SUCCESS, cnt, successProps));
+                }
+            }
+        }
+
+        // Flush dropped counts
+        for (code in _droppedCounts) {
```
MSNev: Looks like these could be collapsed as well into a single helper function

AISKU/src/AISku.ts
JacksonWeber: Moved the initialization.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Changed `ISdkStatsConfig.trk: (item: ITelemetryItem) => void` to `ISdkStatsConfig.core: IAppInsightsCore`. All internal calls now use `cfg.core.track(item)` instead of `cfg.trk(item)`. Updated AISku.ts initialization to pass `core: _core` instead of the `trk` wrapper function.

AISKU/src/AISku.ts
JacksonWeber: The interval is updated to be passed through the config and used in the `createSdkStatsNotifCbk`. Since the SDK stats listener is now created/destroyed inside `onConfigChange`, the interval should be updated by destroying and recreating the listener when config changes.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Addressed as part of the above change to make interval dynamic.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Added a `changed` flag in `_incSuccess` to ensure this isn't changed if a new value isn't set.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Replaced the `objHasOwn` + if/else pattern with the simplified pattern in the new `_incBucketed` helper function

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Addressed in above comment.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Merged `_incDropped` and `_incRetry` into a single `_incBucketed(counters, items, code)` helper function.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Removed the `P_LANG`, `P_VER`, `P_COMPUTE`, `P_TEL_TYPE`, `P_DROP_CODE`, and `P_RETRY_CODE` constants. Replaced all usages with inline string literals (`"language"`, `"version"`, `"computeType"`, `"telemetry_type"`, `"drop.code"`, `"retry.code"`)

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Imported `MetricDataType` from `"../telemetry/ai/DataTypes"` and replaced the inline `"MetricData"` string in `_createMetric` with the `MetricDataType` constant.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Created a unified helper to handle this, `_flushBucketed()`.

shared/AppInsightsCore/src/interfaces/ai/IConfiguration.ts
```diff
@@ -258,6 +259,12 @@ export interface IConfiguration extends IOTelConfig {
      * @defaultValue eTraceHeadersMode.All
      */
     traceHdrMode?: eTraceHeadersMode;
+
+    /**
+     * [Optional] SDK Stats configuration for periodic reporting of telemetry pipeline metrics.
+     * @since 3.3.12
+     */
+    sdkStats?: ISdkStatsConfig;
```
MSNev, line 267: I just realized something :-)

There is no "defaults" being applied as part of this PR (that I can see), either in the AppInsightsCore (applies to both AI and 1DS) or the AISKU (just for AI) -- which means the values won't be dynamic as they won't exist during initialization -- we should set the defaults of this config block to the expected defaults (this is in addition to the "feature" enablement.

MSNev: We should also have tests (I've not gone through the tests yet -- so ignore if you already have this), that test the enabling / disabling of the feature and changing of the settings within the sdkstats to ensure they are being followed.

The "gotya" with the dynamic config is that the easiest way is to use the fake timers and because the dynamic config changes are "notified" via a setTimeout(..., 0) you need to use the `this.clock.tick(1)` to simulate this after changing a config value for it to be applied.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /**
+     * Common helper to increment a bucketed counter (dropped or retry) keyed by code and telemetry type.
+     */
+    function _incBucketed(counters: { [code: string]: { [telType: string]: number } }, items: ITelemetryItem[], code: string) {
+        if (!items || !items.length) {
+            return;
+        }
+        if (!_safeKey(code)) {
```
MSNev: nit: this could be collapsed into the above to return the number of "return" (non-minifiable code) keywords.
Ideally, invert the logic to remove all `return` usages and just put the code below within an if that validates the pre-conditions

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,249 @@
...
+                if (cnt > 0) {
+                    var successProps: { [key: string]: any } = {};
+                    successProps.telemetry_type = telType;
+                    core.track(_createMetric(MET_SUCCESS, cnt, successProps));
+                }
+            }
+        }
+
+        // Flush dropped and retry counts via common helper
+        _flushBucketed(_droppedCounts, MET_DROPPED, "drop.code");
```
MSNev, line 210: is the "." part of the specification?

As this is going to be problematic for 1DS as there is a "mode" where this will get expanded into "get: { code: xxx }" when sent. By default this mode is now off, but there are still some users that need this for their backend storage.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,249 @@
...
+     */
+    function _flushBucketed(counters: { [code: string]: { [telType: string]: number } }, metricName: string, codePropKey: string) {
+        for (var code in counters) {
+            if (objHasOwn(counters, code)) {
+                var bucket = counters[code];
+                for (var telType in bucket) {
+                    if (objHasOwn(bucket, telType)) {
+                        var cnt = bucket[telType];
+                        if (cnt > 0) {
+                            var props: { [key: string]: any } = {};
```
MSNev: We could (I think -- minification comment) collapse this into the `_createMetric` function so that we just pass 3 arguments (code and codePropKey are optional) so that we only create objects and populate these values within the helper function (as the usage of `.telemetry_type` won' be minified.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: image
Unfortunately it is. Do we need to discuss changing the spec on this?

shared/AppInsightsCore/src/interfaces/ai/IConfiguration.ts
JacksonWeber: Added defaults + tests to ensure that dynamic config updates change values upon next flush.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Collapsed the two early-return guard clauses into a single if check and removed the return keywords.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Collapsed the props creation into `_createMetric` so now it takes `telType` and optional `code` & `codePropKey` params instead of a prebuilt object.

No more multiple object creation

AISKU/src/AISku.ts
```diff
@@ -93,8 +95,14 @@ const defaultConfigValues: IConfigDefaults<IConfiguration & IConfig> = {
         [IKEY_USAGE]: {mode: FeatureOptInMode.enable}, //for versions after 3.1.2 (>= 3.2.0)
         [CDN_USAGE]: {mode: FeatureOptInMode.disable},
         [SDK_LOADER_VER]: {mode: FeatureOptInMode.disable},
-        [ZIP_PAYLOAD]: {mode: FeatureOptInMode.none}
+        [ZIP_PAYLOAD]: {mode: FeatureOptInMode.none},
+        [SDK_STATS]: {mode: FeatureOptInMode.enable}
     },
+    sdkStats: cfgDfMerge({
+        lang: "JavaScript",
```
MSNev: Do we really want to make this a config? As this would allow end-users to change this to anything they like

AISKU/src/AISku.ts
```diff
@@ -93,8 +95,14 @@ const defaultConfigValues: IConfigDefaults<IConfiguration & IConfig> = {
...
+        ver: UNDEFINED_VALUE,
```
MSNev: I guess likewise with this one.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+
+    function _createMetric(name: string, value: number, telType: string, code?: string, codePropKey?: string): ITelemetryItem {
+        // Re-read from core.config each flush so dynamic config changes are picked up
+        var statsCfg = (core.config && core.config.sdkStats) || {};
+        var props: { [key: string]: any } = {
+            telemetry_type: telType,
+            language: statsCfg.lang || "JavaScript",
+            version: statsCfg.ver || "unknown",
```
MSNev: Rather than defaulting to "Unknown" we could put a const version at the top of the file, which would be set by the build -- downside, for 1DS this would report the ApplicationInsights version and not the 1DS version.

And I guess these 2 setting are also part of the eariler comment about "letting" the end-user change / set the config -- should both of these be some sort of "internal" config (which we don't currently have). But there is a pluginString that we could use something like this that could expose the SDK version etc

AISKU/Tests/Unit/src/SdkStatsFeature.tests.ts
```diff
@@ -0,0 +1,378 @@
...
+
+    private _findSdkStatsListener(ai: AppInsightsSku): ISdkStatsNotifCbk | null {
+        let core = ai["core"];
+        let notifyMgr = core.getNotifyMgr();
+        let listeners = (notifyMgr as any).listeners;
+        if (listeners) {
+            for (let i = 0; i < listeners.length; i++) {
+                let listener = listeners[i];
+                // The SDK stats listener has a flush and unload method
+                if (listener && typeof listener.flush === "function" && typeof listener.unload === "function" &&
```
MSNev, line 78: general comment (doesn't need to be changed)

This is potentially fragile and might break at some point in the future, as this would find the "first" listener that has all 4 functions and may not actually be the sdkStats listener -- maybe use a "known" symbol to tag the returned instance (object returned by the createSdkStats...) using effectively the Symbol.for (via symbolFor helper) as this returns the same global symbol for the provided string -- so symbolFor("sdkStats") always creates and returns the same symbol regardless of who creates it the first time,

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Added this constant to pull version from the build.

AISKU/src/AISku.ts
JacksonWeber: Removed lang from being dynamically configurable.

shared/AppInsightsCore/src/core/AppInsightsCore.ts
```diff
@@ -109,6 +109,11 @@ const defaultConfig: IConfigDefaults<IConfiguration> = objDeepFreeze({
     loggingLevelConsole: eLoggingSeverity.DISABLED,
     diagnosticLogInterval: UNDEFINED_VALUE,
     traceHdrMode: eTraceHeadersMode.All,
+    sdkStats: cfgDfMerge({
+        lang: "JavaScript",
```
MSNev: These 2 are the 2 items I was talking about -- by having them here they will be populated in the shared config (and become dynamic) -- if they are not needed (because we don't wany to let them be changed) then they should not be included (extra bundle size and CPU / memory usage on initialization)

shared/AppInsightsCore/src/index.ts
```diff
@@ -39,6 +39,7 @@ export { parseResponse } from "./core/ResponseHelpers";
 export { IXDomainRequest, IBackendResponse } from "./interfaces/ai/IXDomainRequest";
 export { _ISenderOnComplete, _ISendPostMgrConfig, _ITimeoutOverrideWrapper, _IInternalXhrOverride } from "./interfaces/ai/ISenderPostManager";
 export { SenderPostManager } from "./core/SenderPostManager";
+export { createSdkStatsNotifCbk, ISdkStatsConfig, ISdkStatsNotifCbk } from "./core/SdkStatsNotificationCbk";
```
MSNev, line 42: Do we need to export "createSdkStatsNotifCbk"? 
As this then becomes part of the public API, as part of removing the 1ds-core-js so it's now part of the applicationinsights-core-js it gives us more flexibility to avoid exposing internal implementations just to share for 1ds.

MSNev: And tests "should" be able to reach into the src files within importing from the package to enable internal unit testing

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    function _createMetric(name: string, value: number, telType: string, code?: string, codePropKey?: string): ITelemetryItem {
+        var props: { [key: string]: any } = {
+            telemetry_type: telType,
+            computeType: "unknown"
+        };
+
+        // version and language are SDK-internal
+        objDefine(props, "language", { v: "JavaScript", w: false, c: false, e: true });
```
MSNev: I don't think we need to use objDefine here as your just creating the "event" (metric) so just adding to the props "should" be fine ....

Unless you want to explicitly stop the end-users changing this value via a telemetry initializer (which is really all this is doing. -- It was just the config / core instance that I was thinking about for this not the reported event.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,238 @@
...
+        objDefine(props, "version", { v: _version, w: false, c: false, e: true });
```
MSNev: This one is also going to always use the application insights version and not the 1ds-version when running from 1ds. -- So as customer stats go this "might" confuse (cause questions / incidents) people.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Creates an INotificationListener that accumulates success/dropped/retry counts and periodically
+ * flushes them as Item_Success_Count, Item_Dropped_Count, and Item_Retry_Count metrics via core.track().
+ * Language is hardcoded to "JavaScript" and version is set at build time.
+ * Reads config.sdkStats.int dynamically from core.config on each flush.
+ * @param core - The IAppInsightsCore instance (provides track() and config)
+ * @returns An INotificationListener with flush and unload methods
+ */
+/*#__NO_SIDE_EFFECTS__*/
+export function createSdkStatsNotifCbk(core: IAppInsightsCore): ISdkStatsNotifCbk {
```
MSNev: Perhaps either pass the "version" from the core instance here -- or add a specific readonly version to the core instance and use that.

shared/AppInsightsCore/src/core/AppInsightsCore.ts
JacksonWeber: Removed `lang: "JavaScript"` and `ver: UNDEFINED_VALUE` from the `sdkStats` block in `defaultConfig`. Only `int: 900000` remains as a shared dynamic default, since that's the only user-configurable property.

shared/AppInsightsCore/src/index.ts
JacksonWeber: The export is kept for now because AISKU imports it via `@microsoft/applicationinsights-core-js` (standard node_modules resolution). Removing it would break the AISKU build.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Replaced `objDefine(props, "language", { v: ..., w: false })` and `objDefine(props, "version", { v: ..., w: false })` with regular property assignments on the props object.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Changed `createSdkStatsNotifCbk` to accept a `sdkVersion: string` parameter instead of using a hardcoded `var _version = "#version#"` build-time constant. The caller (AISKU or 1DS) now passes the correct version for their SDK. AISKU passes `_sdkVersion` which is set from the `#version#` build-time placeholder (replaced with the AISKU package version during build).

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,235 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { ITimerHandler, objCreate, objHasOwn, scheduleTimeout } from "@nevware21/ts-utils";
+import { IAppInsightsCore } from "../interfaces/ai/IAppInsightsCore";
+import { INotificationListener } from "../interfaces/ai/INotificationListener";
+import { ITelemetryItem } from "../interfaces/ai/ITelemetryItem";
+import { MetricDataType } from "../telemetry/ai/DataTypes";
+
+var FLUSH_INTERVAL = 900000; // 15 min default
+var MET_SUCCESS = "Item_Success_Count";
```
MSNev, line 10: For 1DS, I'm not sure that these event names will work, we should also define the supported events in common schema

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,235 @@
...
+var MET_RETRY = "Item_Retry_Count";
+var DROP_CLIENT_EXCEPTION = "CLIENT_EXCEPTION";
+
+// Guard against prototype-polluting keys
+function _safeKey(key: string): boolean {
+    return key !== "__proto__" && key !== "constructor" && key !== "prototype";
+}
+
+// Map baseType to spec telemetry_type values
+var _typeMap: { [key: string]: string } = {
```
MSNev, line 21: There are also a couple of additional Types in 1DS
- ContentUpdateData
- PageActionData
- PageUnloadData (I think this exists in AI as well)

Exceptions while the "baseType" is still ExceptionData the event name is called "Ms.Web.ClientError"

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
MSNev: There is also no specifically defined "Metric" in common schema

MSNev: They probably should also be scoped to avoid clashing end-user custom events with the same name -- this is probably also true for all SDK's and not just the web (but the web is probably where this might be more of an issue -- especially for 1ds)

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -1388,6 +1414,37 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
                 }
             }
 
+            /**
+             * Extracts minimal ITelemetryItem objects from IInternalStorageItem[] for notification dispatch.
+             * Uses the stored baseType (bT) to reconstruct telemetry items.
+             */
+            function _extractTelemetryItems(payload: IInternalStorageItem[]): ITelemetryItem[] {
```
MSNev, line 1422: **[Copilot Performance] Hot-path allocation: _extractTelemetryItems creates N+1 throwaway objects per batch**

This function is called on every _onSuccess, _onError, and _notifyRetry path. For a batch of 500 events it allocates 1 array + 500 { name, baseType } objects that are immediately discarded after the notification listener reads .baseType.

Since the only consumer (SdkStatsNotifCbk) just needs the aseType to map to a telemetry type and increment a counter, consider one of:

1. **Pre-aggregate here** — iterate payload once, build a Map<bT, count>, and pass a single summary object to the notification. This turns 500 allocations into ~5–8 (one per unique baseType).
2. **Short-circuit when no listeners** — _getNotifyMgr() always returns the manager even when zero listeners are registered. Adding a quick mgr.listeners.length check (or similar) before calling _extractTelemetryItems would avoid the allocation entirely when SDK stats is disabled.

The same pattern appears at lines 813–818 (_onError), 867–872 (_onSuccess), and 1439–1443 (_notifyRetry).

channels/1ds-post-js/src/PostChannel.ts
```diff
@@ -1024,6 +1024,7 @@ export class PostChannel extends BaseTelemetryPlugin implements IChannelControls
              */
             function _requeueEvents(batches: EventBatch[], reason?: number) {
                 let droppedEvents: IPostTransmissionTelemetryItem[] = [];
+                let requeuedEvents: IPostTransmissionTelemetryItem[] = [];
```
MSNev: **[Copilot Performance] Unconditional array allocation on retry path**

`requeuedEvents` is allocated on every call to `_requeueEvents` even when all events end up dropped (or vice-versa for `droppedEvents`). On the happy-path where every event is requeued, the `droppedEvents` array is wasted, and on the failure path the `requeuedEvents` array is wasted.

Consider lazy-initializing:
```js
let requeuedEvents;
// inside the loop:
if (!requeuedEvents) { requeuedEvents = []; }
requeuedEvents.push(theEvent);
```
Minor savings per call, but `_requeueEvents` runs on every retriable failure.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,235 @@
...
+                    }
+                }
+            }
+            if (changed) {
+                _ensureTimer();
+            }
+        }
+    }
+
+    function _createMetric(name: string, value: number, telType: string, code?: string, codePropKey?: string): ITelemetryItem {
```
MSNev, line 165: **[Copilot Performance] Minor: static properties rebuilt on every _createMetric call**

`language`, `computeType`, and `version` are identical across all metrics in a flush. Two small wins:

1. Pre-compute `var _version = sdkVersion || ""unknown"";` once at factory creation time (line 67) instead of evaluating on each call.
2. On line 214–216, `_flush` replaces all three accumulator objects with fresh `objCreate(null)` calls. An alternative is clearing in-place (`for (var k in obj) { delete obj[k]; }`), which avoids creating new null-prototype objects every 15 minutes.

Both are low-frequency (flush interval), so minor impact — but they're free cleanups.

MSNev (commented):
**Performance-focused review (identified by GitHub Copilot)**
This review focuses on unnecessary object allocations and performance optimization opportunities on the hot send/retry paths.

AISKU/src/AISku.ts
```diff
@@ -436,6 +444,22 @@ export class AppInsightsSku implements IApplicationInsights<IConfiguration & ICo
...
+
+                        // Enable/disable SDK Stats listener dynamically based on config
+                        if (isFeatureEnabled(SDK_STATS, _config, true)) {
+                            if (!_sdkStatsListener) {
+                                _sdkStatsListener = createSdkStatsNotifCbk(_core, _sdkVersion);
+                                _core.addNotificationListener(_sdkStatsListener);
+                            }
+                        } else if (_sdkStatsListener) {
+                            try {
+                                _sdkStatsListener.unload();
```
MSNev, line 456: **[Copilot Correctness] Race condition: unload() enqueues metrics via core.track(), then listener is immediately removed**

_sdkStatsListener.unload() internally calls _flush() which calls core.track() to enqueue the accumulated SDK stats metrics. Immediately after, removeNotificationListener is called. If any part of the core.track() pipeline fires a notification synchronously (e.g. eventsSent from a channel that processes inline), the listener is still registered and could re-enter. Conversely, if core.track() enqueues asynchronously, the listener may be removed before the event is fully processed.

The same pattern appears in the _doUnload path at lines 626-627.

Worth verifying that core.track() inside _flush completes synchronously before the listener is removed, or consider nullifying the accumulator state inside unload() so any late callbacks are no-ops.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -1388,6 +1414,37 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
+                if (payload && payload.length) {
+                    let items: ITelemetryItem[] = [];
+                    arrForEach(payload, (p) => {
+                        if (p) {
+                            let baseType = p.bT || "EventData";
+                            items.push({ name: baseType, baseType: baseType } as ITelemetryItem);
```
MSNev: **[Copilot Correctness] name is set to baseType, losing the original event name**

The reconstructed ITelemetryItem uses baseType for both name and baseType:
`js
items.push({ name: baseType, baseType: baseType } as ITelemetryItem);
`

An original event like { name: "MyCustomEvent", baseType: "EventData" } becomes { name: "EventData", baseType: "EventData" }. The original event name is lost.

Today the only consumer (SdkStatsNotifCbk) uses item.name to filter out SDK stats metrics (Item_Success_Count, etc.) and item.baseType for telemetry type mapping. Since name is now always the baseType string, the self-filtering in _isSdkStatsMetric works only by coincidence - "EventData" !== "Item_Success_Count". But if a future listener relies on name being the actual event name, it will get wrong data.

Consider either:
- Using a distinct sentinel for name (e.g. omit it or set to empty string), or
- Storing the original name alongside bT in IInternalStorageItem if it is needed downstream.

channels/1ds-post-js/src/PostChannel.ts
```diff
@@ -1053,6 +1055,15 @@ export class PostChannel extends BaseTelemetryPlugin implements IChannelControls
                     }
                 });
 
+                // Notify listeners of retried events
+                if (requeuedEvents.length > 0) {
+                    // Extract HTTP status code from the EventBatchNotificationReason if in the ResponseFailure range (9000-9999)
+                    let statusCode = (reason >= EventBatchNotificationReason.ResponseFailure && reason <= EventBatchNotificationReason.ResponseFailureMax)
```
MSNev, line 1063: **[Copilot Correctness] Status code defaults to 0 when reason is outside the ResponseFailure range - ambiguous sentinel**

`js
let statusCode = (reason >= ...ResponseFailure && reason <= ...ResponseFailureMax)
    ? reason - EventBatchNotificationReason.ResponseFailure
    : 0;
`

When reason falls outside the 9000-9999 range, statusCode becomes 0. In HTTP semantics, status code 0 conventionally means "network failure / no response" (e.g. CORS block, offline). Using 0 as the fallback makes it indistinguishable from an actual network-level failure.

Downstream, SdkStatsNotifCbk converts this to a string ("0") and uses it as retry.code. A consumer looking at retry.code: "0" cannot tell if it was a real network failure or just an unmapped reason code.

Consider using -1 or undefined as the fallback to clearly indicate "status code not available / not applicable".

MSNev (commented):
**Correctness review (identified by GitHub Copilot)**
This review covers potential correctness issues around race conditions, data fidelity, and ambiguous sentinel values.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
lzchen: I'm wondering if the existing [metric stream](https://msazure.visualstudio.com/One/_git/CommonSchema?path=/v4.0/PartB/metrics/MetricStream.md) in common schema can be utilized (even though it seems more originating from OpenTelemetry).

channels/applicationinsights-channel-js/src/Sender.ts
JacksonWeber: _Replying as GitHub Copilot 🤖_

Addressed in 691fe55e. I went with option 2 (short-circuit): `_onError`, `_onSuccess`, and `_notifyRetry` now check `mgr.listeners && mgr.listeners.length` before calling `_extractTelemetryItems`. When no notification listeners are registered — the common case when SDK stats is disabled — the N+1 throwaway item allocation is skipped entirely rather than built and discarded.

channels/1ds-post-js/src/PostChannel.ts
JacksonWeber: _Replying as GitHub Copilot 🤖_

Addressed in 691fe55e. `requeuedEvents` and `droppedEvents` are now lazily initialized via `(arr || (arr = [])).push(theEvent)`, so only the array that actually receives events is allocated on a given call. The downstream `length` checks were updated to null-guard the arrays (`if (requeuedEvents && requeuedEvents.length > 0)`).

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: _Replying as GitHub Copilot 🤖_

Addressed in 691fe55e, both suggestions:
1. The version is now computed once at factory creation (`var _version = sdkVersion || \unknown\;`) and reused in `_createMetric` instead of being re-evaluated per metric.
2. `_flush` now resets the three accumulators in place via a small `_clearObj` helper (deletes own keys) rather than allocating fresh `objCreate(null)` objects every interval.

AISKU/src/AISku.ts
JacksonWeber: _Replying as GitHub Copilot 🤖_

Fixed in 08764816. `unload()` now sets an `_unloaded` guard **before** calling `_flush()`. So if `core.track()` synchronously re-enters this listener (e.g. an inline `eventsSent` from a channel that processes events synchronously) while it is still registered, those callbacks (`eventsSent`/`eventsDiscarded`/`eventsRetry`) become no-ops, and `_ensureTimer` will no longer re-arm the flush timer. This makes both this path and the `_doUnload` path at lines 626-627 safe regardless of whether `track()` completes synchronously or asynchronously. Added a regression test ("notifications after unload are no-ops") in `SdkStatsNotificationCbk.Tests.ts`.

channels/applicationinsights-channel-js/src/Sender.ts
JacksonWeber: _Replying as GitHub Copilot 🤖_

Fixed in 08764816 using the "store the original name" option. I added an `iN` field to `IInternalStorageItem` and populate it from `telemetryItem.name` at enqueue time, then `_extractTelemetryItems` rebuilds the item as `{ name: p.iN || baseType, baseType }` instead of using `baseType` for both.

This matters for correctness, not just future-proofing: SDK stats metrics emitted via `core.track()` flow back through this channel, and `SdkStatsNotifCbk._isSdkStatsMetric` filters them out **by name** (`Item_Success_Count`, etc.). With `name` previously overwritten by `baseType` (`"MetricData"`), those emitted stats metrics bypassed the filter and were counted back into the stats — a feedback loop. Preserving the real name closes that loop.

channels/1ds-post-js/src/PostChannel.ts
JacksonWeber: _Replying as GitHub Copilot 🤖_

Fixed in 08764816. The fallback is now `-1` instead of `0` when `reason` falls outside the `ResponseFailure` (9000-9999) range. Downstream `SdkStatsNotifCbk.eventsRetry` stringifies this to `retryCode: "-1"`, which is clearly distinguishable from `"0"` — the conventional HTTP value for a real network-level failure (CORS block, offline, no response). I went with `-1` over `undefined` so the `"" + statusCode` conversion stays clean (`"-1"` rather than `"undefined"`).

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Added the additional types

channels/applicationinsights-channel-js/src/Interfaces.ts
```diff
@@ -14,6 +14,18 @@ export interface IInternalStorageItem {
      * total retry count
      */
     cnt?: number;
+    /**
+     * baseType of the original telemetry item, used for SDK stats telemetryType mapping.
+     * @since 3.3.12
```
MSNev: Is the since tag version correct?

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    function _incSuccess(items: ITelemetryItem[]) {
+        if (_unloaded || !items || !items.length) {
+            return;
+        }
+        var changed = false;
+        for (var i = 0; i < items.length; i++) {
+            if (!_isSdkStatsMetric(items[i])) {
+                var t = _getTelType(items[i]);
+                if (_safeKey(t)) {
```
MSNev: Use the ts-utils [isUnsafePropKey](https://nevware21.github.io/ts-utils/typedoc/functions/isUnsafePropKey.html) instead of duplicating this code and the strings.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,256 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { ITimerHandler, objCreate, objHasOwn, scheduleTimeout } from "@nevware21/ts-utils";
+import { IAppInsightsCore } from "../interfaces/ai/IAppInsightsCore";
+import { INotificationListener } from "../interfaces/ai/INotificationListener";
+import { ITelemetryItem } from "../interfaces/ai/ITelemetryItem";
+import { MetricDataType } from "../telemetry/ai/DataTypes";
+
+var FLUSH_INTERVAL = 900000; // 15 min default
+var MET_SUCCESS = "Item_Success_Count";
```
MSNev, line 10: I still have the same comment about these polluting the users-events and for 1DS (this same code will be used -- or should be), so this needs some flexibility so that for 1DS it's configured to use the "Ms.Web.SdkStat" as the metrics name.

JacksonWeber: Added optional eventName param to the `createSdkStatsNotifCbk`

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: Good callout, updated to pull in the util from the ts-utils

channels/applicationinsights-channel-js/src/Interfaces.ts
JacksonWeber: Updated the version tag

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,268 @@
...
+        var bucket = counters[code];
+        if (!bucket) {
+            bucket = counters[code] = objCreate(null);
+        }
+        var changed = false;
+        for (var i = 0; i < items.length; i++) {
+            if (!_isSdkStatsMetric(items[i])) {
+                var t = _getTelType(items[i]);
+                if (!isUnsafePropKey(t)) {
+                    bucket[t] = (bucket[t] || 0) + 1;
```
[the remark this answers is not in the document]
JacksonWeber: 🤖 _Reply from Copilot_

Addressed this CodeQL `js/prototype-polluting-assignment` alert (#6935) — it is a **false positive** and has been dismissed with that reason. Rationale:

- **Null-prototype targets:** every accumulator written to (`_successCounts`, `_droppedCounts`/`_retryCounts` and each per-code `bucket`) is created with `objCreate(null)`. Assigning to a null-prototype object can never reach `Object.prototype`, so prototype pollution is not possible here.
- **Allow-listed key:** the key `t` comes from `_getTelType()`, which only ever returns a value from the fixed `_typeMap` table (`CUSTOM_EVENT`, `DEPENDENCY`, `EXCEPTION`, …) or the literal `"CUSTOM_EVENT"`. It is never the raw, attacker-influenced `baseType`, so it can never be `__proto__`/`constructor`/`prototype`.
- **Defense in depth:** the assignment is additionally guarded by `!isUnsafePropKey(t)` (and `!isUnsafePropKey(code)` for the bucketed counters) from `@nevware21/ts-utils`.

CodeQL can't track the null-prototype origin through the `counters[code]` indirection nor the allow-list through `objHasOwn`, hence the flag. No functional change was required.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,268 @@
...
+        var evtName = _statsEventName || name;
+        if (_statsEventName) {
+            props.metricName = name;
+        }
+
+        return {
+            name: evtName,
+            baseType: MetricDataType,
+            baseData: {
+                name: evtName,
```
MSNev: This can be the passed `name` (props.metricName) -- it's only the top-level name that will cause the issue.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
MSNev: While this addresses the issue for 1DS, it doesn't address the issue for Application Insights users and their user-events in the custom events / log / metrics table.

I think it's fine for the internal baseData.name to be the defined names, just not the top level event name.

I think the backend is also going to be looking at the passed "name" to determine where / how to route the event (although they might user a combination of the baseType)

JacksonWeber: I see, would unifying under a single dedicated top-level event name to group SDK Stats under work to solve that? Should solve the point about being used to route the event, and would allow all Customer SDK stats to either be filtered in or out based on the top level event name.

JacksonWeber: Change has been implemented.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -0,0 +1,262 @@
...
+import { MetricDataType } from "../telemetry/ai/DataTypes";
+
+var FLUSH_INTERVAL = 900000; // 15 min default
+var MET_SUCCESS = "Item_Success_Count";
+var MET_DROPPED = "Item_Dropped_Count";
+var MET_RETRY = "Item_Retry_Count";
+var DROP_CLIENT_EXCEPTION = "CLIENT_EXCEPTION";
+// Dedicated top-level event name for AI stats. Keeps counter names out of the top-level name so
+// the backend doesn't route stats into the customer's custom events / metrics tables.
+var AI_STATS_EVENT = "Ms.AI.SdkStat";
```
MSNev: I was thinking that it should match something like the main Application Insights events -- so something like
Microsoft.ApplicationInsights.SdkStats?? potentially with the iKey (this want the portal could do something in the future -- not just the workbooks)

MSNev (approved):
Approved with 1 additional suggestion

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
JacksonWeber: I've implemented your suggestion here. Including the iKey component for future Azure portal usage.

MSNev (approved):

JacksonWeber:
@microsoft-github-policy-service rerun

PR #2715, opened 2026-03-07
OTel Web SDK Phase 1

hectorhdzg:
Phase 1: SDK Foundation (Critical)
Component	Location	Description
createOTelWebSdk()	src/otel/sdk/	Main SDK entry point factory
IOTelWebSdkConfig	src/interfaces/otel/config/	SDK configuration interface
Deprecate OTelSdk class	src/otel/sdk/OTelSdk.ts	Remove DynamicProto usage
Deliverable: Functional SDK that can be instantiated with trace+log providers.


shared/otel-core/src/interfaces/otel/config/IOTelWebSdkConfig.ts
```diff
@@ -0,0 +1,136 @@
...
+     * Sampler implementation.
+     * Determines which traces/spans should be recorded and exported.
+     *
+     * @see {@link IOTelSampler}
+     */
+    sampler: IOTelSampler;
+
+    /**
+     * Performance timing function.
+     * Injected for testability — allows tests to control time measurement.
```
MSNev: Is this only used for testing?

If so we should perhaps have a separate "testing" config that extends this one???

MSNev: When using a testing framework (like sinon) they provide the ability to "control" the timings with mocks

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+import { createSpan } from "../api/trace/span";
+import { getContextSpan, setContextSpan } from "../api/trace/utils";
+import { createLoggerProvider } from "./OTelLoggerProvider";
+
+/**
+ * Creates a no-op logger that silently discards all emitted log records.
+ * Used when the SDK has been shut down.
+ * @returns A no-op IOTelLogger instance
+ */
+function _createNoopLogger(): IOTelLogger {
```
MSNev: We should not have any "No-Op" stuff in the SDK

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+ * sdk.shutdown();
+ * ```
+ *
+ * @since 4.0.0
+ */
+export function createOTelWebSdk(config: IOTelWebSdkConfig): IOTelWebSdk {
+    // Validate required dependencies upfront
+    let _handlers: IOTelErrorHandlers = config.errorHandlers;
+
+    if (!config.errorHandlers) {
```
MSNev: This should be using the "defaultConfig" which is passed as the 2nd argument to `createDynamicConfig(config)` which then populates the defaults unless overrwritten by the user.

MSNev: defaults are also required to ensure that EVERY possible config property exists at the point of the call to `createDynamicConfig` so that everything is "dynamic"...

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+            _tracers[key] = _createSdkTracer(tracerName, tracerVersion, schemaUrl);
+        }
+
+        return _tracers[key];
+    };
+
+    _self.getLogger = function (name: string, version?: string, options?: IOTelLoggerOptions): IOTelLogger {
+        if (_isShutdown) {
+            handleWarn(_handlers, "A shutdown OTelWebSdk cannot provide a Logger");
+            return _createNoopLogger();
```
MSNev: We should just return `null` or `undefined` (which I guess changes the "matching" OTel specification), but if we don't we will end up having all of the No-Op code in the distro where most of the time it will never be used.

MSNev: Or (which is what I did for spans) we have a flag like this "isRecording" which we explicitly set to false causing "effectively" a no-op instance but without duplicate code.

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+        }
+
+        return _loggerProvider.getLogger(name, version, options);
+    };
+
+    _self.forceFlush = function (): IPromise<void> {
+        if (_isShutdown) {
+            handleWarn(_handlers, "Cannot force flush a shutdown OTelWebSdk");
+            return createSyncPromise(function (resolve) {
```
MSNev: Small code improvement (minification) there is a [createSyncResolvedPromise](https://nevware21.github.io/ts-async/typedoc/functions/createSyncResolvedPromise.html) which does this already

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+
+        if (operations.length > 0) {
+            return createAllPromise(operations).then(function (): void {
+                // All flushed
+            });
+        }
+
+        return createSyncPromise(function (resolve) {
+            resolve();
+        });
```
MSNev, line 224: The createAllPromise handles dealing with an empty array so you could just have the following instead of both return statements (and all other usages) -- there is also a createSyncAllPromise() version as well.

```ts
  return createAllPromise(operations || []).then(function (): void {
      // All flushed
  });
```

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+                // The sampler may have provided an updated trace state
+                // Note: createDistributedTraceContext handles trace state internally
+                // For now we rely on the context's built-in trace state management
+            }
+
+            // Merge sampler-provided attributes with user-provided attributes
+            let spanAttributes = attributes;
+            if (isRecording && samplingResult.attributes) {
+                // Merge: user attributes take precedence, sampler attributes fill gaps
+                spanAttributes = {};
```
MSNev: We should use the AttributeContainer for this to avoid copying / enumerating all of the attributes unless needed.

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+                newCtx.traceId = _idGenerator.generateTraceId();
+            }
+
+            // Always generate a new spanId for this span
+            newCtx.spanId = _idGenerator.generateSpanId();
+
+            // Run the sampler to decide whether to record this span
+            let attributes = opts.attributes || {};
+            let links = opts.links || [];
+            let samplingResult = _sampler.shouldSample(
```
MSNev: Will there always be a sampler? (no no-op version) in which case this needs to handle null / undefined.

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+            }
+
+            return createSpan(spanCtx, spanName, kind);
+        }
+
+        let tracer: IOTelTracer = setProtoTypeName({
+            startSpan: function (spanName: string, options?: IOTelSpanOptions, context?: IOTelContext): IReadableSpan | null {
+                return _startSpan(spanName, options, context);
+            },
+            startActiveSpan: function <F extends (span: IReadableSpan) => unknown>(
```
MSNev: These should use the helper top level functions (refactored to handle the "context" (We can do this later), to avoid creating "new" function for each element -- and the standalone functions are better for tree-shaking.

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+                }
+
+                // Create the span using the resolved parameters
+                let span = _startSpan(spanNameArg, opts, ctx);
+
+                // Set the span as active in a new context and execute the callback
+                let activeCtx = ctx || _getActiveContext();
+                let contextWithSpan = setContextSpan(activeCtx, span);
+
+                return _contextManager.with(contextWithSpan, function () {
```
MSNev: likewise this this we have "useSpan" and "withSpan" helpers for the same reason.

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+        return tracer;
+    }
+
+    /**
+     * Creates a no-op tracer that does not create any spans.
+     * Used when the SDK has been shut down.
+     *
+     * @returns A no-op IOTelTracer instance
+     */
+    function _createNoopTracer(): IOTelTracer {
```
MSNev: My refactoring to create a otel-noop removed all of this (from memory)

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,496 @@
...
+    return setProtoTypeName(_self, "OTelWebSdk");
+}
+
+/**
+ * Creates a no-op SDK instance that silently discards all operations.
+ * Returned when required dependencies are missing to prevent runtime crashes.
+ * @param config - The original config, used for getConfig()
+ * @returns A safe no-op IOTelWebSdk instance
+ */
+function _createNoopSdk(config: IOTelWebSdkConfig): IOTelWebSdk {
```
MSNev: :-(

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,419 @@
...
+                objDefine(spanCtx, "parentSpanContext", {
+                    v: parentSpanCtx,
+                    w: false
+                });
+            }
+
+            return createSpan(spanCtx, spanName, kind);
+        }
+
+        function _startActiveSpan<F extends (span: IReadableSpan) => unknown>(
```
JacksonWeber, line 373: Looks like `_startSpan` returns null when `_isShutdown` is true, but `_startActiveSpan` passes that null unchecked into `setContextSpan` and the user's callback. Since the callback is typed as (span: IReadableSpan) => unknown, consumers have no type-level warning and will hit a runtime crash (e.g. `span.end()` on null) if they hold a tracer reference obtained before shutdown.

Suggest guarding before the context/callback path:
```typescript
let span = _startSpan(spanNameArg, opts, ctx);
if (!span) {
    return undefined as ReturnType<F>;
}
 
let activeCtx = ctx || _getActiveContext();
let contextWithSpan = setContextSpan(activeCtx, span);
return _contextManager.with(contextWithSpan, fn as (...args: any[]) => ReturnType<F>, undefined, span) as ReturnType<F>;
```

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -0,0 +1,419 @@
...
+const _defaultConfig: IConfigDefaults<IOTelWebSdkConfig> = objDeepFreeze({
+    resource: {
+        isVal: function (v: any) {
+            return !!v;
+        },
+        v: null as any
+    },
+    errorHandlers: cfgDfMerge<IOTelErrorHandlers>({}),
+    contextManager: {
+        isVal: function (v: any) {
```
MSNev, line 55: These are probably not valid "defaults", but we can change these later
- specifically, for the contextManager I thinking that we will want to have the instance of the SDK be the default contextManager (or at the least an internal instance managed by itself) -- we can review this later

MSNev (approved):
There are some possible minor issues, but this is a good start for us to build upon and find the stuff we don't know yet :-)

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
hectorhdzg: Added TODO

JacksonWeber (approved):

PR #2716, opened 2026-03-12
Make URL Redaction more dynamic

rads-1996:


shared/AppInsightsCore/src/interfaces/ai/IConfiguration.ts
```diff
@@ -232,19 +232,34 @@ export interface IConfiguration extends IOTelConfig {
...
+     * [Optional] A flag to enable or disable redaction for username and password in URLs.
      * @defaultValue true
      */
     redactUrls?: boolean;
 
+    /**
+     * [Optional] A flag to enable or disable redaction for query parameters.
+     * @defaultValue true
+     */
+    redactQueryParams?: boolean;
```
rads-1996: I have added an opt in/out option indiviually for username:password and query params. Previously we just had a redactUrls which would completely not redact anything in the url, if set to False

Do we need this level of flexibility?

shared/AppInsightsCore/src/utils/EnvUtils.ts
```diff
@@ -455,8 +455,10 @@ function redactQueryParameters(url: string, config?: IConfiguration): string {
...
     }
 
-    if (config && config.redactQueryParams) {
-        sensitiveParams = DEFAULT_SENSITIVE_PARAMS.concat(config.redactQueryParams);
+    if (config && config.appendRedactQueryParams) {
+        sensitiveParams = DEFAULT_SENSITIVE_PARAMS.concat(config.appendRedactQueryParams);
+    } else if (config && config.replaceRedactQueryParams) {
+        sensitiveParams = config.replaceRedactQueryParams;
     } else {
         sensitiveParams = DEFAULT_SENSITIVE_PARAMS;
```
rads-1996, line 466: The user either gives a `replaceRedactQueryParams` which overwrites the default config or sets `appendRedactQueryParams`. If the user does not set any config then the default config query params are redacted. If the user does not want to redact any of the queryParams, they can pass an empty array to the `replaceRedactQueryParams`

rads-1996:
Will document all the changes once decide all the config changes.

shared/AppInsightsCore/src/interfaces/ai/IConfiguration.ts
MSNev: This seems like too many configs (with long names) which are ( probably ) going to be rarely used.

Suggestion: Introduce a new constant enum and lest change the redactUrls to use that
```ts
redactUrls?: boolean | RedactUrlOption
```

With values like (names are suggestions only so feel free to change them)
- true / false (does what it does today - nothing or use defaults with merged urls)
- AllMerge (same as true)
- AllReplace
- UrlOnly
- UsernamePasswordOnly
- etc

Or use a bitwise value logic for the values
0x00-0x0f (Url)
0x10 add (to Skip Username/Passwword etc)

shared/AppInsightsCore/src/interfaces/ai/IConfiguration.ts
```diff
@@ -232,19 +232,34 @@ export interface IConfiguration extends IOTelConfig {
...
+
     /**
      * [Optional] Additional query parameters to redact beyond the default set.
      * Use this to specify custom parameters that contain sensitive information.
      * These will be combined with the default parameters that are redacted.
      * @defaultValue ["sig", "Signature", "AWSAccessKeyId", "X-Goog-Signature"]
      * @example ["sig", "Signature", "AWSAccessKeyId", "X-Goog-Signature","auth_token", "api_key", "private_data"]
      */
-    redactQueryParams?: string[];
+    appendRedactQueryParams?: string[];
```
MSNev: As we have published the original version, we can't just rename the config values, we would have to go through a multi-year and multi-version process of deprecation. Hence the idea above -- keep this the same and an optional "extension" of the above "flag" the extend the functionality while maintaining backwards compatability.

MSNev (changes requested):

shared/AppInsightsCore/src/interfaces/ai/IConfiguration.ts
rads-1996: Quick question, when the user choose replace/append for query params, the username and password should be redacted as per logic right?

shared/AppInsightsCore/src/enums/ai/UrlRedactionOptions.ts
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Controls how the user can configure which parts of the URL should be redacted. Example, certain query parameters, username and password, etc.
+ * @since <next_release_version>
```
rads-1996: Will replace this with the next version number

shared/AppInsightsCore/src/utils/EnvUtils.ts
```diff
@@ -455,8 +456,12 @@ function redactQueryParameters(url: string, config?: IConfiguration): string {
         return url;
     }
 
-    if (config && config.redactQueryParams) {
+    const option = config ? config.redactUrls : undefined;
```
rads-1996, line 459: I added this check here to avoid having to change the function arguments.

shared/AppInsightsCore/src/enums/ai/UrlRedactionOptions.ts
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Controls how the user can configure which parts of the URL should be redacted. Example, certain query parameters, username and password, etc.
+*/
+
+export const enum UrlRedactionOptions {
```
MSNev: OK, looking good just a couple of comments
- const enum is excellent and it's what we want to use internally (like you have)

minor changes
- the const enum should be named with a leading lowercase "e"
- As this is usable from the `IConfiguration` it also needs to be exported from the index
- Because of the exporting requirement we "should" probably also create a user usable enum using createEnumStyle
- We should also export a type (the name you have now without a leading lowercase "e") and use this in the IConfiguration (so that files doesn't need to change)

Examples of what I'm talking about for the enum definitions are in https://github.com/microsoft/ApplicationInsights-JS/blob/main/shared/AppInsightsCore/src/enums/ai/EventsDiscardedReason.ts

rads-1996: I have addressed the comments. PTAL when you have time. Thanks!

shared/AppInsightsCore/src/utils/EnvUtils.ts
```diff
@@ -455,8 +456,12 @@ function redactQueryParameters(url: string, config?: IConfiguration): string {
...
+    
+    if (option === UrlRedactionOptions.appendToDefault) {
```
MSNev: these ones should be referencing the `eUrlRedactOptions` so that during conversion from TypeScript to JavaScript TS will replace with the "number", by referencing the type the code is not minifyable

shared/AppInsightsCore/src/utils/EnvUtils.ts
```diff
@@ -543,17 +548,30 @@ export function fieldRedaction(input: string, config: IConfiguration): string {
     if (!input || !isString(input) || strIndexOf(input, " ") !== -1) {
         return input;
     }
-    const isRedactionDisabled = config && config.redactUrls === false;
+
+    const option = config ? config.redactUrls : undefined;
+
+    const isRedactionDisabled = option === false || option === UrlRedactionOptions.false;
```
MSNev: Same again use the const enum `eUrlRedactOptions`

MSNev: Add all other references

MSNev (commented):
A couple of minor fixes for minification purposes -- should not affect the functionality

shared/AppInsightsCore/src/utils/EnvUtils.ts
rads-1996: > Add all other references

Did you mean "And all other references"

MSNev (approved):

PR #2717, opened 2026-03-20
Phase 0 implementation

hectorhdzg:
Phase 0: CONTEXT.md Compliance Fixes (Prerequisite)

Fix existing implementations to comply with CONTEXT.md before building new features.

| Component | Changes Required |
|-----------|-----------------|
| `_createTracerProvider` | Rename to `createTracerProvider`, add config object, add error handlers, add `onConfigChange` |
| `createLoggerProvider` | Remove default config, validate required deps, get handlers from config, add `onConfigChange` |
| `createContextManager` | Add config object with error handlers |
| `createSpan` | Add `onConfigChange` if caching config values |
| `createResource` | Add shutdown/cleanup method |
| All | Add `IUnloadHook` management with `.rm()` calls during shutdown |
| All | Add comprehensive TypeDoc documentation |

**Deliverable**: All existing implementations pass CONTEXT.md validation checklist.

### DIFF
--- a/shared/otel-core/src/interfaces/otel/context/IContextManagerConfig.ts
+++ b/shared/otel-core/src/interfaces/otel/context/IContextManagerConfig.ts
@@ -1,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { IOTelErrorHandlers } from "../config/IOTelErrorHandlers";
+import { IOTelContext } from "./IOTelContext";
+
+/**
+ * Configuration interface for creating a context manager.
+ *
+ * @since 4.0.0
+ */
+export interface IContextManagerConfig {
+    /**
+     * The parent / root context to use if there is no active context.
+     */
+    parentContext?: IOTelContext;
+
+    /**
+     * Error handlers for internal diagnostics.
+     *
+     * @see {@link IOTelErrorHandlers}
+     */
+    errorHandlers?: IOTelErrorHandlers;
+}
--- a/shared/otel-core/src/interfaces/otel/context/IContextManagerConfig.ts
+++ b/shared/otel-core/src/interfaces/otel/context/IContextManagerConfig.ts
@@ -1,11 +1,14 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { IOTelErrorHandlers } from "../config/IOTelErrorHandlers";
 import { IOTelContext } from "./IOTelContext";
 
 /**
  * Configuration interface for creating a context manager.
+ *
+ * @remarks
+ * Error handlers are inherited from the SDK/core config (IOTelConfig/IOTelWebSdkConfig)
+ * to avoid duplicating them in each component config.
  *
  * @since 4.0.0
  */
@@ -14,11 +17,4 @@ export interface IContextManagerConfig {
      * The parent / root context to use if there is no active context.
      */
     parentContext?: IOTelContext;
-
-    /**
-     * Error handlers for internal diagnostics.
-     *
-     * @see {@link IOTelErrorHandlers}
-     */
-    errorHandlers?: IOTelErrorHandlers;
 }
--- a/shared/otel-core/src/interfaces/otel/logs/IOTelLoggerProviderConfig.ts
+++ b/shared/otel-core/src/interfaces/otel/logs/IOTelLoggerProviderConfig.ts
@@ -1,13 +1,37 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+import { IOTelErrorHandlers } from "../config/IOTelErrorHandlers";
 import { IOTelResource } from "../resources/IOTelResource";
 import { IOTelLogRecordLimits } from "./IOTelLogRecordLimits";
 import { IOTelLogRecordProcessor } from "./IOTelLogRecordProcessor";
 
+/**
+ * Configuration interface for the OpenTelemetry LoggerProvider.
+ * Provides all configuration options required for LoggerProvider initialization.
+ *
+ * @remarks
+ * - The `resource` and `errorHandlers` properties are required
+ * - Config is used directly — never copied with spread operator
+ * - Supports dynamic configuration via `onConfigChange` callbacks
+ *
+ * @since 3.4.0
+ */
 export interface IOTelLoggerProviderConfig {
-    /** Resource associated with trace telemetry  */
-    resource?: IOTelResource;
+    /**
+     * Resource information for telemetry source identification.
+     * Provides attributes that describe the entity producing telemetry.
+     */
+    resource: IOTelResource;
+
+    /**
+     * Error handlers for internal diagnostics.
+     * Provides hooks to customize how different types of errors and
+     * diagnostic messages are handled.
+     *
+     * @see {@link IOTelErrorHandlers}
+     */
+    errorHandlers: IOTelErrorHandlers;
 
     /**
      * How long the forceFlush can run before it is cancelled.
@@ -15,7 +39,7 @@ export interface IOTelLoggerProviderConfig {
      */
     forceFlushTimeoutMillis?: number;
 
-    /** Log Record Limits*/
+    /** Log Record Limits */
     logRecordLimits?: IOTelLogRecordLimits;
 
     /** Log Record Processors */
--- a/shared/otel-core/src/interfaces/otel/logs/IOTelLoggerProviderConfig.ts
+++ b/shared/otel-core/src/interfaces/otel/logs/IOTelLoggerProviderConfig.ts
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { IOTelErrorHandlers } from "../config/IOTelErrorHandlers";
 import { IOTelResource } from "../resources/IOTelResource";
 import { IOTelLogRecordLimits } from "./IOTelLogRecordLimits";
 import { IOTelLogRecordProcessor } from "./IOTelLogRecordProcessor";
@@ -11,7 +10,8 @@ import { IOTelLogRecordProcessor } from "./IOTelLogRecordProcessor";
  * Provides all configuration options required for LoggerProvider initialization.
  *
  * @remarks
- * - The `resource` and `errorHandlers` properties are required
+ * - The `resource` property is required
+ * - Error handlers are inherited from the SDK/core config (IOTelWebSdkConfig)
  * - Config is used directly — never copied with spread operator
  * - Supports dynamic configuration via `onConfigChange` callbacks
  *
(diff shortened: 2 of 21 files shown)


shared/otel-core/src/otel/sdk/OTelLoggerProvider.ts
```diff
@@ -2,65 +2,109 @@
...
     }
 
-    const sharedState = createLoggerProviderSharedState(
-        resource,
+    let sharedState = createLoggerProviderSharedState(
+        _resource,
         forceFlushTimeoutMillis,
         reconfigureLimits(logRecordLimits),
-        config && config.processors ? config.processors : []
+        config.processors || []
```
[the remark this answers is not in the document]
JacksonWeber: This looks to be a legitimate concern. Should be addressed.

shared/otel-core/src/otel/sdk/OTelLoggerProvider.ts
```diff
@@ -2,65 +2,109 @@
...
+
+    let defaults = loadDefaultConfig();
+    let forceFlushTimeoutMillis: number;
+    let logRecordLimits;
+
+    let _isShutdown = false;
+    let _unloadHooks: IUnloadHook[] = [];
+
+    // Register for config changes — save the returned IUnloadHook
+    let _configUnload = createDynamicConfig(config).watch(function () {
```
JacksonWeber: The `createDynamicConfig().watch()` callback updates `_resource`, `forceFlushTimeoutMillis`, `logRecordLimits`, and `_host`, but `sharedState` is constructed once from initial values and never mutated.

shared/otel-core/src/otel/api/trace/tracerProvider.ts
```diff
@@ -2,38 +2,85 @@
...
     return {
         getTracer(name: string, version?: string): IOTelTracer {
-            const tracerKey = (name|| "ai-web") + "@" + (version || "unknown");
-            
-            if (!tracers[tracerKey]) {
-                tracers[tracerKey] = _createTracer(host);
+            if (_isShutdown) {
+                handleWarn(_handlers, "A shutdown TracerProvider cannot provide a Tracer");
+                return null;
+            }
```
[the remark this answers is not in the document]
JacksonWeber: This concern is also legitimate.

shared/otel-core/src/interfaces/otel/context/IContextManagerConfig.ts
```diff
@@ -0,0 +1,24 @@
...
+     * The parent / root context to use if there is no active context.
+     */
+    parentContext?: IOTelContext;
+
+    /**
+     * Error handlers for internal diagnostics.
+     *
+     * @see {@link IOTelErrorHandlers}
+     */
+    errorHandlers?: IOTelErrorHandlers;
```
MSNev, line 23: I don't think we should have this defined here, the "implementation" should inherit/use the core (Sdk/Distro) config to use the errorHandlers -- which should be provided during initialization...

Otherwise, we could end up passing the same error handlers multiple times

shared/otel-core/src/interfaces/otel/logs/IOTelLoggerProviderConfig.ts
```diff
@@ -1,21 +1,45 @@
...
+    resource: IOTelResource;
+
+    /**
+     * Error handlers for internal diagnostics.
+     * Provides hooks to customize how different types of errors and
+     * diagnostic messages are handled.
+     *
+     * @see {@link IOTelErrorHandlers}
+     */
+    errorHandlers: IOTelErrorHandlers;
```
MSNev, line 34: same comment as context

shared/otel-core/src/interfaces/otel/trace/ITracerProviderConfig.ts
```diff
@@ -0,0 +1,30 @@
...
+     * @see {@link ITraceHost}
+     */
+    host: ITraceHost;
+
+    /**
+     * Error handlers for internal diagnostics.
+     *
+     * @see {@link IOTelErrorHandlers}
+     */
+    errorHandlers?: IOTelErrorHandlers;
```
MSNev, line 29: same comment as context -- and this being defined for context, logs and trace reinforces my original comment

shared/otel-core/src/otel/api/context/contextManager.ts
```diff
@@ -2,16 +2,64 @@
...
+ *
+ * ctxMgr.with(myContext, () => {
+ *   // myContext is now active within this callback
+ * });
+ * ```
+ *
+ * @since 4.0.0
  */
-export function createContextManager(parentContext?: IOTelContext): IOTelContextManager {
+export function createContextManager(config?: IContextManagerConfig): IOTelContextManager {
```
MSNev, line 45: As this is our distro, I'd pass the core sdk config (which is already dynamic)

shared/otel-core/src/otel/api/context/contextManager.ts
```diff
@@ -2,16 +2,64 @@
...
+    let _unloadHooks: IUnloadHook[] = [];
+
+    // Local cached values — updated via onConfigChange
+    let _handlers: IOTelErrorHandlers = {};
+    let parentContext: IOTelContext;
+
+    // Register for config changes — save the returned IUnloadHook
+    // Only set up dynamic config watcher when config is provided to avoid per-instance overhead
+    if (config) {
+        let _configUnload = createDynamicConfig(_cfg).watch(function () {
```
MSNev, line 56: not sure about this, it probably should just be applying defaults (if we want to split the defaults to each component) and then this would just use the onConfigChange helper as it's a single function (some minified) vs the watch not getting minified.

shared/otel-core/src/otel/api/context/contextManager.ts
```diff
@@ -2,16 +2,64 @@
...
+            _handlers = _cfg.errorHandlers || {};
+            parentContext = _cfg.parentContext;
```
MSNev, line 58: Question: For the "parentContext" do we or is it valid to change the parent context after construction -- it really just seems to be (currently) the "fallback" for when there is no "active" or when the current context is disabled?

Based on that I'd almost consider changing this to 

```ts
createContextManager(config?: IOTelConfig, parentContext?: IOTelContextManager): IOTelContextManager 
```

shared/otel-core/src/otel/api/trace/span.ts
```diff
@@ -61,7 +82,7 @@ export function createSpan(spanCtx: IOTelSpanCtx, orgName: string, kind: OTelSpa
 
     function _handleIsEnded(operation: string, extraMsg?: string): boolean {
         if (isEnded) {
-            handleSpanError(errorHandlers, "Span {traceID: " + spanContext.traceId + ", spanId: " + spanContext.spanId + "} has ended - operation [" + operation + "] unsuccessful" + (extraMsg ? (" - " + extraMsg) : STR_EMPTY) + ".", spanName);
+            handleSpanError(_errorHandlers(), "Span {traceID: " + spanContext.traceId + ", spanId: " + spanContext.spanId + "} has ended - operation [" + operation + "] unsuccessful" + (extraMsg ? (" - " + extraMsg) : STR_EMPTY) + ".", spanName);
```
MSNev, line 85: Maybe we should consider updating the `handleErrors.ts` so they could be passed either the `IOTelErrorHandlers` or `IOTelWebSdkConfig` and then we can put the "dereferencing" from the config to the configured error handlers in a single location instead of having the helper dereferencing in all components

shared/otel-core/src/otel/api/trace/tracerProvider.ts
```diff
@@ -2,38 +2,111 @@
...
+    let _tracers: { [key: string]: IOTelTracer } = {};
+    let _isShutdown = false;
+    let _unloadHooks: IUnloadHook[] = [];
+
+    // Local cached values — updated via onConfigChange
+    let _handlers: IOTelErrorHandlers;
+    let _host = config.host;
+
+    // Register for config changes — save the returned IUnloadHook
+    let _configUnload = createDynamicConfig(config).watch(function () {
```
MSNev, line 68: It should use onConfigChange and then it removes the need for "getting" the host first (calling createDynamicConfig by default uses the passed config as a template, so it's possible that the host object returned on line 65 is not the same object on line 70.

shared/otel-core/src/otel/api/trace/tracerProvider.ts
```diff
@@ -2,38 +2,111 @@
...
 
     return {
         getTracer(name: string, version?: string): IOTelTracer {
-            const tracerKey = (name|| "ai-web") + "@" + (version || "unknown");
-            
-            if (!tracers[tracerKey]) {
-                tracers[tracerKey] = _createTracer(host);
+            if (_isShutdown) {
+                handleWarn(_handlers, "A shutdown TracerProvider cannot provide a Tracer");
+                return _NOOP_TRACER;
```
MSNev, line 78: Noooo!!! ;-)

I would prefer that this just return null, and only if we "really" need to support a tracer that does nothing we can add a flag to the _createTracer so it does nothing (as required).

I think the way I did this in the 3.4.0 code base was that it always returns a tracer (which is basically the host itself) and during the "end" processing it then return a span that is not recording etc.

shared/otel-core/src/otel/sdk/OTelLoggerProvider.ts
```diff
@@ -2,65 +2,123 @@
...
+ * });
+ *
+ * const logger = provider.getLogger("my-service", "1.0.0");
+ * ```
+ *
+ * @since 3.4.0
+ */
 export function createLoggerProvider(
-    config: IOTelLoggerProviderConfig = {}
+    config: IOTelLoggerProviderConfig
```
MSNev, line 52: same comment as the context pass the IOTelWebSdkConfig instance and use the applyDefaults() to populate the log provider.

Basically, do we need to support the end-users creating their own LoggerProvider? As I would think we just want them to use ours.

shared/otel-core/src/otel/sdk/OTelLoggerProvider.ts
```diff
@@ -2,65 +2,123 @@
...
         reconfigureLimits(logRecordLimits),
-        config && config.processors ? config.processors : []
+        config.processors || []
     );
 
-    let isShutdown = false;
-    const handlers: IOTelErrorHandlers = {};
+    // Register for config changes — save the returned IUnloadHook
+    // Updates both local cached values and sharedState so dynamic config changes propagate
+    let _configUnload = createDynamicConfig(config).watch(function () {
```
MSNev, line 90: same as previous comments

shared/otel-core/src/otel/sdk/OTelWebSdk.ts
```diff
@@ -149,11 +149,24 @@ export function createOTelWebSdk(config: IOTelWebSdkConfig): IOTelWebSdk {
     // Create a root context for the SDK so that context operations always have a valid base
     let _rootContext: IOTelContext = createContext(_apiAdapter);
 
-    // Create the logger provider using existing factory
-    let _loggerProvider = createLoggerProvider({
+    // Create a shared config object for the logger provider that gets updated
+    // when the SDK config changes, so the provider picks up new values.
+    let _loggerProviderConfig = {
```
MSNev, line 154: I think that this should just be part of the shared Sdk config instead of as an internal reference

PR #2719, opened 2026-03-31
Improve OsPlugin reliability: proactive OS retrieval, unload handling, and session caching

MSNev:
- Refactor OsPlugin to start OS version retrieval during initialize() instead of waiting for the first telemetry event, reducing latency for the initial tracked events
- Add page unload/pagehide/visibilitychange event handlers to flush the queued telemetry when the page is being unloaded before the OS lookup completes
- Respect disableFlushOnUnload config to skip registering unload handlers when disabled
- Properly clean up unload event handlers after OS version is resolved
- Improve session storage caching: validate cached OS data on load and guard writes behind isStorageUseDisabled config check
- Use safeGetLogger and getNavigator() instead of direct navigator/core.logger access for safer operation when core may not be fully initialized
- Fix safeGetLogger to fall back to core.config when no explicit config is provided

### DIFF
--- a/.aiAutoMinify.json
+++ b/.aiAutoMinify.json
@@ -26,6 +26,7 @@
                 "TelemetryUnloadReason",
                 "TelemetryUpdateReason",
                 "eTraceHeadersMode",
+                "eUrlRedactionOptions",
                 "eValueKind",
                 "EventLatencyValue",
                 "eEventPropertyType",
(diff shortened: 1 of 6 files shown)


rads-1996 (approved):
LGTM

JacksonWeber (approved):

PR #2720, opened 2026-04-06
[Release] Increase version to 3.4.1

MSNev:
- Update API documentation to highlight the deprecation (as when published this comes into effect), ensuring to call out the knwo required changes
- Add copilot release instructions and bump the version to the previously released beta release version to prepare for release as we are skipping 3.4.0 and 4.4.0 as the primary and will only be releasing those versions as beta releases


docs/API-reference.md
```diff
@@ -0,0 +1,414 @@
+### trackPageView
```
MSNev, line 1: This file is moved from the root of the repo to the docs folder and updated as it referenced a lot of the items that are now deprecated.

hectorhdzg (approved):

PR #2735, opened 2026-05-19
fix: prevent prototype pollution in extend() and objExtend() via __proto__ key filtering

hectorhdzg:
Both extend() in extUtils.ts and objExtend() in HelperFuncs.ts were vulnerable to prototype pollution when processing untrusted input containing __proto__ keys (e.g. from JSON.parse). In deep extend mode, target['__proto__'] resolves to Object.prototype via the getter, causing recursive merge to directly mutate Object.prototype and pollute all objects globally.

- Add __proto__ key guard in extend() objForEachKey callback

- Add __proto__ key guard in objExtend() for...in loop

- Add prototype pollution unit tests for both functions

### DIFF
--- a/shared/AppInsightsCore/src/ext/extUtils.ts
+++ b/shared/AppInsightsCore/src/ext/extUtils.ts
@@ -6,7 +6,7 @@
 */
 import {
     arrForEach, getInst as getGlobalInst, getNavigator, hasDocument, hasWindow, isArray, isBoolean, isNullOrUndefined, isNumber, isObject,
-    isString, isUndefined, objForEachKey, perfNow, strIndexOf, strLeft
+    isString, isUndefined, isUnsafePropKey, objForEachKey, perfNow, strIndexOf, strLeft
 } from "@nevware21/ts-utils";
 import { STR_EMPTY } from "../constants/InternalConstants";
 import { EventLatency, EventLatencyValue, FieldValueSanitizerType, GuidStyle, eEventPropertyType, eValueKind } from "../enums/ext/Enums";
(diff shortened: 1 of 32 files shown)


shared/AppInsightsCore/src/ext/extUtils.ts
```diff
@@ -283,6 +283,11 @@ export function extend(obj?: any, obj2?: any, obj3?: any, obj4?: any, obj5?: any
     for (; i < length; i++) {
         var obj = theArgs[i];
         objForEachKey(obj, (prop, value) => {
+            // Prevent prototype pollution by skipping unsafe keys
+            if (prop === "__proto__" || prop === "constructor" || prop === "prototype") {
```
MSNev: Should use `objForEachKeySafe` or the new `isUnsafePropKey` helper instead of repeating the "__proto__" and other strings multiple times in the code base

shared/AppInsightsCore/src/utils/HelperFuncs.ts
```diff
@@ -415,6 +415,11 @@ export function objExtend<T1, T2, T3, T4, T5, T6>(obj1?: T1 | any, obj2?: T2, ob
         let isArgArray = isArray(arg);
         let isArgObj = isObject(arg);
         for (let prop in arg) {
+            // Prevent prototype pollution by skipping unsafe keys
+            if (prop === "__proto__" || prop === "constructor" || prop === "prototype") {
```
MSNev: Should use the new `isUnsafePropKey` helper instead of repeating the "__proto__" and other strings multiple times in the code base

MSNev (approved):

PR #2741, opened 2026-06-17
Enable SDK Stats and route to the SDK Stats ingestion endpoint

JacksonWeber:
## Summary

Re-enables the **SDK Stats** manager (`createStatsMgr`) and routes its events to the distro-owned SDK Stats ingestion endpoint used by [opentelemetry-distro-dotnet](https://github.com/microsoft/opentelemetry-distro-dotnet), instead of piggy-backing on the customer's breeze endpoint.

### Endpoints (from `Azure/azure-sdk-for-net` `StatsbeatConstants.cs`
- Non-EU -> `https://stats.monitor.azure.com/v2/track`
- EU -> `https://eu.stats.monitor.azure.com/v2/track`
- Placeholder iKey `00000000-0000-0000-0000-000000000000`, same Breeze envelope, no auth. EU vs non-EU chosen from the customer endpoint region.

### Configurable Destination (new `stats.monitor` endpoint vs. legacy breeze)

The SDK Stats destination is now selectable at runtime via the CDN / dynamic config:

- New config field `config.stats` (`IStatsBeatConfig`), overridable via the CDN / cfgSync dynamic config.
- `config.stats.mode` (`eStatsEndpointType`): `SdkStats` (default) sends to `stats.monitor.azure.com` / `eu.stats.monitor.azure.com`; `Breeze` sends to the legacy breeze endpoint (the customer's own breeze host, using the Microsoft-owned SDK Stats iKey `c4a29126…` / `7dc56bab…` by region).
- The destination iKey and ingestion endpoint are resolved per-event in `_track` from the live config, so CDN changes take effect at runtime. A full `endCfg` key map (with explicit `key` and new `url`) can also be supplied via `config.stats.endCfg` to fully override routing.

Note: in `Breeze` mode the events flow through the normal channel to the customer's breeze endpoint (legacy behaviour); in `SdkStats` mode they bypass the customer buffer and are excluded from request counting.

### Changes
- **`StatsBeat.ts`** — SDK Stats endpoint constants, EU/non-EU region detection (`getStatsEndpoint`), per-event destination + placeholder-iKey stamping, **enabled-by-default** feature gate (opt-out via `featureOptIn: "sdkStats"`), `createSdkStatsMgrConfig()` helper, and configurable mode + key map support.
- **`AppInsightsCore.ts` / `IAppInsightsCore.ts` / `index.ts`** — restore `getStatsBeat`/`setStatsMgr`, fields, unload cleanup, stubs, interface methods, and exports.
- **`Sender.ts`** — restore request-counting hooks + `statsBeatData`; redirect SDK Stats items to the SDK Stats endpoint via a per-item URL override on `_doSend`/`_getPayload`, bypassing the customer buffer; SDK Stats sends are excluded from counting (no feedback loop); optimized property access for minification efficiency.
- **`AISku.ts`** — create + init + set the manager after core init (kept out of core/AISKULight for bundle size).
- **`IConfiguration.ts`** — added `stats?: IStatsBeatConfig` field for dynamic config.
- **`StatsType.ts`** — added `eStatsEndpointType` enum (`SdkStats` / `Breeze`).
- **`IStatsBeat.ts`** — added `IStatsBeatConfig.mode` and `IStatsBeatKeyMap.url` fields.
- **Tests** — re-enabled all SDK Stats unit tests (`StatsBeat.Tests.ts` in both core and channel-js), fixed import paths, updated terminology from "statsbeat" to "SDK Stats" in test names and assertions.

### Architecture note
`createStatsMgr` lives in core, which cannot depend on the channel's Breeze `EnvelopeCreator`. So SDK Stats events flow through `_core.track()` (reusing the channel's envelope creation) and the redirect to the SDK Stats endpoint happens in the **Sender** via a per-item endpoint override.

### Verification
- Verified with `tsc --noEmit` from source using the repo's exact compiler settings — **core, channel, and AISKU compile with 0 errors**.
- Reviewed for minification efficiency — no spread operators, optional chaining, or nullish coalescing; optimized repeated property access.
- Unit tests uncommented and updated with proper terminology.

### Follow-ups (left for review)
- Update bundle-size test thresholds (`AISKULightSize`, `AppInsightsCoreSize`) if needed.
- Run `npm run lint-fix` + `npm run fullCleanBuild` before marking ready (requires environment outside VS Code due to Rush/ESLint plugin version conflicts).

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


channels/applicationinsights-channel-js/Tests/Unit/src/StatsBeat.tests.ts
```diff
@@ -1,322 +1,322 @@
...
-//             feature: "StatsBeat",
-//             getCfg: (core, cfg) => {
-//                 return cfg?._sdk?.stats;
-//             }
-//         });
+        let statsMgr = createStatsMgr();
+        // Initialize
+        let unloadHook = statsMgr.init(this._core, {
+            feature: "StatsBeat",
+            getCfg: (core, cfg) => {
```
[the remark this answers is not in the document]
JacksonWeber: Fixed in a56505dc. `initializeCoreAndSender()` now initializes the *same* `core` first and then calls `statsMgr.init(core, ...)` against that initialized instance (previously it init'd against `this._core` before `core.initialize()`, leaving the manager disabled). The stats `endCfg` match list now uses the Sender's `config.endpointUrl` instead of the hard-coded `https://example.endpoint.com`, so the endpoint matches and metrics are tracked. An `IStatsBeatState` (with the same endpoint) is now passed to `getStatsBeat()` when wiring the count spy, so the spy wraps the same cached instance the Sender uses.

channels/applicationinsights-channel-js/Tests/Unit/src/StatsBeat.tests.ts
```diff
@@ -1,322 +1,322 @@
...
+        name: "SDK Stats increments success count for xhr sender",
+        useFakeTimers: true,
+        useFakeServer: true,
+        stepDelay: 100,
+        fakeServerAutoRespond: true,
+        steps: [
+            () => {
+                let window = getWindow();
+                let fakeXMLHttpRequest = (window as any).XMLHttpRequest; // why we do this?
+                let config = this.createSenderConfig(TransportType.Xhr) && {disableSendBeaconSplit: true};
```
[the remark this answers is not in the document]
JacksonWeber: Fixed in a56505dc. The config is now built without the `&&`, which had discarded the Sender config and evaluated to just `{ disableSendBeaconSplit: true }` (dropping `endpointUrl` and the other required fields):

```ts
let config: any = this.createSenderConfig(TransportType.Xhr);
config.disableSendBeaconSplit = true;
```

shared/AppInsightsCore/Tests/Unit/src/ai/StatsBeat.Tests.ts
```diff
@@ -1,366 +1,366 @@
...
+            }
+        });
+
+        this.testCase({
+            name: "SDK Stats: test dynamic configuration changes",
+            useFakeTimers: true,
+            test: () => {
+                // Setup core with statsbeat enabled
+                this._core.initialize(this._config, [new ChannelPlugin()]);
+                // Initialize SDK Stats manager for a specific endpoint
```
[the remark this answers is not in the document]
JacksonWeber: Fixed in a56505dc. Since the core is now initialized in `testInitialize()`, this test guards the initialization so it won't initialize a second time:

```ts
if (!this._core.isInitialized()) {
    this._core.initialize(this._config, [new ChannelPlugin()]);
}
```

shared/AppInsightsCore/Tests/Unit/src/ai/InternalSdkStats.Tests.ts
```diff
@@ -1,366 +1,366 @@
...
-//     public registerTests() {
-
-//         this.testCase({
-//             name: "StatsBeat: Initialization",
-//             test: () => {
-//                 // Test with no initialization
-//                 Assert.equal(false, this._statsMgr.enabled, "StatsBeat should not be initialized by default");
+        // Create spy for tracking telemetry
+        _self._trackSpy = this.sandbox.spy(_self._core, "track");
+    }
```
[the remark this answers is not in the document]
JacksonWeber: Fixed in a56505dc. `testInitialize()` now initializes `_core` once (using `_config` plus a minimal `ChannelPlugin`) before the tests run, so `createStatsMgr().init()` can hook the config change and enable the manager. The "Initialization" test still asserts the manager is disabled before `init()` is called and enabled afterwards. I also added `featureOptIn["StatsBeat"]` to `_config` so the dynamic-config test can toggle the mode without hitting an undefined `featureOptIn`.

shared/AppInsightsCore/src/core/StatsBeat.ts
```diff
@@ -424,3 +542,39 @@ export function createStatsMgr(): IStatsMgr {
...
+ * Create the default {@link IStatsMgrConfig} used to enable the SDK Stats collection. By default the
+ * resulting events are routed to the distro-owned SDK Stats ingestion endpoint
+ * (`stats.monitor.azure.com` / `eu.stats.monitor.azure.com`). The destination can be changed at
+ * runtime via the CDN / dynamic config by setting `config.stats` (an {@link IStatsBeatConfig}) - for
+ * example setting `config.stats.mode` to {@link eStatsEndpointType.Breeze} routes SDK Stats to the
+ * legacy breeze endpoint instead. SDK Stats are enabled by default and can be opted-out using the
+ * `featureOptIn` configuration with the {@link STATS_SDK_FEATURE} name.
+ * @returns The {@link IStatsMgrConfig} to pass to {@link IStatsMgr.init}.
+ */
+export function createSdkStatsMgrConfig<CfgType extends IConfiguration = IConfiguration>(): IStatsMgrConfig<CfgType> {
```
MSNev: This breaks dynamic config support, this should only reference the single global config where the "Sku's" (AISKU and 1DS versions) can set their own defaults

JacksonWeber: Fixed this to use a live reference to the global config so we don't become detached from the dynamic config's state.

shared/AppInsightsCore/src/core/StatsBeat.ts
```diff
@@ -4,25 +4,130 @@
...
+ */
+export const STATS_SDK_IKEY = "00000000-0000-0000-0000-000000000000";
+
+/**
+ * Distro-owned SDK statistics ingestion endpoints. SDK Stats events are routed here instead of
+ * to the customer's breeze endpoint. The EU endpoint is used when the customer's endpoint maps
+ * to an EU data-boundary region, otherwise the non-EU endpoint is used.
+ * @see https://github.com/microsoft/opentelemetry-distro-dotnet
+ */
+export const STATS_SDK_ENDPOINT_NON_EU = "https://stats.monitor.azure.com/v2/track";
```
MSNev: It's not stats.monitor.azure.com any more -- we removed that. It's now `data.stats.monitor.azure.com`.
And we should not be embedding this url directly, you should be reading from 
- https://data.stats.monitor.azure.com/cfg/v1.json and
- https://eu-data.stats.monitor.azure.com/cfg/v1.json
To read the json file which includes
```json
{
  "ver": 1,
  "enabled": false,
  "url": "eu-data.stats.monitor.azure.com"
}
```
And check the `enabled` flag, and if `enabled` is true then enable the component and if it's not then the feature should be disabled.
Also when enabled you should use the provided `url` as the domain for the endpoint to send the requests (as this could be redirected to breeze instead of ourselves) so
`https://data.stats.monitor.azure.com/v2/track` (the /v2/track should be coded and I believe there is a constant for this already.

And for 1ds this would become `https://data.stats.monitor.azure.com/OneCollector/1.0` -- although, we probably should have either another config, or add fields to this config so that the 1ds SdkStats can be independently enabled / disabled and have it's own url... Especially as the "iKey" to use will be different for 1DS and AppInsights.

shared/AppInsightsCore/src/core/StatsBeat.ts
```diff
@@ -4,25 +4,130 @@
...
 import { _throwInternal, safeGetLogger } from "../diagnostics/DiagnosticLogger";
 import { _eInternalMessageId, eLoggingSeverity } from "../enums/ai/LoggingEnums";
-import { eStatsType } from "../enums/ai/StatsType";
+import { eStatsEndpointType, eStatsType } from "../enums/ai/StatsType";
 import { IAppInsightsCore } from "../interfaces/ai/IAppInsightsCore";
 import { IConfiguration } from "../interfaces/ai/IConfiguration";
 import { INetworkStatsbeat } from "../interfaces/ai/INetworkStatsbeat";
-import { IStatsBeat, IStatsBeatConfig, IStatsBeatState, IStatsEndpointConfig } from "../interfaces/ai/IStatsBeat";
-import { IStatsMgr, IStatsMgrConfig } from "../interfaces/ai/IStatsMgr";
+import { IStatsBeat, IStatsBeatConfig, IStatsBeatKeyMap, IStatsBeatState, IStatsEndpointConfig } from "../interfaces/ai/IStatsBeat";
```
MSNev: We should rename all of the "StatsBeat" to SdkStats something that represents that this is the internal SdkStats (as we already have the customer SdkStats using sdkStats...

shared/AppInsightsCore/src/core/StatsBeat.ts
```diff
@@ -4,25 +4,130 @@
...
+
+/**
+ * Determine the distro-owned SDK Stats ingestion endpoint for the provided customer endpoint.
+ * When the region maps to an EU region the EU endpoint is returned, otherwise (including unknown
+ * regions) the non-EU endpoint is returned.
+ * @param endpoint - The customer breeze endpoint that the SDK Stats are being collected for.
+ * @returns The SDK Stats ingestion endpoint URL.
+ */
+export function getStatsEndpoint(endpoint: string): string {
+    return _isEuEndpoint(endpoint) ? STATS_SDK_ENDPOINT_EU : STATS_SDK_ENDPOINT_NON_EU;
```
MSNev: This should be using the cfg file to determine whether it's enabled and where to send it.

shared/AppInsightsCore/src/core/StatsBeat.ts
```diff
@@ -349,16 +456,18 @@ export function createStatsMgr(): IStatsMgr {
...
-                    // Call the getCfg function to get the latest configuration for the statsbeat instance
-                    // This should also evaluate the throttling level and other settings for the statsbeat instance
-                    // to determine if it should be enabled or not.
-                    _statsBeatConfig = statsConfig.getCfg(core, details.cfg);
+                if (isFeatureEnabled(featureName || STATS_SDK_FEATURE, details.cfg, true) === true) {
+                    // Seed the SDK Stats defaults into the single global config so they remain dynamic and
+                    // can be overridden via the CDN / dynamic config or by the SKU, then reference the live
+                    // nested stats config so the dependency is registered and runtime changes are picked up.
+                    details.setDf(details.cfg, _sdkStatsDefaults as IConfigDefaults<CfgType>);
+                    _statsBeatConfig = details.ref<CfgType, IStatsBeatConfig>(details.cfg, "stats");
```
MSNev: Lets remove the `_statsBeatConfig` and just assigned local variables, that way we avoid the `_statsBeatConfig.xxxxx` (where the xxxx becomes unminifable), and then we also don't need to do this `.ref()` hack (which actually has side effects -- that we probably don't want)

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -349,16 +456,18 @@ export function createStatsMgr(): IStatsMgr {
...
-                if (statsConfig && isFeatureEnabled(statsConfig.feature, details.cfg, false) === true) {
-                    // Call the getCfg function to get the latest configuration for the statsbeat instance
-                    // This should also evaluate the throttling level and other settings for the statsbeat instance
-                    // to determine if it should be enabled or not.
-                    _statsBeatConfig = statsConfig.getCfg(core, details.cfg);
+                if (isFeatureEnabled(featureName || STATS_SDK_FEATURE, details.cfg, true) === true) {
+                    // Seed the SDK Stats defaults into the single global config so they remain dynamic and
+                    // can be overridden via the CDN / dynamic config or by the SKU, then reference the live
+                    // nested stats config so the dependency is registered and runtime changes are picked up.
+                    details.setDf(details.cfg, _sdkStatsDefaults as IConfigDefaults<CfgType>);
```
MSNev: You "shouldn't" need to do the cast (as IConfigDefaults<CfgType>), if you do we are probably missing somethings that we should address.

shared/AppInsightsCore/src/interfaces/ai/IInternalSdkStats.ts
```diff
@@ -80,6 +80,12 @@ export interface IStatsBeatKeyMap {
      */
     key?: string;
 
+    /**
+     * The SDK Stats ingestion endpoint URL that matching events should be redirected to. When
+     * omitted, matching events are sent to the customer's configured (breeze) endpoint instead.
+     */
+    url?: string;
```
MSNev: Further on my ramblings...
This is the stuff that I don't think we need anymore with the new endpoint....
Specifically, `IStatsEndpointConfig`, `IStatsBeatKeyMap` and `IStatsBeatConfig`...
These really existed so that when we fetch the CDN config we can configured the combination of key -> endpoints etc, so that they get resolved correctly (as every instance would get these and select where to send the old StatsBeat data). But as we are now fetching the config from the `data.stats.monitor.azure.com` endpoint separately for this, I'm thinking we can dump all of this and any code that existed

shared/AppInsightsCore/src/core/AppInsightsCore.ts
```diff
@@ -626,47 +626,47 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
...
-            //         // Disable any previously created statsbeat instance
-            //         if (_statsBeat) {
-            //             _statsBeat.enabled = false;
-            //             _statsBeat = null;
-            //         }
-            //     }
-
-            //     _statsMgr = statsMgr;
-            // };
+            _self.getStatsBeat = (statsBeatState: IStatsBeatState) => {
```
MSNev: This probably just needs to be renamed to not call it StatsBeat anymore.

shared/AppInsightsCore/src/core/AppInsightsCore.ts
```diff
@@ -1736,14 +1736,14 @@ export class AppInsightsCore<CfgType extends IConfiguration = IConfiguration> im
         return null;
     }
 
-    // public getStatsBeat(statsBeatState: IStatsBeatState): IStatsBeat {
-    //     // @ DynamicProtoStub -- DO NOT add any code as this will be removed during packaging
-    //     return null;
-    // }
+    public getStatsBeat(statsBeatState: IStatsBeatState): IStatsBeat {
```
MSNev: general comment, the `type` field in `IStatsBeatState` with the `StatsType` and `sStatsType` was originally created to unify the way customer and internal stats are handled where the "connectionString" (really the `endpoint`) might change so that it could keep each set of network stats isolated based on the endpoint it was attempting to send to when it got the errors.

This was to support
- dynamicConfig (user changes their connectionString and therefore the endpoint)
- We changed the ikey -> endpoint mapping in the config (which I'm thinking we shoulc just remove -- other comment)
- The endpoint (breeze) sends a redirect request, so they send to endpoint 1 and breeze redirects to endpoint 2 etc...

I think this is probably overkill now with our new specific config and the customer stats has already been implemented differently, than this design

shared/AppInsightsCore/src/core/AppInsightsCore.ts
MSNev: So the returned `IStatsBeat` was / is supposed to be the collection of the actual reported "Stats" (I've not checked this PR for the usage of this yet), this was supposed to then handle customer and internal stats in the same way -- but we now have the `createSdkStatsNotifCbk` handle the customer ones internally -- so this might not be needed either (depending how you have the internal ones being counted).

shared/AppInsightsCore/src/core/StatsBeat.ts
JacksonWeber: • Endpoints updated to `data.stats.monitor.azure.com` / `eu-data.stats.monitor.azure.com`
• No longer embedding the ingestion URL. I now fetch `cfg/v1.json` from those hosts, read the `enabled` flag, and gate collection on it.
• When enabled I build the endpoint from the returned `url` plus the existing `DEFAULT_BREEZE_PATH` which covers your point about it being possibly redirected to breeze.
• Added a `STATS_SDK_ONECOLLECTOR_PATH` constant as scaffolding but deferred the full 1DS-specific config to a follow-up. Let me know if this is acceptable.

shared/AppInsightsCore/src/core/StatsBeat.ts
JacksonWeber: Changed to InternalSdkStats. Doesn't look like it impacted bundle sizes negatively, but let me know what you think.

shared/AppInsightsCore/src/core/StatsBeat.ts
JacksonWeber: Updated to pull this from cfg.

shared/AppInsightsCore/src/core/StatsBeat.ts
JacksonWeber: Removed `_statsBeatConfig` and replced with minifiable locals.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Fixed the typing and removed the cast.

shared/AppInsightsCore/src/interfaces/ai/IInternalSdkStats.ts
JacksonWeber: Removed the interfaces, routing logic for the breeze endpoint routing, etc. as a part of addressing above comments being addressed. Cleaning up dead comments too.

shared/AppInsightsCore/src/core/AppInsightsCore.ts
JacksonWeber: I believe these are both still needed, and `getStatsBeat` has been renamed to match the rest of the project changes.

shared/AppInsightsCore/src/core/AppInsightsCore.ts
JacksonWeber: Makes sense, I've cleaned up the dead code related to this.

shared/AppInsightsCore/src/core/StatsBeat.ts
```diff
@@ -299,37 +406,37 @@ function _getEndpointCfg(statsBeatConfig: IStatsBeatConfig, type: eStatsType): I
...
  */
-function _getIKey(endpointCfg: IStatsEndpointConfig, endpoint: string): string | null {
-    let statsKey: string = null;
+function _getKeyMap(endpointCfg: IStatsEndpointConfig, endpoint: string): IStatsBeatKeyMap | null {
+    let matched: IStatsBeatKeyMap = null;
     if (endpointCfg.keyMap) {
         arrForEach(endpointCfg.keyMap, (keyMap) => {
             if (keyMap.match) {
                 arrForEach(keyMap.match, (url) => {
                     if (_isMatchEndpoint(url, endpoint)) {
```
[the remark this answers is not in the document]
JacksonWeber: Resolved in previous changes.

shared/AppInsightsCore/Tests/Unit/src/ai/StatsBeat.Tests.ts
```diff
@@ -1,366 +1,343 @@
...
+
+        // Create spy for tracking telemetry
+        _self._trackSpy = this.sandbox.spy(_self._core, "track");
+    }
+
+    public testCleanup() {
+        super.testCleanup();
+        this._core = null as any;
+        this._statsMgr = null as any;
+    }
```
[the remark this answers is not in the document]
JacksonWeber: Updated.

shared/AppInsightsCore/src/core/SdkStatsNotificationCbk.ts
```diff
@@ -109,29 +101,34 @@
...
-        }
+    // Iterates items (skipping our own emitted stats) and increments target[telType]; returns true if any counted.
+    function _incCounts(target: { [telType: string]: number }, items: ITelemetryItem[]): boolean {
         var changed = false;
         for (var i = 0; i < items.length; i++) {
             if (!_isSdkStatsMetric(items[i])) {
                 var t = _getTelType(items[i]);
                 if (!isUnsafePropKey(t)) {
-                    _successCounts[t] = (_successCounts[t] || 0) + 1;
+                    target[t] = (target[t] || 0) + 1;
```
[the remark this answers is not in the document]
JacksonWeber: Updates should resolve this warning.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -0,0 +1,565 @@
...
+ * server-side. This matches the convention used by the Microsoft OpenTelemetry distros.
+ */
+export const STATS_SDK_IKEY = "00000000-0000-0000-0000-000000000000";
+
+/**
+ * SDK Stats config endpoints (`cfg/v1.json`). The `{ ver, enabled, url }` JSON is read at runtime to
+ * gate collection and resolve the ingestion host. EU endpoint is used for EU data-boundary regions.
+ */
+export const STATS_SDK_CFG_URL_NON_EU = "https://data.stats.monitor.azure.com/cfg/v1.json";
+export const STATS_SDK_CFG_URL_EU = "https://eu-data.stats.monitor.azure.com/cfg/v1.json";
```
MSNev: We should minimize the duplication here as strings can't be compressed.
We should have a 
const scheme = "https://";
const baseUrl = "data.stats.monitor.azure.com/cfg/v1.json";
and then create the above from these :-)

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -0,0 +1,565 @@
...
+const STATS_COLLECTION_SHORT_INTERVAL: number = 900000; // 15 minutes
+const STATS_MIN_INTERVAL_SECONDS = 60; // 1 minute
+const STATS_LANGUAGE = "JavaScript";
+const STATS_TYPE = "Browser";
+
+/**
+ * The placeholder instrumentation key used when reporting SDK statistics to the distro-owned
+ * SDK Stats ingestion endpoint. The endpoint does not require authentication, the placeholder
+ * key only satisfies the connection-string / envelope iKey requirement and is ignored
+ * server-side. This matches the convention used by the Microsoft OpenTelemetry distros.
```
MSNev: "ignored server-side" is not always true.

If the config returns breeze as the URL then no stats will be collected, so we should not send any telemetry (regardless of the backend) without a instrumentation key -- this can be in the config and we can have dynamic config provide / replace the value.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -0,0 +1,565 @@
...
+import { IStatsMgr } from "../interfaces/ai/IStatsMgr";
+import { ITelemetryItem } from "../interfaces/ai/ITelemetryItem";
+import { IPayloadData } from "../interfaces/ai/IXHROverride";
+import { IConfigDefaults } from "../interfaces/config/IConfigDefaults";
+import { MetricDataType } from "../telemetry/ai/DataTypes";
+import { getJSON, isFetchSupported, isXhrSupported } from "../utils/EnvUtils";
+import { getResponseText, isFeatureEnabled, openXhr } from "../utils/HelperFuncs";
+
+const STATS_COLLECTION_SHORT_INTERVAL: number = 900000; // 15 minutes
+const STATS_MIN_INTERVAL_SECONDS = 60; // 1 minute
```
MSNev: Do we really want to send the events every minute (assuming the SDK is actively sending data)?
I believe that .Net only sends every 30/60 minutes...
Which means we would need to cache this in session / local storage or something.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -0,0 +1,565 @@
...
+export const STATS_SDK_ENDPOINT_KEY = "_sdkStatsEndpoint";
+
+/**
+ * The default feature name used to gate the SDK Stats manager. SDK Stats is enabled by default
+ * and can be opted-out via the featureOptIn configuration using this name.
+ */
+export const STATS_SDK_FEATURE = "sdkStats";
+
+// EU data-boundary regions, mirrors the EU region set used by the Azure Monitor OpenTelemetry exporter
+const STATS_EU_REGIONS = [
```
MSNev: We can probably reduce this eg.
"france" vs "francecentral", "francesouth" etc.

or we extract the duplicate portions "south", "central", "east", "west" and the countries and then create the merge strings within the array.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -0,0 +1,565 @@
...
+    }
+
+    return isEU;
+}
+
+/**
+ * Returns the SDK Stats config URL (`cfg/v1.json`) for the endpoint (EU vs non-EU).
+ */
+export function getStatsCfgUrl(endpoint: string): string {
+    return _isEuEndpoint(endpoint) ? STATS_SDK_CFG_URL_EU : STATS_SDK_CFG_URL_NON_EU;
```
MSNev: We should cache the "endpoint" resolution to avoid running the _isEuEndpoint for every time `_track` is called.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -0,0 +1,565 @@
...
+            // Add extra properties
+            let baseProperties = {
+                "rp": "unknown",
+                "attach": "Manual",
+                "cikey": internalSdkStatsStats.cKey,
+                "os": STATS_TYPE,
+                "language": STATS_LANGUAGE,
+                "version": internalSdkStatsStats.sdkVer || "unknown",
+                "endpoint": "breeze",
+                "host": _networkCounter.host
```
MSNev, line 375: Some additional properties that are web specific that would be good to have reported as part of the SDK Stats is whether the SDK was loaded via the snippet and what version (` ai.context.internal.snippetVer``) and if it was loaded via CDN or npm (`ai.context.internal.sdkSrc` <- the CDN name it was loaded from) Set via the `_findSdkSourceFile()` in AISku.ts (not currently setfor 1DS) -- we just have not have anywhere to report this before

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -0,0 +1,565 @@
...
+            }
+
+            internalSdkStatsEvent.iKey = STATS_SDK_IKEY;
+            // Carry the SDK Stats ingestion endpoint so the sending channel can redirect the event
+            // away from the customer's breeze endpoint. This marker is removed by the channel before
+            // the event is serialized.
+            internalSdkStatsEvent.data = internalSdkStatsEvent.data || {};
+            internalSdkStatsEvent.data[STATS_SDK_ENDPOINT_KEY] = url;
+
+            _core.track(internalSdkStatsEvent);
```
MSNev: I don't think this is going to work??

As it looks like you are passing in the core from the AISku instance (which is the users configured instance), which will be batched and sent to the users endpoint (channel control).

It also means that when the end-users have their own telemetry initializers attached to the core to add / remove properties they may add / remove PII to the internal SDK events, also if they have not written their initializer correctly (they expect fields / properties that don't exist for a metric (the sdk stats event) it could cause an unhandled JavaScript exception. And I've seen some sites throw a complete page load failure error (causing their site to be declared "offline") on any unhandled exception.

While we need the root `_core` instance to monitor it for any changed config (loaded via the dynamic config or users changes their endpoint url), we should create instance and initialize our own instance of the SDK configured with the cfg.json endpoint.

which means that the `init` function should have an additional callback to support the creating of the SDK Instance (as not everyone uses core, properties and sender), so that for the main operation (SKU) we can control what is included (so AISku Lite there is no properties, for 1DS there is no Sender or AI Properties, so the SKU's in 1DS have to initialize themselves) etc.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Let's discuss this further, my understanding was that we can't expect the session to live that long so lowering the timing threshold would give us a better chance of collecting telemetry. Let's discuss further in the meeting Monday.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: That's a great idea. Happy to brainstorm some more values that we could pass here. I can propose a spec PR for SDK Stats from there. How would you feel about that coming as a part of a follow up PR since any change should be additive?

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: We can avoid these strings by not sending in the event that we can't find the URLs from the config service.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Should put the ikey in the config as well. We could be sending to breeze, so this is still required.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: We should utilize session storage. We should start this at daily, but pull from the config so we can change this on the fly later.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Just create local map for this (cache result)

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: We could use feature/instrumentation stats for the collection of this information to reduce size over the wire.

Follow up discussion with Hector/Leighton/myself to determine the best approach.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Start with 3p (but constructed to work with both), in particular the JSLL shim might be a good place to start.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -0,0 +1,565 @@
...
+        } catch (e) {
+            // Unparseable -> no config available
+        }
+    }
+
+    return result;
+}
+
+/** Default SDK Stats config fetch (fetch, else XHR); calls oncomplete with the parsed config or null. */
+function _defaultStatsCfgFetch(cfgUrl: string, oncomplete: (result: IInternalSdkStatsCfgResult) => void): void {
```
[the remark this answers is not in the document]
JacksonWeber: Good catch - fixed in dbb7b7b. Both `_defaultStatsCfgFetch()` and `_parseStatsCfg()` now declare `IInternalSdkStatsCfgResult | null` so the internal signatures match the public `InternalSdkStatsCfgFetchFn` type (which already declared the nullable result) and the actual null-on-failure behaviour.

channels/applicationinsights-channel-js/src/Sender.ts
```diff
@@ -1074,13 +1125,13 @@ export class Sender extends BaseTelemetryPlugin implements IChannelControls {
...
+            function _getPayload(payload: IInternalStorageItem[], urlOverride?: string): IInternalPayloadData {
                 if (isArray(payload) && payload.length > 0) {
                     let batch = _self._buffer.batchPayloads(payload);
                     let headers = _getHeaders();
                     let payloadData: IInternalPayloadData = {
                         data: batch,
-                        urlString: _endpointUrl,
+                        urlString: urlOverride || _endpointUrl,
                         headers: headers,
                         disableXhrSync: _disableXhr,
```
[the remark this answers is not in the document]
JacksonWeber: Agreed, this is a real leak risk. Fixed in dbb7b7b - `_getPayload()` now uses an empty header set whenever `urlOverride` is supplied, so customer-configured `customHeaders` (Authorization etc.) are never forwarded to the Microsoft-owned SDK Stats endpoint. The SDK Stats POST does not need any of the customer headers; `SenderPostManager` still adds the transport-level headers it requires.

shared/AppInsightsCore/Tests/Unit/src/ai/InternalSdkStats.Tests.ts
```diff
@@ -0,0 +1,435 @@
...
+                
+                // Create mock payload data with timing information
+                const payloadData = {
+                    urlString: "https://example.endpoint.com",
+                    data: "testData",
+                    headers: {},
+                    timeout: 0,
+                    disableXhrSync: false,
+                    statsData: {
+                        startTime: "2023-10-01T00:00:00Z" // Simulated start time
```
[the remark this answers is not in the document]
JacksonWeber: Correct - fixed in dbb7b7b. `startTime` is now `Date.now()`, so `utcNow() - startTime` produces a valid number instead of NaN and `Request_Duration` is emitted with a real value.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Changed this to pull from `cfgUrl` and construct the EU endpoint based on that.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Added the session storage for the internal SDK Stats events storage. Set the default collection interval to 1 hr but it should accept a `stats.shrtInt` value from the dynamic config to allow us to determine how often internal SDK stats are sent.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: I removed the hardcoded old `STATS_SDK_IKEY` and source it now from the dynamic config using `config.stats.iKey`. Should not work for sending this to breeze as well in case that ever happens in the future.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Replaced 12 full region names with 8 shared prefixes. Also moved to using a regex parse. Copilot suggested this and looks like it saves an addition 71 bytes.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Added a cache mapping customer endpoints to resolved SDK Stats config URLs, and updated `_isEuEndpoint()` to now run only once per endpoint instead of for each metric passed to `_track()`. Cache will clear automatically upon `stats.cfgUrl` change.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: I've added the snippet version to the SDK version in place of something like a distro or exporter per the naming convention established in the spec. As for collecting NPM load vs. CDN load, I'd like to address this in a follow up PR for feature SDK stats if possible since this PR is already quite large.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: The customer’s core is now only used to observe dynamic configuration and endpoint changes. SDK Stats no longer passes through the customer’s telemetry initializers, plugins, batching, headers, or ingestion endpoint. 

Each SKU now provides a callback to create the appropriate isolated SDK Stats pipeline. I'll do some more testing with this new approach, but from what I've done so far it looks to be working.

shared/AppInsightsCore/src/core/InternalSdkStats.ts
```diff
@@ -0,0 +1,789 @@
...
+        // Start a new window.
+    }
+
+    return null;
+}
+
+function _saveStore(logger: IDiagnosticLogger, key: string, store: _IStatsStore) {
+    let json = getJSON();
+    if (json) {
+        utlSetSessionStorage(logger, key, json.stringify(store));
```
MSNev: nit: This is unlikely, but it's probably worth putting this in a try / catch block as well, as historically, I have seen some sites added overrides (prototype pollution) to `Number`s and string that has caused JSON.stringify() to fail (mostly because recursion -- ie. they had a library that some internal values that referenced themselves)

shared/AppInsightsCore/src/interfaces/ai/IInternalSdkStats.ts
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { IPayloadData } from "./IXHROverride";
+
+/**
+ * The interface for the stats beat plugin, which is responsible for collecting and sending statistics about the SDK.
+ * It is used to track the performance and usage of the SDK, and to identify any issues or errors that may occur.
+ * @since 3.3.7
```
MSNev: nit: the @ since tags are probably out of date now (unless they have just been moved -- didn't check that)

channels/applicationinsights-channel-js/Tests/Unit/src/InternalSdkStats.tests.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    private initializeCoreAndSender(config: any, instrumentationKey: string) {
+        const sender = new Sender();
+        const core = new AppInsightsCore();
+        const coreConfig = {
+            instrumentationKey,
+            stats: {
+                shrtInt: 900,
+                // The config url gates collection, without it nothing is collected or sent
+                cfgUrl: "https://data.stats.monitor.azure.com/cfg/v1.json",
```
MSNev: Probably should be the test endpoint to be safer `tst-data.stats.monitor.azure.com`, just so we don't have test data accidently going to the prod endpoint

MSNev: Which should eventually be the pure scripting test url

shared/AppInsightsCore/Tests/Unit/src/ai/InternalSdkStats.Tests.ts
```diff
@@ -0,0 +1,1026 @@
...
+import { IConfiguration } from "../../../../src/interfaces/ai/IConfiguration";
+import { createStatsMgr, getStatsCfgUrl } from "../../../../src/core/InternalSdkStats";
+import { IInternalSdkStatsState } from "../../../../src/interfaces/ai/IInternalSdkStats";
+import { ITelemetryItem } from "../../../../src/interfaces/ai/ITelemetryItem";
+import { IPlugin } from "../../../../src/interfaces/ai/ITelemetryPlugin";
+import { IAppInsightsCore } from "../../../../src/interfaces/ai/IAppInsightsCore";
+import { FeatureOptInMode } from "../../../../src/enums/ai/FeatureOptInEnums";
+
+const STATS_COLLECTION_SHORT_INTERVAL: number = 900; // 15 minutes
+const STATS_TEST_CFG_URL = "https://data.stats.monitor.azure.com/cfg/v1.json";
```
MSNev: Same as previous test endpoint comment

shared/AppInsightsCore/src/core/InternalSdkStats.ts
JacksonWeber: Added the try/catch here.

shared/AppInsightsCore/src/interfaces/ai/IInternalSdkStats.ts
JacksonWeber: Updated the tag versions here

channels/applicationinsights-channel-js/Tests/Unit/src/InternalSdkStats.tests.ts
JacksonWeber: Makes sense, updated this as well.

shared/AppInsightsCore/src/diagnostics/ThrottleMgr.ts
```diff
@@ -67,6 +71,39 @@ export class ThrottleMgr {
...
 
+        /**
+         * Check whether a feature is enabled and sampled in by its named throttle configuration.
+         * A missing or disabled throttle configuration leaves the feature opt-in decision unchanged.
+         * The sampling decision is cached until the configured sampling rate changes.
+         * @param feature - The featureOptIn and throttleMgrCfg key.
+         * @param sdkDefaultState - The default state when featureOptIn does not define the feature.
+         * @returns true when the feature is enabled and allowed by its throttle configuration.
+         */
+        _self.canUseFeature = (feature: string, sdkDefaultState?: boolean): boolean => {
```
MSNev: To be consistent with the rest of the throttle manager, we should also persist the details and support the max number of events per day, month, year etc.

As this was part of the major effort with the throttle manager so that we could control the number of events that any single browser would send. We could probably just reuse the concept of "msgId" by refactoring to so that it's either stored as a string or number and when it's a string with say some sort of prefix like "ft:" we know this is a feature matrix of values. the problem with this is that the existing deployed SDK's may not handle loading strings as numbers (this includes when running different versions of the SDK in the same runtime).
The alternate would be that when it's a string we persist into a separate "feature" setting in the config etc.

MSNev: This way we can reuse the existing "canThrottle" function, but we will still need something like this.

I'm thinking more like instead of storing "messages" in a queue we store callbacks, so the existing sendMessage would wrap a callback that performs the flush message, which the useFeature (canUeFeature) would call the passed callback to then perform the operation.... But this way it can still use the internal counting / persisting of the values.

JacksonWeber: Refactored feature throttling to use the existing throttle flow. The queue now stores callbacks, so SDK Stats sends honor sampling, intervals, daily limits, and `maxSendNumber`, with feature state stored separately from message throttles. I also verified a real send against the test endpoint and confirmed subsequent and 0%-sampled sends are suppressed.

AISKU/src/AISku.ts
```diff
@@ -66,7 +66,7 @@ const CDN_USAGE = "CdnUsage";
 const SDK_LOADER_VER = "SdkLoaderVer";
 const ZIP_PAYLOAD = "zipPayload";
 const SDK_STATS = "SdkStats";
-var _sdkVersion = "#version#";
+var _sdkVersion = '3.4.3';
```
MSNev: 😃

JacksonWeber: Fixed the pinned version here

shared/AppInsightsCore/src/interfaces/ai/IConfig.ts
```diff
@@ -382,9 +382,9 @@ export interface IConfig {
     addIntEndpoints?: boolean;
 
     /**
-     * [Optional] Sets throttle mgr configuration by key
+     * [Optional] Sets throttle manager configuration by numeric message ID or string feature name.
      */
-    throttleMgrCfg?: {[key: number]: IThrottleMgrConfig};
+    throttleMgrCfg?: {[key: string]: IThrottleMgrConfig};
```
MSNev: This is public, so we should have `[key: number | string]` just in case.

JacksonWeber: Updated the typing here.

JacksonWeber:
PR was going slightly over the bundle size limit, so I achieved some minification savings via using a regex for the EU region parsing instead of the functions that existed before.

AISKU/src/AISku.ts
```diff
@@ -88,7 +88,7 @@ const sdk_stats_throttle_config = {
     disabled: false,
     limit: cfgDfMerge<IThrottleLimit>({
         samplingRate: 100,
-        maxSendNumber: 1
+        maxSendNumber: 8832 // Longest three-month window: 92 days * 24 hours * 4 reports per hour
```
MSNev: This would also cause every browser to send up to 8832 on a single day / hour / minute.

So is would be better to set the interval to daily as this will still inherit the default which is to only send on the 28th day every 3 months

JacksonWeber: Makes sense, I've updaed this to use the `dayInterval` value and `maxSendNumber` to just 24 * 4 per hour

MSNev (approved):

MSNev (approved):

PR #2754, opened 2026-08-12
fix(test): stabilize flaky SpanLifeCycle custom endTime test

hectorhdzg:
The "EndTime: custom endTime is respected" test used Date.now() as the custom span end time and asserted the recorded endTime round-tripped within 10ms. The span start time is derived from performance.timeOrigin + performance.now(), not Date.now(). On loaded/virtualized CI hosts (observed on the Windows build container, Node 20) the system wall clock can drift slightly behind the span's perf-based start time. When Date.now() lands before the span start, span.end() clamps the duration to zero and sets endTime === startTime, so the round-trip difference exceeds the 10ms tolerance and the test intermittently fails (763 tests, 1 failed).

Derive the custom end time from the span's own startTime so the comparison stays within the same time domain the SDK uses internally. This guarantees end > start (no zero-duration clamping) and keeps the value in the epoch range for an exact millisToHrTime round-trip, making the assertion deterministic.

### DIFF
--- a/AISKU/Tests/Unit/src/SpanLifeCycle.Tests.ts
+++ b/AISKU/Tests/Unit/src/SpanLifeCycle.Tests.ts
@@ -570,7 +570,16 @@ export class SpanLifeCycleTests extends AITestClass {
             test: () => {
                 // Arrange
                 const span = this._ai.startSpan("endtime-custom");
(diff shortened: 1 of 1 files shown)


JacksonWeber (approved):
LGTM after copilot comment is resolved.

